PR #1, opened 2026-06-12
Let Lemonade auto-select its llama.cpp backend

rominf:
## Summary

- The Lemonade adapter hard-forced `lemonade load --llamacpp rocm` with no fallback. On GPUs the bundled ROCm build does not support (e.g. the Radeon 780M / gfx1103 iGPU), Lemonade reports `llamacpp:rocm` as "Unsupported GPU", so every managed load returned a body-less 500 and spun in a load/evict/unload loop. Standalone Lemonade serves the same model because it falls back to Vulkan.
- Now query `lemonade backends`, pick the best supported llama.cpp backend (`rocm` > `vulkan` > `cpu`), install it on demand, and load with it instead of hardcoding rocm.
- The backend is resolved at install and re-resolved at serve time, so existing installs self-heal without a reinstall. Health verification, persisted state, and adapter messaging all follow the selected backend rather than assuming rocm.

**Why:** managed loads were completely broken on GPUs without a supported ROCm llama.cpp build, even though the model runs fine via Lemonade's own Vulkan fallback.

**Risk:** low — behavior is unchanged on GPUs where `rocm` is supported (it stays the top priority); the fallback only engages where the previous hardcoded path already failed.

## Test plan

- [x] `cargo test -p rocm-engine-lemonade` (incl. new backend-parser/selection tests)
- [x] `cargo build --release -p rocm`
- [x] Manual: restarted a previously-500-looping Qwen3-4B service; log shows `Installing backend: llamacpp:vulkan`, `Using LlamaCpp Backend: vulkan`, model loaded on `AMD Radeon 780M Graphics`, chat smoke `200 OK` (~10.5 tok/s); `services list` reports `ready`.

### DIFF
--- a/engines/lemonade/src/lib.rs
+++ b/engines/lemonade/src/lib.rs
@@ -31,8 +31,11 @@ const DEFAULT_HOST: &str = "127.0.0.1";
 const DEFAULT_MODEL: &str = "Qwen3-4B-Instruct-2507-GGUF";
 const DEFAULT_MODEL_REPO_DIR: &str = "models--unsloth--Qwen3-4B-Instruct-2507-GGUF";
 const DEFAULT_MODEL_GGUF: &str = "Qwen3-4B-Instruct-2507-Q4_K_M.gguf";
-const ROCM_BACKEND_RECIPE: &str = "llamacpp";
+const LLAMACPP_RECIPE: &str = "llamacpp";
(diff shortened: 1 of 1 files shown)


juhovainio (approved):

michaelroy-amd (approved):
Approving: clean, conflict-free (lemonade auto-backend). Landing as part of incremental PR-stack merge.

PR #2, opened 2026-06-12
Enable native-certs for ureq across all crates

rominf:
## Summary

- Enable ureq's `native-certs` feature in every crate that makes HTTPS requests (`rocm`, `rocmd`, `rocm-core`, `rocm-engine-lemonade`), so TLS validates against the operating system's certificate store instead of the bundled `webpki-roots` static set.

**Why:** in environments that terminate TLS with a custom/private root CA or a corporate MITM proxy, the static webpki root set rejects the re-signed certificate chain and downloads fail. Using the system trust store lets those certificates (already installed at the OS level) be honored, while remaining a no-op on machines with a standard public CA chain.

**Risk:** low — only changes which trust anchors ureq consults; no API or behavior change for standard public endpoints.

## Test plan

- [x] `cargo build` across the workspace with the feature enabled.
- [x] Verified downloads succeed behind a TLS-intercepting proxy with a custom root CA, where they previously failed certificate verification.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -98,7 +98,7 @@ checksum = "a7be2f067ccd8d4b4d4a66ddafe0f32a5dff31732f32dbff85fefc40929b1f72"
 dependencies = [
  "keyring-core",
  "log",
- "security-framework",
+ "security-framework 3.7.0",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


r0x0r (approved):

r0x0r:
TLS error is gone when downloading Python on macOS with this pr

rominf:
Thanks for a review, @r0x0r. Let's wait for https://github.com/ROCm/rocm-cli/pull/4 to be merged, to make all CI checks pass.

michaelroy-amd (approved):
Approving: clean native-certs enablement across crates. Incremental PR-stack merge.

PR #3, opened 2026-06-12
Migrate Python environment management from pip to uv

rominf:
## Summary

- Replace `pip` with `uv` for all Python package installations (ComfyUI dependencies, TheRock SDK/torch wheels, managed venv setup).
- Switch managed Python provisioning from `python-build-standalone` downloads to `uv python install` / `uv python find` (uv sources the same astral-sh builds but handles download, caching, extraction, and path resolution internally).
- Rename the internal runtime format tag from `pip` to `wheel` to better reflect what it installs (`--format wheel`, on-disk `runtimes/wheel/`, enums, JSON tool schemas, and tests).

**Why:** uv is dramatically faster than pip for dependency resolution and installation, and unifies venv creation + Python installation in a single tool. This removes several bespoke helpers (pip cache, pip timeout/retry knobs, pip arg builders, the python-build-standalone fetch/extract chain) in favour of uv's built-in equivalents.

**Non-obvious:** `uv pip install` targets the venv via `--python <venv_python>` rather than activating it; `uv_pip_install_base()` encodes this so call-sites don't have to. The `pip_cache_dir` field on `ComfyUiManifest` is retained as `Option<PathBuf>` (set to `None`) for forwards/backwards serialisation compatibility with manifests already on disk.

**Risk:** medium — touches all managed-runtime provisioning paths. Mitigated by the retained manifest compatibility field and updated smoke/unit tests.

## Test plan

- Smoke test updated to assert `rocm-cache/uv` (not `rocm-cache/pip`) is absent on first run.
- `therock_sdk_install_test.py` updated to pass `--format wheel` and assert `format: wheel` in output.
- Unit test `uv_install_base_targets_venv_python` verifies `uv_pip_install_base` produces a correctly targeted arg list.

### DIFF
--- a/apps/rocm/src/comfyui.rs
+++ b/apps/rocm/src/comfyui.rs
@@ -2,10 +2,10 @@ use crate::{format_structured_tool_call, runtime_usability_status, therock};
 use anyhow::{Context, Result, bail};
 use flate2::read::GzDecoder;
 use rocm_core::{
-    AppPaths, RocmCliConfig, download_file_to_path, format_http_base_url, managed_pip_cache_dir,
+    AppPaths, RocmCliConfig, download_file_to_path, ensure_uv_binary, format_http_base_url,
     runtime_is_cosmopolitan_windows, runtime_is_linux, runtime_is_windows,
     runtime_path_for_windows_child, runtime_path_list_join, runtime_path_list_split,
-    runtime_paths_equivalent, unix_time_millis,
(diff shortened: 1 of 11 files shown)


juhovainio (approved):

michaelroy-amd (approved):
Approving: pip→uv migration incl. InstallFormat pip→wheel rename. Incremental PR-stack merge.

PR #4, opened 2026-06-12
Fix CI: prek hooks, headless/flaky test fixes, and drop vendored Codex

rominf:
## Summary

Add [prek](https://github.com/j178/prek) (a fast, drop-in `pre-commit` replacement) so contributors run the same checks locally that CI enforces, catching fmt/clippy/lint failures before they reach a PR.

`.pre-commit-config.yaml` wires up:
- **Hygiene** via prek's native `repo: builtin` hooks (trailing-whitespace, end-of-file-fixer, check-yaml, check-merge-conflict, mixed-line-ending) — zero-setup, no external clone.
- **Python**: `ruff-check --fix` + `ruff-format` (`scripts/`, engine workers).
- **Shell**: `shellcheck` (`install.sh`, `scripts/*.sh`).
- **Rust** (local hooks mirroring CI): `cargo fmt` on commit; `cargo clippy -D warnings` and `cargo test` on **pre-push** to keep commits fast.
- **PowerShell**: `scripts/check_powershell_syntax.py` — cross-platform, runs where `pwsh` exists and skips cleanly otherwise (CI covers Windows).

Vendored `third_party/` is excluded. Setup/usage documented in the README.

This PR also includes the mechanical cleanup needed to make the hooks pass on a clean checkout (repo-wide ruff format + import sort, three shellcheck fixes) and fixes pre-existing fmt/clippy failures the new hooks surface — notably gating `rocm-core`'s Windows-only `OsStr` import and the Windows GPU/doctor inventory helpers behind `cfg` so the workspace is clippy-clean on all targets.

## CI test fixes

Wiring `cargo test` into the workflow surfaced pre-existing test failures (red on `main` too) that only reproduce in CI's environment. Fixed here so the suite is green:

- **Headless terminal size (chat follow):** the chat view resolved its follow-the-bottom scroll against the real `size()` before the first paint; with no TTY (CI) this fell back to the follow sentinel / bailed out, leaving the view at the top. Added an `effective_terminal_size` fallback used by both scroll paths — also fixes auto-follow when the CLI runs without a controlling terminal.
- **Long test artifact paths:** TUI snapshot tests rooted their install dirs under the workspace; on CI checkouts the paths were long enough to wrap inside the panels and break the assertions. Moved the test artifact dir under the system temp dir.
[...]

PR #5, opened 2026-06-12
Speed up CI

r0x0r:
This pull request optimizes the CI pipeline and acceptance test workflow by reusing prebuilt release binaries, particularly the large `rocm-codex` binary, to avoid redundant compilation steps. It introduces environment variable handling and script logic to detect and use existing binaries, improving build efficiency and reducing CI runtime.

**CI/CD pipeline improvements:**
- Added a step in `.github/workflows/ci.yml` to pre-build release binaries and the vendored Codex binary during push events, making them available for later acceptance steps and reducing duplicate compilation.
- Set the `CARGO_TARGET_DIR` environment variable in the CI workflow to ensure consistent target directory usage across steps.
- Acceptance install lifecycle step now runs only on push events, aligning with when prebuilt binaries are available.
- Removed redundant `cargo build` steps from both Linux and Windows jobs, as builds are now handled earlier or reused. [[1]](diffhunk://#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fL37-R46) [[2]](diffhunk://#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fL73-L76)

**Acceptance and packaging script enhancements:**
- Updated `scripts/acceptance-install-upgrade-tui-uninstall.sh` to resolve and export the location of the prebuilt `rocm-codex` binary, enabling downstream scripts to reuse it if present.
- Modified `scripts/build-vendored-codex.sh` to support a fast path: if the `ROCM_CODEX_PREBUILT_BINARY` environment variable is set and points to an existing binary, the script installs it directly instead of rebuilding, further reducing redundant builds in CI and local development.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,12 @@ jobs:
       - name: Test
         run: cargo test --workspace --all-targets
 
+      - name: Pre-build release binaries and codex for acceptance step
+        if: github.event_name == 'push'
+        run: |
+          cargo build --release -p rocm -p rocmd -p rocm-engine-pytorch -p rocm-engine-llama-cpp -p rocm-engine-lemonade -p rocm-engine-atom -p rocm-engine-vllm -p rocm-engine-sglang
+          bash scripts/build-vendored-codex.sh release
+
       - name: Local no-fallback smoke
         run: python scripts/smoke_local.py --skip-build
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,7 @@ permissions:
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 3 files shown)


rominf:
Superseded by dropping the vendored Codex tree entirely. Instead of prebuilding `rocm-codex` to avoid the CI timeout, we're removing the ~70-crate vendored Codex workspace outright — it ships no essential functionality (reachable only via the hidden `--experimental-codex-tui` flag) and removing it fixes the timeout at the source while dropping the upstream-sync burden. The removal is being folded into #4 so CI is fixed in one PR. Thanks @r0x0r for the investigation here.

PR #9, opened 2026-06-15
Add channel tui option

r0x0r:
This pull request adds support for selecting the ROCm SDK installation channel (release or nightly) in both the onboarding and install manager flows of the TUI. Users can now switch between stable (release) and preview (nightly) channels via the UI, and the selected channel is used when installing the SDK. The UI, keyboard handling, and test code have all been updated to support this new capability.

### DIFF
--- a/apps/rocm/src/therock.rs
+++ b/apps/rocm/src/therock.rs
@@ -1365,9 +1365,7 @@ fn channel_rocm_candidates(versions: &[String], channel: TheRockChannel) -> Vec<
 }
 
 fn is_stable_runtime_version(version: &str) -> bool {
-    parse_version(version)
-        .map(|parsed| parsed.stage == VersionStage::Stable)
-        .unwrap_or(false)
+    parse_version(version).is_some_and(|parsed| parsed.stage == VersionStage::Stable)
 }
 
 fn select_latest_stack_package(
--- a/apps/rocm/src/therock.rs
+++ b/apps/rocm/src/therock.rs
@@ -3504,21 +3504,21 @@ mod tests {
     #[test]
     fn pip_runtime_selects_latest_common_rocm_suffix_not_latest_rocm_package() {
         let rocm_versions = vec![
-            "7.13.0a20260512".to_owned(),
-            "7.13.0a20260513".to_owned(),
-            "7.14.0a20260602".to_owned(),
+            "7.13.0".to_owned(),
+            "7.13.1".to_owned(),
+            "7.14.0".to_owned(),
         ];
         let torch_versions = vec![
-            "2.9.1+rocm7.13.0a20260513".to_owned(),
-            "2.10.0+rocm7.13.0a20260513".to_owned(),
+            "2.9.1+rocm7.13.1".to_owned(),
+            "2.10.0+rocm7.13.1".to_owned(),
         ];
         let torchvision_versions = vec![
-            "0.24.0+rocm7.13.0a20260513".to_owned(),
-            "0.25.0+rocm7.13.0a20260513".to_owned(),
+            "0.24.0+rocm7.13.1".to_owned(),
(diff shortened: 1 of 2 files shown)


rominf (dismissed):
## Review

Thanks for the PR — the toggle semantics, Release default, and correct threading of the channel value through to `install sdk --channel` are all sound. Tests are well updated and replacing `handle_key(Up)` navigation with direct `select_onboarding_choice()` calls is a clear improvement.

A few issues need addressing before merge.

### Critical: channel label always shows "Release"

Both `install_sdk_choice_label` and `onboarding_menu_label` hardcode the string `"Release"` as the left-side label regardless of the selected channel:

```rust
InstallSdkChoice::Channel => format!(
    "{:<14} {}",
    "Release",        // ← always "Release"
    if matches!(channel, InstallSdkChannel::Release) {
        "Stable (default)"
    } else {
        "Nightly (preview)"  // when nightly: displays "Release         Nightly (preview)"
    }
),
```

When the user switches to Nightly the row reads `Release         Nightly (preview)` — contradictory. Compare with the Folder row: `Install folder: /path/`. The channel row label should be a neutral descriptor like `"Channel:"` so it stays coherent in both states: `Channel:        Nightly (preview)`.

### `toggle_install_sdk_channel` updates status unconditionally

```rust
fn toggle_install_sdk_channel(&mut self) {
    if let Some(state) = self.install_manager.as_mut()
        && let InstallManagerScreen::Sdk { channel, .. } = &mut state.screen
    {
        *channel = channel.toggle();
        state.message = None;
        state.detail_scroll = 0;
    }
    self.status = "ROCm channel changed. ...".to_owned(); // ← runs even when the if let fails
}
```

If the install manager isn't on the SDK screen, the toggle silently does nothing but the status still claims the channel changed. Move `self.status =` inside the `if let` block.

### Duplicate Left/Right arms in `handle_onboarding_key`

The Left and Right handlers for `OnboardingMenuChoice::Channel` are identical — same guard conditions, same two-line body. Either collapse them or extract a helper (consistent with how `toggle_install_sdk_channel` already abstracts the install manager side).

### Minor

- Status messages use lowercase `"release is stable, nightly is preview."` but UI labels use title-case `"Stable (default)"` / `"Nightly (preview)"`. Align the casing.
- Detail text says `"On the Release row, Enter or Left/Right switches channel."` — will mislead users who have Nightly selected. Should be `"On the Channel row, ..."`.
- The `Char('2')` test block is testing pre-existing number-key-jump behavior — a short comment explaining what `'2'` does and why the assertion follows would help future readers.
- It's worth confirming the terminal height change from 32 to 36 rows in the install-failed render test reflects real additional content rather than masking an overflow.

r0x0r:
All review comments have been addressed:

**Critical: channel label always shows "Release"**
Fixed — both `install_sdk_choice_label` and `onboarding_menu_label` now use `"Channel:"` as the neutral left-side label, so the row reads `Channel:        Stable (default)` or `Channel:        Nightly (preview)` consistently.

**`toggle_install_sdk_channel` updates status unconditionally**
Fixed — `self.status =` is now inside the `if let` block so the status only updates when the channel actually toggles.

**Duplicate Left/Right arms in `handle_onboarding_key`**
Fixed — collapsed into a single `KeyCode::Left | KeyCode::Right` arm.

**Minor items:**
- Casing aligned: detail text now says `"Release is stable..."` and `"switches Release and Nightly"` to match UI title-case.
- Detail text already said `"On the Channel row, ..."` in the pushed branch; confirmed correct.
- Added a comment to the `Char('2')` test block explaining that `'2'` jumps directly to the second actionable row (Channel) and activates it.
- Investigated the 32→36 height change: all assertions pass at height 32 as well, so the extra rows were unnecessary. Reverted to 32 to avoid masking overflow.

juhovainio (approved):

PR #19, opened 2026-06-17
Multi-target builds: native per-OS CI and Cosmopolitan removal

volen-silo:
Replaces the Cosmopolitan universal-binary approach with native per-OS Rust builds, and adds CI to match.

## Commits
1. **Remove Cosmopolitan; build native Rust per-OS** — eliminate every Cosmopolitan/APE code path and tool, collapsing `target_vendor="cosmo"` conditional compilation back to native `cfg(windows)`/`cfg(not(windows))`. Strips cosmo runtime detection, TCP I/O, curl download, the APE argv0 resolver, and PATH-join helpers from rocm-core; collapses cosmo branches in the pytorch/lemonade engines and across apps/rocm; makes keyring deps plain per-OS targets; deletes the cosmo spike scripts and plan docs.
2. **Add per-OS CI: native hosted builds + optional GPU/WSL hardware tests** — build and test natively on GitHub-hosted runners (Ubuntu + Windows) as the required gate, uploading each job's debug binaries as artifacts. Optional self-hosted hardware tests (gfx1151 Windows, Instinct bare-metal, WSL Ubuntu/Fedora) download those artifacts and exercise what hosted runners cannot (real AMD GPU, real WSL).

## CI behavior
- Required gate is fully hosted: `build-and-test` (ubuntu), `windows-build-and-test` (windows), `fmt`, `clippy`, `coverage`.
- Every build/test job is gated on `fmt` + `clippy`, so expensive cycles run only when lint is green.
- Hardware-test jobs are opt-in per fleet via repo/org variables (`ENABLE_WIN11_GPU_CI`, `ENABLE_INSTINCT_CI`, `ENABLE_WSL_UBUNTU_CI`, `ENABLE_WSL_FEDORA_CI`), are `continue-on-error` (never block a PR), and are guarded against untrusted fork PRs.

## Validation
- `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and the workspace test suite pass locally on Linux (one known pre-existing flaky TUI test passes in isolation).
- Windows / GPU / WSL paths require their respective runners.


.github/actionlint.yaml
```diff
@@ -0,0 +1,8 @@
+self-hosted-runner:
+  labels:
+    - windows-gfx1151-gpu-rocm
+    - wsl
+    - ubuntu
+    - fedora
+    - instinct
+    - rocm
```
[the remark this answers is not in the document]
volen-silo: False positive — `linux` is a built-in actionlint runner label, so it doesn't need to be listed here.

Verified by running actionlint (v1, latest) against this exact `ci.yml` + `actionlint.yaml`: clean pass, exit 0, no warning about `linux`. When actionlint *does* flag a genuinely unknown label it prints its known-label set, which explicitly includes the built-ins:

> "self-hosted", "x64", "arm", "arm64", "linux", "macos", "windows"

The custom `labels:` list only needs the non-built-in labels — which is why `wsl`, `ubuntu`, `fedora`, `instinct`, `rocm`, and `windows-gfx1151-gpu-rocm` are listed but `linux` is not (`ubuntu`/`fedora` are listed precisely because, unlike `linux`, they are not built-in). Adding `linux` would be misleading clutter, so leaving as-is.

.github/workflows/ci.yml
```diff
@@ -26,6 +32,22 @@ jobs:
       - name: Build
         run: cargo build --workspace --all-targets
 
+      - name: Upload Linux binaries
```
rominf: The artifact is uploaded **before** the `Test` step (line 51). If tests fail, `build-and-test` is marked failed so downstream hardware jobs won't run — but the artifact already exists and is accessible. Semantically it's cleaner to upload **after** all test steps so the stored artifact always corresponds to a passing build. Hardware runners that want artifacts for debugging failed builds can still use `if: always()` on the download step.

Same issue on `windows-build-and-test` (line 90).

.github/workflows/ci.yml
```diff
@@ -26,6 +32,22 @@ jobs:
...
+        uses: actions/upload-artifact@v4
+        with:
+          name: linux-binaries
+          if-no-files-found: error
+          retention-days: 3
```
rominf: `retention-days: 3` may be too short for self-hosted runners. If a hardware runner is offline over a weekend or short maintenance window, the artifact expires before the job can pick it up. `7` days gives a safer buffer without a significant storage cost. (Same on line 95 for the Windows artifact.)

.github/workflows/ci.yml
```diff
@@ -157,58 +197,160 @@ jobs:
...
+        run: python .\scripts\smoke_local.py --skip-build
 
-      - name: ATOM adapter detection
-        run: .\target\debug\rocm-engine-atom.exe detect
+      - name: PyTorch GPU end-to-end
+        run: python .\scripts\pytorch_therock_gpu_test.py --engine .\target\debug\rocm-engine-pytorch.exe
+
+  instinct-gpu-test:
+    name: Instinct bare-metal GPU test (artifact)
+    runs-on: [self-hosted, linux, instinct, rocm]
```
rominf: The artifact downloaded here was built on `ubuntu-latest` (currently Ubuntu 24.04, glibc 2.39). If the Instinct bare-metal runner is on RHEL/Rocky/an older Ubuntu, the binary will fail at startup with `GLIBC_2.XX not found` — no error from Cargo, just a silent runtime crash.

`wsl-fedora-test` correctly addresses this by building inside WSL Fedora. Worth either:
- Documenting the minimum glibc version required of the `instinct` runner label, or
- Building on `ubuntu-22.04` (glibc 2.35) to target a lower baseline.

.github/workflows/ci.yml
```diff
@@ -157,58 +197,160 @@ jobs:
...
+        with:
+          name: linux-binaries
+          path: target/debug
 
-      - name: SGLang adapter capabilities
-        run: .\target\debug\rocm-engine-sglang.exe capabilities
+      - name: Test prebuilt binaries in WSL Ubuntu
+        run: |
+          $ErrorActionPreference = "Stop"
+          $wp = (wsl -d Ubuntu wslpath "'$($PWD.Path)'").Trim()
```
rominf: The WSL path `$wp` is embedded inside single quotes in the bash heredoc:
```powershell
cd '$wp'
```
If the Windows working-directory path contains a single quote (unusual but possible), this breaks bash parsing without a useful error. Consider using `printf '%q'` to safely quote the path, or at minimum add a guard:
```powershell
if ($wp -match "'") { throw "Working directory path contains a single quote; cannot safely pass to bash" }
```
Same pattern on line 339 in `wsl-fedora-test`.

.github/workflows/ci.yml
```diff
@@ -157,58 +197,160 @@ jobs:
...
+    runs-on: [self-hosted, wsl, fedora]
+    # Builds in-WSL, so gate on the lint check too (don't burn the scarce
+    # Fedora runner if fmt/clippy are red).
+    needs: [fmt, clippy]
+    # Secondary: builds inside WSL Fedora instead of reusing the ubuntu-latest
+    # artifact, to avoid cross-distro glibc mismatch. Non-blocking.
+    continue-on-error: true
+    if: vars.ENABLE_WSL_FEDORA_CI == 'true' && (github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository)
+    timeout-minutes: 60
+    defaults:
```
rominf: `wsl-fedora-test` is missing the `permissions:` block that all other hardware test jobs include:
```yaml
permissions:
  contents: read
  actions: read
```
`actions: read` is needed for `download-artifact` to access artifacts via the GitHub API. Without it the download step will fail with a 403 when the job eventually runs.

.github/workflows/ci.yml
```diff
@@ -157,58 +197,160 @@ jobs:
...
-        run: .\target\debug\rocm-engine-pytorch.exe resolve-model qwen32b
+      - name: Build and GPU-test in WSL Fedora
+        run: |
+          $ErrorActionPreference = "Stop"
+          $wp = (wsl -d Fedora wslpath "'$($PWD.Path)'").Trim()
+          $script = @"
+          set -euo pipefail
+          cd '$wp'
+          export CARGO_TARGET_DIR="`$HOME/.cache/rocm-cli-target"
+          if ! command -v cargo >/dev/null 2>&1; then
```
rominf: The Rust toolchain is downloaded from `sh.rustup.rs` on every CI run (~400–600 MB). `CARGO_TARGET_DIR` is redirected to `~/.cache/rocm-cli-target` for incremental builds, but the toolchain itself and the Cargo registry are not cached across runs. Consider adding:
```yaml
- name: Cache Cargo registry and toolchain
  uses: actions/cache@v4
  with:
    path: |
      ~/.cargo/registry
      ~/.cargo/git
      ~/.cargo/bin
      ~/.rustup
    key: fedora-cargo-${{ hashFiles('**/Cargo.lock') }}
    restore-keys: fedora-cargo-
```
Since this is a self-hosted WSL runner the cache directory persists on disk, so even a simple directory-level check before the `curl | sh` would help.

docs/implementation-completion-audit.md
```diff
@@ -121,14 +120,13 @@ python3 scripts/llama_cpp_therock_gpu_test.py --self-test
...
-For WSL, run the same universal binary through the APE/Linux path, not as a
-Windows process:
+For WSL/Linux, build and run the native Linux binary directly:
 
 ```bash
-sh /home/jam/rocm-cli-e2e/bin/rocm doctor
-python3 scripts/local_assistant_therock_gpu_test.py --rocm /home/jam/rocm-cli-e2e/bin/rocm --rocm-launch-prefix sh --engine lemonade --model qwen --require-tool-call
-python3 scripts/local_assistant_therock_gpu_test.py --rocm /home/jam/rocm-cli-e2e/bin/rocm --rocm-launch-prefix sh --engine pytorch --model qwen --require-tool-call
-python3 scripts/comfyui_therock_gpu_test.py --rocm /home/jam/rocm-cli-e2e/bin/rocm --rocm-launch-prefix sh
+/home/jam/rocm-cli-e2e/bin/rocm doctor
```
rominf: Personal username `jam` embedded in documentation paths (`/home/jam/rocm-cli-e2e/bin/rocm`). This appears in a public repo and will rot as soon as that developer changes machines. Suggest replacing with a generic placeholder like `/home/user/rocm-cli-e2e/bin/rocm` or `<your-path>/rocm`.

rominf (changes requested):
Overall the direction is right — removing the Cosmopolitan path eliminates a lot of fragile complexity, and the CI structure is solid. A few items worth fixing before merge, plus some lower-priority suggestions.

.github/workflows/ci.yml
volen-silo: Fixed in 1ebaab9 — moved the upload to after Build+Test on both jobs. (Note: the hardware jobs gate on this job's success via `needs`, so a test-failed build never reaches them anyway — but agreed the post-test position is semantically cleaner.)

.github/workflows/ci.yml
volen-silo: Fixed in 1ebaab9 — bumped `retention-days` to 7 on both artifacts.

.github/workflows/ci.yml
volen-silo: Fixed in 1ebaab9 — `build-and-test` now runs on `ubuntu-22.04` (glibc 2.35), so `linux-binaries` targets a lower baseline. If the Instinct runner turns out to be RHEL/Rocky 8 (glibc 2.28) we'd need an even lower baseline (container or older build); the 22.04 choice is documented in a comment for now.

.github/workflows/ci.yml
volen-silo: Fixed in 1ebaab9 — added a guard that throws if `$wp` contains a single quote, on both the WSL Ubuntu and WSL Fedora jobs.

.github/workflows/ci.yml
volen-silo: Not applicable — `wsl-fedora-test` builds in-WSL and has no `download-artifact` step, so it doesn't need `actions: read`. The three hardware jobs that *do* download artifacts already carry the permissions block.

.github/workflows/ci.yml
volen-silo: Leaving as-is — these are persistent self-hosted WSL runners, so `~/.cargo`/`~/.rustup` survive across runs, and the job already guards with `if ! command -v cargo` before the rustup install, so the toolchain isn't re-downloaded every run. `actions/cache` targets ephemeral runners; not needed here.

docs/implementation-completion-audit.md
volen-silo: Fixed in 1ebaab9 — replaced `/home/jam` with `/home/user` throughout the doc (5 occurrences).

.github/workflows/ci.yml
volen-silo: Correction: the `ubuntu-22.04` pin (1ebaab9) broke CI, so 2337278 reverts it and documents the baseline instead.

The pin failed because the shared cargo cache is keyed on `runner.os` (`"Linux"`) and can't distinguish 22.04 from 24.04 — the job restored its own prior 24.04 cache onto the 22.04 runner, so 24.04-built proc-macro `.so` files (`GLIBC_2.39 not found`) and aws-lc-sys C objects (`undefined symbol: __isoc23_sscanf`) failed to load/link. Pinning the image also wouldn't actually guarantee portability (RHEL/Rocky 9 ship glibc 2.34, below even 22.04's 2.35).

So the artifact stays built on `ubuntu-latest`, with a documented glibc-baseline contract: a hardware-test runner that downloads it must have a compatible (>=) glibc, otherwise it should build in-place as `wsl-fedora-test` does. When the Instinct runner is actually provisioned we can check its glibc and, if needed, add a proper low-baseline build (container/manylinux) then.

.github/workflows/ci.yml
```diff
@@ -26,6 +32,22 @@ jobs:
...
         run: cargo build --workspace --all-targets
 
+      - name: Upload Linux binaries
+        uses: actions/upload-artifact@v4
+        with:
+          name: linux-binaries
+          if-no-files-found: error
+          retention-days: 3
+          path: |
+            target/debug/rocm
```
rominf: These hardware GPU tests run against **debug** binaries. Debug builds skip optimizations and may behave differently from what users actually ship — LLVM can mask undefined behaviour in debug mode that surfaces under `-O2`, and performance-sensitive code paths (e.g. inference throughput, startup latency) won't reflect production reality.

Consider building with `--release` before uploading artifacts, and adjusting the paths to `target/release/*`. The compile time penalty on hosted runners is real (~3–5× longer), but the hardware tests validate user-facing behaviour — testing them against a debug binary gives weaker coverage guarantees.

If keeping debug builds is intentional (e.g. for richer panic messages during early runner bring-up), at least document that explicitly in the job comment so future maintainers know it is a deliberate tradeoff. Same applies to the Windows artifact (line 97).

volen-silo: Deliberate — documented it explicitly in a983406 (on both the Linux and Windows upload steps).

Reasoning for keeping debug here: every job that consumes this artifact is a *functional* GPU smoke test — `rocm doctor` (device detection), engine `detect`/`capabilities`, `smoke_local.py` (CLI + no-CPU-fallback policy), and the PyTorch GPU end-to-end run. None of them measure throughput or latency, so the debug-vs-release difference doesn't change what they assert. The UB-under-`-O2` angle is essentially nil for these Rust smoke checks too.

Against that, building `--release` on `build-and-test` would add ~3-5x to the **required** hosted gate on every PR, for marginal coverage gain on functional tests. Release-fidelity / optimized validation already lives in the nightly/release pipeline, which builds `manylinux2014` (glibc 2.17) binaries — so the production binary does get exercised, just not on the per-PR path.

If we later want optimized GPU validation per-PR, it's a clean follow-up (a separate opt-in release artifact + hardware job) rather than slowing the required gate. For now the job comments call out the tradeoff so it's not a silent default.

.github/workflows/ci.yml
```diff
@@ -157,58 +208,162 @@ jobs:
...
+        with:
+          name: linux-binaries
+          path: target/debug
 
-      - name: SGLang adapter capabilities
-        run: .\target\debug\rocm-engine-sglang.exe capabilities
+      - name: Test prebuilt binaries in WSL Ubuntu
+        run: |
+          $ErrorActionPreference = "Stop"
+          $wp = (wsl -d Ubuntu wslpath "'$($PWD.Path)'").Trim()
```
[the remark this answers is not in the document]
volen-silo: Fixed in 36e0287 — switched to `wslpath -u "$($PWD.Path)"` (no literal single quotes; PowerShell double-quotes handle spaces, `-u` forces Windows→WSL). Applied to both the Ubuntu and Fedora jobs. The existing `$wp` single-quote guard stays as belt-and-suspenders.

.github/workflows/ci.yml
```diff
@@ -157,58 +208,162 @@ jobs:
...
+        shell: pwsh
+    steps:
+      - uses: actions/checkout@v6
 
-      - name: Resolve large-model recipe
-        run: .\target\debug\rocm-engine-pytorch.exe resolve-model qwen32b
+      - name: Build and GPU-test in WSL Fedora
+        run: |
+          $ErrorActionPreference = "Stop"
+          $wp = (wsl -d Fedora wslpath "'$($PWD.Path)'").Trim()
```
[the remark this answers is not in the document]
volen-silo: Fixed in 36e0287 — switched to `wslpath -u "$($PWD.Path)"` (no literal single quotes; PowerShell double-quotes handle spaces, `-u` forces Windows→WSL). Applied to both the Ubuntu and Fedora jobs. The existing `$wp` single-quote guard stays as belt-and-suspenders.

volen-silo:
**Scope update:** the self-hosted GPU/WSL hardware-test jobs have been removed from this PR and deferred to a follow-up (commit 356a95e).

Rationale: those jobs never actually ran (no self-hosted runners are wired into the repo yet), and the artifact-upload steps that fed them were uploading binaries nothing consumed. This PR now contains only what CI actually exercises: native per-OS build/test on hosted runners (lint-gated) plus fmt/clippy/coverage/lint — on top of the Cosmopolitan removal.

The planned hardware-test design (build-once-on-hosted → download artifacts → test on dedicated gfx1151 / Instinct / WSL-Ubuntu / WSL-Fedora runners, opt-in + non-blocking + fork-safe) is captured in `docs/ci-hardware-testing.md` and will land as a reviewed unit in the follow-up, validated against real runners.

This resolves the earlier review threads that targeted the hardware jobs (debug-vs-release binaries, the `wslpath` quoting, the Fedora `permissions` block, and the Instinct glibc baseline) — that code now moves to the follow-up where it can be tested for real. The `.github/actionlint.yaml` self-hosted label config was dropped too, since nothing references those labels anymore.

rominf (approved):
Approving. The Cosmopolitan removal is complete and clean (no residual `target_vendor`/`cosmo` references in any Rust/TOML), all round-1 feedback was addressed (fixed or declined with sound reasoning), and the scope reduction — deferring hardware tests to a follow-up with a design doc in `docs/ci-hardware-testing.md` — makes this a focused, reviewable unit. CI is fully green.

Non-blocking, out of scope for this PR: the Windows-builds-release / Linux-builds-debug asymmetry in the CI matrix predates this PR (it lives on `main`), and the personal `jam` username still appears in a few untouched docs; both are worth a follow-up sweep but don't belong here.

PR #29, opened 2026-06-21
§2: wire `rocm daemon` to the real background-helper loop + on-demand autostart

michaelroy-amd:
**Stack 1/10** · base: `main`

Fixes ROADMAP §2 (daemon regression). `rocm daemon` only printed a status panel; the real loop lived in the separate `rocmd` crate.

- Adds `rocmd` path-dep to `apps/rocm` (acyclic); `rocm daemon` now runs the real foreground loop via `rocmd::run_from_args(["rocmd","run","--automations-enabled"])`.
- `rocm daemon --status` preserves the old status panel.
- `ensure_background_helper_running()`: on-demand **detached** autostart (double-spawn-guarded via `AutomationRuntimeState.daemon_pid` + `process_is_running`) on `automations enable` and `serve --managed`.

**Test plan:** `cargo build/test/clippy -D warnings/fmt --check` (workspace, all-targets) green; integration test `daemon_runs_real_foreground_loop`. Reviewed: rust + security + maintainability, zero unresolved CRITICAL/HIGH.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3230,6 +3230,7 @@ dependencies = [
  "rocm-engine-pytorch",
  "rocm-engine-sglang",
  "rocm-engine-vllm",
+ "rocmd",
  "rpassword",
  "serde",
  "serde_json",
--- a/apps/rocm/Cargo.toml
+++ b/apps/rocm/Cargo.toml
@@ -30,6 +30,7 @@ rocm-engine-llama-cpp = { path = "../../engines/llama-cpp" }
 rocm-engine-pytorch = { path = "../../engines/pytorch" }
 rocm-engine-protocol = { path = "../../crates/rocm-engine-protocol" }
 rocm-engine-sglang = { path = "../../engines/sglang" }
+rocmd = { path = "../rocmd" }
 rocm-engine-vllm = { path = "../../engines/vllm" }
 rpassword.workspace = true
 serde.workspace = true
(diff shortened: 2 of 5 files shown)


rominf (commented):
Approve with minor comments. CI green; daemon now drives the real `rocmd` foreground loop, verified by the new `daemon_run.rs` integration test (matches the real `rocmd run` banner). Two things worth a follow-up:

- **`ensure_background_helper_running` is a non-atomic guard.** `running && pid_alive` has a TOCTOU window and there's no flock/singleton in `rocmd`, so two concurrent `automations enable` / `serve --managed` calls could each spawn a daemon writing the same state file. Low probability for an interactive CLI, but an advisory file lock around startup would close it.
- `ensure_background_helper_running` itself is untested (neither the already-running no-op nor the autostart paths). Worth a unit test on the guard.

Branch will want a rebase — `main` has moved 4 commits ahead of the base.

michaelroy-amd:
rebased onto main. The subsequent 10 PRs will be rebased cascading in order (i.e. 30 off 29, 312 off 30, etc).

rominf (approved):

rominf (changes requested):

```
git config gpg.format ssh
git push --force-with-lease
```

rominf (approved):
Signing gate cleared — all commits now show **Verified**. Re-reviewed the current diff against `main` after the re-sign.

Approving. The daemon wiring is correct: the sequential double-spawn guard works via persisted `daemon_pid` + `process_is_running`, spawning is argv-only (no shell), and setup errors still propagate.

Non-blocking follow-up (unchanged from before): there's no advisory lock in `rocmd` startup, so a direct `rocm daemon` invocation or a narrow concurrent-autostart race could spawn a duplicate. Worst case is a transient duplicate daemon, not corruption — worth a follow-up flock/named-lock at `rocmd run` startup, but not a merge blocker.

PR #30, opened 2026-06-21
§1: rocm-core-free execution seam (RocmToolExecutor boundary)

michaelroy-amd:
**Stack 2/10** · base: `supergoal/phase-1-daemon-wiring`

Load-bearing foundation for §1 chat parity: a plain-data tool-call boundary the dash chat uses so the bin executes ROCm tools.

- `crates/rocm-dash-tui/src/tool_exec.rs`: `RocmToolExecutor` trait + `RocmToolOutcome`/`ApprovalIntent` — **no rocm-core/tokio/ratatui** in signatures.
- Bin `dash_seam.rs::BinToolExecutor` delegates to the existing engine (`validate_chat_tool_call` → `run_internal_mcp_call`); injected via `ResolvedArgs` (None for demo/replay/mock).

Plumbing only — no behavior change yet. **Invariants:** rocm-dash-tui keeps 0 rocm-core edges; agent.rs stays the sole `rig` namer. All mandatory cargo gates green.

### DIFF
--- a/apps/rocm/src/dash.rs
+++ b/apps/rocm/src/dash.rs
@@ -208,10 +208,9 @@ async fn run_async(
     // Inject the bin-side tool-execution seam for a live dash only. Demo/replay
     // and the offline chat mock keep `tool_executor = None` and behave as today.
     if !chat_mock && replay.is_none() {
-        args.tool_executor = Some(std::sync::Arc::new(crate::dash_seam::BinToolExecutor::new(
-            paths.clone(),
-        ))
-            as rocm_dash_tui::tool_exec::SharedRocmToolExecutor);
+        let executor: rocm_dash_tui::tool_exec::SharedRocmToolExecutor =
+            std::sync::Arc::new(crate::dash_seam::BinToolExecutor::new(paths.clone()));
+        args.tool_executor = Some(executor);
     }
     // A live daemon is only needed when connecting; replay/demo feeds events
     // straight into the TUI, so skip the embedded daemon in that mode.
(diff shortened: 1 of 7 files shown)


rominf (approved):
LGTM. Clean `RocmToolExecutor` seam — object-safe trait, right supertraits for `Arc<dyn>`, plain-data args, no behavior change. Two hermetic tests cover the read-only and approval-required outcomes.

One optional note: the read-only classification is computed twice on differently-normalized args (seam vs `run_internal_mcp_call`); they agree for all current name-based tools, but a one-line comment noting the double-classification is intentional would help future readers.

rominf (changes requested):

```
git config gpg.format ssh
git push --force-with-lease
```

rominf (approved):

volen-silo (approved):
Approve — clean, tightly-scoped plumbing.

The dependency inversion is done right: rocm-dash-tui keeps **0** rocm-core edges (only doc-comment mentions in `tool_exec.rs`), and the bin impl delegates to the real engine (`validate → is_read_only → run_internal_mcp_call | chat_tool_approval_request`) instead of reimplementing classification. Genuinely behavior-neutral — injected for live dash only, `None` for demo/replay/mock. Gates green locally: `rocm-dash-tui --lib` 363 passed, `rocm dash_seam` 2 passed.

Non-blocking notes for Phase 3/4:
- The sync `execute()` blocks on the **read-only** path too (`examine`, `bridge_snapshot`, `load_managed_services`), not just mutation — route it through `spawn_blocking` when wiring into the async loop.
- Envelope asymmetry: `run_internal_mcp_call` unwraps a nested `{"arguments": {...}}` via `internal_mcp_arguments`, but the read-only check and the approval branch operate on raw args — worth a one-line comment that `execute()` expects unwrapped args.

Neither blocks merge. Ship it.

r0x0r (approved):

PR #31, opened 2026-06-21
§1: read-only ROCm tools + nav/session slash commands + parity map

michaelroy-amd:
**Stack 3/10** · base: `supergoal/phase-2-execution-seam`

- 12 read-only ROCm tools (doctor/engines/services/gpu/…) via a `rocm_read_tool!` macro, delegating through the Phase-2 seam; registered in both Rig + ChatGPT backends.
- Nav/session + read-only slash commands (`/home /help /clear /quit /doctor /runtimes /config /logs /gpu /model /daemon`).
- Seeds `docs/dash-parity-map.md` (30 legacy commands).

All cargo gates green; mock round-trip + slash-dispatch tests. Reviewed clean.

### DIFF
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -7135,6 +7135,10 @@ pub(crate) fn validate_chat_tool_call(call: &providers::ChatToolCall) -> Result<
     }
     match call.name.as_str() {
         "examine"
+        // `doctor` is the dash-side LLM tool + `/doctor` overlay name for the
+        // same machine inspection the bin exposes as `examine`. Accept it as an
+        // alias so a model-issued `doctor` call resolves end-to-end.
+        | "doctor"
         | "bridge_snapshot"
         | "gpu_snapshot"
         | "engines"
@@ -7706,6 +7710,7 @@ pub(crate) fn chat_tool_call_is_read_only(call: &providers::ChatToolCall) -> boo
     matches!(
         call.name.as_str(),
         "examine"
+            | "doctor"
             | "bridge_snapshot"
             | "gpu_snapshot"
             | "engines"
@@ -8334,7 +8339,7 @@ pub(crate) fn run_internal_mcp_call(
     }
 
     match name {
-        "examine" => {
+        "examine" | "doctor" => {
             let examine = ExamineSummary::gather()?;
             let text = render_examine_text()?;
             Ok(internal_mcp_tool_success(text, serde_json::json!(examine)))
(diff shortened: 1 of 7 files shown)


rominf (approved):
LGTM. Strong test coverage for a TUI PR — read-only round-trip, None-executor no-panic, full slash-command matrix, tool-name disjointness.

One note: `rocm_command`'s "mutating subcommands are rejected" guarantee is enforced across the seam in the bin (correct layer) but isn't verifiable/tested in this PR — worth ensuring the executor impl lands a test rejecting e.g. `["serve"]`.

rominf (changes requested):

```
git config gpg.format ssh
git push --force-with-lease
```

rominf (approved):

volen-silo (dismissed):
Request changes — one blocking bug: the `doctor` read-only tool is dead-on-arrival.

`agent.rs` registers a tool named `doctor` (first in `ROCM_READ_TOOL_NAMES`), but the bin engine has no `doctor` — the equivalent is `examine`. There is zero `doctor` handling in `validate_chat_tool_call`, `chat_tool_call_is_read_only`, or the `run_internal_mcp_call` dispatch in `apps/rocm/src/main.rs`.

So when the model actually calls it:

```
DoctorRocmTool::call → run_rocm_read_tool → BinToolExecutor::execute("doctor", …)
  → validate_chat_tool_call → bail!("local assistant requested unsupported ROCm tool `doctor`")
  → RocmToolOutcome::Error
```

Every invocation returns `{"error":"…unsupported ROCm tool `doctor`"}`. The other 11 names all match the bin's accept-list — `doctor` is the only break.

Why no test caught it: `read_only_tool_round_trips_to_json` uses `FakeExec`, which echoes a fixed value and ignores the name, so it never exercises the real name-validation path. The name-completeness test only checks the dash-side array against itself.

Fix — pick one:
- Register the tool as `examine` to match the bin (simplest), or
- Add a `doctor` → `examine` alias in the bin's three sites (`validate`, `is_read_only`, dispatch) so the `doctor` name works end-to-end and matches the `/doctor` overlay naming.

Either way, please add a guard test asserting `ROCM_READ_TOOL_NAMES` ⊆ the bin's accepted tool set (or an integration test driving the real `BinToolExecutor`), so a name-only tool can't silently ship again.

Non-blocking (carryover from the seam PR): the LLM read-only tools call `e.execute()` synchronously inside the async `Tool::call`, but `execute()` does blocking I/O. The slash-tool path correctly uses `spawn_blocking`; the LLM tool-call path does not, so a model-issued `engines`/`services`/`bridge_snapshot` blocks a tokio worker. Consider wrapping `execute()` in `run_rocm_read_tool` for parity.

The rest is solid — slash routing, the decoupled `SlashToolReply`, and the concise summarizer are all clean and test-locked.

michaelroy-amd:
Thanks @volen-silo — addressed in the latest push (force-pushed after a rebase onto current `main` + re-sign; all commits show **Verified**).

**Blocking `doctor` bug — fixed (your option B, the alias).**
The dash side intentionally registers the tool as `doctor` to match the `/doctor` overlay, so rather than rename, I added `doctor` as an alias of the bin's `examine` in all three sites you identified:
- `validate_chat_tool_call` — `"examine" | "doctor"` accepted
- `chat_tool_call_is_read_only` — `"examine" | "doctor"` treated read-only
- `run_internal_mcp_call` dispatch — `"examine" | "doctor" =>` runs the same `ExamineSummary::gather()` path

So a model-issued `doctor` now resolves end-to-end instead of returning `unsupported ROCm tool \`doctor\``.

**Guard test — added.**
`read_tool_names_are_subset_of_bin_accept_list` iterates `ROCM_READ_TOOL_NAMES` and asserts each name is in the bin's accept-list (i.e. `validate_chat_tool_call` never rejects it as `unsupported ROCm tool`). It's deliberately hermetic — it checks name acceptance only, not arg validation, so it doesn't drive the real `BinToolExecutor::execute` / live I/O but still fails loudly if a name-only tool is ever added again. (Confirmed it goes RED on the pre-fix tree and GREEN after.)

**Non-blocking `spawn_blocking` carryover — not in this round.**
Tracking the blocking-`execute()`-inside-async-`Tool::call` parity separately rather than folding it into this PR; flagging so it's not silently dropped.

Re-review when you have a moment — happy to adjust the alias-vs-rename call if you'd prefer the simpler rename.

rominf (approved):
Re-reviewed the updated diff after the doctor-fix push.

The blocking `doctor` bug @volen-silo flagged is now resolved in `apps/rocm/src/main.rs`: `examine` | `doctor` is aliased across all three sites — `validate_chat_tool_call`, `chat_tool_call_is_read_only`, and the `run_internal_mcp_call` dispatch — so a model-issued `doctor` call now resolves end-to-end instead of bailing with `unsupported ROCm tool`. A regression test asserting the dash tool names are accepted by the bin allowlist was added, closing the name-only-tool gap that let it ship silently.

All 7 CI checks pass (build-and-test, windows-build-and-test, clippy, fmt, coverage, hawkeye, lint). Commits are signed/Verified.

Approving. @volen-silo — when you have a moment, please re-review so the prior CHANGES_REQUESTED can be cleared; this is the bottom of the stack and #32–#38 are queued behind it.

PR #32, opened 2026-06-21
§1: in-chat approval modal + core mutating tools (install/engine/serve/services)

michaelroy-amd:
**Stack 4/10** · base: `supergoal/phase-3-readonly-and-nav`

The defining capability: mutating ROCm actions behind explicit human approval.

- Reuses `ui/approval.rs`; mutating tools surface `ApprovalRequired` via `ClientMsg::ChatApprovalRequired` → app-loop modal → on Approve `execute_approved` (runs as a **captured subprocess**, no TUI corruption) → exactly one follow-up turn.
- Wires install/engine/serve/services + safety validators reused unchanged (no public bind / loopback-only / no CPU device).

Reviewed: rust + **security** + maintainability — approval gate is unbypassable. All cargo gates green.

### DIFF
--- a/apps/rocm/src/dash_seam.rs
+++ b/apps/rocm/src/dash_seam.rs
@@ -53,10 +53,23 @@ impl RocmToolExecutor for BinToolExecutor {
                         }
                         b
                     },
-                    args: req.args,
+                    name: name.to_owned(),
+                    arguments: args.clone(),
                 }),
                 Err(e) => RocmToolOutcome::Error(e.to_string()),
             }
+        }
+    }
+
+    fn execute_approved(&self, name: &str, args: &serde_json::Value) -> RocmToolOutcome {
+        // Replay the approved mutating call via the captured-subprocess path
+        // (`allow_mutation = true`). `run_internal_mcp_call` re-validates the
+        // call first, so the safety validators stay the single gate; it runs the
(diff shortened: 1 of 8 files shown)


rominf (commented):
Approve with minor comments. Reviewed this as the security-critical PR and the gate holds: the approval seam is enforced **centrally** — the LLM tool loop can only return `ApprovalRequired`, execution happens solely via `execute_approved` from the event-loop Approve verdict, and both paths re-validate before running. No bypass, no TOCTOU on the validated args, argv-vector spawning (no shell injection). Excellent test coverage proving approval-is-required from both entrypoints.

Minors:

- `launch_server` arg-building pushes positional user values as bare argv; a value like `--managed` would be parsed as a flag by the child clap parser (arg-confusion, not injection — same gated binary). A `--` separator before positionals would be good defense-in-depth (pre-existing, low risk).
- `approval.rs` defaults the modal cursor to `Approve`, so Enter on a freshly-opened destructive modal approves. The two-key flow mitigates it, but defaulting destructive modals to Deny is the safer convention.

rominf (changes requested):

```
git config gpg.format ssh
git push --force-with-lease
```

rominf (approved):

PR #33, opened 2026-06-21
§1: lifecycle mutating commands — /update /comfyui /uninstall /setup

michaelroy-amd:
**Stack 5/10** · base: `supergoal/phase-4-approval-core-mutating`

Routes the remaining lifecycle commands through the Phase-4 approval modal via the `rocm_command` seam; read paths stay read-only.

- `/update` (report=read, `--apply`=approval), `/comfyui` (status/logs=read, install/start/stop=approval), `/uninstall` (**bare = dry-run safe default**, real=approval), `/setup` (status=read, reset=approval).
- One contained bin addition: `setup` classifier + in-process status renderer. DRY `rocm_cmd_request` helper.

Reviewed: rust + security (uninstall dry-run default + approval-gated) + maintainability. All cargo gates green.

### DIFF
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -8740,6 +8740,15 @@ fn run_rocm_read_only_in_process(paths: &AppPaths, args: &[String]) -> Result<St
         {
             render_uninstall_dry_run(paths)
         }
+        [command] if command.eq_ignore_ascii_case("setup") => {
+            render_setup_status_text(paths, &config)
+        }
+        [command, subcommand]
+            if command.eq_ignore_ascii_case("setup")
+                && subcommand.eq_ignore_ascii_case("status") =>
+        {
+            render_setup_status_text(paths, &config)
+        }
         _ => bail!(
             "unsupported in-process read-only rocm command: {}",
             format_structured_tool_call("rocm", args)
@@ -16149,6 +16158,27 @@ model recipes
             Some("rocm")
         );
         assert!(mcp_tool_result_text(&result).contains(env!("CARGO_PKG_VERSION")));
+    }
+
+    #[test]
+    fn setup_status_renders_in_process() {
+        let (_root, paths) = test_paths("setup-status-in-process");
+
+        for args in [
+            vec!["setup".to_owned()],
+            vec!["setup".to_owned(), "status".to_owned()],
+        ] {
+            let text = run_rocm_read_only_in_process(&paths, &args)
+                .expect("setup status read path should render in-process");
+            assert!(
+                !text.trim().is_empty(),
+                "setup status text should be non-empty for {args:?}"
+            );
+            assert!(
+                text.contains("ROCm setup"),
+                "setup status text should be reached for {args:?}: {text}"
+            );
+        }
     }
 
     #[test]
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -7539,6 +7539,16 @@ fn chat_rocm_command_action_from_args(mut args: Vec<String>) -> Result<ChatRocmC
                 command_title: "ComfyUI".to_owned(),
             })
         }
+        Some("setup") if second.as_deref().is_none_or(|value| value == "status") => {
+            Ok(ChatRocmCommandAction::ReadOnly(args))
+        }
+        Some("setup") if second.as_deref() == Some("reset") => {
+            Ok(ChatRocmCommandAction::Approval {
+                args,
+                pending_title: "Reset first-time setup".to_owned(),
+                command_title: "Setup".to_owned(),
+            })
+        }
         Some(command) => bail!("local assistant cannot use unsupported rocm command `{command}`"),
         None => bail!("rocm_command requires at least one argument"),
     }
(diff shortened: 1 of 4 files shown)


rominf (approved):
LGTM. Verified the security-relevant paths: all four lifecycle commands construct argv client-side then re-classify server-side, so the TUI can't force execution. `/uninstall` defaults to `--dry-run` (real uninstall is approval-gated + auto-`--yes`); `/setup reset` is approval-gated and correctly non-destructive. Tests cover the read/mutate split end-to-end including the uninstall dry-run default. No findings.

rominf (changes requested):

```
git config gpg.format ssh
git push --force-with-lease
```

rominf (approved):

rominf (dismissed):
Requesting changes — one safety issue in the `/uninstall` flag handling (confirmed against this head). Note: I'll re-review the final diff once the rebase onto main lands and the PR is mergeable; flagging now so it isn't lost.

**`/uninstall` resolves conflicting flags to the destructive interpretation (`app.rs:881-889`).**
The real-vs-dry-run decision only inspects the second token: `matches!(rest.split_whitespace().nth(1), Some("--apply" | "now" | "real"))`. So `/uninstall --apply --dry-run` sets `real = true` and dispatches `["uninstall"]` (a real uninstall) even though `--dry-run` is present. This runs counter to the command's own documented SAFE-default intent in the comment just above it.

It is mitigated — the real uninstall is approval-gated, so this surfaces an approval modal rather than executing silently — but for a destructive command the parse shouldn't quietly drop an explicit `--dry-run`. Please parse all tokens and either prefer `--dry-run` when both are present, or reject the conflicting combination with a guiding error; keep the accepted real trigger aligned with the documented `--apply` (the undocumented `now`/`real` aliases are also worth dropping or documenting). A regression test that `/uninstall --apply --dry-run` does not dispatch a real uninstall would lock it in.

Once that's addressed (and the branch is mergeable) I'll re-review.

michaelroy-amd:
Thanks @rominf — this is addressed in the current head (now rebased onto `main`, mergeable, all commits Verified). The `/uninstall` parse was reworked (via the committed review suggestion + a clippy/fmt follow-up), mapping to each of your points:

- **Parse all tokens, not just the 2nd** — now `let flags: Vec<&str> = rest.split_whitespace().skip(1).collect();` then `flags.contains(&"--apply")` / `flags.contains(&"--dry-run")` (`app.rs:865-867`).
- **Reject the conflicting combination** — `/uninstall --apply --dry-run` no longer dispatches; it pushes a guiding error: *"conflicting /uninstall flags: choose either --dry-run (safe) or --apply (real uninstall)"* (`app.rs:868-872`).
- **Align the real trigger with documented `--apply`; drop `now`/`real`** — done; only `--apply` triggers a real uninstall now (`real = saw_apply`).
- **Regression test** — added `slash_uninstall_conflicting_flags_is_guided` (asserts no dispatch + an error turn), alongside the existing dry-run/apply tests. `cargo test -p rocm-dash-tui uninstall` green; fmt + `clippy -D warnings` clean.

Ready for your re-review.

rominf (dismissed):
Re-reviewing the updated head — the `/uninstall` flag-safety fix resolves my earlier request.

`/uninstall` now collects all flags (`rest.split_whitespace().skip(1)`), rejects `--apply` + `--dry-run` together with a guiding error and no dispatch, and drops the undocumented `now`/`real` aliases so only the documented `--apply` triggers a real (approval-gated) uninstall. The regression test `slash_uninstall_conflicting_flags_is_guided` locks in that the conflicting combination does not dispatch. This matches the command's SAFE-default intent.

LGTM — approving.

crates/rocm-dash-tui/src/app.rs
```diff
@@ -816,6 +819,95 @@ impl AppState {
                     }
                 }
             }
+            // --- Group D-rest: lifecycle ops (read/mutate split via rocm_command) ---
+            // Read paths classify as ReadOnly in the bin and return a result turn;
+            // mutating paths classify as ApprovalRequired and open the Phase-4 modal.
+            "update" => {
+                // `/update` reports available updates (read-only); `/update --apply`
+                // (or `/update apply`) applies them (approval-gated).
+                let apply = matches!(rest.split_whitespace().nth(1), Some("--apply" | "apply"));
```
rominf: `/update` accepts both `apply` and `--apply` here, but the documented usage (`/update --apply`), `/uninstall --apply`, and the real `rocm update` flag (`--apply`) all use the dash form — the bare `apply` alias is undocumented. Suggest dropping it so `--apply` is the single consistent mutating trigger across the lifecycle commands. (Keep `setup`/`comfyui` as subcommands — those mirror the real CLI subcommand grammar, so `--apply` for the apply toggle vs. subcommands for operation selection is the right split.)

rominf: Done in 3b7afc3 — dropped the bare `apply` alias so `--apply` is the sole mutating trigger, matching the documented usage, `/uninstall --apply`, and the `rocm update --apply` flag. `setup`/`comfyui` stay as subcommands.

apps/rocm/src/main.rs
```diff
@@ -7539,6 +7539,16 @@ fn chat_rocm_command_action_from_args(mut args: Vec<String>) -> Result<ChatRocmC
...
+        Some("setup") if second.as_deref().is_none_or(|value| value == "status") => {
+            Ok(ChatRocmCommandAction::ReadOnly(args))
+        }
+        Some("setup") if second.as_deref() == Some("reset") => {
+            Ok(ChatRocmCommandAction::Approval {
+                args,
+                pending_title: "Reset first-time setup".to_owned(),
+                command_title: "Setup".to_owned(),
+            })
+        }
```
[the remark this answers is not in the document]
rominf: Fixed in 94adece. Added a `setup` arm to `ensure_rocm_command_is_read_only` (bare/`status` read-only, `reset` mutating), mirroring the bin classifier, plus a regression test (`rocm_command_helper_treats_setup_status_as_read_only_and_reset_as_mutating`). The read-only allowlist is now consistent across binaries.

volen-silo (approved):
LGTM — approving. All prior findings (the `/uninstall` conflicting-flags safety, the undocumented `/update` bare `apply` alias, and the `rocmd` `setup` guard gap) are resolved in the current head, and CI is fully green (7/7 incl. windows).

I did an independent cross-binary consistency pass — the property that the same `rocm_command` classifies identically across all three surfaces — and it holds for every lifecycle command:
- Bin classifier (`chat_rocm_command_action_from_args`): `update` (read unless `--apply`), `comfyui status/logs`, `uninstall --dry-run`, `setup`/`setup status` read; mutating variants → Approval.
- In-process exec (`run_rocm_read_only_in_process`): matching arms; bare `["update"]` is handled by `[command, rest @ ..]` (rest empty) → `render_update_text`.
- rocmd guard (`ensure_rocm_command_is_read_only`): `update => !--apply`, `comfyui`, `uninstall => --dry-run`, `setup => bare|status`.

All three line up — no `setup`-style divergence remains. `lifecycle_read_mutate_split_is_honest` is a good both-directions tripwire.

Minor nit (non-blocking): `/update` parses its flag positionally (`rest.split_whitespace().nth(1) == Some("--apply")`), whereas `/uninstall` was hardened to scan all tokens. So `/update <extra> --apply` would silently fall back to the read-only report. It fails safe (degrades to read-only, never to a destructive apply), so not a safety issue — just worth aligning with `/uninstall`'s all-flags parsing for predictability. Fine to leave or fold into a follow-up.

rominf:
Folded in the @volen-silo nit in 4e8c910: \`/update\` now scans all tokens for \`--apply\` (matching \`/uninstall\`), so the trigger is no longer position-sensitive (`/update <extra> --apply` applies). Added a `slash_update_apply_is_position_independent` regression test. Still fails safe either way.

PR #34, opened 2026-06-21
§1: automations/reviews/approve/reject/edit + permissions

michaelroy-amd:
**Stack 6/10** · base: `supergoal/phase-5-lifecycle-mutating-rest`

- `/automations` (list=read; enable/disable via `watcher_enable`/`watcher_disable`, validate-before-approval), `/reviews /approve /reject /edit` via a new `proposal_action` bin tool (show=read, approve/reject=mutate → `update_automation_proposal_status`).
- `/permissions` (status=read; full-access/ask via new `ConfigCommand::SetPermissions`, approval-gated).

Reviewed: rust + **security** (full-access grant cannot bypass approval — traced) + maintainability. All cargo gates green.


rominf (approved):
LGTM. Permissions model is default-deny (`ask`), and enforcement is tested in both directions (approve requires `allow_mutation`; show is read-only). Confirmed the dash executor always opens the modal for mutating tools and never consults `full_access` (the full-access bypass lives only in the untouched binary TUI).

Two notes:

- `/automations enable` arg parsing is positional and lossy — `--mode` is only honored as the token right after the watcher id; trailing/unknown tokens are silently dropped. UX, not security.
- `/approve` flips proposal status to `approved` but doesn't dispatch the underlying action — safe (approval can't trigger an unreviewed action), but confirm that matches intended Phase-6 scope if operators expect it to *run* the proposal.

rominf (changes requested):

```
git config gpg.format ssh
git push --force-with-lease
```

rominf (approved):

rominf (approved):

Re-reviewed: automations/reviews + approve/reject/edit + ; mutating watcher/proposal tools stay approval-gated via the Phase-4 modal and bin-side validators.

michaelroy-amd:
Pruned committed Copilot suggestions that were over-engineering / a false positive (simplicity-first / ponytail pass). Force-pushed; kept the one genuine fix.

**Kept**
- `/edit` slash-tool label fix (`review {id}` → matches the invoked command) — real user-facing accuracy.

**Dropped (with reasons)**
- **config `set-permissions` ValueEnum "fix"** — *false positive*. `PermissionsModeArg` already has `#[value(rename_all = "snake_case")]` (so `FullAccess` → `full_access`, matching what the dash sends) plus an explicit `alias = "full-access"`; `config_set_permissions_classifies_as_approval` already covers `full_access`. No bug to fix.
- **`proposal_id` control-char rejection** — defensive hardening (YAGNI) for local single-user LLM/user input; not worth the added validation surface.
- **watcher-id control-char rejection** — same.
- **`/automations enable --mode` missing-value handling** — the existing default-mode fallback is reasonable; special-casing a missing flag value adds branch complexity for an unlikely input.

PR #36, opened 2026-06-21
§1: provider switch (/provider) + /chat + Anthropic backend (rig)

michaelroy-amd:
**Stack 8/10** · base: `supergoal/phase-7-nl-plan`

- `AnthropicAgentClient` via rig (no Cargo feature needed; **agent.rs stays the sole rig namer**), registering the same read+mutating ROCm tools as the other backends.
- `/provider local|openai|anthropic` rebuilds the active backend (local restored from a saved clone); `/chat` entry/passthrough.
- Keys env-first → OS secure store via `provider_keys`, **never argv/logs**. ureq/reqwest partition intact (bin has no rig).

Reviewed: rust + **security** (no key in argv/logs — confirmed) + maintainability. All cargo gates green.


rominf (commented):
Approve with minor comments. `/provider` + Anthropic backend via `rig` looks good; default endpoint is hardcoded to `https://api.anthropic.com` (ignores `cfg.base_url`, can't be pointed elsewhere), key rides `x-api-key` and never hits argv/base_url, Debug redacts it.

One real bug:

- **Failed-switch state desync** (event-loop `provider_switch` drain): on a failed non-Local switch the code reverts `active_provider` to `Local` but leaves `agent` unchanged. If a prior `/provider openai` succeeded and a later `/provider anthropic` fails, the UI shows `local` while chat still routes to OpenAI. Fix: also restore `agent` (revert to the previously-active provider, not hardcoded `Local`). This path is untested. (Looks like #38's "honest provider revert" addresses exactly this — worth confirming #36 doesn't ship independently of that fix.)
- Minor: `AnthropicAgentClient::complete` triplicates the prompt/timeout/annotate tail of the other two backends; could extract a shared helper.

rominf (changes requested):

```
git config gpg.format ssh
git push --force-with-lease
```

rominf (approved):
Signing gate cleared — all commits now show **Verified**. Re-reviewed the current diff against `supergoal/phase-7-nl-plan` after the re-sign.

Approving as a stacked PR. The Anthropic provider wiring is solid: key is env-first then secure store, carried in-process (never in argv), `base_url` pinned to the provider default, key in `x-api-key` (won't leak into error strings). Tool/approval parity is preserved across all three backends.

One known caveat, accepted because it's fixed downstream: the failed-switch arm reverts `active_provider` to `Local` unconditionally, so a failed switch *while already on a remote provider* mislabels the active provider (display + routing, not a crash or capability hole). This is exactly the desync fixed by #38's honest provider revert. Acceptable in stack order.

rominf (approved):

Re-reviewed: provider switch /provider + /chat + Anthropic (rig) backend; API key resolved env→secure-store across the dash seam, tool/approval parity preserved.

rominf (changes requested):
Requesting changes — retracting my earlier approval. Two High-severity correctness bugs in the provider-switch path (both confirmed against the code at this head, not just the Copilot flags):

**1. `/provider openai` with no key silently builds a dead backend (`agent.rs:897-905`, `app.rs:1513-1520`).**
`RigAgentClient::new` falls back to a `"sk-no-key"` bearer when `cfg.api_key` is `None` and returns `Ok`, so `build_chat_agent(ChatProvider::Openai, …)` returns `Some(...)` even with `OPENAI_API_KEY` unset. The switch reports "switched to openai" and then 401s at request time. This violates `build_chat_agent`'s own documented contract ("`None` signals 'couldn't build' (e.g. a missing key) so the caller surfaces an actionable error turn instead of switching to a dead backend"), and it's inconsistent with the Anthropic path, which correctly fails fast (`AnthropicAgentClient::new` returns a Build error naming the env var when the key is absent).

Fix: gate the OpenAI build on a non-empty `args.chat_api_key` and return `None` when it's missing, so the event loop surfaces the actionable error turn. (The `sk-no-key` dummy should remain only for the managed-gateway/custom-header case that genuinely ignores the bearer.)

**2. Failed provider switch leaves `active_provider` and the live `agent` inconsistent (`app.rs:2025-2035`).**
On build failure the code sets `state.active_provider = ChatProvider::Local` but leaves `agent` unchanged. If the previous backend was remote (e.g. switching Anthropic → OpenAI and the OpenAI build fails), the UI then claims "local" while chat requests still route to the old remote backend.

Fix: in that failure branch, also restore `agent` to the local agent so the displayed provider and the live backend stay consistent (or, if local can't be rebuilt here, keep `active_provider` on the prior backend and word the error accordingly — but the two must not diverge).

Non-blocking (carryover Copilot nits): the parity-map row (`docs/dash-parity-map.md:42`) says keys are "env-first then secure store" but only the Anthropic key consults the secure store — the OpenAI key is env-only; and the argv-scan assertion (`app.rs:~3889`) isn't a reliable "no key in argv" proof. Worth tidying but not blocking.

Apologies for the earlier approval — I'd relied on Copilot's non-blocking summary and missed these inline findings. Once (1) and (2) are addressed I'll re-review promptly.

michaelroy-amd:
Pruned one committed Copilot suggestion (simplicity-first / ponytail pass). Force-pushed; kept the doc-accuracy fix.

**Kept**
- parity-map wording fix: only the Anthropic key consults the OS secure store; the OpenAI-compatible key is env-only — doc accuracy.

**Dropped (with reason)**
- **OpenAI keyless fail-fast** (gate provider build on non-empty `OPENAI_API_KEY`, return `None` otherwise) — would break the intended keyless flow for local OpenAI-compatible servers (vLLM / llama.cpp), which accept the dummy bearer by design. Keeping the current keyless-tolerant build behavior.
- (Also dismissed, not committed: provider-switch-failure leaves `agent` unchanged, and the `std::env::args()` argv-scan test — keeping the existing security-intent test as-is.)

rominf (dismissed):
Thanks for the update, but both blocking issues from my previous review are still open — the new commit only changes one line in `docs/dash-parity-map.md`; `app.rs` and `agent.rs` are unchanged. Keeping changes requested.

1. `/provider openai` with no key still builds a dead backend. `build_chat_agent`'s OpenAI arm (`app.rs:1514`) still passes `api_key: args.chat_api_key.clone()`, and `RigAgentClient::new` still falls back to the `sk-no-key` bearer (`agent.rs:899`) and returns `Ok` — so the switch reports success and then 401s at request time. (Note: the `chat_api_key.is_none()` check around `app.rs:1642` is the *default-launch* ChatGPT-OAuth fallback path, not the provider-switch build, so it doesn't cover this.) Please gate the OpenAI build on a non-empty key and return `None` when absent.

2. Failed provider switch still leaves state inconsistent. The failure branch (`app.rs:2021-2023`) sets `active_provider = ChatProvider::Local` but doesn't reset `agent`, so a failed switch from a remote backend leaves the UI claiming "local" while requests still route to the old remote backend.

Once these two are addressed I'll re-review.

michaelroy-amd:
@rominf both addressed. (1) `build_chat_agent`'s OpenAI arm now requires a non-empty `chat_api_key` and returns `None` otherwise, so a keyless `/provider openai` no longer builds the `sk-no-key` backend — the switch fails fast with the actionable error and stays on the current backend. New test `build_chat_agent_openai_requires_key` covers no-key→None and key→Some. (2) The failed-switch branch now also does `agent = local_agent.clone()` alongside `active_provider = Local`, so a failed switch from a remote backend no longer leaves requests routed to the old remote while the UI claims local. Ready for re-review.

michaelroy-amd:
Rebased onto current `main` (post-#35 squash) and re-signed; the only conflict was `docs/dash-parity-map.md` (kept main's `plan` row + this PR's `provider`/`chat` rows).

Both blocking issues from your last review are addressed in `1dac2c7` (`fix(dash): fail OpenAI provider switch fast without a key; reset agent on failed switch`) — that commit landed at 08:40 PDT on 6/25, **after** your 06:58 re-review of the docs-only `8aff7e0`, which is why they still looked open:

1. **OpenAI build with no key** — `build_chat_agent`'s OpenAI arm now gates on a non-empty key: `let api_key = args.chat_api_key.clone().filter(|k| !k.trim().is_empty())?;` and passes `Some(api_key)` to `RigAgentClient::new`, returning `None` when absent so the switch surfaces an actionable error and stays on the current backend instead of building a dead `sk-no-key` one. Covered by `build_chat_agent_openai_requires_key`.
2. **Failed switch left state inconsistent** — the failure branch now also resets the live agent (`agent = local_agent.clone();`) alongside `active_provider = ChatProvider::Local`, so a failed switch from a remote backend no longer routes requests to the old remote while the UI claims local.

`cargo fmt --check`, `clippy --all-targets --all-features -D warnings`, and `cargo test -p rocm-dash-tui` all clean. Re-requesting review.

juhovainio (approved):

PR #48, opened 2026-06-24
Improve CLI help 

r0x0r:
This pull request significantly improves the usability and user guidance of the `rocm` CLI by enhancing command documentation, providing better help messages, and introducing intelligent suggestions for mistyped commands. The changes focus on making the CLI more approachable for new users and easier to navigate, especially when encountering errors or seeking help.

**Enhanced CLI documentation and help:**

* Expanded the main CLI help and individual command help texts, including detailed `long_about` and `after_help` sections with usage examples for commands like `chat`, `update`, `model`, `serve`, `logs`, `install sdk`, and `engines install`. This provides users with clear instructions and practical command usage patterns. [[1]](diffhunk://#diff-85923bafe93886cd11cf9fc2a11c96d715facfd36cea9a7018b79526241f40a1L59-R75) [[2]](diffhunk://#diff-85923bafe93886cd11cf9fc2a11c96d715facfd36cea9a7018b79526241f40a1L131-R160) [[3]](diffhunk://#diff-85923bafe93886cd11cf9fc2a11c96d715facfd36cea9a7018b79526241f40a1R178-R184) [[4]](diffhunk://#diff-85923bafe93886cd11cf9fc2a11c96d715facfd36cea9a7018b79526241f40a1L179-R236) [[5]](diffhunk://#diff-85923bafe93886cd11cf9fc2a11c96d715facfd36cea9a7018b79526241f40a1L238-R286) [[6]](diffhunk://#diff-85923bafe93886cd11cf9fc2a11c96d715facfd36cea9a7018b79526241f40a1R341-R349) [[7]](diffhunk://#diff-85923bafe93886cd11cf9fc2a11c96d715facfd36cea9a7018b79526241f40a1R390-R392)

* Added descriptive comments to enum variants such as `InstallFormat`, `Provider`, `WatcherModeArg`, and `TelemetryModeArg` to clarify their intent and possible values.

**Improved error handling and user guidance:**

* Implemented logic to detect unrecognized single-word commands, providing a helpful error message that suggests the closest known command (using Levenshtein distance) and directs users to available help resources. This includes a new `unknown_command_hint` function and supporting logic.

* Updated the fallback notes for unmatched freeform requests to direct users to `rocm --help` and suggest including an action word in their request.

**Testing and reliability:**

* Added new tests to verify that the unknown command hinting logic works as expected, including suggestions for close typos and correct handling of natural language requests.

### DIFF
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -337,9 +337,24 @@ rocm logs --search error timeout")]
         #[arg(value_name = "QUERY")]
         query: Vec<String>,
     },
-    /// Start the background helper in the foreground.
+    /// Run the background supervisor that powers automations, managed servers, and the dashboard.
+    ///
+    /// The daemon is a multi-role helper that, while running:
+    ///   - executes enabled automation watchers (update checks, driver-plan
+    ///     checks, artifact prefetch) on a 5s tick in a sandboxed subprocess
+    ///   - health-checks and auto-recovers managed local model servers
+    ///     (vLLM, SGLang, Lemonade, llama.cpp)
+    ///   - collects GPU thermal/VRAM metrics every 60s for the TUI dashboard
+    ///   - listens on a local webhook port for automation events from other
+    ///     `rocm` commands
+    ///
+    /// It is normally started on demand by `rocm automations enable` and
+    /// `rocm managed serve`, so you rarely need to launch it yourself. Run it
+    /// directly only when you want to observe its behavior in the foreground,
+    /// e.g. for debugging.
+    #[command(verbatim_doc_comment)]
     Daemon {
-        /// Print the automation status panel instead of running the helper loop.
+        /// Print a one-shot snapshot of automation/watcher status, then exit without starting the supervisor loop.
         #[arg(long)]
         status: bool,
     },
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -349,7 +349,7 @@ rocm logs --search error timeout")]
     ///     `rocm` commands
     ///
     /// It is normally started on demand by `rocm automations enable` and
-    /// `rocm managed serve`, so you rarely need to launch it yourself. Run it
+    /// `rocm serve --managed`, so you rarely need to launch it yourself. Run it
     /// directly only when you want to observe its behavior in the foreground,
     /// e.g. for debugging.
     #[command(verbatim_doc_comment)]
@@ -863,37 +863,24 @@ fn run_freeform(request: String, approve: bool) -> Result<()> {
 /// Returns clap's own error when a freeform request looks like a botched
 /// command invocation rather than natural language, so the caller can surface
 /// clap's "did you mean"/usage message and exit instead of feeding it to the
-/// planner. Triggers when the first token is a command-like word and either it
-/// stands alone with a near match, or it is followed by flag-style arguments
-/// (for example `doctorgdfg --help`). Returns `None` for genuine prose.
+/// planner. clap's error is surfaced when it carries a subcommand suggestion (a
+/// near-miss typo such as `instal` or `automatios list`) or when the request
+/// includes flag-style arguments (such as `doctorgdfg --help`). Genuine prose,
+/// which yields neither, is left to the planner.
 fn command_invocation_error(request_args: &[String]) -> Option<clap::Error> {
-    let first = request_args.first()?;
-    let looks_like_command_token = !first.is_empty()
-        && first
-            .chars()
-            .all(|c| c.is_ascii_alphanumeric() || c == '-' || c == '_');
-    if !looks_like_command_token {
-        return None;
-    }
-    // A multi-word request with no flags reads as natural language; leave it to
-    // the planner.
-    let has_flag = request_args.iter().any(|arg| arg.starts_with('-'));
-    if request_args.len() > 1 && !has_flag {
+    if request_args.is_empty() {
         return None;
     }
     let argv = std::iter::once("rocm".to_owned()).chain(request_args.iter().cloned());
-    match Cli::try_parse_from(argv) {
-        Err(err) if err.kind() == clap::error::ErrorKind::InvalidSubcommand => {
-            // A lone unknown word with no near match may genuinely be a
-            // natural-language ask, so leave that case to the planner.
-            if request_args.len() == 1 && !err.to_string().contains("similar") {
-                None
-            } else {
-                Some(err)
-            }
-        }
-        _ => None,
+    let err = Cli::try_parse_from(argv).err()?;
+    if err.kind() != clap::error::ErrorKind::InvalidSubcommand {
+        return None;
     }
+    let has_suggestion = err
+        .get(clap::error::ContextKind::SuggestedSubcommand)
+        .is_some();
+    let has_flag = request_args.iter().any(|arg| arg.starts_with('-'));
+    (has_suggestion || has_flag).then_some(err)
 }
 
 fn setup(command: Option<SetupCommand>) -> Result<()> {
@@ -15550,6 +15537,16 @@ mod tests {
             .expect("a command-like token followed by a flag should yield a clap error");
         assert_eq!(err.kind(), clap::error::ErrorKind::InvalidSubcommand);
         assert!(err.to_string().contains("doctorgdfg"));
+    }
+
+    #[test]
+    fn mistyped_command_with_trailing_argument_yields_suggestion() {
+        // A near-miss subcommand followed by a normal (non-flag) argument should
+        // still surface clap's suggestion instead of falling to the planner.
+        let err = command_invocation_error(&["automatios".to_owned(), "list".to_owned()])
+            .expect("a near-miss subcommand with a trailing arg should yield a clap suggestion");
+        assert_eq!(err.kind(), clap::error::ErrorKind::InvalidSubcommand);
+        assert!(err.to_string().contains("automations"));
     }
 
     #[test]
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -62,7 +62,23 @@ use std::time::Duration;
 static BUILTIN_ENGINE_ENV_LOCK: OnceLock<Mutex<()>> = OnceLock::new();
 
 #[derive(Parser, Debug)]
-#[command(name = "rocm", about = "ROCm AI Command Center CLI", version)]
+#[command(
+    name = "rocm",
+    about = "ROCm AI Command Center CLI",
+    long_about = "ROCm AI Command Center CLI: install ROCm, manage local inference engines, \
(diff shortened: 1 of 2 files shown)


apps/rocm/src/main.rs
```diff
@@ -56,7 +56,23 @@ use std::time::Duration;
...
+#[command(
+    name = "rocm",
+    about = "ROCm AI Command Center CLI",
+    long_about = "ROCm AI Command Center CLI: install ROCm, manage local inference engines, \
+and run OpenAI-compatible model servers on AMD GPUs.\n\n\
+Run `rocm` with no subcommand to open the interactive dashboard (TUI). Use `rocm doctor` \
+to check that your GPU and ROCm install are ready, then `rocm serve <model>` to start a server.",
+    version,
+    after_help = "EXAMPLES:\n  \
+rocm doctor                       Check GPU, ROCm install, and engines\n  \
```
rominf: Here and everywhere else: the CLI doesn't have `doctor` command anymore, it was renamed split to `examine`, `diagnose`, and `fix`. Please update your repository copy.

apps/rocm/src/main.rs
```diff
@@ -15166,6 +14968,26 @@ mod tests {
...
+        let hint = unknown_command_hint("instal").expect("a close typo should produce a hint");
+        assert!(hint.contains("Did you mean 'install'?"));
+    }
+
+    #[test]
+    fn natural_language_request_has_no_unknown_command_hint() {
+        assert!(unknown_command_hint("please install rocm").is_none());
+        assert!(unknown_command_hint("is rocm installed?").is_none());
+    }
+
```
rominf: This is unnecessary, as `clap` has suggestions engine built-in, it can be enabled like this (note I added the patch for clap to actually handle mistyped commands instead of passing it to freeform planner; I believe it's non-obvious how to use it, and I would drop the freeform planner at all):
```
diff --git a/Cargo.toml b/Cargo.toml
index 6a06b46..80ad082 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ rust-version = "1.96"
 anyhow = "1.0"
 async-stream = "0.3"
 axum = "0.8"
-clap = { version = "4.5", features = ["derive", "string"] }
+clap = { version = "4.6", features = ["derive", "string", "suggestions"] }
 clap_complete = "4.6"
 crossterm = { version = "0.29", features = ["libc", "use-dev-tty"] }
 directories = "6.0"
diff --git a/apps/rocm/src/main.rs b/apps/rocm/src/main.rs
index ca9c117..fe1a90d 100644
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -681,6 +681,21 @@ fn main() -> Result<()> {

     let freeform_invocation = parse_freeform_invocation(&raw_args);
     if should_treat_as_freeform(&freeform_invocation) {
+        // Single-word unknown inputs are almost certainly a typo of a known
+        // subcommand rather than natural language. Let clap surface its
+        // "did you mean" suggestion when it finds a near match.
+        if freeform_invocation.request_args.len() == 1 {
+            let argv: Vec<String> = std::iter::once("rocm".to_owned())
+                .chain(raw_args.iter().cloned())
+                .collect();
+            if let Err(e) = Cli::try_parse_from(&argv) {
+                if e.kind() == clap::error::ErrorKind::InvalidSubcommand
+                    && e.to_string().contains("similar")
+                {
+                    e.exit();
+                }
+            }
+        }
```

rominf (changes requested):
I haven't yet check all the commands in the examples. I recommend you verify they all operational.

apps/rocm/src/main.rs
r0x0r: Good catch. I rebased this branch onto current `main`, which already replaces `doctor` with `examine`/`diagnose`/`fix`, and updated the help text (top-level `long_about`/`after_help`) to reference `rocm examine` instead of `rocm doctor`. I also re-checked every example command against its actual subcommand/flags. Done in fdbf447.

apps/rocm/src/main.rs
r0x0r: Switched to clap's built-in suggestions engine as suggested: enabled the `suggestions` feature and removed the hand-rolled Levenshtein matcher. A mistyped subcommand now surfaces clap's "did you mean"/usage error and exits — both when it stands alone (`instal` → `tip: ... 'install'`) and when it's followed by flags (`doctorgdfg --help`) — while genuine natural-language requests still flow to the planner. Done in fdbf447.

r0x0r:
@rominf Verified using an agent that all commands in the example are operational

rominf (commented):
### Review: typo-routing has a gap, and one help string names a non-existent command

CI is green and the leak scan is clean. Two concrete issues, plus a refactor that fixes the bigger one and removes a fragility in the same move.

#### 1. Mistyped subcommand + argument bypasses clap suggestions (logic bug)

`rocm automatios list` falls through to the natural-language planner instead of surfacing clap's "did you mean `automations`":

```
$ rocm automatios list
request plan
  ...
  note: No ROCm action matched this request.
  note: Run `rocm --help` to see available commands, or rephrase ...
```

Root cause in `command_invocation_error` (`apps/rocm/src/main.rs:855`):

```rust
let has_flag = request_args.iter().any(|arg| arg.starts_with('-'));
if request_args.len() > 1 && !has_flag {
    return None;   // <- "automatios list" exits here, never reaching clap
}
```

A near-miss subcommand followed by a normal argument (no flag) is treated as prose. It should still get a suggestion. `instal` works only because it is a single token; `instal list` would fail the same way.

#### 2. `Daemon` help references a command that does not exist

`apps/rocm/src/main.rs:347` says the daemon is "normally started on demand by `rocm automations enable` and `rocm managed serve`". There is no `rocm managed serve` command — the managed path is the `--managed` flag on `serve` (`serve --managed`, cf. `main.rs:3760` and `main.rs:10810`). Should read `rocm serve --managed`.

#### 3. Suggested refactor (fixes #1 and drops a fragility)

The lone-token branch keys off `err.to_string().contains("similar")` (`main.rs:864`), matching clap's human-readable wording. clap exposes this structurally via `err.get(ContextKind::SuggestedSubcommand)`. Switching to it lets the helper collapse to a single rule: surface clap's error when it carries a suggestion, **or** when a flag is present (e.g. `doctorgdfg --help`); otherwise fall through to the planner. That removes both the `len() > 1 && !has_flag` early-return (cause of #1) and the string match, and keeps the existing tests green — `instal` still suggests, prose like `please install` still routes to the planner.

rominf:
One more, on commit hygiene: both commits on this branch are **unsigned** (`git log --format=%G?` → `N`), whereas every commit on `main` is signed (`E`). If signed commits are expected here, this needs the two commits re-signed, which means a history rewrite + `--force-with-lease` rather than a follow-up commit. Not CI-enforced now, I will create a PR for this soon.

r0x0r:
Thanks for the detailed review — all three addressed in 6fac2ab.

**1 & 3 (typo-routing gap + refactor).** Adopted the suggested refactor. `command_invocation_error` now surfaces clap's error whenever clap carries a subcommand suggestion (read structurally via `err.get(ContextKind::SuggestedSubcommand)`) **or** the request contains a flag; otherwise it falls through to the planner. This drops both the `len() > 1 && !has_flag` early-return and the fragile `to_string().contains("similar")` match. `rocm automatios list` now suggests `automations`, and `instal list` suggests `install`:

```
$ rocm automatios list
error: unrecognized subcommand 'automatios'

  tip: some similar subcommands exist: 'status', 'automations'
```

Prose (`please install a model`) and lone non-command words still route to the planner. Added regression tests for the trailing-arg case.

**2 (daemon help).** Fixed `rocm managed serve` → `rocm serve --managed`.

`cargo clippy --workspace` is clean and `cargo test -p rocm` is green (764 tests).

**On commit signing:** good flag — happy to align if signed commits are expected here. As you noted it needs a history rewrite to re-sign the existing commits rather than a follow-up, so I'll leave the existing history in place pending the separate PR you mentioned; let me know if you'd prefer I rewrite + `--force-with-lease` instead.

michaelroy-amd:
**Review — approve with nits.** Help/doc-only + clap 4.5→4.6 (`suggestions`); low risk and a real UX improvement.

- **[MEDIUM]** `--engine` and `--device` are `Option<String>` (free-form), but the help now hand-writes `[possible values: lemonade, pytorch, llama.cpp, vllm, sglang]` and `[possible values: gpu_required, gpu_preferred, cpu_only]`. That text will drift from the actual accepted values over time. Either make them `ValueEnum` (clap then renders + validates the list for free) or add a test asserting the help list matches the real accept-set.
- **[LOW]** Branch is **BEHIND main** — rebase needed; currently CHANGES_REQUESTED.

The per-variant and `ValueEnum` doc comments and the `EXAMPLES` after-help are a nice touch.

r0x0r:
Thanks @michaelroy-amd — both addressed in 21f8873.

**[MEDIUM] `--engine`/`--device` possible-values drift.** Went with the test option rather than `ValueEnum`, because making these enums would change accepted input:

- `--engine` is intentionally free-form: `select_serve_engine` passes any value through (recipe/config can also supply engines), so a `ValueEnum` would start rejecting inputs that work today.
- `--device` accepts aliases beyond the displayed canonical names — `parse_device_policy` maps `auto`/`gpu` → `gpu_required`, and `cpu`/`cpu_only` is deliberately rejected with the "CPU is not a fallback" message (strict-GPU policy). A `ValueEnum` would drop those aliases and that targeted error.

So instead there are now two sync tests that render the actual `--help` and assert the advertised lists match their source of truth:
- `serve_engine_help_lists_match_engine_inventory` → `builtin_engine_inventory()`
- `serve_device_help_lists_match_device_policy_names` → the `DevicePolicy` variant names

Writing the test surfaced an existing drift: the engine list was missing `atom` (present in the inventory), now added — so `serve --help` lists `lemonade, pytorch, llama.cpp, vllm, sglang, atom`. These tests fail if either list drifts again.

**[LOW] Branch behind main.** Rebased onto current `main`; PR is now `MERGEABLE`. Commits are SSH-signed and DCO-signed.

`cargo clippy --workspace` clean; `cargo test -p rocm` green (788 tests).

r0x0r:
@rominf follow-up on the signing note — I went ahead with the rewrite rather than waiting for a separate PR.

The branch history is now rewritten: all four commits are SSH-signed and carry DCO `Signed-off-by` trailers (`git log --format='%G?'` → `G` on each), and the branch is rebased onto current `main`.

Full-transparency caveat: GitHub still shows them **Unverified** (`reason: unknown_key`) right now, because I signed with a locally-generated SSH key and that public key isn't registered to the account yet. They will flip to **Verified** as soon as the key is added under the account's *SSH signing keys*. If you'd rather these be signed by a specific canonical key/identity as the trust anchor, tell me which one and I'll re-sign + `--force-with-lease` to that instead — I didn't want to assume an ad-hoc key should be the trust root.

rominf (approved):

PR #50, opened 2026-06-24
Skip duplicate managed service launch when already alive

volen-silo:
## Problem

`rocm serve <model> --managed` spawns a managed-service process unconditionally. Running it a second time for the same engine+model — e.g. the chat assistant re-issuing `rocm serve <path> --engine <e> --managed` via the `rocm_command` tool, or re-confirming the serve wizard — spawns a **second** process for the same model/port. Both run simultaneously, causing port conflicts and a doubled entry in the dashboard.

The TUI job-bridge has an idempotency guard, but it only covers TUI job state — not the OS process. Once the first job completes (the managed daemon detaches quickly), the guard clears and a second launch succeeds.

## Fix

`start_managed_service` is the single choke point for every managed launch. Added an idempotency guard there:

- `existing_live_managed_service(paths, service_id)` reuses the existing `load_managed_service` (which refreshes liveness) + `managed_service_is_live`. Since `service_id = generate_service_id(engine, canonical_model_id)`, the same engine+model always collides and is detectable.
- If a live service is found, surface it (endpoint, status), log a `managed_service_launch_skipped` audit event, and return `Ok(())` — **no second process spawned**. Treated as idempotent success so the chat assistant sees the repeat request as satisfied, not failed.
- Stale/dead manifests refresh to `stopped` and yield `None`, so a genuine relaunch still proceeds.

The file-based manifest persists across the job-bridge race, so this is robust where the TUI guard wasn't.

## Tests

- `duplicate_managed_launch_is_detected_when_live` — live PID ⇒ detected.
- `dead_managed_service_allows_relaunch` — dead PID ⇒ relaunch allowed.
- `missing_manifest_allows_launch` — no manifest ⇒ launch proceeds.

`cargo test -p rocm` (managed suite) and `cargo clippy -p rocm --all-targets` both green.

## Known limitation

Out of scope: two truly-concurrent launches within ~200ms before either writes a live manifest. The realistic "asked twice" sequence is fully serialized, so the file guard catches it.

### DIFF
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -11972,11 +11972,14 @@ fn existing_live_managed_service(
     engine: &str,
     canonical_model_id: &str,
 ) -> Option<ManagedServiceRecord> {
-    load_managed_services(paths).ok()?.into_iter().find(|record| {
-        record.engine == engine
-            && record.canonical_model_id == canonical_model_id
-            && managed_service_is_live(record)
-    })
+    load_managed_services(paths)
+        .ok()?
+        .into_iter()
+        .find(|record| {
+            record.engine == engine
+                && record.canonical_model_id == canonical_model_id
+                && managed_service_is_live(record)
+        })
 }
 
 fn managed_service_running_state(status: &str) -> &'static str {
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -3714,10 +3714,14 @@ fn start_managed_service(
     // Idempotency guard: if a managed service for this engine+model is already
     // alive, surface it and spawn nothing. A second `serve --managed` (e.g. the
     // chat assistant re-issuing the same request) is treated as satisfied, not
-    // an error. Stale/dead services fall through and relaunch normally.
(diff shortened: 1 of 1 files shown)


volen-silo:
Good catch from Copilot — this was a real defect. `generate_service_id` appends `unix_time_millis()`, so every launch produces a unique `service_id`; the original guard keyed on it would never match an existing service, and the duplicate would still spawn. The unit tests only passed because they reused a hardcoded id for both write and lookup.

Fixed in 05e7180:
- `existing_live_managed_service` now keys on `(engine, canonical_model_id)` and scans `load_managed_services` (which refreshes liveness and returns records newest-first), skipping on the first live match.
- The skip path now reports/audits the **existing** service_id, not the freshly generated one.
- Tests reworked: `duplicate_managed_launch_detected_across_distinct_service_ids` writes an old dead + new live manifest with **different** service_ids but the same engine+model and asserts the newest live one is detected; added `live_service_for_other_model_does_not_block` to guard against over-matching.

Note: the License-header CI failure is unrelated and pre-existing on `main` (6 files lack the SPDX header) — fixed separately in #52.

r0x0r (approved):

PR #52, opened 2026-06-24
Add missing Apache-2.0 license headers

volen-silo:
## Problem

The hawkeye **License header check** is red across the repo. Six source files already on `main` are missing the SPDX header that `licenserc.toml` mandates, so the check fails on `main` and on every open PR (hawkeye scans the whole tree, not just the diff).

Affected files:
- `apps/rocm/src/dash_seam.rs`
- `apps/rocm/tests/daemon_run.rs`
- `crates/rocm-core/src/diagnose.rs`
- `crates/rocm-core/src/examine.rs`
- `crates/rocm-core/src/fix.rs`
- `crates/rocm-dash-tui/src/tool_exec.rs`

## Fix

Prepend the standard header (per `licenserc.toml`) to each file:

```rust
// Copyright Advanced Micro Devices, Inc.
//
// SPDX-License-Identifier: Apache-2.0
```

Header-only change — no code touched. After merge, rebasing open PRs clears the license check for them too.

### DIFF
--- a/apps/rocm/src/dash_seam.rs
+++ b/apps/rocm/src/dash_seam.rs
@@ -1,3 +1,7 @@
+// Copyright Advanced Micro Devices, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
 //! Bin-side implementation of the dash execution seam.
 //!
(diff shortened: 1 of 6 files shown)


volen-silo:
#51

PR #54, opened 2026-06-24
Enforce signed and signed-off commits via hooks and CI

rominf:
## What

Adds blocking enforcement that every commit carries a cryptographic signature and a DCO `Signed-off-by` trailer, at two layers:

- **Local (prek):** a pre-commit hook checks signing is configured; a pre-push hook verifies outgoing commits are signed and signed-off.
- **CI:** a new `commit-signatures` job fails a pull request if any of its commits is not GitHub-"Verified" or is missing a sign-off.

The check is a `cargo xtask verify-commits` subcommand (`--base`, `--require-verified`, `--check-config`), so identical logic runs locally and in CI.

## Why

The contributing guidelines already require signed, signed-off commits, but nothing enforced it. This makes the requirement self-checking and keeps unsigned commits out of the history.

## Notes

- `docs/commit-signatures.md` documents local signing setup and the complementary native GitHub "Require signed commits" ruleset — and why a repo-side check is still needed: rulesets don't apply to pull requests from forks and can't enforce DCO sign-off.
- The CI job becomes blocking once it is marked a required check in branch protection / rulesets.

## Test plan

- `cargo build`, `test`, `clippy -D warnings`, `fmt --check` all clean; 16 unit tests cover the decision logic (signed / unsigned / bad-sig / missing sign-off / strict-unverified / empty range).
- Dogfoods itself: this PR's own commit is signed + signed-off and passes `cargo xtask verify-commits`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -228,25 +228,37 @@ jobs:
   commit-signatures:
     name: Commit signatures + sign-off
     runs-on: ubuntu-latest
-    # Only meaningful for PRs, where a base branch defines the commit range.
-    if: github.event_name == 'pull_request'
+    # Runs on pull requests and in the merge queue. Both define a base that
+    # bounds the commit range. Handling `merge_group` matters once this is a
+    # *required* check: the queue fires `merge_group` (not `pull_request`), so a
+    # job gated to PRs only would never produce the required check and the queue
+    # entry would stall forever.
+    if: github.event_name == 'pull_request' || github.event_name == 'merge_group'
     steps:
       - uses: actions/checkout@v6
         with:
           fetch-depth: 0
-          # Check out the PR head commit, not the synthetic refs/pull/N/merge
-          # commit (which is unsigned and lacks a sign-off, and would otherwise
-          # appear in origin/<base>..HEAD and fail the gate on every PR).
-          ref: ${{ github.event.pull_request.head.sha }}
+          # PR: check out the PR head, not the synthetic refs/pull/N/merge commit
+          # (which is unsigned and lacks a sign-off, and would otherwise appear
+          # in origin/<base>..HEAD and fail the gate on every PR).
+          # merge_group: check out the queue head — the rebased/squashed commits
+          # GitHub built for the group. Merge commits are disabled on this repo,
+          # so these are normal commits that preserve their Signed-off-by trailer
+          # and are GitHub "Verified" (web-flow signed).
+          ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.event.merge_group.head_sha }}
 
       - uses: actions-rust-lang/setup-rust-toolchain@v1
 
       - name: Verify commits are signed and signed-off
         # `--require-verified` shells out to the `gh` CLI; pass the job token so
         # `gh api` is authenticated (avoids the unauthenticated rate limit).
+        # Base: the PR's base branch, or — in the queue — the exact commit the
+        # merge group was built on (`base_sha`), so the range is just the queued
+        # commits and not whatever else has since landed on the base branch.
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: cargo xtask verify-commits --base origin/${{ github.base_ref }} --require-verified
+          BASE: ${{ github.event_name == 'pull_request' && format('origin/{0}', github.base_ref) || github.event.merge_group.base_sha }}
+        run: cargo xtask verify-commits --base "$BASE" --require-verified
 
   # Hardware (real AMD GPU / WSL) smoke tests on dedicated self-hosted runners
   # are intentionally not part of this workflow yet. See
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -225,6 +225,29 @@ jobs:
       - name: Check license headers
         uses: korandoru/hawkeye@v6
 
+  commit-signatures:
+    name: Commit signatures + sign-off
+    runs-on: ubuntu-latest
+    # Only meaningful for PRs, where a base branch defines the commit range.
+    if: github.event_name == 'pull_request'
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+          # Check out the PR head commit, not the synthetic refs/pull/N/merge
+          # commit (which is unsigned and lacks a sign-off, and would otherwise
+          # appear in origin/<base>..HEAD and fail the gate on every PR).
+          ref: ${{ github.event.pull_request.head.sha }}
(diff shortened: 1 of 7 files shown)


r0x0r (approved):

michaelroy-amd:
**Review — LGTM** (already approved; light pass). Thorough, well-documented enforcement of signed + signed-off commits via xtask + prek hooks + a blocking CI gate.

- ✅ The `commit-signatures` job correctly checks out `pull_request.head.sha` (not the synthetic `refs/pull/N/merge` commit, which is unsigned and would false-fail every PR), and passes `GH_TOKEN` so `--require-verified`'s `gh api` calls aren't rate-limited.
- ✅ pre-commit does the cheap config assertion; the full-range verification is deferred to pre-push + mirrored in CI — good split.
- New `docs/commit-signatures.md` is `.md`, so no SPDX header needed (hawkeye scans rs/py/sh only).
- [LOW] pre-push `verify-commits` compiles xtask — slightly slow per push, but acceptable.

No blocking issues.

michaelroy-amd (dismissed):
Formal review (requested). I read the substance, primarily `xtask/src/verify_commits.rs`. The design and test coverage are excellent and I'm happy with the logic — one licensing blocker has to be fixed first, hence request-changes rather than approve.

### Blocking
- **New file `xtask/src/verify_commits.rs:3` has `SPDX-License-Identifier: Apache-2.0`, but `main` relicensed to MIT in #53.** This branch predates #53 — its own `licenserc.toml` still says Apache-2.0, which is why the hawkeye job passes on the PR branch today. But #54 doesn't modify `licenserc.toml`, so when the merge queue rebases onto current `main` (MIT `licenserc`), hawkeye will fail on this file's header — and since #56 runs CI on `merge_group`, that blocks the queue merge. Fix: rebase onto current `main` and change the header on new file(s) to `SPDX-License-Identifier: MIT` (only `verify_commits.rs` is a hawkeye-scanned new file; `docs/commit-signatures.md` is fine).

### Non-blocking (LOW)
- `collect_commits` spawns two extra `git log -1` (`%s`, `%B`) per commit, and `verified_via_github` one `gh api` per commit — O(n) subprocess/API calls. Fine for normal PR sizes (GH_TOKEN avoids the rate limit); if it ever matters, a single `git log` with a record separator would collapse the per-commit git calls.

### What's good (so it's on record)
- `evaluate_commits` is pure and cleanly separated from I/O, with thorough unit coverage (empty range, local N/B-fail vs lenient G/U/E/X/Y/R-pass, strict GitHub-Verified incl. the `None`-verdict-must-fail case, multi-failure-per-commit, case/whitespace-insensitive sign-off, https/ssh slug parsing).
- Strict mode correctly treats a missing/`null` `verification.verified` as distinct from `false` (commit not pushed yet) instead of silently passing.
- The CI job checks out `head.sha` (not the synthetic unsigned merge ref) and passes `GH_TOKEN` — both right.

Once the header is MIT after the rebase, this is an approve from me.

volen-silo:
Reviewed the full change (all 7 files) and ran build / tests / `clippy -D warnings` / `fmt --check` locally — all green, 16 tests pass. Overall this is a clean, well-structured PR: a single pure `evaluate_commits` with git/`gh` I/O kept at the edges, good test coverage of the decision logic, and accurate docs. Nice touch that local mode trusts `%G?` (dev has their own key) while CI trusts GitHub "Verified" (runners don't) — that split is the right call.

A few things to consider, none blocking the logic itself:

**[MEDIUM] Merge-queue interaction once this is a required check** (`ci.yml`)
The job is gated `if: github.event_name == 'pull_request'`, so it never runs on `merge_group`. CI was deliberately wired to run on `merge_group` (#56 / commit 12aada2). If `commit-signatures` is marked *required* (as the docs suggest) while a merge queue is active, the `merge_group` run won't produce this check and the queue entry can stall waiting on a required check that never starts. Either also handle `merge_group` (with the queue base rather than a PR base), or document that it must not be a required check while the merge queue is in use. Worth settling before flipping it to required.

**[LOW] `check_config` only accepts the literal `true` for `commit.gpgsign`** (`verify_commits.rs:295-300`)
Git treats `1`/`yes`/`on` as truthy and will sign with them, but the check compares the lowercased string `== "true"`, so e.g. `commit.gpgsign = 1` yields a false "signing is not configured". Suggest `git config --type=bool --get commit.gpgsign` to normalize.

**[LOW] Pre-push hook hardcodes `origin/main` as base** (pre-push hook / `run()` default)
The hook ignores the refs git passes on stdin and always checks `origin/main..HEAD`. A stale/missing `origin/main`, or a branch targeting a different base, can give a wrong range or a confusing failure. Consider documenting the fetch-first expectation, or reading the pushed range from stdin.

**[NIT] Per-commit subprocess fan-out** (`verify_commits.rs:154-159`): `git log` x2 + `gh api` x1 per commit. Fine for typical PRs; batchable if large ranges ever matter.

**[NIT] `--require-verified` run locally hard-fails on commits not yet pushed** (`verified_via_github:222-225`) — intended for CI, just a sharp edge with a clear message.

Verified as fine (not issues): action versions are consistent with the rest of `ci.yml`; `default_install_hook_types: [pre-commit, pre-push]` is declared so the pre-push hook installs; CI checks out `head.sha` (not the merge ref) with `fetch-depth: 0` so the base/merge-base resolve and the unsigned merge commit isn't falsely flagged; `GH_TOKEN` is passed and works for fork PRs; clap conflict rules are symmetric; docs match the implemented flags.

rominf:
Thanks for the thorough pass @volen-silo. Addressed in aca1025:

- **[MEDIUM] merge-queue interaction** — the `commit-signatures` job now runs on `merge_group` as well, verifying the queued commits against `merge_group.base_sha` (the queue base) rather than a PR base. Merge commits are disabled on this repo, so the queued commits retain their `Signed-off-by` and are web-flow "Verified" — the strict check passes in the queue, and it can now be marked required without stalling it (also called out in `docs/commit-signatures.md`).
- **[LOW] `check_config` truthy values** — now uses `git config --type=bool --get commit.gpgsign`, so `1`/`yes`/`on` are accepted, not just the literal `true`.
- **[LOW] pre-push hardcoded base** — documented the fixed `origin/main` base (keep it fetched; pass `--base <ref>` for a branch targeting a different base). Reading the pushed range from stdin is a reasonable future improvement; kept the simpler hook for now.

The two NITs are intentional and left as-is: the per-commit `git log`/`gh api` fan-out is fine for typical PR sizes (`GH_TOKEN` avoids the rate limit) and is batchable if large ranges ever matter; and the local `--require-verified` hard-fail on unpushed commits is the CI-only mode, which fails with a clear message. Happy to revisit either.

volen-silo (approved):

PR #57, opened 2026-06-25
ci: cache pinned, checksum-verified hawkeye for license headers

rominf:
## What

Make the `license-headers` CI job install a pinned, checksum-verified hawkeye binary instead of pulling the Docker image on every run.

Previously the job used `korandoru/hawkeye@v6`, a Docker action that re-pulls the container image each run (effectively uncached). Now it:

- **Pins** hawkeye to `v6.5.1`.
- **Downloads** the prebuilt `x86_64-unknown-linux-gnu` binary directly from the GitHub release.
- **Verifies** it against a SHA256 recorded in the workflow before use.
- **Caches** the verified binary via `actions/cache@v5`, keyed on OS + version + hash; downloads only on a cache miss.
- Runs `hawkeye check` from the cached path.

## Why

Warm runs skip the network entirely; cold runs fetch a few-MB binary instead of a Docker image, and the pinned version makes the gate reproducible.

The checksum is recorded **in-repo**, out of band from the release. GitHub release assets are mutable, so verifying only against the checksum a release publishes alongside the artifact would still trust a republished or tampered binary (the attacker republishes a matching checksum too). Pinning the hash here makes that case fail instead.

## Notes

- No behavior change to the check itself — same `licenserc.toml`, same `hawkeye check`.
- To upgrade: bump the version and replace the hash with the new release's published `.sha256` (confirm the bytes first).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -220,10 +220,43 @@ jobs:
   license-headers:
     name: License header check (hawkeye)
     runs-on: ubuntu-latest
+    env:
+      # Install a pinned prebuilt hawkeye binary instead of pulling the
+      # korandoru/hawkeye Docker image on every run, and verify it against a
+      # sha256 recorded here. The hash lives in-repo, out of band from the
+      # release: GitHub release assets are mutable, so a republished/tampered
+      # artifact (and its co-published .sha256) would pass an upstream-only
+      # check — pinning the hash here makes that fail instead. To upgrade, bump
+      # the version and replace the hash with the new release's published
+      # .sha256 (confirm the bytes first).
+      HAWKEYE_VERSION: v6.5.1
+      HAWKEYE_SHA256: d6eb0505a45a15244f4f789158aafe5e3f1a7dc86c9dc1d7651f3cb1e1b321e0
+      HAWKEYE_INSTALL_DIR: ${{ github.workspace }}/.hawkeye-bin
     steps:
       - uses: actions/checkout@v6
+
+      - name: Cache hawkeye binary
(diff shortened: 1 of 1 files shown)


michaelroy-amd:
**Review — approve.** Clean perf win: install a pinned, cached hawkeye binary instead of pulling the `korandoru/hawkeye` Docker image on every run. `actions/cache` key tracks `HAWKEYE_VERSION` (v6.5.1) automatically; installer uses `--proto '=https' --tlsv1.2`.

- **[LOW]** The `curl … | sh` installer isn't checksum-verified — consider pinning a sha256 of the installer/binary for supply-chain hardening.
- **[LOW]** Touches the same `license-headers`/CI block as #54 and #55 — will conflict; coordinate merge order.

No blocking issues.

rominf:
Thanks for the review — both points addressed.

**[LOW] checksum verification** — implemented, and pinned out of band. The job now downloads the exact prebuilt tarball and verifies it against a SHA256 recorded in the workflow before use, rather than trusting a checksum fetched from the release at run time.

I evaluated `taiki-e/install-action` first (cleaner), but hawkeye isn't in its native manifest, so it falls back to cargo-binstall, which verifies against the `.sha256` hawkeye co-publishes in the same release. Since GitHub release assets are mutable, a republished/tampered artifact (with a matching republished checksum) would still pass that path. Recording the hash in-repo makes that case fail the build. Caching is kept; the hash is part of the cache key, so a version/hash bump invalidates it.

**[LOW] conflict with #54/#55** — noted. Both are still open; I'll rebase this onto whichever lands first.

juhovainio (approved):

PR #59, opened 2026-06-25
Sync Lemonade ROCm channel with installed ROCm

volen-silo:
## Motivation

The Lemonade engine always defaulted to its `stable` ROCm llama.cpp backend, regardless of whether the user installed ROCm from the **release** or **nightly** channel via `rocm install`. A user who deliberately picked the nightly ROCm channel still got Lemonade's stable ROCm build.

## What this does

Make Lemonade's ROCm backend **channel** follow the channel the user picked during `rocm install`:

- **rocm-core**: surface the active managed runtime's channel. `TheRockFamilyManifest` now deserializes the `channel` field (already written to the registry), and a new `active_managed_therock_channel(paths, config)` helper resolves it using the same active-key → most-recent selection as `active_managed_therock_environment` (extracted into a shared selector).
- **lemonade engine**: at engine install time, *before* backend selection, set Lemonade's `rocm_channel` to match the install (`release → stable`, `nightly → nightly`) via the bundled `lemonade config set` CLI. The applied channel is surfaced in the install warnings.

## Scope / decisions

- The backend binary is kept at Lemonade's tested default (`rocm_bin = builtin`). There is no reliable mapping from a ROCm SDK version (e.g. `6.3.0`) to a Lemonade llama.cpp build tag (e.g. `b1260`), and pinning a tag is the main source of hard `404` download failures. Mapping the channel while keeping the tested binary is the stable slice.
- Aligning the channel is **best-effort** — a failure warns and continues rather than aborting the engine install.
- No managed ROCm runtime (system/legacy ROCm) → Lemonade's own default is left untouched.

## Testing

- `cargo test -p rocm-core -p rocm-engine-lemonade` — passes, including 4 new unit tests (channel readback, channel mapping, `config set` arg builder).
- `cargo clippy` on both crates — clean.
- `cargo check --workspace` — clean.

### Not verified in CI (deferred to hardware testing)

- The exact `lemonade config set rocm_channel=<value>` argument syntax of the bundled embeddable CLI (taken from Lemonade docs; worth confirming with `lemonade config set --help`).
- End-to-end confirmation that a nightly ROCm install causes Lemonade to fetch its nightly ROCm backend (needs real ROCm + GPU + network).

## Follow-ups (out of scope)

- Re-sync an already-installed Lemonade when the user later switches ROCm channel (would require a backend reinstall).
- Make the Linux direct-serve fast path (hardcoded `rocm-stable` segment) channel-aware.

### DIFF
--- a/crates/rocm-core/src/lib.rs
+++ b/crates/rocm-core/src/lib.rs
@@ -2356,10 +2356,31 @@ pub fn active_managed_therock_environment(
     paths: &AppPaths,
     config: &RocmCliConfig,
 ) -> Result<Option<ManagedRuntimeEnvironment>> {
+    Ok(select_active_managed_therock_record(paths, config)
+        .map(|record| managed_therock_environment_from_record(&record)))
+}
+
+/// Channel (`"release"`/`"nightly"`) of the active managed TheRock runtime.
+///
+/// Reflects the channel recorded at install time. Returns `None` when there is no managed
+/// runtime (system or legacy ROCm) or the registry record predates channel recording.
+pub fn active_managed_therock_channel(
+    paths: &AppPaths,
+    config: &RocmCliConfig,
+) -> Result<Option<String>> {
+    Ok(select_active_managed_therock_record(paths, config).and_then(|record| record.channel))
+}
+
+/// Pick the active managed TheRock runtime record: the one matching
+/// `config.active_runtime_key`, falling back to the most recently installed.
+fn select_active_managed_therock_record(
+    paths: &AppPaths,
+    config: &RocmCliConfig,
+) -> Option<TheRockFamilyManifest> {
     let registry_dir = paths.data_dir.join("runtimes").join("registry");
     let mut records = managed_therock_environment_records(&registry_dir);
     if records.is_empty() {
-        return Ok(None);
+        return None;
     }
 
     if let Some(active_key) = config.active_runtime_key.as_deref()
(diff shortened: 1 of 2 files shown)


michaelroy-amd:
**Review — approve with nits.** Well-scoped. Surfaces the active managed TheRock channel via a new `active_managed_therock_channel()` (cleanly extracted into a shared `select_active_managed_therock_record` selector) and maps `release→stable` / `nightly→nightly`, applied via `lemonade config set` before backend install. Behavior-preserving refactor, defensive parsing, and "no managed runtime → leave defaults untouched" are all the right calls; no injection (args are hardcoded keys + validated enum mapping, passed as a Vec, no shell).

- **[MEDIUM]** Test gap: the new tests cover single-record readback and no-runtime→None, but not the two selector branches the channel helper now shares — active-key match and most-recent (`installed_at_unix_ms`) fallback. Add a multi-record test asserting the newer channel wins, and one asserting `active_runtime_key` overrides recency.
- **[LOW]** Two sequential `config set` calls aren't transactional (partial-apply on second failure) — best-effort semantics make this minor; a one-line comment would suffice. `AppPaths::discover()` is called twice in the install path — thread `&paths` through.
- Rebase onto post-#53 (MIT) main before merge (no new files, so no header violation today).

juhovainio (approved):
lgtm - let's track in a follow-up ticket the ability to pin the exact installed ROCm version that would also be used by Lemonade once we have alignment with the Lemonade team

PR #64, opened 2026-06-29
Keep the telemetry daemon alive when its socket lives in /tmp

rominf:
## Symptom

Running `rocm` (or `rocm dash`) could abort the auto-started telemetry daemon at startup with:

```
rocm: embedded telemetry daemon exited: restricting socket directory /tmp to mode 0700 — check that the directory is owned by the current user (EPERM means it is owned by another user or root): Operation not permitted (os error 1)
```

This happens whenever the daemon's Unix socket is configured to live directly under `/tmp` (for example from a config file written by an older build, or an explicit override).

## Root cause

When binding the Unix socket, the daemon creates the socket's parent directory and then unconditionally `chmod`s it to `0700` as defense-in-depth, treating a failure as fatal. `/tmp` is shared and root-owned (mode `1777`), so a normal user cannot `chmod` it and gets `EPERM` — taking the whole daemon down. The socket itself is already bound with mode `0600`, which is the actual access control, so the parent-directory tightening should not be fatal here.

## Fix

Two layers:

1. **Don't crash on a directory we don't own.** The parent-directory hardening is extracted into a helper that only treats a failed `chmod` as fatal when it *created* the directory (which it therefore owns). For a pre-existing directory it does not own (such as `/tmp`), it logs a warning and continues — the `0600` socket still restricts access to the owning user.

2. **Pick a safer default.** The default socket location no longer sits directly in `/tmp`; it uses a per-user subdirectory (`/tmp/rocm-<user>/rocmdashd.sock`) that the daemon creates and owns, so it can be secured without touching `/tmp` and without colliding with other users on shared hosts.

## Tests

Adds regression tests that fail before and pass after: the `/tmp`-parented socket no longer aborts the daemon (plus coverage for the create/tighten/empty-parent paths), and the default socket is asserted to use a per-user subdirectory rather than the bare temp dir.

### DIFF
--- a/crates/rocm-core/src/lib.rs
+++ b/crates/rocm-core/src/lib.rs
@@ -3836,7 +3836,7 @@ fn default_dashboard_socket() -> String {
     //
     // 1. $XDG_RUNTIME_DIR  — already mode 0700 on systemd systems, ideal.
     // 2. $HOME/.rocm/data/telemetry — standard per-user data dir.
-    // 3. temp_dir()/rocmdashd-<user> — user-named subdir so the parent is
+    // 3. temp_dir()/rocm-<user> — user-named subdir so the parent is
     //    something we create and own, not /tmp itself.
     use std::path::PathBuf;
     let path = std::env::var_os("XDG_RUNTIME_DIR")
@@ -3874,7 +3874,7 @@ fn default_dashboard_socket() -> String {
                 user
             };
             std::env::temp_dir()
-                .join(format!("rocmdashd-{user}"))
+                .join(format!("rocm-{user}"))
                 .join("rocmdashd.sock")
         });
     format!("unix:{}", path.display())
--- a/crates/rocm-dash-core/src/config.rs
+++ b/crates/rocm-dash-core/src/config.rs
@@ -104,36 +104,72 @@ pub struct DaemonConfig {
 
 /// Default Unix-socket address for the telemetry daemon.
 ///
-/// The socket lives in a *per-user* subdirectory of the temp dir
-/// (`/tmp/rocm-<user>/`) rather than directly in `/tmp`. The daemon creates and
-/// owns that subdirectory, so it can tighten it to mode `0o700` without the
-/// `EPERM` that results from trying to `chmod` a shared, root-owned `/tmp`
-/// (mode `1777`). Keeping it per-user also avoids collisions on multi-user hosts.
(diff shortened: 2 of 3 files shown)


michaelroy-amd (approved):
Reviewed and tested locally — this resolves the embedded-telemetry-daemon EPERM crash. Approving.

## What works well

- **`prepare_socket_dir` is the right fix.** Treating a failed `chmod` as fatal *only* when we created the directory, and downgrading to a warning for a pre-existing dir we don't own (`/tmp`, mode 1777), is correct: the `0o600` socket is the real access control, so the parent-dir tightening is genuinely defense-in-depth here. Good call extracting it into its own function.
- **Test coverage is thorough** — the create / tighten-owned / empty-parent / unowned-parent paths are all covered, and the unowned-parent regression test carefully skips when run as root or as the temp-dir owner so it never mutates a shared `/tmp`. Nice.
- Repro confirmed against my own machine: a `~/.rocm/config.json` migrated from an older rocm-dash TOML had `unix:/tmp/rocmdashd.sock` baked in, which is exactly the crash this fixes.

## One follow-up I pushed as a commit

I added c311877 — please review. The new `default_socket()` in `rocm-dash-core` only used `temp_dir()/rocm-<user>`, dropping the `$XDG_RUNTIME_DIR` and `$HOME/.rocm/data/telemetry` tiers that `rocm-core::default_dashboard_socket` already prefers. On systemd hosts the two configs then disagreed (canonical `rocm` → `/run/user/<uid>/…`, standalone rocm-dash → `/tmp/rocm-<user>/…`, the worse location even when the ideal 0700 runtime dir exists).

The commit:
- mirrors the full 3-tier precedence (XDG_RUNTIME_DIR → HOME → temp), and
- factors it into a pure `socket_path(...)` helper so precedence is tested deterministically without `set_var` (unsafe + racy under parallel tests in edition 2024).

Heads up: your `default_socket_uses_per_user_subdir_not_bare_tmp` test asserted the parent starts with `rocm-`, which would now fail whenever `$XDG_RUNTIME_DIR` is set (CI included) since the parent becomes `/run/user/<uid>`. I replaced it with an env-independent invariant test plus three `socket_path` precedence tests. All lib tests + clippy pass.

@rominf could you take a look at c311877 and confirm you're happy with it? Happy to adjust if you'd rather keep dash-core's default minimal.

PR #62, opened 2026-06-27
feat(dash): IA redesign — 5-tab shell, inline Details, live TTFT/TPOT, Observe

michaelroy-amd:
## Summary

Redesigns the `rocm-dash-tui` dashboard information architecture and serving telemetry. This is the dash IA-redesign work that sat on top of the §1+§2 supergoal stack (PRs #31–#38); it has been rebased clean onto `main` so the diff contains **only** the redesign.

### Highlights
- **5-tab IA** — `Home / ROCm / Serving / Observe / Chat`, replacing the legacy `Overview/Hardware/Instances/Bench/Chat` shell, with a `PaneFocus` model and `action.rs` retired.
- **Inline → re-modal manager evolution** — managers first rendered inline in the ROCm/Serving Details pane (`de-modal`), then returned to overlays (`re-modal`) with real log scrolling.
- **Live serving telemetry** — TTFT/TPOT sourced from vLLM histograms; Observe redesigned around AI-serving metrics (node throughput hero, activity feed).
- **btop-style chrome** — outlined tabs, rounded bento boxes with per-box color, clickable legend, wide triptych shell with a LOGS/CONTEXT right dock, and a minimal launcher front door for bare `rocm`.
- **Esc menu / command palette / tabbed Options / global Help.**
- Design mockups + responsive/surface docs under `docs/design/`.

### Rebase reconciliation
The branch predated `main`'s `doctor_manager`→`examine_manager` rename (PR #31) and `main`'s hawkeye license-header enforcement. A single `fix(dash): reconcile…` commit adapts the merged tree: renames retired symbols (`OpenDoctor`/`doctor_manager`/`DoctorManagerState`) to the `examine_*` API, adds MIT SPDX headers to the new files, fixes a few fmt/clippy nits, and normalizes mockup file endings.

## Test plan
- [x] `cargo build --workspace --all-targets`
- [x] `cargo fmt --check`
- [x] `cargo clippy --locked --workspace --all-targets -- -D warnings`
- [x] `hawkeye check`
- [x] `rocm-dash-tui` suite — 545 tests (lib + characterization + job-bridge)
- [x] All commits signed + signed-off (`cargo xtask verify-commits`)
- [x] CI commit-signatures (`--require-verified`) — runs post-push against GitHub
- Note: `tui::tests::running_models_hides_stopped_services_from_main_list` fails locally only when a server occupies `127.0.0.1:11435`; it is an env-dependent legacy-`tui.rs` test (passes on `main` with the port free).

PR #67, opened 2026-06-29
Retire legacy tui.rs: route `rocm bootstrap setup` to the dash onboarding, then delete the ~45k-line module

michaelroy-amd:
## Summary

Retires the legacy ~45k-line monolith TUI (`apps/rocm/src/tui.rs`), which until
now was kept compiling only via the `_RETAINED_TUI_ENTRY` dead-code anchor. The
one live consumer — `rocm bootstrap setup` (via `run_bootstrap_setup`) — is
re-routed to the modern, App-free onboarding wizard that already lives in the
dash crate (`rocm-dash-tui` `ui::onboarding` over `jobs`/`job_console`).

Follows up @rominf's review on #62 ("legacy code isn't fully retired"), split
into its own change as requested so #62 didn't carry the giant delete.

## Commit 1 — route `rocm bootstrap setup` to the dash onboarding
- `ResolvedArgs::start_onboarding`; honored at app startup by opening the
  onboarding overlay.
- `dash::run_bootstrap` (+ testable `bootstrap_args` helper) reusing the
  `run`/`run_chat` runtime path; `bootstrap.rs` calls it.

## Commit 2 — delete tui.rs + prune the dead-code cascade
- Delete `apps/rocm/src/tui.rs` (incl. its test module); drop `mod tui;`, the
  retention anchor, and its guard test.
- Compiler/clippy-driven prune of helpers only tui.rs used (driver-install
  TUI-approval cluster, `render_sidebar_text`, `friendly_provider_label`,
  `freeform_plan_uses_provider`, the `*_for_tui` log renderers,
  `logs_browser_page_count`, `tui_help_text`); `build_freeform_plan` is now
  `#[cfg(test)]` over the live `build_freeform_plan_with_recipes`.
- `comfyui.rs`: drop `render_tui_status`/`render_tui_logs`/`is_installed`/
  `running_url` (+ their tests); live `render_logs`/`models_folder`/
  `render_models_path` and the unmanaged-URL helpers stay.
- `docs/testing.md`: remove the obsolete legacy-TUI navigability test list.

## Commit 3 — correct the non-interactive setup guidance
- `bootstrap.rs`: the non-interactive fallback previously told users to run bare
  `rocm` "to choose an install folder". Post-reroute that's wrong on both counts
  (bare `rocm` opens the dash chat; the wizard has no folder picker). Now points
  at `rocm bootstrap setup` and drops the folder-picker wording. (Addresses
  @rominf's review.)

## Commit 4 — chat `/v1` 404 + Esc closes a running serve modal (#68)
- #68 was stacked on this branch and merged into it, so its fix rides to `main`
  with this PR: `DEFAULT_CHAT_BASE_URL` gains the required `/v1` suffix (Rig
  appends `chat/completions` to `base_url`), and `Esc` now closes a job console
  while the job keeps running. Includes the review follow-ups from #68 (a
  revert-proof `/v1` regression test + softened doc wording).

## UX note (decision + rationale)
Bootstrap adopts the dash onboarding UX (install release wheels to the default
location + adopt-existing-folder).

- **Release/Nightly channel toggle:** *not* dropped — it was restored in the
  onboarding wizard by #73 (now on `main`), so `bootstrap setup` retains channel
  choice.
- **Install-folder picker (`--prefix`):** deferred, not removed by design. The
  dash onboarding wizard doesn't yet have an install-location step, so the guided
  flow installs to the default location only. The `--prefix` plumbing still exists
  (`install_manager.rs`, and the chat tool path), so restoring it is a
  wizard-UI addition rather than a re-plumb. Tracked in #75.

## Test plan
- [x] `cargo fmt --all --check`
- [x] `cargo clippy --locked --workspace --all-targets -- -D warnings`
- [x] `cargo xtask manifest --check`
- [x] `hawkeye check --config licenserc.toml`
- [x] `cargo llvm-cov … --fail-under-lines 70` (78.53% lines)
- [x] `cargo test --workspace --all-targets -- --test-threads=1`
- [x] `cargo test --workspace --all-targets` (parallel)
- [x] `cargo xtask verify-commits --base origin/main` (linear onto main; all commits signed + signed-off)
- [ ] Manual: `rocm bootstrap setup` launches the dash onboarding and installs / adopts (needs an interactive terminal + GPU host)

---
_Note: rebased linearly onto `main` to drop a GitHub web merge commit that lacked a `Signed-off-by` trailer (the sole CI failure). #68's squash-merge commit was re-applied here as a signed + signed-off cherry-pick, so its fix is retained on the path to `main`; the final tree matches the pre-rebase merged state._


rominf (changes requested):
The retirement and the dead-code prune look clean — compiler/clippy-driven, no dangling references to the removed helpers, and the bootstrap reroute is wired correctly (the onboarding overlay renders and takes input regardless of the active tab, and nothing else shadows it on the bootstrap path). The split into reroute-then-delete makes a 45k-line removal reviewable. One thing holds me back from approving:

**The dropped flags are asserted as intentional but not justified.** The UX note says:

> The legacy install-folder picker (`--prefix`) and Release/Nightly channel toggle are intentionally dropped.

"Intentionally dropped" tells us it wasn't an accident, but not *why* — and these are user-facing capabilities being removed. Could you record the reasoning, for reviewers now and readers later? Specifically:

- Why drop them rather than carry them into the dash onboarding wizard — a current limitation of the wizard, a deliberate UX simplification, or deferred work?
- Is losing the install-location choice and the nightly channel acceptable for the `bootstrap setup` audience, or is a follow-up planned to restore them?

A sentence or two of rationale (plus a tracking issue if it's deferred) would make this tradeoff reviewable rather than just declared.

**Related concrete consequence:** the non-interactive fallback in `run_setup` still tells users to "choose an install folder":

```
"ROCm setup needs an interactive terminal. Run `rocm` from a terminal to choose an install folder and set up ROCm/TheRock."
```

With the folder picker gone this is now inaccurate, and bare `rocm` opens the dash chat rather than the setup wizard — so the guidance points at the wrong command too. Worth updating in this PR to point at `rocm bootstrap setup` and drop the folder-picker wording.

michaelroy-amd:
It was intended but only until my next sprint. Restored in PR #73

michaelroy-amd:
Thanks @rominf — both points addressed, plus a history cleanup.

- **Non-interactive message (concrete consequence):** fixed in a new commit — the fallback now says `Run \`rocm bootstrap setup\` from a terminal to set up ROCm/TheRock.`, dropping the folder-picker wording and the wrong bare-`rocm` guidance.
- **Rationale for the dropped flags:** recorded in the UX note.
  - The **Release/Nightly channel toggle** turned out *not* to be dropped after all — #73 restored it in the onboarding wizard (now on `main`), so I corrected the stale claim.
  - The **install-folder picker (`--prefix`)** is **deferred**, not removed by design: the wizard has no install-location step yet, but the `--prefix` plumbing still exists, so it's a UI addition rather than a re-plumb. Tracked in #75.

Separately, this branch's CI was red only on **Commit signatures + sign-off**: a GitHub web merge commit (resolving conflicts) landed without a `Signed-off-by` trailer. I rebased the branch linearly onto `main`, so it's now three signed + signed-off commits with no merge commit — the local `verify-commits` / clippy / test hooks pass. One side effect worth flagging: that web merge had folded in #68's `/v1` chat fix; I dropped it here so #67 is purely the tui.rs retirement, and the `/v1` fix stays in #68 (with its review fixes). #68 will re-target `main` once this lands.

Re-review when you have a moment?

michaelroy-amd:
Correction to my previous comment: I had it backwards on #68.

#68 was **merged into this branch** (`retire-legacy-tui`), not into `main` — so its fix reaches `main` *through* this PR. My initial rebase dropped that squash-merge commit; I've since **re-applied it as a signed + signed-off cherry-pick**, so #68's `/v1` + Esc fix (with its review follow-ups) is retained here.

Final branch: 4 commits, linear on `main`, all signed + signed-off. Verified the resulting tree matches the pre-rebase merged state exactly (aside from the Commit 3 message fix). No functional change was lost.

rominf (approved):
Approving — this addresses my earlier review, and the split-out was the right call.

Verified during review:
- The dead-code prune is complete and correct: every removed symbol (comfyui `render_tui_status`/`render_tui_logs`/`is_installed`/`running_url`, the driver-install TUI-approval cluster, `render_sidebar_text`, `friendly_provider_label`, `freeform_plan_uses_provider`, the `*_for_tui` renderers, `logs_browser_page_count`, `tui_help_text`, `_RETAINED_TUI_ENTRY`) has zero remaining references tree-wide, and every "kept" helper still has live callers. `build_freeform_plan`'s `#[cfg(test)]` gate is correct (its only caller is also test-gated). `cargo build` + `clippy -D warnings` clean.
- Surviving shipping logic kept its own tests, so no silent coverage regression — the coverage gate (78.53% > 70%) holds.
- The chat `/v1` fix is correct and the regression test is genuinely revert-proof (asserts the resolved probed default `ends_with("/v1")`, matching Rig's `base_url + "/" + path` join to `.../v1/chat/completions`).

I pushed one follow-up commit (docs cleanup): `docs/testing.md` still invoked ~40 test names by exact name that lived in the deleted `tui.rs` test module — `cargo test <name>` matches those silently with 0 tests, so the doc read as covered when it wasn't. Removed the stale references, keeping only tests that still exist.

Non-blocking, for a follow-up (not gating this PR):
- The production `run_setup -> dash::run_bootstrap` dispatch has no test; only `bootstrap_args` is covered.
- Esc-closes-console now fans out to the other manager screens with no regression test guarding that behavior.
- Tradeoff worth noting: for an explicitly `--foreground` serve, Esc drops the wizard state with no in-TUI path back to that still-running job (fine for the default managed path via the Services tab). Consistent with the stated "keeps running in the background" intent — flagging so it's a conscious choice.

PR #71, opened 2026-06-30
ci: cache rustc with sccache on compile-heavy jobs

rominf:
## Problem

CI time is dominated by compilation, not tests. On the Windows job the workspace build is ~560s versus ~150s for the test step, and it pays that even on a **full cache hit**: `rust-cache` restores the registry and dependency artifacts but deliberately does not cache workspace crates, so all 15 crates and their targets recompile from scratch every run.

## Change

Add `sccache` as a `RUSTC_WRAPPER` on the compile-heavy jobs (`build-and-test`, `windows-build-and-test`, `clippy`), backed by the GitHub Actions cache. sccache keys each `rustc` invocation on its source, dependencies, and flags, so unchanged crates are restored instead of recompiled across runs and branches. It layers on top of `rust-cache` (which still handles the registry/deps) rather than replacing it. `CARGO_INCREMENTAL=0` is already set, which sccache requires. Coverage is left out (instrumented builds). The action is pinned to a commit SHA.

A `sccache --show-stats` step reports the hit rate so the effect is visible in the logs.

## Status: draft — measuring

This is an experiment to validate the win before merging. sccache helps on the **second** run (the first populates the cache), and the GitHub Actions cache backend has real limits (10 GB/repo shared with rust-cache, per-request chattiness). I'll post the cold-vs-warm build-time delta and hit rate from two runs, then mark ready only if the numbers justify it.

This is independent of the affected-test-selection work (#70).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,6 +94,14 @@ jobs:
 
   build-and-test:
     runs-on: ubuntu-latest
+    # sccache caches each rustc invocation (keyed by source + deps + flags) in the
+    # GitHub Actions cache, so unchanged crates are not recompiled across runs.
+    # This complements setup-rust-toolchain's rust-cache, which restores the
+    # registry and dependency target/ but deliberately does not cache workspace
+    # crates — the bulk of the build time here.
+    env:
+      RUSTC_WRAPPER: sccache
+      SCCACHE_GHA_ENABLED: "true"
     # Don't spend per-OS build/test cycles unless the cheap lint gate is green.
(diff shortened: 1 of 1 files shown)


michaelroy-amd:
Superseded by #125, rebuilt from current `main` rather than resolving this stale workflow branch.

The replacement preserves the original sccache intent while retaining current affected-crate CI, merge-queue checks, signing gates, and E2E/reporting workflows. It remains draft until GitHub Actions provides a cold/warm cache comparison and hit-rate evidence.

PR #72, opened 2026-06-30
xtask: generate THIRD_PARTY_NOTICES.txt and keep it current in CI

rominf:
## What

The repository distributes hundreds of third-party crates in its binaries but had **no Third-Party Notices file** attributing them or reproducing their licenses. This adds one, generated from the dependency tree, and a CI gate that keeps it accurate.

- **`cargo xtask tpn`** generates `THIRD_PARTY_NOTICES.txt` at the repo root by driving [`cargo-about`](https://github.com/EmbarkStudios/cargo-about) with a committed config (`about.toml`, accepted licenses) and template (`about.hbs`, output format). `--check` regenerates in memory and fails if the committed file is stale — the same pattern as `cargo xtask manifest --check`.
- **CI** installs a SHA-pinned `cargo-about` and runs `cargo xtask tpn --check`, gated on a dedicated paths filter (dependency tree, config/template, generator, or the notices file).
- The committed `THIRD_PARTY_NOTICES.txt` groups components by license, listing each license text once with every crate it covers.

## Supporting changes

- **First-party crates marked `publish = false`** (via `[workspace.package]` inheritance). They are an application and its internal libraries, never crates.io packages — and this excludes them from the notices via cargo-about's `private.ignore`.
- **SHA-pinned GitHub Actions**: recorded the full-SHA-with-version-comment convention as a guardrail and applied it to the remaining `install-action` still on a moving tag.

## Notes for reviewers

- The notices file is excluded from the whitespace-fixing pre-commit hooks so its verbatim license texts stay byte-identical to cargo-about's output (otherwise `--check` would fail).
- The cargo-about version is pinned in CI and in the regenerate docs, because its output is version-sensitive and the check is a byte-for-byte comparison.
- The header notice is a generic, factual paragraph isolated in a template comment block, so a specific required format can be dropped in later without touching the generated component list.
- Determinism: cargo-about's default all-targets union is host-independent, so the file regenerates identically on Linux or Windows.

## Verifying

```bash
cargo install cargo-about@0.9.1 --locked --features cli
cargo xtask tpn --check   # passes against the committed file
```

### DIFF
--- a/THIRD_PARTY_NOTICES.txt
+++ b/THIRD_PARTY_NOTICES.txt
@@ -1,0 +1,10775 @@
+THIRD-PARTY SOFTWARE NOTICES
+
+This file lists the third-party open-source components distributed with
+rocm-cli, together with their versions and the full text of their licenses.
+The components are grouped by license; each license text appears once and lists
+every component it covers. This file is generated from the project's dependency
+lockfile by `cargo xtask tpn`; do not edit it by hand.
+
+================================================================================
+Apache License 2.0
+================================================================================
+
+The following component(s) are licensed under Apache License 2.0:
+  - num_threads 0.1.7 <https://github.com/jhpratt/num_threads>
+
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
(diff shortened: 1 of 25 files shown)


THIRD_PARTY_NOTICES.txt
juhovainio, line 1: @rominf could we please use SPDX shortcode for the licenses instead of embedding the full license text

juhovainio (dismissed):

THIRD_PARTY_NOTICES.txt
rominf: The reason the file embeds full license texts is that most permissive licenses legally require it when you redistribute binaries:
- MIT / ISC — "the above copyright notice and this permission notice shall be included in all copies…"
- BSD-2/3-Clause — must reproduce the copyright notice + disclaimer in binary distributions
- Apache-2.0 — must include a copy of the License

A list of SPDX identifiers alone (MIT, Apache-2.0, …) does not satisfy these redistribution obligations — that's exactly why cargo-about emits verbatim texts, and why the PR is built around it. Replacing full text with shortcodes would make the file smaller but legally insufficient as an attribution/notices artifact.

juhovainio (approved):

PR #73, opened 2026-06-30
feat(dash): restore Release/Nightly channel toggle + version pin in onboarding

michaelroy-amd:
## Summary

The dash onboarding's **Install ROCm SDK** path hardcoded `--channel release --format wheel`, dropping the Release/Nightly channel choice the legacy `tui.rs` onboarding used to offer (intentionally dropped when bootstrap adopted the dash onboarding UX in #67). The underlying CLI (`rocm install sdk`) and the TheRock resolver already fully support `--channel`, `--version`, and `--build-date` (mutually exclusive, validated) — only the onboarding **UI** was missing. This re-exposes it.

## What changed

A new **Configure** sub-view on the InstallSdk path (`crates/rocm-dash-tui/src/ui/onboarding.rs`):

- `←/→` (or Tab) toggle **channel** — Release / Nightly
- `↑/↓` cycle an optional **version pin** — None / Build date / Version
- typing edits the pin value; **Enter** stages the install for approval; **Esc** returns to the menu

The pin maps to the CLI's mutually-exclusive `--build-date` / `--version` flags and is available on **both** channels. A `None` pin (the default) leaves the command at `--channel <ch> --format wheel`, so the default **Release path is byte-identical** to the prior behavior.

Scope is limited to `onboarding.rs`. No CLI or resolver changes were needed.

## Test plan

All in `cargo test -p rocm-dash-tui` (run `--test-threads=1` per the known tui env-race flake):

- [x] Default Release confirm → args unchanged (`--channel release --format wheel`)
- [x] Nightly toggle → `--channel nightly`; toggle is reversible
- [x] Nightly + Build date pin → appends `--build-date <YYYY-MM-DD>`
- [x] Version pin → appends `--version <v>` (and not `--build-date`)
- [x] Empty pin value is omitted
- [x] Chars ignored until a pin mode is selected; Backspace edits the value
- [x] Esc returns to the Choose menu
- [x] Existing InstallSdk tests updated for the extra Configure-confirm step
- [x] `cargo clippy -p rocm-dash-tui --all-targets` clean; `cargo build -p rocm` green; 531 lib tests pass

## Notes

- Independent of the `dash-ui-polish-work` launcher-routing work; based on `main`.
- The launcher front-door (bare `rocm`) does not yet expose this — a separate follow-up if desired.

### DIFF
--- a/crates/rocm-dash-tui/src/ui/onboarding.rs
+++ b/crates/rocm-dash-tui/src/ui/onboarding.rs
@@ -8,8 +8,10 @@
 //! version the phase-3 plan calls for: get a clean machine to a working ROCm
 //! runtime two ways —
 //!
-//! - **Install ROCm SDK** — a one-shot gated `rocm install sdk --channel release
-//!   --format wheel`.
+//! - **Install ROCm SDK** — a one-shot gated `rocm install sdk` with a Configure
+//!   step to pick the channel (Release/Nightly) and an optional version pin
+//!   (`--build-date` / `--version`); defaults to `--channel release --format
+//!   wheel`.
 //! - **Adopt existing folder** — pick an existing ROCm env with the Wave-0
 //!   [`FolderBrowser`], then approve `rocm runtimes adopt`.
 //!
@@ -65,7 +67,7 @@ pub const CHOICES: &[OnboardingChoice] = &[
 impl OnboardingChoice {
     const fn label(self) -> &'static str {
         match self {
-            Self::InstallSdk => "Install ROCm SDK (release · pip)",
+            Self::InstallSdk => "Install ROCm SDK (pip)",
             Self::AdoptExisting => "Adopt an existing ROCm folder",
         }
     }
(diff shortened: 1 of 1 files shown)


rominf (approved):
Approve. The default Release path is provably unchanged (byte-identical args, pinned by a dedicated test), the new Configure sub-view is self-contained and well covered, and the generated commands match the real `rocm install sdk` contract.

Verified the flag contract end-to-end against the CLI, not just the unit tests:
- `--channel release|nightly` (default release), `--version`, `--build-date <YYYY-MM-DD>` all match — the UI's date placeholder even mirrors the CLI metavar.
- `rocm install sdk --channel nightly --build-date <date>` parses and reaches the TheRock resolver correctly (a subsequent "no compatible packages for that date" is a data-level result, not a flag issue).
- `PinMode` being a single enum means the UI can only ever emit one of `--version`/`--build-date`, so the mutually-exclusive pair can't be sent together.

Two optional, non-blocking follow-ups:
1. `pin_value` carries over when the Pin mode is switched (e.g. type a build date, switch to Version, and the date becomes the version value). Clearing it on mode change would re-show the placeholder and avoid feeding stale input to a different flag.
2. Pin values are passed through with no client-side validation (spaces / arbitrary non-control chars accepted); the user only learns of an unusable value after approving and running. Delegating to the resolver is a reasonable scoped choice — just noting there's no pre-flight feedback.

Neither blocks merge.

PR #76, opened 2026-07-03
feat(remote): add `rocm remote` — serve a model on a remote GPU host over SSH

volen-silo:
## Summary

`rocm remote serve <host> <model>` brings up an OpenAI-compatible model server on a remote AMD GPU box and makes it callable from your machine with one command: connect over SSH, ensure the CLI + ROCm are present (pushing the CLI if missing), start a **managed** `rocm serve` on the remote, open a loopback tunnel, and print a ready-to-use base URL. Sessions are persistent and observable — `rocm remote status` / `attach` / `stop` manage them afterward.

Built in two commits (one per phase).

## Commit 1 — `rocm remote serve`

- New `rocm remote` command group and `apps/rocm/src/remote/` module (`transport`, `bootstrap`, `mod`).
- A `Transport` trait with an `SshTransport` that shells out to the system `ssh`/`scp` — reuses the user's keys / `~/.ssh/config`, adds no new dependency, and stays synchronous. A Tailscale-native transport can slot in behind the same trait later.
- **Reuses the remote's own `rocm serve` / `rocm services`** — no reimplementation of engine/runtime/device selection; the server runs where the GPU is.
- Bootstrap probes `rocm --version` + ROCm and: **auto-pushes the local binary** if the CLI is missing (the whole runtime is one self-contained binary — `rocm daemon` is the supervisor, there is no separate `rocmd` to install), and **detects-and-fails** if ROCm is absent (never installs the multi-GB SDK).
- Adds **`rocm services list --json`** — the machine-readable listing the remote orchestration reads back.
- Loopback-only forward by default.

## Commit 2 — persistent, observable sessions

- `serve` is now **async and always-persistent**: it starts the managed server (which loads the model in the background), opens a **detached** tunnel that survives the command exiting, records the session locally, and returns immediately with a `Pending` status and the base URL.
- **`rocm remote status [<id>] [--watch]`** reconciles two independent lifecycles into a table — remote-server health (polled over SSH via `services list --json`) and local tunnel liveness (PID check) — shown as separate columns. `attach` rebuilds a dropped tunnel without redeploying; `stop` tears down both. `list` is intentionally reserved for listing configured hosts (distinct from active sessions).
- **No local reconciler daemon** — the `Pending → Healthy` progression comes from the remote's own managed-service lifecycle; `status` polls on demand.
- The tunnel keeper is a standalone detached `ssh` process using **`ControlMaster=no`**, so killing its PID deterministically closes the forward.
- New `RemoteSessionRecord` (one JSON file per session under `remote-sessions/`), mirroring `ManagedServiceRecord`.

## Design decisions

- **System `ssh`/`scp`, not a Rust SSH crate** — synchronous, no new dependency, and the tunnel is a child process that fits the existing detach + PID-tracking model.
- **Auto-push the local binary** (rather than a public download) because the release repo is currently private; the pushed binary is verified on the remote before use, which cleanly catches an arch/libc mismatch.
- **Transport-agnostic** registry/status/attach/stop, so a future Tailscale/mesh transport only swaps the forward mechanic (the `Tunnel` column becomes `via mesh`).

## Test plan

- `cargo test -p rocm -p rocm-core` — green, incl. ~19 new tests: ssh/scp argv construction (incl. `ControlMaster=no` teardown option), session-id + host resolver (incl. ambiguous-host), status classification + table rendering via an injected status stub (no network), bootstrap probe branches, and the natural-language-planner allowlist routing for `remote`.
- `cargo clippy --all-targets -- -D warnings` clean; `cargo fmt --check` clean.
- **Validated end-to-end over an SSH loopback** against a stubbed remote `rocm`: serve → `status` (Remote Healthy / Tunnel Up) → reachable *through the tunnel* → kill the tunnel PID → endpoint refused (confirms the `ControlMaster=no` teardown) → `status` shows degraded (Tunnel Down / Remote Healthy) → `attach` restores it → `stop` tears down both server and tunnel.

## Not included / follow-ups

- **E2E on a real AMD GPU box** (real model load, live `Pending → Healthy`) — deferred; only verifiable against live hardware.
- Auto-reconnect for a dropped tunnel, `rocm remote list` for configured hosts, and a Tailscale transport for NAT traversal are future increments.

### DIFF
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -9,6 +9,7 @@ mod dash;
 mod dash_seam;
 mod provider_keys;
 mod providers;
+mod remote;
 mod serve_summary;
 mod therock;
 mod uninstall;
@@ -331,6 +332,11 @@ rocm serve qwen2.5-7b-instruct --verbose --device gpu_required")]
         #[command(subcommand)]
         command: Option<ServicesCommand>,
     },
+    /// Serve a model on a remote GPU host and forward it to this machine.
+    Remote {
+        #[command(subcommand)]
+        command: remote::RemoteCommand,
+    },
     /// Manage optional background checks and review requests.
     Automations {
         #[command(subcommand)]
@@ -602,6 +608,9 @@ enum ServicesCommand {
         /// Include failed, stopped, and old service records.
         #[arg(short, long)]
         all: bool,
+        /// Emit the machine-readable service records as JSON.
+        #[arg(long)]
+        json: bool,
     },
     /// Show logs for a local model server.
     Logs {
@@ -1505,6 +1514,7 @@ fn dispatch(cli: Cli) -> Result<()> {
         }),
         Some(Command::Comfyui { command }) => comfyui(command),
         Some(Command::Services { command }) => services(command),
+        Some(Command::Remote { command }) => remote::remote(command),
         Some(Command::Automations { command }) => automations(command),
         Some(Command::Config { command }) => config(command),
         Some(Command::Logs {
@@ -4444,9 +4454,20 @@ fn run_foreground_service(
 
 fn services(command: Option<ServicesCommand>) -> Result<()> {
     let paths = AppPaths::discover()?;
-    match command.unwrap_or(ServicesCommand::List { all: false }) {
-        ServicesCommand::List { all } => {
-            print!("{}", render_services_text(&paths, all)?);
+    match command.unwrap_or(ServicesCommand::List {
+        all: false,
+        json: false,
+    }) {
+        ServicesCommand::List { all, json } => {
+            if json {
+                let records = load_managed_services(&paths)?
+                    .into_iter()
+                    .filter(|record| all || managed_service_is_live(record))
+                    .collect::<Vec<_>>();
+                println!("{}", serde_json::to_string_pretty(&records)?);
+            } else {
+                print!("{}", render_services_text(&paths, all)?);
+            }
             Ok(())
         }
         ServicesCommand::Logs { service_id } => {
@@ -15446,6 +15467,7 @@ fn treat_as_natural_language(args: &[String]) -> bool {
         "comfyui",
         "comfy",
         "services",
+        "remote",
         "automations",
         "config",
         "logs",
@@ -16244,6 +16266,21 @@ mod tests {
         assert!(structured.approve);
         assert!(!treat_as_natural_language(&structured.request_args));
         assert!(!should_treat_as_freeform(&structured));
+    }
+
+    #[test]
+    fn remote_subcommand_routes_to_clap_not_the_planner() {
+        // `rocm remote serve …` must reach the structured command, not be
+        // swallowed by the natural-language planner. Guards the STRUCTURED
+        // allowlist against dropping `remote`.
+        let invocation = parse_freeform_invocation(&[
+            "remote".to_owned(),
+            "serve".to_owned(),
+            "gpu-box".to_owned(),
+            "qwen2.5-7b-instruct".to_owned(),
+        ]);
+        assert!(!treat_as_natural_language(&invocation.request_args));
+        assert!(!should_treat_as_freeform(&invocation));
     }
 
     #[test]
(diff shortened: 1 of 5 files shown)


volen-silo:
### CodeQL `rust/path-injection` findings — assessment (false positives, non-blocking)

The 6 CodeQL alerts on this PR are all the same rule — *Uncontrolled data used in path expression* — and all trace to inputs that aren't an untrusted boundary for a local CLI. CodeQL isn't a required check here, so these don't block merge, but recording the analysis for whoever triages the Security tab:

| Alert | Sink | "user-provided value" is… | Verdict |
|---|---|---|---|
| 604–606 | `remove_dir_all` in `remote/mod.rs` **tests** | the process's own `TMPDIR` (`std::env::temp_dir()`) | test-only cleanup of our own temp dir |
| 608–609 | `read_dir` / `is_dir` on `remote_sessions_dir()` | the app's own `ROCM_CLI_DATA_DIR` config env var | reading our own data dir |
| 607 | `fs::write` in `RemoteSessionRecord::write` | `session_id`, derived from the `host` CLI arg | `session_id_for()` sanitizes `host` to `[A-Za-z0-9-]` (no `/`, no `.`), so it can't escape the sessions dir |

This is the same pattern the existing managed-service code uses (`load_managed_services`, `ManagedServiceRecord::write`), which is why the repo already has ~87 open alerts of this exact class. Nothing here widens the trust boundary: the "inputs" are the tool's own config env var, its own `TMPDIR`, and a sanitized CLI argument.

Recommendation: **dismiss as "won't fix / false positive"**, consistent with the existing alerts of this class. Happy to add a defensive `[A-Za-z0-9._-]`-only sanitizer at the `remote_session_path` boundary if we'd prefer belt-and-suspenders — though it would only address alert 607 and diverge from how the rest of the codebase treats the data dir.

volen-silo:
### How to test `rocm remote` locally

**Build:** `cargo build -p rocm`

**Against a reachable GPU host** (the real path):
```bash
# host must be SSH-reachable and have ROCm; the CLI is auto-pushed if missing
rocm remote serve gpu-box qwen2.5-7b-instruct     # returns immediately: Status: Pending + local URL
rocm remote status --watch                        # watch Pending -> Healthy as the model loads
OPENAI_BASE_URL=<printed-url> rocm chat --provider openai --prompt "hi"
rocm remote stop <session-id>                     # tears down tunnel + remote server
```
Worth verifying: the session survives closing the terminal; after `kill`-ing the tunnel pid, `status` shows `Tunnel: Down` (degraded) and `rocm remote attach <id>` rebuilds it; a host without ROCm bails with a clear "install ROCm first" message (no SDK auto-install).

**No GPU box? Loopback smoke (no ROCm needed).** The command surface only asks the remote for four things — `rocm --version`, `rocm serve --managed`, `rocm services list --json`, `rocm services stop` — so you can exercise the whole ssh/scp/tunnel/status/attach/stop lifecycle against `localhost`:
1. Ensure passwordless `ssh localhost` works.
2. Put a small **fake `rocm`** (plus a stub `rocminfo` so the ROCm probe passes) earlier on the remote `PATH` that: prints a version; on `serve --managed` writes a `ManagedServiceRecord` JSON and launches a throwaway `python3 -m http.server` on the port; emits that record for `services list --json`; kills it on `services stop`.
3. Run the normal `rocm remote serve localhost <model>` → `status` → `attach` → `stop` against it.

This confirms the tunnel actually forwards, that killing the tunnel pid closes the forward, and that stop tears both down — all without a GPU. (I have a self-contained `fake-rocm` + driver script that automates exactly this with pass/fail checks; happy to share on request.)

**Note:** `rocm services list --json` was added as the machine-readable listing the remote status polling reads back.

mikeroySoft:
We would likely use TCP for remote connectivity since it's already wired in.

volen-silo:
🔴 Automated review · pr-review-watcher · 2533b457

## Summary
Adds `rocm remote serve/status/attach/stop` — brings up a managed `rocm serve` on a remote GPU box over the system `ssh`/`scp`, opens a detached loopback tunnel, and tracks the session locally; a clean, well-scoped feature that reuses the remote's own serve/services machinery rather than reimplementing it. Verdict: **Approve** (non-blocking polish only). Verified: `cargo build -p rocm -p rocm-core` and `cargo test -p rocm -p rocm-core` both green (513 tests incl. all 19 new remote tests), `cargo clippy --all-targets` and `cargo fmt --check` clean — matching the PR's claims. Confirmed the load-bearing safety property: user-supplied `model`/`engine`/`device`/`gpu` are POSIX-quoted before landing in the remote shell (`shell_quote`), and the only unquoted interpolations (`remote_service_id`, `remote_cli`) are trust-boundary-safe — `service_id` is sanitized by `generate_service_id` (alnum/dash) and `remote_cli` is either `"rocm"` or a hardcoded path; `--managed` serve returns quickly (spawns detached) so the synchronous `run()` orchestration is correct; `ControlMaster=no` + `setsid` make the tunnel PID a real teardown handle. Blocking: 0 · Non-blocking: 4.

**None.**

## Non-blocking
- `apps/rocm/src/remote/mod.rs:369` — `query_remote_status` runs `services list --json` without `--all`; the remote handler filters that path to *live* services only, so a crashed server (status `failed`/`stopped`) is dropped from the listing and reported as **"Gone"**, not "Failed" — `remote_status_label`'s `failed`/`stopped` arms are unreachable here. Pass `--all` so the headline "surface remote health" behavior can actually distinguish a crash from a removed service.
- `apps/rocm/src/remote/mod.rs:168` — re-serving the same host+`remote_port` reuses the deterministic session id and overwrites the record with no teardown of the previous session, orphaning the old detached `ssh -L` process (and, with auto-picked ports, its bound local port). Consider loading any existing record and killing its `tunnel_pid` before overwrite.
- `apps/rocm/src/remote/mod.rs:163` — the tunnel PID is captured immediately after spawn with no post-bind check; with `ExitOnForwardFailure=yes` a bind failure exits the child instantly, so a session can be persisted with an already-dead tunnel. A brief liveness re-check (or a one-shot forward probe) before printing the ready URL would avoid handing back a dead endpoint.
- `apps/rocm/src/remote/transport.rs:120` — `destination` is placed as an argv element *before* `--`, so a `host` beginning with `-` is parsed as an ssh option (e.g. `-oProxyCommand=…`). Self-inflicted only (user's own arg, same as bare `ssh`), but a `--`/`user@`-prefix guard would close it cheaply.

michaelroy-amd:
I'm going to actually close this for now to clean up the PRs. If we intend to do remote, we have another implementation already via the daemon and TCP.

PR #77, opened 2026-07-03
feat(serve): deployment summary by default instead of a log stream

volen-silo:
## What

`rocm serve <model>` streamed raw engine logs and blocked the terminal. It now defaults to backgrounding the server and printing a progressive **deployment summary**:

- An animated spinner while the server starts up.
- A summary table once ready: deployment status, full inference endpoint with port (e.g. `http://127.0.0.1:1337/v1/chat/completions`), the API-qualified model name, and a best-effort inference smoke test (time-to-first-token, tokens/sec).
- Control returns to the shell with the server running in the background; stop via `rocm services stop`, logs via `rocm logs`.

## Flags

- `--verbose` (or `--foreground`) — attach to this terminal and stream every engine log line, as before. For debugging/diagnostics.
- `--no-smoke-test` — skip the post-startup inference probe (metrics show `n/a`).
- `--managed` — now an explicit synonym for the default (background) behavior.

`--verbose` conflicts with `--managed` (no foreground stream to attach to a backgrounded server — use `rocm logs`).

## Works the same across every engine

All new logic lives at the CLI/HTTP layer above the engine adapters, so the summary shape is identical for lemonade, vLLM, SGLang, PyTorch, llama.cpp, and atom. Readiness already normalizes per-engine health paths; the smoke test uses the OpenAI-compatible `/v1/chat/completions` every serving engine exposes. TTFT is measured exactly; throughput is approximated from the token stream and degrades to `n/a` where unavailable.

## Notes

- Rich output is gated on an interactive TTY. Piped / CI / chat-assistant output keeps the plain line-by-line form unchanged.
- `start_managed_service` now returns a report rather than printing, so the rich and plain renderings share one launch path.
- Default lifecycle changes from foreground to background (the old attached behavior is preserved under `--verbose`/`--foreground`).

## Testing

- [x] `cargo fmt --all --check` clean
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean
- [x] `cargo test --workspace --all-targets` green (12 new tests: summary rendering, cross-engine row parity, metric math, `n/a` fallback, flag parsing/conflicts)
- [ ] Live GPU check (deferred, needs AMD hardware): `rocm serve qwen-smoke` and a vLLM model to confirm spinner → table → real TTFT/tok-s, plus `--verbose` streaming

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,12 +138,22 @@ rocm serve <model> [--engine lemonade|pytorch|llama.cpp|atom|vllm|sglang]
                    [--gpu auto|<index>]
                    [--runtime-id KEY | --env-id ID]
                    [--host HOST] [--port PORT]
-                   [--foreground | --managed]
+                   [--verbose] [--foreground | --managed]
+                   [--no-smoke-test]
                    [--allow-public-bind]
 ```
 
-`--managed` runs the server in the background under rocm-cli's supervision.
-`--foreground` attaches it to the current terminal.
+By default the server runs in the background under rocm-cli's supervision and
+prints a deployment summary — a progress indicator while it starts, then a table
+with the status, the full inference endpoint, the API-qualified model name, and a
+quick smoke test (time to first token and approximate tokens/sec). Control
+returns to your shell with the server still running; manage it later with `rocm
+services` (below).
+
+`--verbose` (or `--foreground`) instead attaches the server to the current
+terminal and streams every engine log line — use it to debug a startup problem.
+`--managed` is the explicit form of the default background behavior.
+`--no-smoke-test` skips the post-startup inference probe.
 
 `--gpu` selects which AMD GPU the server runs on. `auto` (the default) probes
 per-GPU VRAM with `amd-smi` and picks the lowest-numbered GPU that is idle and
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -9,6 +9,7 @@ mod dash;
 mod dash_seam;
 mod provider_keys;
 mod providers;
+mod serve_summary;
 mod therock;
 mod uninstall;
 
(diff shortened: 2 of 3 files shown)


rominf (commented):
Nice change — backgrounding by default with a readable summary is a real UX win, and the tests are thorough (the clap conflict is genuinely enforced and covered, and `run_smoke_test` returning `SmokeMetrics` rather than `Result` means a failed probe can't take down the serve command). A few things worth considering, none blocking:

### Spinner — worth a look at `indicatif` before hand-rolling

The custom `Spinner` is small and correct (good call using `crossterm` so it's cross-platform, and writing the animation to stderr while the table goes to stdout keeps piped output clean). That said, the idiomatic way to avoid hand-rolling frame arrays + tick timing + TTY detection is `indicatif`'s `ProgressBar::new_spinner()`, which gives all of that (including a steady-tick background thread) for free.

The honest tradeoff: we don't currently depend on `indicatif` (or any spinner crate — `crossterm`/`ratatui` are the only terminal libs in the tree, and neither is a drop-in inline spinner), so this would be a new dependency, and it only replaces the spinner — the summary table stays hand-rolled either way. So it's a judgment call rather than a clear win. Flagging it mostly so the choice is deliberate; if we expect more progress/spinner UI later, pulling in `indicatif` once may pay off.

### README is now stale

`README.md` (the `rocm serve` section) still documents `[--foreground | --managed]`, describes `--managed` as "background" and `--foreground` as "attach", but doesn't mention that **background is now the default**, and omits the new `--verbose` and `--no-smoke-test` flags. Since this PR changes the default lifecycle, the user-facing docs should move with it.

### Summary can read as success on a readiness timeout

When the 45s HTTP-ready wait times out, `status` becomes `"starting"`, but `render_summary` only special-cases `already_running` — so a launch that never became ready still prints a positive "Deployment summary" and the process exits 0. Given the whole point of the summary is a clear at-a-glance status, it'd be good to visibly distinguish "not ready yet / timed out" from a healthy deployment.

### Minor: `tok/s` is approximate

`gen_tps` is derived from streamed SSE chunk counts, not actual tokens, so it's really ~chunks/s (backend-dependent). A one-line doc/label note ("approximate, assumes ~1 token per streamed chunk") would keep users from over-trusting the exact number.

### Pre-existing, not introduced here

The `try_wait` post-spawn liveness check that surfaces exact startup failures is `#[cfg(not(windows))]`; on Windows the code just sleeps. Not this PR's doing, but since `rocm serve` now backgrounds by default, every Windows user hits that path now instead of only `--managed` users — might be worth a follow-up.

volen-silo:
Thanks @rominf — addressed in 413b9cb:

- **Summary reads as success on timeout** — fixed. A launch that hits the startup timeout (`status = "starting"`) now renders a **"Deployment summary (not ready yet)"** heading plus a note pointing at `rocm logs` / `rocm services list`, so it can't be mistaken for a healthy deployment. Covered by a new test (`readiness_timeout_does_not_read_as_success`).
- **`tok/s` is approximate** — the row is now labelled **`throughput (approx)`**, with a code comment noting it's derived from SSE chunk counts (~1 token/chunk), not the engine's token accounting.
- **README stale** — updated the `rocm serve` section: background is now documented as the default, and `--verbose` / `--no-smoke-test` are added.

- **Spinner vs `indicatif`** — kept hand-rolled for now, deliberately: it'd be a new dependency that only replaces the spinner (the table stays hand-rolled either way). Happy to revisit if we grow more progress UI.
- **Windows `try_wait` liveness check** — agreed it's pre-existing and now on the default path for Windows users; leaving it out of this PR. Worth a follow-up — I can file one if you'd like.

rominf (approved):

PR #80, opened 2026-07-06
feat(lemonade): serve models by their canonical Hugging Face name

rominf:
## Summary
- `rocm serve --engine lemonade <owner/repo:variant>` (e.g. `LiquidAI/LFM2.5-230M-GGUF:Q4_0`) now works. The model is downloaded and served under its **exact canonical Hugging Face name** — the `/v1/models` id, the name in `rocm services`, and the `model` field for `rocm chat` all match what you typed (parity with the vLLM engine).
- **Why:** previously only a handful of built-in Lemonade aliases could be served. Any other name was passed straight to `lemonade load`, which only resolves already-registered models, so an arbitrary Hugging Face checkpoint failed.

## How
- A canonical checkpoint can't be served under its exact name through Lemonade's model router: Lemonade renames registered models (a bare `pull owner/repo:variant` derives a `repo-variant` name; custom registration forces a `user.` prefix). So instead the engine pulls once to download the GGUF, then runs the packaged llama.cpp `llama-server` on that file directly with `--alias <canonical>`, which serves it under exactly that name.
- The `llama-server` binary is discovered under `bin/llamacpp/<backend>/`, so this works with whatever backend Lemonade installed — `vulkan` (e.g. WSL2) or `rocm-stable` (native ROCm). The existing direct-serve path (previously ROCm-only) is generalized and reused for the built-in model too.
- The GGUF is located in the Hugging Face hub cache using Lemonade-style variant matching (exact filename, else a quantization-label match such as `Q4_0`).

## Test plan
- Unit tests: canonical-name detection, GGUF cache-path/variant selection, pull-arg construction, alias pass-through.
- Manual (ROCm or Vulkan GPU host with the Lemonade engine installed):
  - `rocm serve --engine lemonade LiquidAI/LFM2.5-230M-GGUF:Q4_0`
  - `curl -s localhost:11435/v1/models` → the model `id` is `LiquidAI/LFM2.5-230M-GGUF:Q4_0`
  - a chat completion against that model returns `200`
- Exercised against a WSL2 + Lemonade (vulkan) install during development.

## Risk
Low–medium. New behavior is gated to `owner/repo:variant` references; built-in aliases and the default-model path are unchanged (the direct-serve helper was generalized, not rewritten). There is no GPU CI, so the end-to-end serve path is verified manually.

### DIFF
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -15387,9 +15387,21 @@ fn model_list_ready_for_model(
                     .into_iter()
                     .filter_map(|field| model.get(field).and_then(serde_json::Value::as_str))
                     .any(|loaded| service_model_names_match(loaded, canonical_model_id));
-                name_matches && (!require_rocm_backend || service_model_reports_rocm_backend(model))
+                name_matches && (!require_rocm_backend || service_model_gpu_ready(model))
             })
         })
+}
+
+/// Whether a lemonade `/v1/models` entry is served on GPU. A stock `llama-server`
+/// (direct-serve) entry has no `recipe_options` and is accepted (that path only runs
+/// GPU backends); a Lemonade-router entry carries `recipe_options`, so it must report a
+/// ROCm backend — which keeps a registered-but-unloaded model (empty `recipe_options`)
+/// from reading as ready.
+fn service_model_gpu_ready(model: &serde_json::Value) -> bool {
+    match model.get("recipe_options") {
+        None => true,
+        Some(_) => service_model_reports_rocm_backend(model),
+    }
 }
 
 fn service_model_reports_rocm_backend(model: &serde_json::Value) -> bool {
@@ -15825,6 +15837,36 @@ mod tests {
             200,
             "OK",
             "Qwen3-0.6B-GGUF"
+        ));
+    }
+
+    #[test]
+    fn lemonade_direct_serve_model_reads_ready_without_recipe_options() {
+        // The HF direct-serve path runs a stock llama-server whose `/v1/models` entry
+        // has no `recipe_options`. It must read as ready by name (that path is GPU-only),
+        // while a registered-but-unloaded lemonade entry (empty `recipe_options`) must not.
+        let direct = json!({
+            "data": [{ "id": "LiquidAI/LFM2.5-230M-GGUF:Q4_0", "object": "model" }]
+        })
+        .to_string();
+        assert!(service_http_readiness_response_ready(
+            "lemonade",
+            "/v1/models",
+            200,
+            &direct,
+            "LiquidAI/LFM2.5-230M-GGUF:Q4_0"
+        ));
+
+        let registered = json!({
+            "data": [{ "id": "LiquidAI/LFM2.5-230M-GGUF:Q4_0", "recipe_options": {} }]
+        })
+        .to_string();
+        assert!(!service_http_readiness_response_ready(
+            "lemonade",
+            "/v1/models",
+            200,
+            &registered,
+            "LiquidAI/LFM2.5-230M-GGUF:Q4_0"
         ));
     }
 
--- a/engines/lemonade/src/lib.rs
+++ b/engines/lemonade/src/lib.rs
@@ -619,8 +619,8 @@ fn serve_http(request: ServeHttpRequest) -> Result<()> {
     // exact name through Lemonade's model router (Lemonade renames registered models and
     // its registry has several naming quirks). Download the GGUF and run a packaged
     // llama-server on it directly with `--alias`, which serves it under exactly that name.
-    if parse_hf_checkpoint(&request.model_ref).is_some() {
-        return serve_hf_checkpoint(&request, &runtime, &process_env, log_path);
+    if let Some(checkpoint) = parse_hf_checkpoint(&request.model_ref) {
+        return serve_hf_checkpoint(&request, &runtime, &process_env, log_path, &checkpoint);
     }
     if runtime_is_linux() && direct_llama_server_path(&runtime.manifest).is_file() {
         ensure_direct_llama_model_available(&request, &runtime, &process_env, log_path)?;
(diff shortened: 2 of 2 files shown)


volen-silo:
Reviewed at `1d61cb6` (fan-out review: parallel file/coupling reviewers + synthesis, findings verified against source). Verification gates green: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace --all-targets` all pass. Nice feature, clean refactor — but two blocking items remain.

Credit where due: `1d61cb6`'s `is_safe_hf_component` correctly closes a path-traversal hole in the cache-path construction (rejects `..`, `/`, `\` while preserving valid names like `LiquidAI/LFM2.5-230M-GGUF`). Good hardening.

## 🔴 Blocking

**B1 — Silent CPU fallback under `gpu_required` (AGENTS.md §6).**
`DIRECT_LLAMA_SERVER_BACKENDS` (line 1905) includes `"cpu"`, and `find_llama_server_binary` (line 1911) also falls back to *any* directory under `bin/llamacpp/`. On a host where Lemonade installed only the `cpu` backend (reachable via `select_best_llamacpp_backend` when rocm/vulkan are unsupported), the HF path selects the cpu `llama-server` and runs it with `--n-gpu-layers 999` (silently clamped to CPU). `require_gpu_required` only checks the *policy*, not the selected binary — so this is a silent CPU serve under `gpu_required`, which §6 forbids.
*Suggested fix:* drop `"cpu"` from `DIRECT_LLAMA_SERVER_BACKENDS` and skip a `cpu` dir in the fallback scan (return `None` → bail with an actionable message), or add a post-selection guard that bails when the chosen binary's backend is `cpu` and the policy isn't `CpuOnly`.

**B2 — Bare `owner/repo` (no `:variant`) is accepted, then fails opaquely.**
`parse_hf_checkpoint("owner/repo")` returns `Some { variant: None }`, so it routes to `serve_hf_checkpoint`. But `run_lemonade_pull`'s own doc (lines ~1655–1658) says a variantless pull triggers Lemonade's interactive variant menu, and it sets `stdin(Stdio::null())` — so the code deliberately enters a case it documents as unanswerable. It exits non-zero with a message that never mentions the missing variant; and if the repo is already cached, `select_gguf_for_variant(ggufs, None)` returns the alphabetically-first GGUF (usually the lowest quant) — a silent wrong-file pick. The PR description says the path is "gated to `owner/repo:variant`", which the code doesn't enforce.
*Suggested fix:* bail early when `variant.is_none()` with a clear message (`use 'owner/repo:variant', e.g. ':Q4_K_M'`); add a regression test.

## 🟠 Non-blocking (important — cross-file)

**N1 — App-level readiness always reports "starting" for the HF path, even on a healthy ROCm server.**
The direct path runs stock `llama-server`, which emits neither Lemonade's `all_models_loaded` (`/v1/health`) nor `recipe_options.llamacpp_backend` (`/v1/models`). `apps/rocm`'s lemonade readiness (`service_http_readiness_response_ready` → `model_list_ready_for_model(..., require_rocm_backend=true)`) therefore never sees ready — regardless of the real backend. Result: the deployment summary shows "not ready yet" and the smoke test is skipped, though the server actually works. `healthcheck_service` has the analogous issue for `rocm services`. vLLM avoids this by using `require_rocm_backend=false`.
*Suggested fix:* for the direct-serve (raw llama-server) mode, accept a name-only `/v1/models` match without the ROCm-backend requirement.

## 🟡 Non-blocking (minor)
- **N2** `select_gguf_for_variant` substring match is unanchored (`contains`, line ~2056): variant `Q4` matches both `…-Q4_0.gguf` and `…-Q4_K_M.gguf`, alphabetically-first wins silently. Anchor on a separator boundary or warn on multiple matches; add a test.
- **N3** `foo/bar.gguf` (repo with extension, no colon) still parses as a checkpoint — minor; a non-existent one yields a confusing "pull failed" rather than "file not found".
- **N4** No direct tests for `find_llama_server_binary` / `collect_gguf_files` — both pure filesystem logic, testable with `tempfile` (no GPU), and central to B1.
- **N5** llama-server child is leaked on startup-timeout / smoke-test failure in `serve_direct_llama_server` (state stays `running`); pre-existing, but more likely now with large first-run GGUFs. Kill the child and mark `failed` on those bail paths.
- **N6/N7** doc nits: `select_gguf_for_variant` says "exact filename match" but is `eq_ignore_ascii_case` (say case-insensitive); `serve_hf_checkpoint` doc implies a variant is always present.

## Tradeoffs (confirm deliberate, no change requested)
- **T1** `find_hf_checkpoint_gguf` sorts GGUFs by full path (revision hash), not mtime, so after a re-pull an older snapshot revision may sort first. Worth a one-line comment.
- **T2** Error-propagation asymmetry (`?` in `ensure_hf_checkpoint_downloaded` vs `let _` in `ensure_direct_llama_model_available`) is correct — flagging only so it's visibly intentional.

## Positives
Refactor renames are complete (no stale refs; `builtin_serve_http` contract untouched across all engines + `apps/rocm`). Process args built via individual `Command::arg()` — injection-safe. `generate_service_id` sanitizes `/`/`:` so HF names are safe as service/file keys. Built-in aliases don't regress (`resolve_lemonade_model_ref` collapses them before `parse_hf_checkpoint`). No OSS leaks; commit is signed-off with no AI footer.

rominf:
Thanks for the thorough review. Addressed in `53043e9`:

**Blocking**
- **B1 (silent CPU under gpu_required)** — dropped `cpu` from `DIRECT_LLAMA_SERVER_BACKENDS` (now GPU-only: `rocm-stable`/`rocm-nightly`/`rocm`/`vulkan`), so a cpu-only host returns `None` → bails rather than serving on CPU. The "scans any dir" fallback you flagged was already removed in `cd27ae8` (it's now an explicit allowlist, no `read_dir`).
- **B2 (bare `owner/repo`)** — `serve_hf_checkpoint` now bails early when there's no `:variant` with an actionable message (`use owner/repo:variant, e.g. :Q4_K_M`), and never enters the interactive-pull path. Regression coverage added.

**N1 (direct-serve never reads ready)** — fixed in both places. A direct-serve stock `llama-server` `/v1/models` entry has no `recipe_options`, so it's now accepted by name (that path is GPU-only after B1), while Lemonade-router entries still require a ROCm/expected backend — so a registered-but-unloaded model (empty `recipe_options`) still correctly reads not-ready. Applied to `apps/rocm`'s `model_list_ready_for_model` and the engine's `query_loaded_model_endpoint`/`healthcheck_service`.

**N2 (unanchored variant match)** — instead of silently taking the alphabetically-first, an ambiguous variant (one matching multiple cached GGUFs, e.g. `:Q4` → `Q4_0` + `Q4_K_M`) now bails listing the matches; a full label (`:Q4_0`) or exact filename resolves to exactly one. Test added.

**N3 (`foo/bar.gguf`)** — now covered by the B2 variant-required bail (it has no `:variant`), so it gets the clear "specify a variant" message rather than an opaque pull failure.

**N4** — added dependency-free tests for `find_llama_server_binary` (prefers GPU, skips cpu) and `collect_gguf_files` (recursion/filtering), using a temp dir so no `tempfile`/notices change.

**N5** — `serve_direct_llama_server` now kills the child and marks the service `failed` if it never becomes ready / fails the smoke test, instead of leaking it with state stuck at `running`.

**N6/N7** — doc fixes (`select_gguf_for_variant` says case-insensitive; `serve_hf_checkpoint` notes a variant is required).

**T1** — added a comment that GGUFs are sorted by path (not mtime); the new ambiguity gate means we never rely on that ordering to disambiguate.
**T2** — confirmed intentional; no change.

Interactive variant selection (the nicer alternative to the B2/N2 errors) is tracked as a follow-up: #82.

juhovainio (approved):
LGTM, tested locally on Ubuntu WSL

PR #81, opened 2026-07-06
fix(dash): rebuild live chat agent when a local endpoint is accepted

michaelroy-amd:
## Problem

In the dash TUI chat, accepting a **detected local endpoint** updated the config (`chat_llm`) but never rebuilt the live `agent` object that actually issues completions. The agent is built once at startup and only ever reassigned on a `/provider` switch. So after a user selected a local server (e.g. a Vulkan-backed lemonade running `qwen3-0.6b`), chat kept routing to the **startup** backend.

When that startup backend was a cloud gateway (an Azure API Management endpoint configured via `chat_url` + `tui.chat_auth_header = "Ocp-Apim-Subscription-Key"` + an env API key), every submit hit the gateway and returned:

```
chat request failed: ... 401 Unauthorized "Access denied due to invalid subscription key"
```

The UI showed the local endpoint as selected; the request physically went to Azure. `chat_llm` and the live `agent` had silently diverged.

## Fix

Close the divergence with a one-shot `chat_endpoint_rebuild` edge (mirrors the existing `chat_persist_dispatch` / `provider_switch` edge-drain pattern):

- **`accept_detect_offer`** (and `save_detect_offer` via delegation) raises the edge and selects `ChatProvider::Local`.
- A **new `event_loop` drain** — placed **after** the `/provider` switch drain and **before** the `chat_dispatch` drain (ordering is load-bearing: otherwise the first post-accept submit would still use the stale agent) — rebuilds the live `agent` from the auth-free local `chat_llm` and refreshes the `local_agent` restore snapshot, so a later `/provider local` restores the accepted backend rather than the stale startup one. On build failure the agent is left unchanged and an error turn surfaces (mirrors `/provider` failure semantics).

### Defense-in-depth

`resolve_llm_config` now strips `api_key`/`auth_header` when the resolved `base_url` host is loopback (`127.0.0.0/8`, `localhost`, `::1`). This prevents a *startup-configured* local endpoint from leaking a cloud credential — a distinct code path the detect-accept fix does not cover.

> Note: this is an intentional behavior change. A user pointing at a local authenticating proxy that genuinely requires an auth header (e.g. a corp gateway bound to `127.0.0.1`) will have that header stripped. This matches `detected_llm_config`'s existing no-auth-for-local stance and is documented in-code.

## Tests

- Extended `detect_offer_lifecycle_accept_switches_chat` and `save_detect_offer_accepts_and_raises_persist_edge` to assert the rebuild edge is raised and the provider realigns to `Local` — **starting from a non-Local provider** so the realignment assertion is load-bearing rather than tautological.
- Added `loopback_base_url_strips_cloud_auth` and `remote_base_url_keeps_cloud_auth` unit tests for `resolve_llm_config`.

## Verification

- `cargo build -p rocm-dash-tui` ✅
- `cargo clippy -p rocm-dash-tui --all-targets -- -D warnings` ✅ (clean)
- `cargo test -p rocm-dash-tui -- --test-threads=1` ✅ — 533 lib + 16 + 5 tests pass, 0 failed

## Notes for reviewers

The fix was developed with an adversarial multi-agent review pass. The rebuild-drain **ordering** (before `chat_dispatch`) and the `local_agent` snapshot refresh are the subtle, load-bearing pieces. The event-loop drain itself is covered indirectly (consistent with the pre-existing `/provider` drain, which is also not unit-tested beyond `build_chat_agent` invariants).


juhovainio (commented):
The auto-detect-and-switch flow is a nice UX addition. Several correctness issues in the new drain path and the loopback auth stripping below.

crates/rocm-dash-tui/src/app/mod.rs
```diff
@@ -963,13 +971,23 @@ impl AppState {
...
         if let Some(cfg) = self.chat_detect_offer.take() {
             self.chat_llm = Some(cfg);
             self.chat_consent = ChatConsent::Accepted;
             self.chat_focused = true;
+            // The accepted endpoint is local; align the displayed provider and
+            // raise the rebuild edge so `event_loop` swaps the live agent to it
+            // (the startup agent may be a cloud gateway — see
+            // `chat_endpoint_rebuild`).
+            self.active_provider = ChatProvider::Local;
+            self.chat_endpoint_rebuild = true;
```
juhovainio: **Bug: `active_provider` not reverted on rebuild failure.** This sets `active_provider = Local` unconditionally before the rebuild attempt, but the `Err` arm (further down) never restores the previous value. If `RigAgentClient::new` fails, the tab displays "Local" while `agent` still points at the old backend, silently routing submits to the wrong place. The existing `provider_switch` drain explicitly captures `previous` and restores it on error — this path should do the same:
```rust
let previous_provider = self.active_provider;
// ... set Local ...
// in Err arm:
state.active_provider = previous_provider;
```

crates/rocm-dash-tui/src/app/mod.rs
```diff
@@ -1880,6 +1900,43 @@ async fn event_loop(terminal: &mut Tui, args: &ResolvedArgs) -> color_eyre::Resu
...
+                            std::sync::Arc::new(c) as std::sync::Arc<dyn crate::agent::AgentClient>;
+                        agent = Some(arc.clone());
+                        // Refresh the restore snapshot so a later `/provider
+                        // local` restores THIS accepted backend, not the stale
+                        // startup one.
+                        local_agent = Some(arc);
+                        state
+                            .chat
+                            .push(ChatTurn::agent("switched to local".to_string()));
+                    }
```
juhovainio: **Bug: system notification corrupts LLM context window.** `ChatTurn::agent(...)` uses `ChatRole::Agent`, which `build_messages` maps to `Message::assistant(content)`. This means the literal string `"switched to local"` is sent to the model as a prior assistant turn on every subsequent submit — text the model never generated, corrupting its context. Consider a `ChatRole::System` / `::Info` variant that is rendered visually but filtered out in `build_messages`.

crates/rocm-dash-tui/src/app/mod.rs
```diff
@@ -1880,6 +1900,43 @@ async fn event_loop(terminal: &mut Tui, args: &ResolvedArgs) -> color_eyre::Resu
...
+        // `/provider local` restore snapshot — at the new local backend.
+        // `accept_detect_offer` swaps `chat_llm` to the auth-free local config
+        // but stays I/O-free, so without this the stale startup agent keeps
+        // routing chat to the cloud gateway (wrong-backend 401 bug). On failure
+        // `agent`/`local_agent` are left unchanged and an error turn surfaces.
+        // Construction only — no network until the next submit.
+        if state.chat_endpoint_rebuild {
+            state.chat_endpoint_rebuild = false;
+            if let Some(cfg) = state.chat_llm.clone() {
+                match crate::agent::RigAgentClient::new(
```
juhovainio: **Bug: silent no-op when `chat_llm` is None.** `chat_endpoint_rebuild` is cleared to `false` on line 84 before this guard. If `chat_llm` is `None` at drain time, the flag is consumed, no agent is built, no error turn is pushed, and `active_provider` stays stuck at `Local` with no feedback to the user. An `else` branch reverting `active_provider` and pushing an error turn would cover this.

crates/rocm-dash-tui/src/app/mod.rs
```diff
@@ -1880,6 +1900,43 @@ async fn event_loop(terminal: &mut Tui, args: &ResolvedArgs) -> color_eyre::Resu
...
+        // Drain the endpoint-rebuild edge (Phase 8 sibling). An accepted
+        // detected-local offer must re-point the LIVE `agent` — and the
+        // `/provider local` restore snapshot — at the new local backend.
+        // `accept_detect_offer` swaps `chat_llm` to the auth-free local config
+        // but stays I/O-free, so without this the stale startup agent keeps
+        // routing chat to the cloud gateway (wrong-backend 401 bug). On failure
+        // `agent`/`local_agent` are left unchanged and an error turn surfaces.
+        // Construction only — no network until the next submit.
+        if state.chat_endpoint_rebuild {
+            state.chat_endpoint_rebuild = false;
```
juhovainio: **Design: rebuild drain duplicates the `/provider local` construction path.** This block re-spells `RigAgentClient::new + Arc cast + local_agent refresh + message` that already exists in the `ProviderSwitch/Local` arm. Any future change to local-agent construction must be applied in both places. The cleaner fix would be to raise a `ProviderSwitch { previous, target: Local }` edge from `accept_detect_offer`, letting the existing drain handle it — and getting the `previous`-capture revert (finding above) for free.

crates/rocm-dash-tui/src/llm.rs
```diff
@@ -84,11 +84,24 @@ pub fn resolve_llm_config(
         .or(cfg_model)
         .map_or_else(|| DEFAULT_CHAT_MODEL.to_string(), str::to_string);
 
+    // Loopback base_urls talk to a local server that needs no gateway auth;
+    // never attach a cloud subscription-key header/api_key to them (prevents a
+    // startup-configured local endpoint from leaking a cloud credential).
+    let is_loopback = parse_host_port(&base_url).is_some_and(|(host, _)| is_loopback_host(&host));
+
```
juhovainio, line 91: **UX issue: silent auth stripping with no warning or escape hatch.** A user running an authenticating local proxy (e.g. a corporate API gateway on `127.0.0.1:8080`) will silently receive 401s with no log line and no config knob to opt out. The PR description acknowledges the trade-off but there is no runtime signal at the point of stripping. At minimum, a `warn!()` log when credentials are discarded would make the behaviour debuggable.

crates/rocm-dash-tui/src/llm.rs
```diff
@@ -120,6 +133,17 @@ pub fn parse_host_port(base_url: &str) -> Option<(String, u16)> {
     }
```
juhovainio, line 159: **Bug: portless bracketed IPv6 URL bypasses auth stripping.** For `http://[::1]/v1` (no explicit port), `parse_host_port` calls `rsplit_once(':')` on the authority `[::1]`, which splits *inside* the brackets — host = `"["`, port_str = `"1]"` — the port parse fails and the function returns `None`. Therefore `is_loopback = false` and credentials are forwarded. All common forms (`localhost`, `127.x.x.x`, `[::1]:port`) are handled correctly; only portless bracketed IPv6 is affected. Fix: detect a leading `[` and handle the bracketed IPv6 form explicitly before falling through to `rsplit_once`.

crates/rocm-dash-tui/src/llm.rs
```diff
@@ -120,6 +133,17 @@ pub fn parse_host_port(base_url: &str) -> Option<(String, u16)> {
...
 }
 
+/// True when `host` is a loopback address.
+///
+/// Matches `localhost` (case-insensitive), the IPv6 loopback (`::1`, with or
+/// without brackets), and any `127.0.0.0/8` IPv4 address.
+fn is_loopback_host(host: &str) -> bool {
+    host.eq_ignore_ascii_case("localhost")
+        || host == "::1"
+        || host == "[::1]"
```
juhovainio: **Nit: dead code branch.** `parse_host_port` preserves brackets for bracketed IPv6 URLs, returning `"[::1]"` as the host, so this `|| host == "::1"` branch is never reached from this call path. (The bare form is reachable in other copies of this helper called with raw CLI host strings, but this private `llm.rs` copy is only called via `parse_host_port`.)

crates/rocm-dash-tui/src/app/mod.rs
michaelroy-amd: Fixed in a6fb302. The edge now carries the previous provider (`chat_endpoint_rebuild: Option<ChatProvider>`); `accept_detect_offer` captures it before the optimistic switch to `Local`, and the drain reverts `active_provider` to it on failure via a shared `revert` closure — mirroring the `provider_switch` drain as you suggested.

crates/rocm-dash-tui/src/app/mod.rs
michaelroy-amd: Fixed in aa03a6d. Added a `ChatRole::System` variant that `build_messages` drops (alongside `Error`), so operational notices are rendered but never sent to the model. Routed all `"switched to …"` notices — this one plus the two pre-existing `provider_switch` ones, same bug — through `ChatTurn::system`.

crates/rocm-dash-tui/src/app/mod.rs
michaelroy-amd: Fixed in 827fad9. The `None` case now folds into the same `revert` path: it restores the previous provider and pushes an actionable error turn instead of clearing the flag and stranding the tab on `Local`.

crates/rocm-dash-tui/src/app/mod.rs
michaelroy-amd: Addressed in 9af1c60. I extracted `build_local_agent()` as the single local-construction path, now shared by the startup build and the rebuild drain — so `RigAgentClient::new` + the `Arc` cast live in one place.

I kept the rebuild as its own drain rather than routing through `ProviderSwitch { target: Local }`: that arm *restores* the `local_agent` snapshot (`agent = local_agent.clone()`), whereas accepting a detected offer must *rebuild* from the new `chat_llm` and refresh the snapshot. Reusing it would have conflated restore-vs-rebuild. The `previous`-capture revert you wanted came in via the edge instead (a6fb302).

crates/rocm-dash-tui/src/llm.rs
michaelroy-amd: Fixed in e91f7af. `resolve_llm_config` now emits a `tracing::warn!` naming the `base_url` when it discards a credential for a loopback host, fired only when an `api_key`/`auth_header` was actually present. Makes the stripping debuggable at the point it happens.

crates/rocm-dash-tui/src/llm.rs
michaelroy-amd: Fixed in ebcdf0c. `parse_host_port` now handles the bracketed IPv6 authority (`[host]` / `[host]:port`) explicitly before the `rsplit_once(':')` fallback, returning the bare host. `http://[::1]/v1` now parses to `("::1", 80)` and gets stripped; added a regression test plus the portless URL to `loopback_base_url_strips_cloud_auth`. Bonus: it also fixes `probe_endpoint`, since `to_socket_addrs` rejects bracketed literals.

crates/rocm-dash-tui/src/llm.rs
michaelroy-amd: Fixed in f053322. Since `parse_host_port` now strips the brackets, `is_loopback_host` only sees a bare host, so I dropped the unreachable `host == "[::1]"` branch and kept the now-live `host == "::1"`, updating the doc to state the bare-host expectation.

juhovainio (approved):
LGTM! Tested locally on WSL Ubuntu with Radeon 780M iGPU

Local model was detected:
Image

And responses are received as expected:
Image

fredespi:
Cross-reference: #88 (EAI-7220) touches the same chat-endpoint flow but at a different layer, so the two are complementary rather than overlapping.

- **This PR (#81)** fixes the *wiring*: when a detected local endpoint is accepted, the live `agent` is rebuilt (and loopback base_urls no longer carry cloud auth), so requests actually go to the accepted backend instead of the stale startup one.
- **#88** fixes *discovery*: `detect_local_endpoint()` only probed Lemonade `:13305` and vLLM `:8000`, so a non-managed `rocm serve` instance (its own default `:11435`) was never found in the first place. It adds `:11435` to the probe list and updates the two UI copy strings.

End to end you need both: #88 makes `rocm serve` discoverable, #81 makes an accepted endpoint get used.

Overlap is minimal — the shared files (`llm.rs`, `ui/tabs/chat.rs`) touch different functions/regions and don't conflict. The one spot to watch is `app/mod.rs`: #88's edit to the `refresh_detect` failure message sits a couple of lines above this PR's `accept_detect_offer` rewrite, so whichever merges second will need a trivial adjacency fixup (no semantic conflict).

PR #85, opened 2026-07-07
feat(serve): detach from an attached serve session with Ctrl-D

volen-silo:
## Summary

`rocm serve --verbose` / `--foreground` ran the engine **in-process**, so the deployment was tied to the terminal — the only way out was `Ctrl-C`, which killed the model. (The default background `rocm serve` was already detached; this closes the gap on the attached/log-streaming path.)

The attached path now spawns the **same detached managed child** the background path uses and tails its log, so the server outlives the session:

- **Ctrl-D → detach**: stops the stream, returns to the shell, leaves the server running. It becomes an ordinary managed service (`rocm services` / `rocm logs` / `rocm services stop`).
- **Ctrl-C → stop**: shuts the server down (with a direct-signal fallback so the model is never orphaned).
- Non-TTY (piped/CI) falls back to a plain log follow.

## Implementation notes

- Split the engine spawn out of `start_managed_service` into `spawn_managed_engine_child` (+ `ManagedSpawn`) so the attached path streams startup logs live instead of blocking on the readiness wait.
- Rewrote the in-process `run_foreground_service` as `run_attached_service` + `stream_attached_logs`, using crossterm raw mode to capture the keys (in raw mode Ctrl-C arrives as a key event, not SIGINT). A `RawModeGuard` restores the terminal on every exit path.
- The key→action mapping is a pure `const fn` (`detach_key_outcome`), unit-tested without a TTY.

## Acceptance criteria

- [x] A user can detach from the ongoing deployment started by `rocm serve`.
- [x] Detaching does not stop the model deployment.
- [x] The user can continue working in the same terminal after detaching.

## Testing

- `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test -p rocm` all pass (incl. new `detach_key_*` tests).
- The interactive Ctrl-D/Ctrl-C behavior needs a real TTY + GPU, so it's covered by a new step in `docs/manual-testing.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,10 +190,13 @@ quick smoke test (time to first token and approximate tokens/sec). Control
 returns to your shell with the server still running; manage it later with `rocm
 services` (below).
 
-`--verbose` (or `--foreground`) instead attaches the server to the current
+`--verbose` (or `--foreground`) instead attaches to the server in the current
 terminal and streams every engine log line — use it to debug a startup problem.
-`--managed` is the explicit form of the default background behavior.
-`--no-smoke-test` skips the post-startup inference probe.
+The server still runs as a managed background process, so while streaming you can
+press **Ctrl-D to detach** — the log stream stops, your shell comes back, and the
+server keeps running (manage it afterward with `rocm services`). Press **Ctrl-C**
+to stop the server instead. `--managed` is the explicit form of the default
+background behavior. `--no-smoke-test` skips the post-startup inference probe.
 
 Use full HuggingFace model IDs (e.g., `Qwen/Qwen2.5-1.5B-Instruct`) for
 reliable cross-engine compatibility. Short aliases from `rocm model` may not
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -4052,6 +4052,7 @@ enum ManagedSpawn {
 /// ([`run_attached_service`]).
 #[allow(clippy::too_many_arguments)]
 fn spawn_managed_engine_child(
+    paths: &AppPaths,
     engine: &str,
     service_id: &str,
     requested_model: &str,
@@ -4064,7 +4065,6 @@ fn spawn_managed_engine_child(
     env_id: Option<&str>,
     engine_recipe: Option<&EngineRecipeHint>,
 ) -> Result<ManagedSpawn> {
-    let paths = AppPaths::discover()?;
     paths.ensure()?;
     fs::create_dir_all(paths.services_dir())?;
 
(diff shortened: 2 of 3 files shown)


rominf (commented):
Nice refactor — pulling `spawn_managed_engine_child` out of `start_managed_service` so the attached path spawns the same detached child and tails its log fixes this at the right layer, and the `RawModeGuard` + pure `detach_key_outcome` split keeps the terminal handling testable. One thing worth addressing before merge, plus two minor nits.

### Attached path doesn't start the supervisor daemon

The background path spawns the child and then calls `ensure_background_helper_running_quiet(...)` (`apps/rocm/src/main.rs:3865`). Per the `Daemon` doc comment, that daemon is what "health-checks and auto-recovers managed local model servers (Lemonade, vLLM)" and feeds the dashboard. `run_attached_service` spawns the identical managed child but never ensures the daemon — so a server started with `--verbose`/`--foreground` and detached with Ctrl-D gets no health-check, no auto-recovery, and no dashboard metrics, unless the daemon already happens to be running from an earlier `serve --managed` / `automations enable`.

That's a gap against "It becomes an ordinary managed service": `rocm services` / `rocm logs` / `rocm services stop` still work (they're file/PID-based), but the supervision a background-managed server gets is missing. It also affects the `AttachOutcome::ServerExited` case — with no daemon, a self-exited server's record stays `running` and is never reconciled.

Looks like a one-liner: call `ensure_background_helper_running_quiet(true)` right after the `ManagedSpawn::Spawned` arm (`apps/rocm/src/main.rs:4406`), matching the background path's ordering so the server is supervised even while attached.

### Minor: raw-mode startup race

The key-reader thread is spawned before `enable_raw_mode()`. In that window a terminal Ctrl-C raises SIGINT and kills the CLI abruptly (the detached child survives, but with no detach/stop message). Enabling raw mode before spawning the reader closes the window.

### Minor: redundant `AppPaths::discover()`

`start_managed_service` calls `AppPaths::discover()` again (`apps/rocm/src/main.rs:4216`) right after `spawn_managed_engine_child` already discovered it. Harmless, just noise.

volen-silo:
Thanks @rominf — all three addressed in d1b11ec:

- **Supervisor daemon on the attached path**: `run_attached_service` now calls `ensure_background_helper_running_quiet(true)` right after the `Spawned` arm, matching the background path's ordering. A Ctrl-D-detached server now gets the same health-check / auto-recovery / record reconciliation / dashboard metrics, and a self-exited server's record gets reconciled.
- **Raw-mode startup race**: `enable_raw_mode()` + the `RawModeGuard` now run *before* the key-reader thread is spawned, so there's no window where a terminal Ctrl-C kills the CLI with no detach/stop message.
- **Redundant `AppPaths::discover()`**: `spawn_managed_engine_child` now takes `&AppPaths`; `start_managed_service` discovers once and threads it in.

`cargo fmt`, `clippy --workspace --all-targets -D warnings`, and `cargo test -p rocm` all pass.

rominf (approved):
The follow-up commit resolves all three points:

- **Supervisor daemon** — `run_attached_service` now calls `ensure_background_helper_running_quiet(true)` right after spawning the child, so a Ctrl-D-detached (or self-exited) server gets the same health-check, auto-recovery, record reconciliation, and dashboard metrics as a background-managed one. Quiet, and placed after the already-running early return, so it only fires when a child was actually spawned.
- **Raw-mode race** — raw mode is now entered before the key reader thread starts, closing the SIGINT window.
- **Redundant discover** — `AppPaths` is discovered once and threaded into `spawn_managed_engine_child`.

All CI green. LGTM.

PR #88, opened 2026-07-08
fix(dash): detect rocm serve's default port in chat auto-detect

fredespi:
## Summary
- `rocm serve` starts vLLM on its own default port (11435, `rocm_core::DEFAULT_LOCAL_PORT`), but the dash chat tab's TCP-probe fallback (used when no `ManagedServiceRecord` exists, i.e. `rocm serve` without `--managed`) only checked `:8000` (vLLM's own default) and `:13305` (Lemonade).
- Extends the fallback probe list to include port 11435, mirrored as a local constant (`rocm-dash-tui` deliberately carries no `rocm-core` dependency, same pattern already used in `serve_wizard.rs`), and updates the two UI copy strings that hardcoded the old port list so they stay in sync.

## Test plan
- [x] `cargo test` passes for `rocm-dash-tui`
- [x] `cargo clippy --workspace --all-targets -- -D warnings` passes
- [x] `cargo fmt --check` passes

### DIFF
--- a/crates/rocm-dash-tui/src/app/mod.rs
+++ b/crates/rocm-dash-tui/src/app/mod.rs
@@ -995,8 +995,9 @@ impl AppState {
             self.chat_detect_offer = Some(cfg);
         } else {
             self.chat_detect_offer = None;
-            self.chat_detect_msg =
-                Some("no local engine found (Lemonade :13305 / vLLM :8000)".into());
+            self.chat_detect_msg = Some(
+                "no local engine found (Lemonade :13305 / vLLM :8000 / rocm serve :11435)".into(),
+            );
         }
     }
 
--- a/crates/rocm-dash-tui/src/llm.rs
+++ b/crates/rocm-dash-tui/src/llm.rs
@@ -139,7 +139,9 @@ pub fn probe_endpoint(base_url: &str, timeout: Duration) -> bool {
     false
 }
 
-/// Probe the known local serving endpoints in priority order (Lemonade, then vLLM) and return the first reachable one.
+/// Probe the known local serving endpoints in priority order (Lemonade, then
+/// vLLM, then `rocm serve`'s non-managed default) and return the first
+/// reachable one.
 ///
 /// Used by the TUI's in-app "detect a local engine" action so the user need not run the CLI skill first.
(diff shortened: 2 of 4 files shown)


juhovainio (approved):

michaelroy-amd (commented):
## Review — Approve ✅

Small, correct, well-scoped fix. Core claims verified against the codebase.

### Verified
- **Port value is correct**: `rocm_core::DEFAULT_LOCAL_PORT = 11_435` (`crates/rocm-core/src/lib.rs:65`) — the mirrored literal matches exactly.
- **No-dependency rationale is real**: `rocm-dash-tui/Cargo.toml` genuinely has no `rocm-core` dep, so duplicating the literal (with a comment pointing back to the source of truth) is consistent with the existing pattern in `serve_wizard.rs`.
- **Detect path is clean**: `detect_local_chat` → `detect_local_endpoint` → `detected_llm_config` (`llm.rs:171`), which carries only `base_url`/`model`/auth (no engine classification), so surfacing `:11435` here needs nothing more.
- `llm::` tests pass; UI copy strings updated in lockstep; doc-comment first paragraph stays short (respects the repo's `too_long_first_doc_paragraph` clippy lint).

### Notes (non-blocking)

**LOW — latent engine-inference gap.** `auto_config_change` (`skills.rs:227`) infers `default_engine` only from `LEMONADE_PORT`/`VLLM_PORT`; a `:11435` endpoint would fall into the `else` → `"local"`, even though `rocm serve` *is* vLLM. This is **not** on this PR's path, and `auto_config_plan`/`auto_config_change` currently have no runtime caller (tests only), so it's dormant — not a regression. Since this PR establishes "`:11435` = vLLM", adding an `ROCM_SERVE_PORT` arm there would prevent the two mappings from drifting when that skill is eventually wired up. Fine to defer.

**LOW — test isolation.** `detect_local_endpoint_probes_rocm_serve_default_port` skips if `:11435` is already bound, but doesn't guard against `:13305`/`:8000` being occupied on the test host. Since the probe returns the *first* reachable endpoint in order, a stray listener on those ports would make the `assert_eq!` fail rather than skip. Low real-world risk in CI; the skip-on-bind instinct is right.

PR #92, opened 2026-07-09
feat(dash): scrollbars on scrolling panels, with click-and-drag

michaelroy-amd:
## What

Panels gave no visual sign when their content overflowed, and there was no way to grab a scrollbar with the mouse.

- Reusable `panel::vertical_scrollbar` / `horizontal_scrollbar` helpers: no-op when content fits, otherwise draw a bar and return the shrunk content rect.
- Wired into the job console (vertical + horizontal), the LOGS dock, bench detail, chat transcript, and all seven list managers.
- **Click-and-drag:** each drawn bar records a per-frame hit-test handle; a click on a track jumps to that position and arms a drag that keeps tracking the pointer until release. The tail-anchored LOGS dock offset is inverted so dragging down goes toward older lines.
- Also clamps the job console's horizontal offset against the widest line.

Excluded on purpose: the list managers' bars are drag-to-jump-selection would be awkward (they're driven by `ListState` selection), so those show a bar but aren't draggable. Easy follow-on.

## Tests

- `panel::tests`: no-op-when-fits, shrink-on-overflow (V and H).
- `app::tests`: proportional position mapping, full grab → drag → release cycle, dock inversion.

## Notes

Part 2 of 3. **Stacked on #91** (both touch the `draw_job_console` signature / call site) — please merge #91 first, then this retargets to `main`.


rominf (approved):
LGTM. Shared scrollbar helper (no duplicated math), correct clamping (`saturating_sub`; div-by-zero guarded by `max_pos == 0` / `span <= 1` early returns), proportional offset math, no panics. Verified drag-when-bar-disappears, DockLogs tail-anchored inversion, and last-track-cell → `max_pos` mapping against the included unit tests.

Two minor notes:
1. Horizontal drag clamps to `content_w - vbody.width` while the keyboard allows one extra column (`content_w - 1`) — a 1-cell inconsistency in max horizontal offset worth aligning.
2. The manager-list scrollbars (engine/config/runtime/services/update/automations, folder_browser) render but aren't hit-tested (no `record_scrollbar`) and track `selected` rather than the scroll offset — looks intentional for this PR's scope, but the mix of grabbable/non-grabbable bars is a UX rough edge worth a note in the description.

Approving.

volen-silo:
🔴 Automated review · pr-review-watcher · 2162f28c

## Summary
Adds reusable vertical/horizontal scrollbar helpers and wires them into the job console, LOGS dock, bench detail, chat transcript, and the seven list managers, with click-and-drag hit-testing for the first four. Verdict: **Approve** — this is a clean, well-tested change. Verified: `cargo build`/`test`(571 pass, `--test-threads=1`)/`clippy -D warnings`/`fmt --check` all clean; confirmed the drag position math (`position_at` proportional map + `DockLogs` tail inversion), the per-frame RefCell registry lifecycle (cleared+rebuilt every `draw`/`draw_focused`, mutually exclusive, no stale handles or borrow conflicts), and that the two initially-suspected "blocking" issues de-verify to pre-existing behavior / non-crashing visual imprecision. Blocking: 0 · Non-blocking: 4.

**None.**

## Non-blocking
- `ui/tabs/chat.rs:240`, `ui/tabs/bench.rs:458` — scrollbar is told `content_len = lines.len()` (pre-wrap) while the `Paragraph` renders with `.wrap(Wrap{trim:false})`, so when lines wrap the thumb under-reports extent (may not appear on real overflow) and chat drag maps to an imprecise offset; either measure wrapped rows for `content_len` or drop `.wrap()`. (Inherits a pre-existing scroll imprecision, now made visible.)
- 7 list managers (e.g. `ui/config_manager.rs:272`) — these bars are draw-only (no `record_scrollbar`/`ScrollTarget`), so the PR summary bullet reading as "click-and-drag" for them is slightly ambiguous vs the body's noted exclusion; worth tightening the wording.
- 7 list managers — pass `selected` (selection index) as the bar `position`, so the thumb tracks the cursor rather than the viewport top and sits mid-track when it should be near an edge; cosmetic only since these bars aren't hit-tested.
- Test gaps: no unit test for the `ScrollbarHandle::new` no-op branch (`drawn == area → None`, guards against a phantom clickable region), and no drag/release test for the horizontal `ConsoleH` grab path (only vertical targets are exercised end-to-end).

PR #93, opened 2026-07-09
feat(dash): make the ROCm/Serving detail pane state-aware

michaelroy-amd:
## What

The Actions → Details pane on the ROCm and Serving tabs rendered fixed copy regardless of system state. Thread `&AppState` into `pane::draw` / `draw_detail` and render a live status block per verb:

- **Set up / Install ROCm** — detected ROCm + driver versions and installed runtimes.
- **Serve a model** — the known-model catalog.
- **Engines** — Lemonade install status (vLLM deferred until the daemon surfaces detection).
- **Running instances** — live instances inline (model / port / gen tps), **no popup**.

Falls back to honest `not detected` / `none running` when telemetry is absent, so it degrades correctly on machines without ROCm.

## Tests

- Tab tests for the populated install (detected ROCm version) and running-instances (inline, no modal) paths.

## Notes

Part 3 of 3 for today's dash UX work. Independent of #91 / #92 (disjoint files) — mergeable in any order.

Follow-on (not in this PR): real downloaded-model detection and vLLM install detection need a new background fetch + `AppState` fields.

### DIFF
--- a/crates/rocm-dash-tui/src/ui/tabs/pane.rs
+++ b/crates/rocm-dash-tui/src/ui/tabs/pane.rs
@@ -20,7 +20,10 @@ use ratatui::style::{Modifier, Style};
 use ratatui::text::{Line, Span};
 use ratatui::widgets::{Paragraph, Wrap};
 
-use crate::app::{KeyAction, PaneFocus};
+use rocm_dash_core::metrics::InstanceStatus;
+
+use crate::app::{AppState, KeyAction, PaneFocus};
+use crate::ui::format;
 use crate::ui::panel::{self, BoxRole};
 use crate::ui::theme::Theme;
 
@@ -82,12 +85,13 @@ pub fn draw(
     list_title: &str,
     verbs: &[Verb],
     sel: usize,
-    focus: PaneFocus,
+    state: &AppState,
     theme: &Theme,
 ) {
+    let focus = state.pane_focus;
     let cols = split_columns(area);
     draw_verb_list(f, cols[0], list_title, verbs, sel, focus, theme);
-    draw_detail(f, cols[1], verbs, sel, focus, theme);
+    draw_detail(f, cols[1], verbs, sel, focus, state, theme);
 }
 
 /// Map a left-click in a domain tab's body to an action.
@@ -196,6 +200,7 @@ fn draw_detail(
     verbs: &[Verb],
     sel: usize,
     focus: PaneFocus,
+    state: &AppState,
     theme: &Theme,
 ) {
     let sel = sel.min(verbs.len().saturating_sub(1));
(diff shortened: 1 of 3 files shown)


rominf (approved):
LGTM. The state→verb mapping is correct and the `_ => Vec::new()` catch-all is appropriate (documented; most `KeyAction` variants have no live view), no panics, lifetimes are fine (`Vec<Line<'static>>` coerces cleanly), both callers (rocm.rs, serving.rs) are updated, and regression tests are added.

Minor (pre-existing, not introduced here): the `OpenServices` live view iterates `state.instances.values().take(5)`, so which 5 of N running instances appear is non-deterministic frame-to-frame — a stable sort by model/port would be a nice follow-up. The count header is always correct.

Approving.

volen-silo:
🔴 Automated review · pr-review-watcher · 56b57ab5

## Summary
Threads `&AppState` into the shared ROCm/Serving `pane::draw` and renders a live status block per verb (installed ROCm/driver + runtimes, engine install status, model catalog, running instances inline) with honest `not detected` / `none running` fallbacks. Verdict: **Approve** — nits only. Verified: `cargo build -p rocm-dash-tui` clean and `cargo test -p rocm-dash-tui --lib -- --test-threads=1` green (545 passed); confirmed all field/type references (`GpuSystemInfo.{rocm,driver,lemond}_version`, `Instance.{model_name,port,gen_tps,status}`, `RuntimeSummary`, `ModelRecipeSummary.id`, `format::tps_opt`) resolve, the `live_lines` match arms are correct with `_ => Vec::new()` covering the non-live verbs, the `.take(4/5)` + "…and N more" arithmetic is correct, both `pane::draw` call sites (rocm.rs, serving.rs) are updated with no missed callers, and the two new tests' selection indices map to the intended verbs. Blocking: 0 · Non-blocking: 4.

**None.**

## Non-blocking
- `pane.rs` (`OpenServices` arm) — running instances come from `state.instances.values()` unsorted, then `.take(5)`; `HashMap` order is nondeterministic, so with >5 running instances *which* 5 (and their order) can shuffle frame-to-frame. The codebase already solved this with `tabs/instances.rs::sorted_instances()` (used in 4 places) — reuse it / sort by `container_name` before `.take(5)`.
- `draw_detail` (pane.rs) — the detail `Paragraph` has no `.scroll()` and no height check; the added live lines (up to ~8 before "What you'll do") make it materially more likely the Start affordance / trailing steps clip off-screen on short terminals, with no scrollbar or "more below" cue. Consider capping the live block or making the pane scroll-aware.
- Test gaps — the `OpenEngineManager` and `OpenServeWizard` live blocks, the `not detected` / `none running` fallbacks, and the "…and N more" truncation are all untested; and `serving.rs:206` `out.contains("8000")` is a weak assertion (matches a bare digit run, not the rendered `:8000`).
- `pane.rs:91` — `draw` passes both `focus` and `state` into `draw_detail` even though `focus == state.pane_focus`; drop the redundant param.

PR #94, opened 2026-07-09
Reuse system-installed ROCm in Lemonade; drop channel mirroring

volen-silo:
## What

Lemonade now detects and reuses the ROCm that rocm-cli already installs, instead of downloading its own runtime. Two changes:

1. **Drop channel-mirroring.** rocm-cli used to run `lemonade config set rocm_channel=<stable|nightly>` + `llamacpp.rocm_bin=builtin` at engine-install time to nudge Lemonade toward the matching channel — a workaround that still let Lemonade download a second ROCm runtime. Lemonade now honors the injected `ROCM_PATH` / `rocm-sdk` (already exported into its process environment), so the mirroring is removed. `install_response` no longer mutates Lemonade config.

2. **Bump bundled embeddable 10.6.0 → 10.10.0.** 10.10.0 is the first release containing the system-ROCm detection. Removing the mirroring without this bump would regress nightly users on 10.6.0, which cannot reuse system ROCm.

3. **Resolve the direct-serve backend dynamically.** The Linux direct-serve fast path hardcoded `bin/llamacpp/rocm-stable`, so it only engaged for a stable build and silently fell through otherwise (nightly-only or vulkan-only embeddables). All three `serve_http` sites now use the existing `find_llama_server_binary()` (priority `rocm-stable → rocm-nightly → rocm → vulkan`, CPU excluded), matching how `serve_hf_checkpoint` already resolves it. The unused `direct_llama_server_path` helper is removed.

## Testing

- `cargo test -p rocm-engine-lemonade` — 26 pass (added `find_llama_server_binary_follows_backend_priority` covering the nightly-only and stable-precedence cases)
- `cargo clippy -p rocm-engine-lemonade --all-targets -- -D warnings` — clean
- `cargo build --workspace` — clean
- Manual smoke on hardware: nightly + release ROCm install, `engines install lemonade` (confirms 10.10.0, no channel-set line, reuses system ROCm), serve + chat.

## Behavior notes

- On hosts where ROCm GPU compute is unavailable (e.g. WSL2), the direct-serve fast path now engages on `vulkan` where before it fell through to the router path. Same end backend, earlier path.

### DIFF
--- a/engines/lemonade/src/lib.rs
+++ b/engines/lemonade/src/lib.rs
@@ -530,16 +530,18 @@ fn serve_http(request: ServeHttpRequest) -> Result<()> {
     if let Some(checkpoint) = parse_hf_checkpoint(&request.model_ref) {
         return serve_hf_checkpoint(&request, &runtime, &process_env, log_path, &checkpoint);
     }
-    if runtime_is_linux() && direct_llama_server_path(&runtime.manifest).is_file() {
+    if runtime_is_linux()
+        && let Some(server) = find_llama_server_binary(&runtime.manifest)
+    {
         ensure_direct_llama_model_available(&request, &runtime, &process_env, log_path)?;
-        let server = direct_llama_server_path(&runtime.manifest);
+        let backend = llama_server_backend_label(&server);
         return serve_direct_llama_server(
             &request,
             &runtime,
             &process_env,
             &server,
             log_path,
-            &anyhow!("using Lemonade packaged ROCm llama-server directly on Linux"),
+            &anyhow!("using Lemonade packaged {backend} llama-server directly on Linux"),
         );
     }
     let mut child = spawn_lemond(
@@ -587,8 +589,9 @@ fn serve_http(request: ServeHttpRequest) -> Result<()> {
         && query_chat_smoke_endpoint(&request.host, request.port, &request.model_ref)
             .unwrap_or(false);
     if let Err(error) = load_result {
-        let direct_server = direct_llama_server_path(&runtime.manifest);
-        if runtime_is_linux() && direct_server.is_file() {
+        if runtime_is_linux()
+            && let Some(direct_server) = find_llama_server_binary(&runtime.manifest)
+        {
             let _ = terminate_pid(child.id(), true);
             let _ = child.wait();
             return serve_direct_llama_server(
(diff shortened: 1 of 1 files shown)


volen-silo:
🔴 Automated review · pr-review-watcher · 9d05f0c6

## Summary
Reuses system-installed ROCm in Lemonade (drops channel-mirroring), bumps the embeddable 10.6.0 → 10.10.0, and resolves the Linux direct-serve backend dynamically via `find_llama_server_binary()` instead of a hardcoded `rocm-stable` path. **Verdict: Approve.** Verified: `cargo test -p rocm-engine-lemonade` 26/26 pass, `cargo clippy -p rocm-engine-lemonade --all-targets -D warnings` clean, leak scan clean; confirmed the `active_managed_therock_channel` import is fully removed with no dead/orphaned symbols left behind, `RocmCliConfig` is still used (l.1440), all 5 version constants in this file are consistently bumped, GPU-required is preserved (CPU excluded from `DIRECT_LLAMA_SERVER_BACKENDS`, `require_gpu_required` gates before the direct path), the rocm-stable case resolves byte-identically to before, and the re-lookup at the two fallback sites is intentional (a backend can be installed by the router path in between). Blocking: 0 · Non-blocking: 3.

**None.**

## Non-blocking
- `apps/rocm/src/main.rs:3303` — `managed_engine_runtime_id` still returns the stale literal `"lemonade-embeddable-10.6.0"` (asserted at `main.rs:20516`); inert today because the Lemonade engine derives identity from the on-disk manifest `env_id` (`lemonade-embeddable-{LEMONADE_VERSION}`) and ignores the incoming `runtime_id`, but it will silently version-skew on every bump — out of this single-file PR's scope, best fixed by exporting `LEMONADE_VERSION` as `pub` and referencing it from both sites.
- `engines/lemonade/src/lib.rs:817` — `ensure_direct_llama_model_available` hardcodes `ROCM_BACKEND_NAME` ("rocm") as the router hint even when the resolved backend is `vulkan`/`rocm-nightly`; harmless (result is discarded, the real gate is file-existence) but newly reachable on vulkan-only hosts via this PR's fast path — pass the resolved backend for clarity.
- `engines/lemonade/src/lib.rs:2793` — new test covers `find_llama_server_binary` priority/CPU-exclusion well, but the three `serve_http` call sites that consume it are untested (acceptable — no process/HTTP harness exists); a vulkan-only winning-case assertion would close the loop on the PR's stated WSL2 motivation.

EugeneKSiloAi:
@volen-silo things lgtm but there are still outstanding copilot comments that could probably be reviewed, if you've already reviewed them let me know before closing.

EugeneKSiloAi (approved):

clydehoang1 (approved):

PR #96, opened 2026-07-10
fix(serve): pass vLLM tool-choice flags so TUI chat works

volen-silo:
## Summary
- `rocm serve` now launches vLLM with `--enable-auto-tool-choice` and a model-matched `--tool-call-parser`, so the TUI chat tab works against vLLM-backed models out of the box.
- Flags are injected into the vLLM engine recipe hint at serve time, independent of catalog coverage: existing recipe hints are extended (recipe-authored flags win, no duplication), and a minimal hint is synthesized when none exists — so arbitrary HF repos and catalog models forced onto a non-preferred engine are covered too.
- Parser is chosen per model family (`llama` → `llama3_json`, `mistral`/`mixtral` → `mistral`), defaulting to `hermes` for the Qwen-heavy catalog.

## Why
The TUI chat tab attaches tool definitions to every non-streaming chat request (`tool_choice: "auto"`). vLLM rejects those with HTTP 400 unless it was started with `--enable-auto-tool-choice` **and** a matching `--tool-call-parser`:

```
chat request failed 400 tool choice requires --enable-auto-tool-choice and --tool-call-parser to be set
```

## Root cause
The launch plumbing to forward `EngineRecipeHint.required_flags` into the `vllm serve` command already existed and worked — but nothing ever populated those flags. No catalog recipe set them, and the recipe-derived hint is gated by `model_recipe_supports_engine()`, which drops to `None` for a model whose preferred engine isn't the one being served (exactly the reported repro: `Qwen/Qwen2.5-1.5B-Instruct` forced with `--engine vllm`). So a catalog-only fix wouldn't have fixed the reported case; the injection has to happen in `serve()` regardless of recipe presence.

## Non-obvious decisions
- **Default parser `hermes`.** A mismatched-but-valid parser only degrades tool-call *parsing* for that model — it never 400s and never breaks plain chat — so a default is strictly safer than the status quo where every chat request failed. A specific model can still override via its recipe's `required_flags` (recipe wins).
- **Injected for all vLLM serves, not gated on `task == chat`.** vLLM's OpenAI server is chat/completions regardless, and the flags are inert when no tools are sent.
- Scope: no change to the TUI/providers tool-sending behavior, and no `model_catalog.json` edits — the central injection supersedes per-recipe authoring.

## Test plan
- New unit tests: parser selection by model family, hint synthesis when no recipe exists, appending to existing recipe flags, preserving recipe-authored tool-choice (no dup), and non-vLLM engines left untouched.
- `cargo test -p rocm` (new tests + existing serve/recipe/provider suites) and `cargo clippy` pass.
- Full end-to-end confirmation (TUI chat against a live vLLM server) requires a ROCm GPU and is deferred to that hardware.

## Risk
Low — additive; non-vLLM engines and recipe-authored flags are untouched, and the flags are inert for requests that don't send tools.

### DIFF
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -3845,6 +3845,70 @@ fn protocol_engine_recipe_hint(
         })
 }
 
+/// The vLLM launch flags required for OpenAI tool-calling to work.
+///
+/// The TUI chat tab always attaches tool definitions to non-streaming chat
+/// requests (`tool_choice: "auto"`). vLLM rejects those with HTTP 400 unless it
+/// was started with `--enable-auto-tool-choice` *and* a matching
+/// `--tool-call-parser`. The parser value is model-specific, so it is derived
+/// from the model ref. A mismatched-but-valid parser only degrades tool-call
+/// parsing for that model (chat still works and never 400s), so `hermes` — the
+/// parser used by the Qwen family that dominates the built-in catalog — is a safe
+/// default for unrecognized families.
+fn vllm_tool_choice_flags(model_ref: &str) -> Vec<String> {
+    let lower = model_ref.to_ascii_lowercase();
+    let parser = if lower.contains("llama") {
+        "llama3_json"
+    } else if lower.contains("mistral") || lower.contains("mixtral") {
+        "mistral"
+    } else {
+        // Qwen and every other family fall back to the hermes parser.
+        "hermes"
+    };
+    vec![
+        "--enable-auto-tool-choice".to_owned(),
+        "--tool-call-parser".to_owned(),
+        parser.to_owned(),
+    ]
+}
+
+/// Ensures a vLLM serve carries the tool-choice launch flags so TUI chat works
+/// out of the box (EAI-7223).
+///
+/// Only vLLM needs this — other engines either accept tools without extra flags
+/// or are not affected. Recipe-authored flags win: if the hint already requests
+/// `--enable-auto-tool-choice`, it is left untouched. When no recipe hint exists
+/// (arbitrary HF repos, or a catalog model forced onto a non-preferred engine),
+/// a minimal hint is synthesized carrying just the tool-choice flags.
+fn engine_recipe_with_tool_choice(
+    engine: &str,
+    model_ref: &str,
+    hint: Option<EngineRecipeHint>,
+) -> Option<EngineRecipeHint> {
+    if !engine.eq_ignore_ascii_case("vllm") {
+        return hint;
+    }
+    match hint {
+        Some(mut hint) => {
+            let already_set = hint
+                .required_flags
+                .iter()
+                .any(|flag| flag == "--enable-auto-tool-choice");
+            if !already_set {
+                hint.required_flags
+                    .extend(vllm_tool_choice_flags(model_ref));
+            }
+            Some(hint)
+        }
+        None => Some(EngineRecipeHint {
+            contract_version: ENGINE_RECIPE_CONTRACT_VERSION.to_owned(),
+            engine: engine.to_owned(),
+            required_flags: vllm_tool_choice_flags(model_ref),
+            ..EngineRecipeHint::default()
+        }),
+    }
+}
+
 /// Parsed `rocm serve` arguments. Grouped into a struct to keep the dispatcher
 /// and `serve()` readable now that the verb carries verbose/smoke-test controls.
 struct ServeArgs {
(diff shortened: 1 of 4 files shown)


volen-silo:
🔴 Automated review · pr-review-watcher · 628cb2e4

## Summary
Injects vLLM `--enable-auto-tool-choice` + a model-matched `--tool-call-parser` into the engine recipe hint at `serve()` time so the TUI chat tab stops getting HTTP 400s — a correct, well-targeted fix for the stated root cause (recipe-gated hint drops to `None` for non-preferred engines). Verdict: **Needs work** (one blocking policy violation; core logic is sound). Verified: `cargo test -p rocm --bin rocm` builds clean and all 5 new tests pass; traced `required_flags` end-to-end into `engines/vllm/src/lib.rs` (`engine_recipe_launch_args` → raw argv append) and confirmed the dedup, `Default`-synthesized hint, and non-vLLM passthrough behave as the tests assert. Blocking: 1 · Non-blocking: 3.

`apps/rocm/src/main.rs:3690` and `:3796` — the doc comment and inline comment both embed the internal ticket ID `EAI-7223`. The repo's own `AGENTS.md` §2 (lines 58–59) explicitly forbids "internal ticket identifiers ... in code comments" on upstream surfaces, and applies the rule to code comments specifically. This is a new violation — `EAI-7223` appears nowhere else in the repo, so it is not an established pattern. A project-instruction violation introduced by the PR is blocking. Fix: drop `(EAI-7223)` from both comments and keep the surrounding descriptive text.

## Non-blocking
- `apps/rocm/src/main.rs:3708` — dedup only skips injection when `--enable-auto-tool-choice` is already present; a recipe that authored `--tool-call-parser <x>` *alone* would get a second `--tool-call-parser` appended, which vLLM argparse resolves to the last (auto-derived) value — silently overriding the recipe author and inverting the docstring's "recipe-authored flags win" claim. Latent today (the sole vLLM catalog recipe has empty `required_flags`); harden by also checking for `--tool-call-parser` presence.
- `apps/rocm/src/main.rs:19982` — `vllm_tool_choice_flags_selects_parser_by_model_family` never exercises the `mistral`/`mixtral` → `mistral` branch; add a `mistralai/Mistral-7B-Instruct` case so a regression there is caught.
- `apps/rocm/src/main.rs:3672` — parser family detection is a full-string substring match, so an incidental `llama`/`mistral` in an org/repo path (e.g. a personal HF namespace) is misclassified; matching the leading `model_ref.split('/').next()` segment would reduce false positives. Documented graceful-degradation tradeoff (wrong parser degrades tool parsing, never 400s), so optional.

apps/rocm/src/main.rs
```diff
@@ -3659,6 +3659,70 @@ fn protocol_engine_recipe_hint(
...
+///
+/// The TUI chat tab always attaches tool definitions to non-streaming chat
+/// requests (`tool_choice: "auto"`). vLLM rejects those with HTTP 400 unless it
+/// was started with `--enable-auto-tool-choice` *and* a matching
+/// `--tool-call-parser`. The parser value is model-specific, so it is derived
+/// from the model ref. A mismatched-but-valid parser only degrades tool-call
+/// parsing for that model (chat still works and never 400s), so `hermes` — the
+/// parser used by the Qwen family that dominates the built-in catalog — is a safe
+/// default for unrecognized families.
+fn vllm_tool_choice_flags(model_ref: &str) -> Vec<String> {
```
rominf: **Blocking:** this heuristic assigns incompatible parsers to valid models. For example, catalog entries from GLM and Gemma families fall through to `hermes`, while vLLM has dedicated parsers; Llama variants also do not all share `llama3_json`. A wrong parser can silently lose structured tool calls or fail during request parsing, so successful startup does not establish that TUI tool use works. Please make parser compatibility explicit model metadata rather than infer it from a free-text ref. The vLLM project already publishes model-specific `features.tool_calling.args` in https://github.com/vllm-project/recipes and via https://recipes.vllm.ai/<org>/<model>.json. Catalog/recipe metadata should supply the flags, an explicit user override should handle arbitrary refs, and unknown models should not guess. This also avoids the current merge path appending a second parser when a recipe already supplies only `--tool-call-parser`.

rominf (dismissed):
The HTTP 400 needs fixing, but selecting a parser from free-form model-name substrings is not safe. vLLM requires an explicit parser and does not auto-detect one. The vLLM project maintains model-specific launch arguments in https://github.com/vllm-project/recipes (also exposed as machine-consumable JSON at https://recipes.vllm.ai/). Please source tool-calling flags from explicit per-model recipe metadata and provide an explicit override for arbitrary model references; unknown models should not receive a guessed parser. Follow-up engine-truth sourcing is tracked in EAI-7204.

michaelroy-amd (approved):
Re-reviewed the current head and prior requested changes. Free-form model-name parser guessing is gone; explicit catalog metadata and the CLI override select parsers, unknown models are not guessed, duplicate parser flags are prevented, and required flags reach the vLLM argv. Focused tests and CI are green.

PR #97, opened 2026-07-10
fix(dash): discover served model for configured chat endpoint

volen-silo:
## Summary
- The dash TUI chat now queries `/v1/models` at startup to discover the served model name when a chat endpoint is configured via `OPENAI_BASE_URL` / `--chat-url` but no model is set.
- **Why:** the documented `rocm serve … && OPENAI_BASE_URL=… rocm` flow was broken — sending a message returned `404 Not Found: The model 'local-model' does not exist`, making chat unusable out of the box.
- **Root cause:** the managed-service and in-TUI detect-offer paths already discover the served model, but the startup env/CLI path fell straight through to `resolve_llm_config`'s `DEFAULT_CHAT_MODEL = "local-model"` placeholder and built the agent with it. Accepting the consent prompt doesn't rebuild, so the placeholder reached the first request.
- **Fix:** after resolving the config, when the endpoint is reachable and no explicit model was given, fetch `/v1/models` and adopt the served id (via the existing `fetch_first_model` / `pick_first_model` helpers), mirroring the managed-service path.

## Non-obvious decisions
- Gated on a successful TCP probe (`probe_ok`) so an unreachable configured endpoint never adds the 3s fetch timeout to startup.
- An explicit `--chat-model` / config value always wins — discovery only fills the placeholder (enforced by the pure `with_discovered_model` helper).
- Kept the `local-model` fallback when `/v1/models` can't be read — some endpoints ignore the model field or don't expose the listing.

## Test plan
- `cargo test -p rocm-dash-tui --lib` — 549 passed (3 new unit tests for `with_discovered_model`: placeholder replaced, explicit model preserved, failed discovery keeps fallback).
- `cargo clippy -p rocm-dash-tui --all-targets` — clean.
- Manual (deferred, needs a live engine): `rocm serve Qwen/Qwen2.5-1.5B-Instruct --engine vllm`, then `OPENAI_BASE_URL=http://127.0.0.1:11435/v1 rocm`, accept the endpoint in the Chat tab, send a message → expect a normal reply and no `local-model` 404.

## Risk
Low — additive best-effort discovery on a single startup path; existing precedence and the neutral fallback are preserved.


volen-silo:
🔴 Automated review · pr-review-watcher · f8c2da16

## Summary
Fixes the broken `rocm serve … && OPENAI_BASE_URL=… rocm` chat flow: when a chat endpoint is configured but no model is set, the startup path now queries `/v1/models` and adopts the served id instead of shipping the `local-model` placeholder that 404s. Verdict: **Approve**. Verified: `cargo test -p rocm-dash-tui --lib` passes (546 passed, 0 failed — incl. the 3 new `with_discovered_model` tests) and `cargo clippy -p rocm-dash-tui --all-targets` is clean on Rust 1.96/edition 2024; confirmed the discovery gate (`!had_managed && probe_ok && args.chat_model.is_none()`) never clobbers an explicit CLI/config model (`args.chat_model` derives from `tui.chat_model`, so a config-file model correctly suppresses it), never fires on a managed endpoint (which already discovers its own model), and safely no-ops when the resolved config is `None` (no-key OAuth path untouched). Blocking: 0 · Non-blocking: 3.

**None.**

## Non-blocking
- `crates/rocm-dash-tui/src/app/mod.rs:1606-1612` — the `explicit_model` arg to `with_discovered_model` is always `None` at this sole call site (the match guard already proved `args.chat_model.is_none()`), so the `explicit_model.is_none()` check is dead here; harmless and still useful as a pure unit-test seam, but could be simplified.
- `crates/rocm-dash-tui/src/app/chat.rs:306` — `probe_ok` is a 300ms TCP-only probe, so a reachable-but-slow endpoint can still stall the first frame up to the 3s `fetch_first_model` timeout; the PR description's "never adds the 3s fetch timeout" only holds for the unreachable case. Pre-existing/mirrored from the managed and local-detect paths (which fetch unconditionally), so strictly narrower than existing behavior — noting, not a regression.
- `crates/rocm-dash-tui/src/llm.rs:211` — discovery silently binds `data[0].id` with no user-visible signal of which model was picked; a one-line `tracing` log of the adopted id would aid debuggability when a multi-model server lists an unexpected entry first.

crates/rocm-dash-tui/src/app/mod.rs
```diff
@@ -1626,6 +1626,20 @@ async fn event_loop(terminal: &mut Tui, args: &ResolvedArgs) -> color_eyre::Resu
...
             )
         });
+        // PR #97 port onto PR #100's startup flow: a *configured* URL (CLI/env)
+        // with no explicit model resolves to the `local-model` placeholder,
+        // which 404s on servers that register the model under its real id. Only
+        // the `Configured` outcome needs this — the `Local` outcome already
+        // carries a `/v1/models`-discovered model from `detect_local_chat`, and
+        // `OAuth` has no config. Discovery is gated inside the helper on
+        // `probe_ok` (an unreachable endpoint is never probed nor replaced) and
+        // on the absence of an explicit model (config precedence wins).
```
EugeneKSiloAi, line 1636: This is quite verbose, would suggest a small cleanup to keep comments at a reasonable size.

EugeneKSiloAi:
@volen-silo lgtm once the tests and any other ci failures are taken care of. Approved.

EugeneKSiloAi (approved):

PR #102, opened 2026-07-11
feat(chat): add /detect slash command for mid-session re-detection

michaelroy-amd:
## Summary

- Adds a `/detect [accept|save|dismiss]` slash command so a local chat
  engine can be re-detected after `ChatConsent::Accepted`, without a
  dedicated keybinding colliding with focused chat text entry.
- Bare `/detect` re-runs the same probe used at startup
  (`detect_local_chat`: managed-services registry, then Lemonade
  :13305 / vLLM :8000 / rocm serve :11435) and echoes the result into
  the chat transcript.
- `/detect accept` / `/detect save` / `/detect dismiss` route through
  the existing `accept_detect_offer` / `save_detect_offer` /
  `dismiss_detect_offer` reducer methods.

## Root cause

Once chat is `ChatConsent::Accepted`, the pre-accept `'d'` gate key is
unreachable: while a text field is focused a bare `d` keypress is
ordinary chat input, and when unfocused the global hotkey table does
not bind `'d'` for the Chat tab. There was no way to react to a local
engine coming online (or changing) after the initial handshake short
of restarting the app.

## Fix

- `slash.rs`: new `"detect"` branch dispatches to the pre-existing
  `AppState` reducer methods used by the pre-accept offer UI, so no
  new detection or endpoint-swap plumbing was introduced.
- `app/mod.rs`:
  - `request_detect()` now also echoes a "Detecting…" message into the
    transcript when already `Accepted` (previously this messaging only
    made sense pre-accept, driving the gate UI).
  - `set_detect_result()` echoes the found/not-found result into the
    transcript when `Accepted`, since `draw()` only renders the
    offer y/s/n prompt pre-accept (confirmed in `ui/tabs/chat.rs`) —
    without this the probe result would be silently discarded once
    accepted.
- Accepting or saving a re-detected endpoint via `/detect accept` /
  `/detect save` goes through the same `accept_detect_offer` /
  `save_detect_offer` calls the initial pre-accept flow uses, so it
  raises the `chat_endpoint_rebuild` dispatch edge (introduced in
  PR #81) exactly like the initial accept does, swapping the live
  agent to the newly detected endpoint.

## Coordination notes

- This touches key/slash routing (`app/mod.rs`, `slash.rs`), a
  different surface than PR #100 (`fix/chat-startup-local-first`,
  which only touches the `event_loop()` startup block) and than
  unmerged upstream PR #97 (which touches `resolve_llm_config`'s
  fallback and the startup block). No overlapping hunks expected.

Relates to EAI-7354.

## Test Plan

- [x] `cargo clippy -p rocm-dash-tui --all-targets --all-features` (clean, `-D warnings`)
- [x] `cargo test -p rocm-dash-tui -- --test-threads=1` (551 passed, 0 failed)
- [x] New tests:
  - `slash_detect_probes_and_echoes_result_while_accepted`
  - `slash_detect_accept_raises_endpoint_rebuild_like_initial_accept`
  - `slash_detect_dismiss_and_unknown_subcommand`

### DIFF
--- a/crates/rocm-dash-tui/src/app/mod.rs
+++ b/crates/rocm-dash-tui/src/app/mod.rs
@@ -4225,6 +4225,41 @@ mod tests {
         assert!(last.content.contains("unknown /detect action"));
     }
 
+    /// Sub-commands are case-insensitive, matching `/permissions` / `/provider`.
+    #[test]
+    fn slash_detect_subcommand_is_case_insensitive() {
+        let mut s = AppState::new("t".into(), "default-dark".into());
+        s.set_detect_result(Some(crate::llm::detected_llm_config(
+            "http://localhost:8000/v1",
+            "m",
+        )));
+        // `DISMISS` (upper) must act exactly like `dismiss`.
+        assert_eq!(
+            s.handle_slash_command("/detect DISMISS"),
+            SlashOutcome::Handled
+        );
+        assert!(
+            s.chat_detect_offer.is_none(),
+            "uppercase sub-command is normalized and handled"
+        );
+    }
+
+    /// `/detect accept` / `/detect save` with nothing pending emits a hint
+    /// turn rather than a silent no-op.
+    #[test]
+    fn slash_detect_accept_without_offer_emits_hint() {
+        let mut s = AppState::new("t".into(), "default-dark".into());
+        assert!(s.chat_detect_offer.is_none());
+        assert_eq!(
+            s.handle_slash_command("/detect accept"),
+            SlashOutcome::Handled
+        );
+        // No endpoint was adopted; a hint explains why.
+        assert_eq!(s.chat_endpoint_rebuild, None);
+        let last = s.chat.last().expect("hint turn");
+        assert!(last.content.contains("no detected endpoint"));
+    }
+
     #[test]
     fn detect_key_available_on_gate_and_offer_keys_take_precedence() {
         // `d` triggers detect from the Unavailable empty-state.
--- a/crates/rocm-dash-tui/src/app/mod.rs
+++ b/crates/rocm-dash-tui/src/app/mod.rs
@@ -1002,6 +1002,13 @@ impl AppState {
     /// `chat_detect_dispatch` edge so `event_loop` runs the probe + `/v1/models`
     /// query off the reducer. No-op while a probe is already in flight or an
     /// offer is awaiting a decision. I/O-free.
+    ///
+    /// Reachable both from the pre-accept gate (`'d'` key) and, once
+    /// `ChatConsent::Accepted`, from the `/detect` slash command (a focused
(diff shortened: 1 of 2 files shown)


rominf (approved):
LGTM. `/detect` dispatch/parsing mirrors sibling slash commands (case-insensitive sub-command via `split_whitespace().nth(1)`, whitespace-robust), and it routes straight through the existing `accept_detect_offer` / `save_detect_offer` / `dismiss_detect_offer` reducers rather than reimplementing detection. `/detect accept` raises `chat_endpoint_rebuild` identically to the initial pre-accept accept (asserted by test). The feature is correctly scoped to post-`Accepted` state.

Minor UX nit: a bare `/detect` while an offer is already pending is a silent no-op (early return on `chat_detect_offer.is_some()`) with no on-screen feedback post-accept — a one-line "an endpoint is already pending — `/detect accept|dismiss` first" hint would close it.

Approving.

volen-silo:
🔴 Automated review · pr-review-watcher · 15979fad

## Summary
Adds a `/detect [accept|save|dismiss]` slash command so a local chat engine can be re-detected mid-session once `ChatConsent::Accepted` (where the pre-accept `'d'` gate key is unreachable), routing accept/save/dismiss through the existing reducer methods so a re-detect+accept raises the `chat_endpoint_rebuild` edge exactly like the initial accept. Verdict: **Approve**. Verified: `cargo test -p rocm-dash-tui` builds clean and passes (569 tests, 0 failed; 5 new detect tests green), `cargo clippy --all-targets --all-features` clean; confirmed `/detect` is only reachable once Accepted (input is unfocusable/unsubmittable before then, so the consent-branching in `request_detect`/`set_detect_result` is correct defensive code), the offer/gate UI (`ui/tabs/chat.rs` `draw_consent`) only renders pre-accept so echoing results into the transcript is the right surface, and `/detect accept` correctly re-raises the endpoint-rebuild edge. Blocking: 0 · Non-blocking: 3.

**None.**

## Non-blocking
- `app/mod.rs:2176-2187` — `/detect save` gives no user-visible feedback once Accepted: the persist success path is `tracing::info!` only, and the failure path sets `chat_detect_msg`, which is a **dead write** post-accept (only rendered by `draw_consent`, gated to `!= Accepted` in `ui/tabs/chat.rs:114`). Consider pushing a `ChatTurn` in both the `Ok`/`Err` arms.
- `app/mod.rs:1066-1080` — `/detect accept`/`save` push no acceptance confirmation into the transcript; the endpoint silently switches. A one-line "switched to <endpoint>" echo would match the discoverability of the rest of the flow.
- `app/mod.rs:1012-1015` — a bare `/detect` while a probe is in flight or an offer is already pending is a silent no-op (early return, no hint), unlike the `/detect accept`-with-no-offer path which does emit a hint. Minor.

PR #103, opened 2026-07-11
fix(dash): wire client-side log file and instrument chat lifecycle

michaelroy-amd:
## Summary

- `tracing-subscriber` was a declared dependency of `rocm-dash-tui` and `rocm-dash-daemon`, but `tracing_subscriber::` was never called anywhere in the codebase — every `info!`/`warn!`/`debug!` call was silently dropped with no subscriber installed.
- Adds a new `apps/rocm/src/logging.rs` module that installs a process-wide, file-only, non-blocking `tracing` subscriber writing daily-rotated files under `~/.rocm/logs` (via a new `AppPaths::client_log_dir()`), with pruning to the 7 most recent files.
- Adds lifecycle tracing (request-start, elapsed/TTFT-equivalent, completion, error, timeout) to the chat request path shared by all three `AgentClient::complete()` backends (`RigAgentClient`, `ChatGptAgentClient`, `AnthropicAgentClient`) in `crates/rocm-dash-tui/src/agent.rs`, tagged with a `backend` field so a hung or failing chat is traceable to a specific provider.

## Root cause

The TUI owns the terminal in raw mode, so no subscriber could safely write to stdout/stderr — but no subscriber was ever wired up at all, meaning EAI-7357's client-side logging requirement was entirely unimplemented despite the dependency being present in `Cargo.toml`.

## Implementation notes

- The writer is strictly file-only and non-blocking (`tracing_appender::non_blocking`) so it never risks corrupting the TUI's raw-mode display.
- `logging::init()` returns `Option<WorkerGuard>`; the guard is held for the process lifetime in `main()` since dropping it flushes and stops the writer.
- A missing/failed `AppPaths::discover()` or log-dir creation degrades to no logging rather than a startup failure.
- Default filter: `warn,rocm=info,rocm_dash_tui=info,rocm_dash_daemon=info`, overridable via `RUST_LOG`.
- Log files are pruned to the 7 most recent daily-rotated files on each `init()` call, bounding disk usage for long-lived installs.

Relates to EAI-7357

## Test Plan

- [x] `cargo build` clean
- [x] `cargo clippy --all-targets --all-features -- -D warnings` clean
- [x] `cargo test -p rocm-dash-tui agent:: -- --test-threads=1` — all 28 existing agent tests pass unchanged
- [x] `cargo test -p rocm` (full suite, 3x repeated to check for flakiness) — 340/340 passed each run
- [x] New unit tests in `apps/rocm/src/logging.rs`: log-dir/file creation, idempotent re-init when a global subscriber already exists, and log pruning retains only the most recent files
---

## Follow-up / out of scope (reviewer note)
This PR wires client-side logging for the **dash/TUI client** (the `rocm` binary), which is the scope of EAI-7357. Instrumenting the standalone `rocmd` daemon binary is intentionally **not** included here and is left as a follow-up.


rominf (approved):
LGTM. The log wiring is solid on every risk axis:
- **File-only**, `with_ansi(false)` + `with_writer(non_blocking)` — the TUI raw-mode terminal is never touched.
- **Non-blocking** via `tracing_appender::non_blocking` (background worker); no blocking I/O on the render/event path.
- **Idempotent** install (`set_global_default(...).ok()`, the only subscriber in the workspace; tested).
- **Guard lifetime**: `_log_guard` bound in `main()` scope, lives for the whole process.
- **Path-validated** (canonicalize + containment under `data_dir`); missing dir degrades to no-logging, never a startup failure.
- **Growth-bounded** (`MAX_RETAINED_LOGS = 7`, prune counts today's active file, keyed on the `rocm-cli.log` prefix so the sibling daemon log is untouched).
- **No secrets**: chat instrumentation logs only `reply_len` / `elapsed_ms` / `skills_fired` / static `backend` — no keys, prompts, or reply bodies.

All three `finish_agent_request` call sites are updated for the new `backend` param; manifest/notices/lock updated consistently for the new deps.

Approving.

volen-silo:
🔴 Automated review · pr-review-watcher · 3a78b426

## Summary
Wires a file-only, non-blocking `tracing` subscriber (7-day-pruned daily logs under `~/.rocm/logs`) that was declared-but-never-installed, and adds backend-tagged chat lifecycle tracing to the shared `finish_agent_request` path. Verdict: **Approve** (no blocking code defects). Verified: `cargo build -p rocm` clean; `cargo test -p rocm logging::` 4/4 pass and `cargo test -p rocm-dash-tui agent::` 28 pass / 4 ignored (unchanged); confirmed against `tracing-appender` 0.2.5 source that `daily` opens `rocm-cli.log.<date>` (date-suffixed) eagerly, so the "prune-after-open" ordering makes `MAX_RETAINED_LOGS` the true on-disk bound (the off-by-one fix and its regression test are correct); confirmed `set_global_default` is installed exactly once in the workspace, the prune prefix filter never matches the daemon's `rocmdashd.log`, the `_log_guard` binding is held for all of `main()`, and the three backend tags match their providers. Dependency/license bookkeeping (`tracing-appender` + transitive `crossbeam-channel`/`symlink`, `ureq` clarify) is complete and reproducible. Blocking: 0 · Non-blocking: 5.

**None.**

## Non-blocking
- Commit messages (dc0d985, 30bda63, 635b986, fd513dc, 2331cfd, 3a78b42) — carry an internal ticket ID (`EAI-7357`); the code diff is clean, but the repo's own `AGENTS.md` §2 forbids internal ticket identifiers in commit-message surfaces. Scrub before/if history is rewritten; harmless to the code itself.
- `crates/rocm-dash-tui/src/agent.rs:169` — the error path logs `error = %e` to a persisted 7-day file; rig/anthropic error `Display` can embed raw HTTP response bodies. Data already appears transiently in the chat UI, so no new exposure, but it becomes durable/aggregatable on disk — consider redacting or lowering to `debug`.
- `crates/rocm-dash-tui/src/agent.rs:140-144` — the "elapsed_ms ≈ TTFT / same instant" doc framing is imprecise for the multi-turn tool path: elapsed spans all tool turns to final reply, not first token. Reword the comment.
- `apps/rocm/src/logging.rs:125` — prune filter matches any file starting with `rocm-cli.log` regardless of a valid date suffix; a stray same-prefix file in the logs dir could be deleted once retention is exceeded. Tighten the suffix check.
- `apps/rocm/src/logging.rs` — no test asserts a given client maps to its expected `backend` tag; `init_is_idempotent…` asserts a tautology given tracing's single-global-subscriber guarantee. Minor coverage gaps.

PR #105, opened 2026-07-11
fix(chat): revalidate a stale persisted chat_url and fall back to a live endpoint

michaelroy-amd:
## Summary

- Revalidates a persisted `tui.chat_url` at startup when it is unreachable and no API key or auth header is configured.
- Offers a freshly detected local engine through the existing detect-offer flow instead of silently switching, so accept, save, and dismiss all work.
- Treats `localhost` and `127.0.0.1` (plus trailing `/` and `/v1`) as the same local endpoint.

## Root cause

`tui.chat_url` survives across launches. If its server stops or moves, `resolve_llm_config` still returns the persisted URL; `probe_ok` only controls the default fallback. Startup then builds against a dead endpoint.

## Fix

- Encapsulates the complete stale-endpoint eligibility guard in `should_revalidate_stale_chat_url`.
- Detects a replacement only for an unreachable persisted URL with no configured credentials.
- Routes a different live endpoint through `AppState::set_detect_result`, preserving the existing `/detect accept|save|dismiss` reducer and persistence behavior.
- Keeps the stale configured endpoint active until the user accepts or saves the detected offer.
- Preserves an explicitly configured model on the existing endpoint; accepting a different server intentionally adopts that server's detected model.
- Corrects the `ResolvedArgs::chat_url` and `chat_model` source documentation.

## Safety and precedence

The offer requires `chat_api_key.is_none() && chat_auth_header.is_none()`, so an unreachable credentialed gateway is never replaced by a keyless local engine. `resolve_llm_config` precedence is unchanged.

## Rebase note

This branch is based on current `main`; the earlier manual overlap with PR #97 has been removed by the rebase.

Relates to EAI-7360.

## Verification

- `cargo test --workspace --all-targets` — 1745 passed, 6 ignored
- `cargo xtask e2e -- -n "A stale saved endpoint offers the live local engine"` — 1 scenario, 6 steps passed

The change adds `@id:chat-stale-endpoint-offers-live-engine` for the user-visible startup behavior.

### DIFF
--- a/crates/rocm-dash-tui/src/app/mod.rs
+++ b/crates/rocm-dash-tui/src/app/mod.rs
@@ -40,8 +40,8 @@ mod summary;
 
 use chat::{
     StartupChatOutcome, build_chat_agent, build_local_agent, detect_local_chat,
-    discover_configured_chat_model, persist_chat_endpoint, stale_chat_url_replacement,
-    startup_chat_outcome,
+    discover_configured_chat_model, persist_chat_endpoint, should_revalidate_stale_chat_url,
+    stale_chat_url_replacement, startup_chat_outcome,
 };
 use summary::{parse_plan_result, summarize_json_value, summarize_slash_tool};
 
@@ -84,7 +84,8 @@ pub struct ResolvedArgs {
     /// There is no `--chat-url` CLI flag today, so this tier is exactly the
     /// persisted config — which is what startup revalidates when it goes stale.
     pub chat_url: Option<String>,
-    /// Chat model, CLI-flag value already merged over config.
+    /// Chat model sourced from persisted config (`tui.chat_model`). There is
+    /// no `--chat-model` CLI flag today.
     pub chat_model: Option<String>,
     /// Custom auth header NAME (CLI-flag value merged over config), e.g.
     /// `Ocp-Apim-Subscription-Key` for Azure APIM gateways.
@@ -1788,44 +1789,29 @@ async fn event_loop(terminal: &mut Tui, args: &ResolvedArgs) -> color_eyre::Resu
             ),
             other => other,
         };
-        // EAI-7360: a persisted `tui.chat_url` can go stale — the server it
-        // pointed at may have been stopped or replaced since the dash last wrote
-        // it. This is the `Configured`-but-unreachable case: the reachable
-        // `Configured` case already ran `discover_configured_chat_model` above,
-        // and a `Local` outcome auto-detected its own live endpoint. Only this
-        // persisted-config tier is revalidated — `chat_env_url` is a separate
-        // tier, and `resolve_llm_config`'s CLI>config>env>probe precedence (the
-        // literal contract unit-tested in `llm.rs`) is untouched: this only
-        // substitutes a *replacement* `LlmConfig` after that call, never changes
-        // how the call itself resolves.
-        //
-        // Require NO configured api_key/auth_header: the replacement is a
-        // keyless `detected_llm_config`, and since `ResolvedArgs` can't
-        // distinguish an explicit `--chat-url` from persisted config, a remote
-        // gateway URL + `OPENAI_API_KEY` that merely TCP-times-out must never
-        // silently swap to a local keyless engine and drop the credential.
-        // Confining the swap to a no-auth `chat_url` keeps it on the true
-        // EAI-7360 target: a dead *persisted* local endpoint.
-        let stale_replacement = if startup_outcome == StartupChatOutcome::Configured
-            && !probe_ok
-            && args.chat_url.is_some()
-            && args.chat_api_key.is_none()
-            && args.chat_auth_header.is_none()
-        {
+        // EAI-7360: a persisted `tui.chat_url` can go stale. Only re-probe an
+        // unreachable configured URL without credentials; a keyless local
+        // offer must never replace a credentialed gateway. `chat_env_url` is a
+        // separate tier, and the resolver's CLI>config>env>probe precedence is
+        // unchanged.
+        let stale_offer = if should_revalidate_stale_chat_url(
+            startup_outcome,
+            probe_ok,
+            args.chat_url.as_deref(),
+            args.chat_api_key.as_deref(),
+            args.chat_auth_header.as_deref(),
+        ) {
             let live = detect_local_chat(state.tool_executor.clone()).await;
-            stale_chat_url_replacement(&probe_target, probe_ok, live)
+            stale_chat_url_replacement(&probe_target, live)
         } else {
             None
         };
-        if let Some(live) = &stale_replacement {
-            state.chat.push(ChatTurn::system(format!(
-                "Configured chat server at {probe_target} is unreachable; switched to the live \
-                 local engine at {} (model: {}). Run `/detect save` to persist it.",
-                live.base_url, live.model
-            )));
-        }
-        let llm = stale_replacement.or(llm);
         state.set_chat_config(llm, args.chat_auto_consent);
+        if let Some(live) = stale_offer {
+            // Reuse the existing offer reducer so accept, save, and dismiss all
+            // work exactly as they do after an explicit `/detect` probe.
+            state.set_detect_result(Some(live));
+        }
         // No reachable local endpoint AND no key/url configured → the no-key
         // ChatGPT OAuth default (device-code login surfaced in the chat tab).
         // This restores the no-key login the vendored Codex path provided; it
@@ -4818,6 +4804,30 @@ mod tests {
             "accept raises the rebuild edge carrying the previous provider"
         );
         assert_eq!(s.active_provider, ChatProvider::Local);
+    }
+
+    #[test]
+    fn stale_startup_endpoint_uses_actionable_detect_offer() {
+        let mut s = AppState::new("t".into(), "default-dark".into());
+        let stale = crate::llm::detected_llm_config("http://127.0.0.1:9/v1", "old");
+        let live = crate::llm::detected_llm_config("http://localhost:13305/v1", "new");
+
+        // Match event_loop ordering: retain the configured endpoint, then offer
+        // the newly detected one through the established reducer.
+        s.set_chat_config(Some(stale.clone()), true);
+        s.set_detect_result(Some(live.clone()));
+
+        assert_eq!(s.chat_llm.as_ref(), Some(&stale));
+        assert_eq!(s.chat_detect_offer.as_ref(), Some(&live));
+        assert!(
+            s.chat
+                .iter()
+                .any(|turn| turn.content.contains("/detect save"))
+        );
+
+        apply_action(&mut s, KeyAction::ChatDetectSave);
+        assert_eq!(s.chat_llm.as_ref(), Some(&live));
+        assert!(s.chat_persist_dispatch);
     }
 
     #[test]
--- a/crates/rocm-dash-tui/src/app/mod.rs
+++ b/crates/rocm-dash-tui/src/app/mod.rs
@@ -40,7 +40,8 @@ mod summary;
 
 use chat::{
     StartupChatOutcome, build_chat_agent, build_local_agent, detect_local_chat,
-    discover_configured_chat_model, persist_chat_endpoint, startup_chat_outcome,
+    discover_configured_chat_model, persist_chat_endpoint, stale_chat_url_replacement,
+    startup_chat_outcome,
 };
 use summary::{parse_plan_result, summarize_json_value, summarize_slash_tool};
 
@@ -79,7 +80,9 @@ pub struct ResolvedArgs {
     /// exit back to the launcher when the overlay is closed at its root. `None`
     /// (the default) is the normal full dashboard — every path stays unchanged.
     pub focus: Option<Focus>,
-    /// Chat endpoint base URL, CLI-flag value already merged over config.
+    /// Chat endpoint base URL, sourced from persisted config (`tui.chat_url`).
+    /// There is no `--chat-url` CLI flag today, so this tier is exactly the
+    /// persisted config — which is what startup revalidates when it goes stale.
     pub chat_url: Option<String>,
     /// Chat model, CLI-flag value already merged over config.
     pub chat_model: Option<String>,
@@ -1748,11 +1751,16 @@ async fn event_loop(terminal: &mut Tui, args: &ResolvedArgs) -> color_eyre::Resu
         let probe_ok = match startup_outcome {
             StartupChatOutcome::Local => true,
             StartupChatOutcome::OAuth => false,
-            StartupChatOutcome::Configured => tokio::task::spawn_blocking(move || {
-                crate::llm::probe_endpoint(&probe_target, crate::llm::PROBE_TIMEOUT)
-            })
-            .await
-            .unwrap_or(false),
+            StartupChatOutcome::Configured => {
+                // Clone so `probe_target` stays available for the EAI-7360
+                // stale-URL recovery and its notice below.
+                let target = probe_target.clone();
+                tokio::task::spawn_blocking(move || {
+                    crate::llm::probe_endpoint(&target, crate::llm::PROBE_TIMEOUT)
+                })
+                .await
+                .unwrap_or(false)
+            }
         };
         let llm = detected.or_else(|| {
             crate::llm::resolve_llm_config(
@@ -1780,6 +1788,43 @@ async fn event_loop(terminal: &mut Tui, args: &ResolvedArgs) -> color_eyre::Resu
             ),
             other => other,
         };
+        // EAI-7360: a persisted `tui.chat_url` can go stale — the server it
+        // pointed at may have been stopped or replaced since the dash last wrote
+        // it. This is the `Configured`-but-unreachable case: the reachable
+        // `Configured` case already ran `discover_configured_chat_model` above,
+        // and a `Local` outcome auto-detected its own live endpoint. Only this
+        // persisted-config tier is revalidated — `chat_env_url` is a separate
+        // tier, and `resolve_llm_config`'s CLI>config>env>probe precedence (the
+        // literal contract unit-tested in `llm.rs`) is untouched: this only
+        // substitutes a *replacement* `LlmConfig` after that call, never changes
+        // how the call itself resolves.
+        //
+        // Require NO configured api_key/auth_header: the replacement is a
+        // keyless `detected_llm_config`, and since `ResolvedArgs` can't
+        // distinguish an explicit `--chat-url` from persisted config, a remote
+        // gateway URL + `OPENAI_API_KEY` that merely TCP-times-out must never
+        // silently swap to a local keyless engine and drop the credential.
+        // Confining the swap to a no-auth `chat_url` keeps it on the true
+        // EAI-7360 target: a dead *persisted* local endpoint.
+        let stale_replacement = if startup_outcome == StartupChatOutcome::Configured
+            && !probe_ok
+            && args.chat_url.is_some()
+            && args.chat_api_key.is_none()
+            && args.chat_auth_header.is_none()
+        {
+            let live = detect_local_chat(state.tool_executor.clone()).await;
+            stale_chat_url_replacement(&probe_target, probe_ok, live)
+        } else {
+            None
+        };
+        if let Some(live) = &stale_replacement {
+            state.chat.push(ChatTurn::system(format!(
+                "Configured chat server at {probe_target} is unreachable; switched to the live \
+                 local engine at {} (model: {}). Run `/detect save` to persist it.",
+                live.base_url, live.model
+            )));
+        }
+        let llm = stale_replacement.or(llm);
         state.set_chat_config(llm, args.chat_auto_consent);
         // No reachable local endpoint AND no key/url configured → the no-key
         // ChatGPT OAuth default (device-code login surfaced in the chat tab).
--- a/tests/e2e-cucumber/features/chat.feature
+++ b/tests/e2e-cucumber/features/chat.feature
@@ -75,3 +75,12 @@ Feature: Chat and endpoint detection
     And the model is registered with the CLI
     When the user sends a one-shot chat prompt through the CLI
     Then the CLI prints the assistant's reply
+
+  @id:chat-stale-endpoint-offers-live-engine @requires-os:linux
+  Scenario: 8 - A stale saved endpoint offers the live local engine
+    Given a running managed model is available locally
+    And chat has a stale saved endpoint
+    When the user opens interactive chat
+    Then the local endpoint is shown for confirmation
+    When the user quits interactive chat
+    Then interactive chat exits successfully
--- a/crates/rocm-dash-tui/src/app/chat.rs
+++ b/crates/rocm-dash-tui/src/app/chat.rs
@@ -292,41 +292,55 @@ pub(super) const fn startup_chat_outcome(
     }
 }
 
-/// Decide whether a persisted `chat_url` should be replaced by a freshly
-/// detected live endpoint (EAI-7360).
+/// Whether startup should offer to replace a stale persisted `chat_url`.
 ///
-/// Pure — takes the already-computed reachability/detection results rather
-/// than doing I/O itself, so it is unit-testable without a tool executor or a
-/// live socket. The caller only feeds this a `live` detection result when
-/// `chat_url` came from persisted config and neither the managed-services
-/// registry nor a direct TCP probe already confirmed it reachable, so this
-/// never revalidates an explicit env override, and `resolve_llm_config`'s own
-/// CLI>config>env>probe precedence (tested directly in `llm.rs`) is
-/// completely untouched — this only ever supplies a *replacement* config
-/// before that call runs.
-///
-/// Returns `Some(live)` only when the configured endpoint is unreachable and
-/// a *genuinely different*, live endpoint was found; `None` otherwise (nothing
-/// to replace, or detection re-found the same URL modulo a trailing `/`/`/v1`).
-pub(super) fn stale_chat_url_replacement(
-    configured_base_url: &str,
+/// The replacement is keyless, so only an unreachable configured URL with no
+/// credentials is eligible. Keeping this decision pure makes the full startup
+/// guard independently testable before the caller performs another probe.
+pub(super) fn should_revalidate_stale_chat_url(
+    startup_outcome: StartupChatOutcome,
     configured_probe_ok: bool,
-    live: Option<crate::llm::LlmConfig>,
-) -> Option<crate::llm::LlmConfig> {
-    if configured_probe_ok {
-        return None;
-    }
-    let configured = normalize_base_url(configured_base_url);
-    live.filter(|l| normalize_base_url(&l.base_url) != configured)
+    chat_url: Option<&str>,
+    chat_api_key: Option<&str>,
+    chat_auth_header: Option<&str>,
+) -> bool {
+    startup_outcome == StartupChatOutcome::Configured
+        && !configured_probe_ok
+        && chat_url.is_some()
+        && chat_api_key.is_none()
+        && chat_auth_header.is_none()
 }
 
-/// Normalize an OpenAI-style base URL for change-detection comparison: drop a
-/// trailing `/` and a trailing `/v1` so a formatting-only difference (e.g.
-/// `http://h:8000` vs `http://h:8000/v1/`) is not mistaken for a server change
-/// and does not emit a spurious "server changed" notice.
-fn normalize_base_url(base_url: &str) -> String {
+/// Return a newly detected endpoint only when it differs from the stale one.
+///
+/// The caller invokes this after `resolve_llm_config` and routes the result
+/// through [`AppState::set_detect_result`], preserving the existing
+/// accept/save/dismiss flow instead of silently replacing the active config.
+pub(super) fn stale_chat_url_replacement(
+    configured_base_url: &str,
+    live: Option<crate::llm::LlmConfig>,
+) -> Option<crate::llm::LlmConfig> {
+    let configured = normalize_base_url(configured_base_url);
+    live.filter(|candidate| normalize_base_url(&candidate.base_url) != configured)
+}
+
+/// Normalize a local OpenAI-style base URL for change detection.
+///
+/// Trailing `/` and `/v1` are formatting-only. `localhost` and `127.0.0.1`
+/// identify the same loopback server; canonicalizing that spelling avoids a
+/// false server-change offer. The common path remains borrowed.
+fn normalize_base_url(base_url: &str) -> std::borrow::Cow<'_, str> {
     let trimmed = base_url.trim().trim_end_matches('/');
-    trimmed.strip_suffix("/v1").unwrap_or(trimmed).to_string()
+    let normalized = trimmed.strip_suffix("/v1").unwrap_or(trimmed);
+    for prefix in ["http://localhost", "https://localhost"] {
+        if let Some(rest) = normalized.strip_prefix(prefix)
+            && (rest.is_empty() || rest.starts_with(':') || rest.starts_with('/'))
+        {
+            let scheme = &prefix[..prefix.len() - "localhost".len()];
+            return format!("{scheme}127.0.0.1{rest}").into();
+        }
+    }
+    normalized.into()
 }
 
 /// Persist an accepted local endpoint to the user's `config.toml`: load the
@@ -698,58 +712,83 @@ mod tests {
     }
 
     #[test]
-    fn stale_chat_url_replacement_none_when_configured_endpoint_reachable() {
-        // Reachable configured endpoint must never be swapped out, regardless
-        // of what a stray `live` detection result would have found.
-        assert_eq!(
-            stale_chat_url_replacement(
-                "http://127.0.0.1:9/v1",
-                true,
-                Some(live("http://127.0.0.1:11435/v1"))
+    fn stale_revalidation_requires_unreachable_uncredentialed_persisted_url() {
+        assert!(should_revalidate_stale_chat_url(
+            StartupChatOutcome::Configured,
+            false,
+            Some("http://127.0.0.1:9/v1"),
+            None,
+            None,
+        ));
+
+        for (outcome, probe_ok, url, key, header) in [
+            (
+                StartupChatOutcome::Local,
+                false,
+                Some("http://x"),
+                None,
+                None,
             ),
-            None
-        );
+            (
+                StartupChatOutcome::Configured,
+                true,
+                Some("http://x"),
+                None,
+                None,
+            ),
+            (StartupChatOutcome::Configured, false, None, None, None),
+            (
+                StartupChatOutcome::Configured,
+                false,
+                Some("http://x"),
+                Some("k"),
+                None,
+            ),
+            (
+                StartupChatOutcome::Configured,
+                false,
+                Some("http://x"),
+                None,
+                Some("Authorization"),
+            ),
+        ] {
+            assert!(!should_revalidate_stale_chat_url(
+                outcome, probe_ok, url, key, header
+            ));
+        }
     }
 
     #[test]
     fn stale_chat_url_replacement_none_when_nothing_live_found() {
         assert_eq!(
-            stale_chat_url_replacement("http://127.0.0.1:9/v1", false, None),
+            stale_chat_url_replacement("http://127.0.0.1:9/v1", None),
             None
         );
     }
 
     #[test]
     fn stale_chat_url_replacement_none_when_live_matches_configured() {
-        // Detection simply re-confirming the same (currently-unreachable-by-
-        // TCP-probe-timing) URL is not a "server changed" situation.
         let url = "http://127.0.0.1:11435/v1";
-        assert_eq!(
-            stale_chat_url_replacement(url, false, Some(live(url))),
-            None
-        );
+        assert_eq!(stale_chat_url_replacement(url, Some(live(url))), None);
     }
 
     #[test]
-    fn stale_chat_url_replacement_swaps_to_a_different_live_endpoint() {
+    fn stale_chat_url_replacement_offers_a_different_live_endpoint() {
         let found = live("http://127.0.0.1:13305/v1");
-        let replacement =
-            stale_chat_url_replacement("http://127.0.0.1:9/v1", false, Some(found.clone()))
-                .expect("a different live endpoint replaces the stale one");
+        let replacement = stale_chat_url_replacement("http://127.0.0.1:9/v1", Some(found.clone()))
+            .expect("a different live endpoint is offered");
         assert_eq!(replacement.base_url, found.base_url);
     }
 
     #[test]
-    fn stale_chat_url_replacement_ignores_trailing_slash_and_v1_formatting() {
-        // A formatting-only difference (trailing `/`, presence/absence of the
-        // `/v1` suffix) is the SAME endpoint — no spurious "server changed".
+    fn stale_chat_url_replacement_normalizes_equivalent_urls() {
         for (configured, found) in [
             ("http://127.0.0.1:8000/v1", "http://127.0.0.1:8000/v1/"),
             ("http://127.0.0.1:8000/v1", "http://127.0.0.1:8000"),
-            ("http://127.0.0.1:8000", "http://127.0.0.1:8000/v1"),
+            ("http://localhost:8000/v1", "http://127.0.0.1:8000/v1"),
         ] {
             assert_eq!(
-                stale_chat_url_replacement(configured, false, Some(live(found))),
+                stale_chat_url_replacement(configured, Some(live(found))),
                 None,
                 "{configured} vs {found} must be treated as the same endpoint"
             );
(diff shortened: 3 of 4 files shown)


rominf (approved):
LGTM. The stale-persisted-URL revalidation is well-guarded: the swap only fires on `!probe_ok` **and** a genuinely-different (trailing-`/` and `/v1` normalized) URL, and is gated on `chat_api_key.is_none() && chat_auth_header.is_none()` so a configured remote gateway that merely times out won't silently fall back to a keyless local engine. No loop (re-detecting the same URL returns `None`, so no spurious "server changed" notice), no panic (`unwrap_or(false)` / `.ok()`), and correct precedence (`managed.or_else(stale).or_else(resolve_llm_config)` — the CLI>config>env>probe contract is untouched). Dropping `chat_model` on swap is correct since it's a different server. Good test coverage including the normalization cases.

Note: on the degraded path (persisted URL dead) startup now does a bounded one-time extra probe (~300ms + `detect_local_chat`) before the first frame — worth a sentence in the PR body.

Approving.

volen-silo:
🔴 Automated review · pr-review-watcher · ed0b00e5

## Summary

- **Change**: Adds a startup revalidation for a stale persisted `tui.chat_url` — when the configured endpoint is unreachable and no auth/key is configured, the dash falls back to a freshly-detected live local engine and surfaces a "server changed" notice. Two files, +147/-3 (`app/chat.rs`, `app/mod.rs`).
- **Overall assessment**: **Approve with minor suggestions.** The core logic is correct and well-scoped. The substitution genuinely prevents building against a dead endpoint, the central safety claims hold (no credential is ever dropped; no explicit-CLI flag is ever swapped), and the new pure functions are cleanly unit-tested.
- **Verification**: built clean and ran the 6 new unit tests locally (`cargo test -p rocm-dash-tui --lib`) — all pass. No blocking findings.

## Verified as correct (the load-bearing claims)

- **Substitution actually fixes the bug**: with `chat_url = Some(dead)`, `resolve_llm_config`'s `cli_url.or(...)` returns the dead URL verbatim regardless of `probe_ok`, so pre-PR the backend was built against the dead endpoint. Because `managed` is always `None` on this path, `managed.or_else(|| stale_replacement)` correctly lets the replacement win over that fallback. Confirmed against `llm.rs:76-80`.
- **No credential drop / no explicit-flag swap**: traced `ResolvedArgs.chat_url` — it is sourced **only** from persisted `t.chat_url` (`apps/rocm/src/dash.rs:188`); there is **no** `--chat-url` clap arg anywhere in `apps/rocm/src/main.rs`. The gate additionally requires `chat_api_key.is_none() && chat_auth_header.is_none()`, and the swap never mutates the key/header in `ResolvedArgs`. The safety invariant the PR is built on holds.
- **Notice accuracy**: because the gate requires `chat_url.is_some()`, `probe_target` here is always the configured URL (never env/default), so "Configured chat server at {probe_target} is unreachable…" is accurate. Interpolation of `live.base_url`/`live.model` is correct.

## Non-blocking suggestions

1. **Dead/redundant guard conjunct** — `crates/rocm-dash-tui/src/app/mod.rs:1606`. `managed.is_none()` is always true when this branch is reached: `managed` (line ~1565) can only be `Some` when `chat_url.is_none() && chat_env_url.is_none()`, but the guard already requires `chat_url.is_some()`. Harmless but misleading — a reader may think it guards a real "managed beats stale persisted URL" case that can't occur. Consider dropping it or adding a one-line "structurally always true" note.

2. **`probe_ok` passed into `stale_chat_url_replacement` is statically `false` at this call site** — `mod.rs:1610`. The guard has `!probe_ok`, so the function's `if configured_probe_ok { return None }` can never fire from here. Not a bug (the function is still correctly tested standalone with `true`), but the parameter is redundant/confusing at the call site.

3. **`normalize_base_url` heuristic gaps** — `crates/rocm-dash-tui/src/app/chat.rs:281`. Verified by executing the exact logic against edge cases:
   - `http://h:8000/v1/v1` → `http://h:8000/v1` (only one `/v1` stripped; non-idempotent).
   - `http://h:8000/V1` → unchanged (case-sensitive strip, unlike `is_loopback_host`'s `eq_ignore_ascii_case` elsewhere) → spurious "server changed" for the same server.
   - `http://h:8000/v1?x=1` → unchanged (query not handled, unlike `parse_host_port`, which splits on `['/', '?']`).
   - `http://h:8000/api/v1` and `http://h:8000/api` normalize equal → a legitimate swap would be suppressed.

   None of these hit the realistic EAI-7360 target (local vLLM/Lemonade on standard `/v1`), so impact is limited to a spurious/suppressed *notice* in the narrow "configured URL unreachable + a live local engine present" window. Worth considering reusing the existing authority-parsing in `llm.rs` (`parse_host_port`) rather than a second, divergent URL canonicalizer, and lowercasing before the `/v1` strip.

4. **Test doesn't assert full-config preservation** — `crates/rocm-dash-tui/src/app/chat.rs` (swap test, ~line 487). `stale_chat_url_replacement_swaps_to_a_different_live_endpoint` asserts only `replacement.base_url`; it never checks `model`/`api_key`/`auth_header`, so a broken reconstruction of the returned `LlmConfig` wouldn't be caught. Add asserts on the other fields, and a case for `/v1/v1` / uppercase `/V1` if you tighten (3).

5. **Stale doc comment adjacent to the change** — `crates/rocm-dash-tui/src/app/mod.rs:82`: `chat_url` is documented as "CLI-flag value already merged over config," which contradicts the actual sourcing (persisted config only) and the very safety invariant this PR relies on. Since it sits in a file this PR touches and is directly relevant to the reviewer's reasoning, fixing it here would be worthwhile. (Related, out-of-PR-scope: `crates/rocm-dash-tui/src/ui/tabs/chat.rs:204` still advertises a `--chat-url` flag that does not exist — flagged for future cleanup, not this PR.)

## Tradeoffs (deliberate choices worth confirming, not change requests)

- **Startup latency on the dead-URL path**: when a configured `chat_url` is unreachable with no key/header, `detect_local_chat` now runs *after* the initial probe — managed detection (blocking executor call + a `fetch_first_model` HTTP GET, up to ~3s) plus up to three sequential TCP probes (≈900ms) plus another model fetch. This stacks onto the existing "one probe before first frame" budget, so the first frame can stall noticeably longer than `PROBE_TIMEOUT` in that specific failure case. Scoped and functionally correct, but a real UX cost.
- **Keyless-remote endpoint indistinguishable from local**: the gate (`chat_url set` + no key + no auth-header) matches not only a dead local endpoint but also a keyless/network-authed **remote** gateway whose key merely isn't in the current shell's env (key is env-only via `OPENAI_API_KEY`/`ROCMDASH_CHAT_API_KEY`). With a 300ms TCP timeout, a real-but-slow remote could probe-fail and get redirected to a local engine. Mitigated: it fires only if a *different live* engine exists, it's visible (system chat turn), and it's non-persistent (only `/detect save` persists). If false-positive swaps on real remotes show up in practice, a "skip swap when configured host is non-loopback" heuristic would tighten it. Noting for awareness — the credential-safety claim itself is not violated.

## Positive signals

- Extracting the decision into a pure, I/O-free `stale_chat_url_replacement` with focused unit tests is the right call — testable without a socket or tool executor.
- The auth-drop gate and its rationale comment (mod.rs:1596-1602) show the credential-leak edge case was reasoned about deliberately, not by accident.
- Existing `resolve_llm_config` precedence tests are genuinely untouched, so the substitution-at-call-site approach keeps the precedence contract intact.

rominf (commented):
Automated review: re-reviewed the delta since my approval at `ed0b00e5` through current head `e8f896c6`. The unreviewed code is the hand-resolved merge of PR #97 into this branch, concentrated in `app/chat.rs` and `app/mod.rs`.

I traced the rewritten startup path: the `llm` rebinding preserves precedence; configured-endpoint model discovery and stale-URL replacement are mutually exclusive on probe success; neither clobbers the other; detection is not repeated; and the previously approved helpers and tests remain intact. I found no blocking code defect in the delta.

Non-blocking:
1. The PR body still says no overlapping hunks were expected, but both files required manual conflict resolution. Please update that coordination note.
2. The helpers are unit-tested independently, but no integration test covers their composed ordering in `event_loop`, where the merge occurred. That would be useful follow-up coverage.

Merge readiness: the PR currently conflicts with `main`, and the GPU and Strix Halo/Windows E2E checks are red. I could not verify those failures as infrastructure flakes from the available log tail, so they should be inspected or rerun before merge.

rominf (changes requested):
The core idea is right and the scope is tight. Two things need addressing before this lands: the remediation the notice tells the user to run doesn't work, and the change hand-rolls a narrower version of a mechanism that already exists and would make it work.

## Blocking

**1. The notice instructs a command that can't work — `crates/rocm-dash-tui/src/app/mod.rs:1820-1826`**

The message ends with "Run `/detect save` to persist it." But `/detect save` is gated on `chat_detect_offer.is_some()` (`app/slash.rs:491-496`), and this path never populates it — it calls `state.set_chat_config(llm, ...)` (`mod.rs:1828`), which only sets `chat_consent`/`chat_llm` (`app/chat.rs:25-32`). A user who follows the instruction gets back "no detected endpoint to act on yet — run `/detect` first to probe for a local engine." So the swap can't actually be persisted, which is the one thing a user hitting this will want.

**2. This duplicates the existing detect-offer flow — `mod.rs:1815-1828` vs `mod.rs:1143-1166`**

`set_detect_result(Some(cfg))` already stores the offer and pushes a transcript message naming endpoint and model with the working `/detect accept | save | dismiss` sub-commands. Routing the detected `live` config through it would reuse the tested reducer, make (1) disappear, and give the user accept/dismiss instead of a silent fait-accompli swap. As written there are two divergent paths for "we found a different local engine", which will drift. (Note `set_detect_result` only pushes the message when consent is `Accepted` — worth checking that fits the startup path, but the offer is set unconditionally, which is what `/detect save` needs.)

**3. Doc comment contradicts the code — `app/chat.rs:305-306`**

"…this only ever supplies a *replacement* config **before that call runs**." The substitution happens after `resolve_llm_config` (`mod.rs:1766` → `1816` → `1827`), and `mod.rs:1799` in this same commit correctly says "after that call." One of the two is wrong.

## Non-blocking

**4. An explicit `chat_model` is silently dropped on the recovery path — `mod.rs:1827`, `chat.rs:311-321`.** `stale_chat_url_replacement` returns the detection result verbatim, whose model comes from `/v1/models` or the default. The sibling reachable path (`discover_configured_chat_model`, `chat.rs:413-426`) goes out of its way to honor an explicit model. Probably the right call for a different server, but it's an undocumented asymmetry and it isn't covered in the long precedence comment above.

**5. The adjacent doc line has the exact bug this PR fixes one line above — `mod.rs:87-88`.** The `chat_url` doc correction is accurate; `chat_model` still says "CLI-flag value already merged over config" and there's no `--chat-model` flag in `apps/rocm/src/main.rs` either. One-line fix, in scope.

**6. Dead condition and dead argument — `mod.rs:1809-1816`.** `should_detect_local_chat` returns `false` whenever `chat_url.is_some()` (`chat.rs:268`), so the `== Configured` check is redundant given the `args.chat_url.is_some()` conjunct. And `stale_chat_url_replacement(&probe_target, probe_ok, live)` passes `probe_ok` from inside an `if !probe_ok` branch, so `chat.rs:316-318`'s early return is unreachable from the only production caller — it exists for one unit test. Defensible as a general helper signature, but worth a note.

**7. `localhost` vs `127.0.0.1` reads as a server change — `chat.rs:327-330`.** `normalize_base_url` compares raw strings; the default base URL uses `127.0.0.1` while the detection candidates use `localhost`. Same server, different spelling, spurious swap plus notice. Unlikely in practice since persisted URLs come from the candidates, but it's precisely the false-positive this helper exists to prevent, and it's untested.

**8. A 300ms TCP-connect probe now has a larger blast radius — `llm.rs:176-188`.** `probe_endpoint` only does `connect_timeout`; it says nothing about `/v1/chat/completions` working. Previously a false "unreachable" just meant "skip the default fallback"; now it substitutes a different endpoint. A server still binding, or an IPv6/IPv4 ordering quirk, can take a user off their real endpoint. Worth a retry or an HTTP-level check before acting on a negative probe.

**9. The risky part is untested — `mod.rs:1809-1828`.** All six new tests exercise `stale_chat_url_replacement`, a two-line string comparison. The four-condition guard, the `detect_local_chat` call, the notice, and the `stale_replacement.or(llm)` sequencing — the wiring that can regress — have none. `stale_chat_url_replacement_none_when_nothing_live_found` (`chat.rs:702-708`) is effectively asserting `Option::filter(None)`.

**10. `normalize_base_url` allocates needlessly — `chat.rs:329`.** Every step yields a sub-slice; returning `&str` avoids the `to_string()`. Cosmetic.

## Worth calling out as good

The `chat_api_key.is_none() && chat_auth_header.is_none()` gate (`mod.rs:1812-1813`) is the right instinct and correctly reasoned — a keyless local swap replacing a credentialed remote gateway would have been a real regression, and it was anticipated rather than found later. Extracting the decision into a pure I/O-free helper is also the right shape; the gap is that the untested part is the part that isn't pure.

## Also noted, not asks

Startup latency on the degraded path grows by a full `detect_local_chat` on top of the probe (bounded, failure-path only, and disclosed in the description). A server dying mid-session still yields a bare `ChatError` via `on_chat_error` (`mod.rs:1257`) with no re-probe — reasonable scope-narrowing, but the failure class is half-covered.

michaelroy-amd:
Addressed the requested changes in cefe5313:

- stale recovery now uses the existing detect-offer reducer, so `/detect save` is actionable and accept/dismiss remain available
- the configured endpoint remains active until the user chooses the detected replacement
- corrected the contradictory helper comment and `chat_model` source documentation
- centralized and tested the complete startup guard
- normalized `localhost` and `127.0.0.1` as the same endpoint
- added `@id:chat-stale-endpoint-offers-live-engine`; the targeted scenario passes

Full workspace tests, clippy, and the local smoke suite pass.

rominf:
Reviewed the current head (cefe5313) as a diff-only read against origin/main.

This revalidates a stale persisted `tui.chat_url` at startup: if it's unreachable and no key/auth header is configured, the dash now probes for a live local engine and routes it through the existing `/detect accept|save|dismiss` offer flow instead of silently swapping the active config. I traced the credential-safety gate (`chat_api_key.is_none() && chat_auth_header.is_none()`), confirmed `resolve_llm_config`'s CLI>config>env>probe precedence is untouched, and checked the `normalize_base_url` equivalence logic (trailing `/`, `/v1`, and `localhost`/`127.0.0.1`) against its unit tests — all consistent with the stated intent.

I also compared this head against the prior formal review threads on this PR: the three blocking issues from the last CHANGES_REQUESTED round (the "run `/detect save`" instruction that couldn't work, the duplicated ad-hoc offer path, and a contradictory doc comment) all appear resolved here — the fix now reuses the existing `set_detect_result` reducer, and the doc comments for `chat_url`/`chat_model` sourcing were corrected. The `localhost` vs `127.0.0.1` normalization gap flagged earlier is also fixed.

I didn't find anything new to add on top of that history. This looks safe to approve after final validation (a fresh CI run on this exact head and the maintainer's own pass) — this comment reflects only a diff-and-history read, not a live run.

crates/rocm-dash-tui/src/app/mod.rs
```diff
@@ -4775,6 +4806,30 @@ mod tests {
...
 
+    #[test]
+    fn stale_startup_endpoint_uses_actionable_detect_offer() {
+        let mut s = AppState::new("t".into(), "default-dark".into());
+        let stale = crate::llm::detected_llm_config("http://127.0.0.1:9/v1", "old");
+        let live = crate::llm::detected_llm_config("http://localhost:13305/v1", "new");
+
+        // Match event_loop ordering: retain the configured endpoint, then offer
+        // the newly detected one through the established reducer.
+        s.set_chat_config(Some(stale.clone()), true);
```
rominf, line 4817: **This passes `true` where production always passes `false` — and that flag is what makes the assertion below reachable.**

The comment says "Match event_loop ordering", but event_loop passes `args.chat_auto_consent` (:1803), and that is hardcoded `false` at the sole production construction site (`apps/rocm/src/dash.rs:207`). Nothing sets it true on any path that reaches the stale-URL branch.

With `false`, `set_chat_config` sets consent to `Pending` (`app/chat.rs:25-32`). And `set_detect_result` (:1146-1156) pushes the `/detect save` transcript turn **only** when `chat_consent == Accepted`:

```rust
let accepted = self.chat_consent == ChatConsent::Accepted;
if let Some(cfg) = offer {
    self.chat_detect_msg = None;
    if accepted {
        self.chat.push(ChatTurn::agent(format!("Detected a local engine: …")));
    }
    self.chat_detect_offer = Some(cfg);
}
```

So the test's load-bearing assertion — `turn.content.contains("/detect save")` — passes only because of the flag being wrong. Flip it to `false` and that assertion fails, because in production this path renders the `chat_detect_offer` panel (`ui/tabs/chat.rs:69-110`, the `[y]`/`[s]`/`[n]` affordances) instead of a transcript turn. Which means the thing users actually see on this path is covered by nothing.

The `chat_detect_offer` and `apply_action(ChatDetectSave)` assertions are fine either way, so the fix is small: pass `false`, and swap the transcript assertion for one on the offer panel. Asserting both branches would be better still, since the accepted-consent path is a real (if not-yet-reachable) mode.

crates/rocm-dash-tui/src/app/mod.rs
```diff
@@ -1748,11 +1752,16 @@ async fn event_loop(terminal: &mut Tui, args: &ResolvedArgs) -> color_eyre::Resu
...
             StartupChatOutcome::Local => true,
             StartupChatOutcome::OAuth => false,
-            StartupChatOutcome::Configured => tokio::task::spawn_blocking(move || {
-                crate::llm::probe_endpoint(&probe_target, crate::llm::PROBE_TIMEOUT)
-            })
-            .await
-            .unwrap_or(false),
+            StartupChatOutcome::Configured => {
+                // Clone so `probe_target` stays available for the EAI-7360
+                // stale-URL recovery and its notice below.
```
rominf, line 1757: **"and its notice below" refers to something this revision removed.**

The earlier revision at `b8a140b0` pushed a `ChatTurn::system` notice after the recovery. The current head replaced that with `set_detect_result` — `git diff origin/main...cefe5313` contains no `ChatTurn::system` at all, so there is no notice below any more.

The clone itself is still needed (`spawn_blocking` wants `'static`, and `probe_target` is used again at :1802), so this is just the comment's tail. Dropping "and its notice below" is the whole fix.

crates/rocm-dash-tui/src/app/mod.rs
```diff
@@ -79,9 +80,12 @@ pub struct ResolvedArgs {
...
     /// (the default) is the normal full dashboard — every path stays unchanged.
     pub focus: Option<Focus>,
-    /// Chat endpoint base URL, CLI-flag value already merged over config.
+    /// Chat endpoint base URL, sourced from persisted config (`tui.chat_url`).
+    /// There is no `--chat-url` CLI flag today, so this tier is exactly the
+    /// persisted config — which is what startup revalidates when it goes stale.
     pub chat_url: Option<String>,
-    /// Chat model, CLI-flag value already merged over config.
+    /// Chat model sourced from persisted config (`tui.chat_model`). There is
+    /// no `--chat-model` CLI flag today.
```
rominf, line 88: **This is the third of three adjacent doc comments with the same wrong claim, and it's the one left unfixed.**

:83-86 and :87-88 were both corrected in this PR to say the value comes from persisted config with no CLI flag. This one still says "(CLI-flag value merged over config)".

There is no `--chat-auth-header` — in fact no `--chat-*` flag at all: `grep 'long = "chat' apps/rocm/src/` returns nothing, and `auth_header` appears in `main.rs` only inside an HTTP request builder. So this is the same defect the PR exists to fix, sitting two lines below the fix.

Same treatment as its neighbours: "Custom auth header NAME, sourced from persisted config (`tui.chat_auth_header`); there is no CLI flag today."

(I raised this class as item #5 last round — flagging again since it's the one instance that didn't get picked up.)

crates/rocm-dash-tui/src/app/mod.rs
```diff
@@ -1780,7 +1789,29 @@ async fn event_loop(terminal: &mut Tui, args: &ResolvedArgs) -> color_eyre::Resu
...
+        // unreachable configured URL without credentials; a keyless local
+        // offer must never replace a credentialed gateway. `chat_env_url` is a
+        // separate tier, and the resolver's CLI>config>env>probe precedence is
+        // unchanged.
+        let stale_offer = if should_revalidate_stale_chat_url(
+            startup_outcome,
+            probe_ok,
+            args.chat_url.as_deref(),
+            args.chat_api_key.as_deref(),
+            args.chat_auth_header.as_deref(),
```
rominf, line 1802: **This runs before the first frame is painted, and its worst case is now multiple seconds of black terminal.**

The whole startup block precedes any `terminal.draw` (first one is at :1860), so this isn't a slow refresh — the user is looking at an unpainted terminal for the duration.

Tracing the cost of `detect_local_chat` on the configured-but-dead path this fix targets:

- `detect_managed_chat` → the `services` tool → `load_managed_services` → `managed_service_endpoint_readiness`, which pays `SERVICE_LIVENESS_CHECK_TIMEOUT` (750ms) plus `rocm_core::INFERENCE_PROBE_TIMEOUT` (8s, `rocm-core/src/lib.rs:614`)
- then `fetch_first_model` with a 3s reqwest timeout (`app/chat.rs:366`)
- then the fallback `detect_local_endpoint` (`PROBE_TIMEOUT`, 300ms)

Before this PR, the configured-but-dead path paid only the ~300ms probe. So this change puts a multi-second worst case specifically on the degraded path it's meant to improve, with nothing on screen.

I flagged the ~300ms cost at `ed0b00e5` and asked for a sentence in the PR body; that estimate is now well short, and the PR body still doesn't mention it.

Two options: paint the first frame first and fold the offer in through the existing reducer once detection resolves (cleanest — the reducer already handles arriving-later offers), or at minimum wrap this call in an explicit short timeout and say so in the PR body.

tests/e2e-cucumber/features/chat.feature
```diff
@@ -75,3 +75,12 @@ Feature: Chat and endpoint detection
     And the model is registered with the CLI
     When the user sends a one-shot chat prompt through the CLI
     Then the CLI prints the assistant's reply
+
+  @id:chat-stale-endpoint-offers-live-engine @requires-os:linux
+  Scenario: 8 - A stale saved endpoint offers the live local engine
+    Given a running managed model is available locally
+    And chat has a stale saved endpoint
+    When the user opens interactive chat
```
rominf, line 83: **This scenario's only assertion is one that scenarios 2 and 3 already make.**

`Then the local endpoint is shown for confirmation` (`dash_steps.rs:431-446`) waits for the mock's ephemeral port to appear anywhere on screen. Scenario 2 (`@id:chat-privacy-notice-accurate`) and scenario 3 (`@id:chat-managed-model-interactive`) both use the identical step.

It does discriminate against `main` today — the plain gate there would render the stale `:9` — so it isn't worthless. But it asserts neither of the two properties this PR actually adds:

1. That the **offer panel** rendered (`Chat — use detected engine?`, `[s] use & save`) rather than the plain consent gate.
2. That the stale endpoint stays active until the user accepts.

`AGENTS.md` §3 wants the user-observable behavior covered, and explicitly rejects proxies that prove the mechanism rather than the behavior. Reusing an assertion two other scenarios already make is that gap one level up: if someone later replaced the offer flow with a silent swap, this scenario would still pass.

Asserting the offer-panel copy plus "the stale URL is still the configured endpoint" would make it fail on exactly the regression it's guarding.

rominf (changes requested):
Re-reviewed at `cefe5313`. All three of the blocking items from my `b8a140b0` review are genuinely fixed: `/detect save` now works because the stale URL routes through the existing `set_detect_result` offer reducer rather than a bespoke swap, the duplicate swap path is gone, and the `stale_chat_url_replacement` doc correctly says "after `resolve_llm_config`". Reusing the offer reducer is the right call — accept/save/dismiss all come for free and the credential guard (`chat_api_key.is_none() && chat_auth_header.is_none()`) holds.

Five things left, none as serious as the last round. The one I'd actually want fixed before merge is the new unit test: it passes `true` for `pre_consent` where production always passes `false`, and that flag is exactly what decides whether the assertion it makes is reachable. Details inline.

The others: a stale comment tail referring to a notice this revision removed, one of three adjacent doc comments left with the wrong claim the PR exists to correct, an unbounded multi-second detection now running before the first frame is painted, and a new scenario whose only assertion is one that two existing scenarios already make.

**On state:** the branch is 96 commits behind `main` (merge-base `ca9f2974`, 2026-07-29) but `mergeable: MERGEABLE` — `git merge-tree` reports zero conflicts, and I checked the merged tree builds and tests green, including the `--test e2e` target that `main` rewrote heavily in the meantime. So no rebase is needed for *correctness*, but `mergeStateStatus` is `BLOCKED` and a rebase is what gets CI to run against something current. All 25 checks that did run are green.

Cleared on re-read: `resolve_llm_config`'s CLI>config>env>probe precedence is untouched; `normalize_base_url` is allocation-free on the common path; and the two `detect_local_chat` calls are correctly mutually exclusive (`should_detect_local_chat` requires `chat_url.is_none()`), so there's no double detection.

Method note: parts of this were read by parallel helper passes that ran degraded (sequential, no fan-out). Everything above I re-verified against the blobs at `cefe5313`.

siloteemu (changes requested):
🔴 Automated review · pr-review-watcher · cefe5313

## Summary
Revalidates an unreachable persisted `tui.chat_url` at startup and routes a freshly detected local endpoint through the existing `/detect` offer flow, with new pure helpers, unit tests, and an e2e scenario. Verdict: **Needs work** — the new e2e scenario is genuinely load-bearing, but the new unit test is not, and the startup path introduces a privacy-gate regression while "use & save" does not actually clear the stale value. Verified by reading the code (not assumed): the e2e scenario **would fail if the production change were reverted** (with `chat_url` set, `should_detect_local_chat` returns false → `Configured` outcome → `resolve_llm_config` returns the stale URL unconditionally since `probe_ok` only gates the `DEFAULT_CHAT_BASE_URL` fallback, so the gate would render port 9, while the `Then` step waits for the mock's OS-assigned port); the new unit test **would still pass if the change were reverted** (it references no new symbol); a previously-reported consent-flag finding is **still true**; a previously-reported "the new scenario only asserts what scenarios 2 and 3 assert" is **not right at the e2e level** (scenario 8 is the only one that plants a persisted unreachable URL, a path scenarios 2 and 3 cannot reach) but **does apply to the new unit test**; the e2e Given's path/schema were verified against `AppPaths::config_path()`, `ROCM_CLI_CONFIG_DIR`, and the serde field names, and the `config/` dir is pre-created. CI (25 passing, 0 failing) relied on for build/lint/test status; no suites were run here. Blocking: 3 · Non-blocking: 5.

- `crates/rocm-dash-tui/src/app/mod.rs:4809-4831` — `stale_startup_endpoint_uses_actionable_detect_offer` does not test the change. Its body calls only pre-existing reducers (`set_chat_config`, `set_detect_result`, `apply_action(ChatDetectSave)`) and references no symbol added by this PR, so it compiles and passes verbatim with the production change reverted; `detect_offer_lifecycle_accept_switches_chat` and `save_detect_offer_accepts_and_raises_persist_edge` already cover everything it asserts except one line. That one line — `assert!(s.chat.iter().any(|t| t.content.contains("/detect save")))` — passes only because the test passes `true` as `pre_consent`, while the sole production construction site (`apps/rocm/src/dash.rs:200`) hardcodes `chat_auto_consent: false`. `set_detect_result` (`mod.rs:1146-1157`) pushes that turn only when consent is `Accepted`, so at real startup it is never pushed and the assertion would be false. Fix: assert against `false` (matching production) and cover the actual wiring — `detect_local_chat_with_probe` (`chat.rs:239`) already exists as an injectable-probe seam, or extract the `stale_offer` block from `event_loop` into a testable function.

- `crates/rocm-dash-tui/src/ui/tabs/chat.rs:69` — the new startup path can surface the accept-and-use affordance before the privacy notice. `draw_consent` picks the offer panel ahead of `consent_gate_lines` whenever `chat_detect_offer` is `Some`, regardless of consent state. Because production startup leaves consent `Pending` and this PR now sets an offer at startup, a first-run user with a stale URL sees the "use detected engine? [y] use now [s] use & save" panel as the first thing on the chat tab and can accept straight into the transcript — the notice `"Your request only leaves this machine after you accept."` (`chat.rs:167`) is never rendered. Previously the offer only appeared after the user pressed `d` at the gate, i.e. after seeing the notice. `chat.feature` scenario 2 encodes "the notice must precede any request" as an invariant, and neither it nor new scenario 8 covers this path (`offer_state_shows_endpoint_and_choices`, `chat.rs:650`, asserts only the panel contents). Fix: include the privacy line in the offer panel when consent is not yet `Accepted`, and add an e2e/unit assertion for it.

- `crates/rocm-dash-tui/src/app/chat.rs:352-366` — "use & save" does not clear the stale value that caused the offer. `persist_chat_endpoint` writes `dirs::config_dir()/rocm-dash/config.toml`, but startup's `chat_url` comes from the unified `config.json` (`AppPaths::config_path()`, `crates/rocm-core/src/lib.rs:906`, surfaced via `apps/rocm/src/dash.rs:192`), and the legacy-TOML migration runs only when `config.json` is absent. So after saving, the stale `tui.chat_url` in `config.json` survives and the same offer reappears on every launch — the PR's stated problem ("a persisted `tui.chat_url` can go stale") is fixed only for the session, contradicting the call-site comment that accept/save/dismiss "work exactly as they do after an explicit `/detect` probe". The split is pre-existing, but this PR is what makes it load-bearing. `DashboardConfig::with_chat_endpoint` (`crates/rocm-core/src/lib.rs:4292`) is the unified-side helper and is currently referenced only from a unit test. Fix: persist through the unified config, or state the session-only scope explicitly in the PR and the offer text.

## Non-blocking
- `crates/rocm-dash-tui/src/app/chat.rs:761-767` — `stale_chat_url_replacement_none_when_nothing_live_found` passes for a trivial implementation (input is `None`); it constrains nothing beyond `Option`'s type.
- `crates/rocm-dash-tui/src/app/chat.rs:798-803` — asserting `Cow::Borrowed` pins an allocation strategy, not behavior; a correctness-preserving refactor (e.g. lowercasing before comparison) would break it.
- `crates/rocm-dash-tui/src/app/chat.rs:332-344` — `normalize_base_url` does not canonicalize `[::1]`, `0.0.0.0`, or scheme/host case, so those spellings still produce a spurious "different server" offer; worth a doc caveat if out of scope.
- `crates/rocm-dash-tui/src/app/chat.rs:319-325` — when the live endpoint has the same URL but a different served model, the freshly fetched config is discarded, so the stale model id is kept even though a verified one was in hand.
- `tests/e2e-cucumber/tests/e2e/dash_steps.rs:80-99` — add a one-line rationale (port 9 as a canonically closed port; this writes the CLI's `config.json`, distinct from the rocm-dash TOML that `/detect save` writes), matching the sibling steps' commenting style and flagging the JSON/TOML trap above.

siloteemu:
🔴 Automated review · pr-review-watcher · cefe5313

## Summary

Forced re-check of the standing change request at the **same commit** we reviewed — the branch has not moved, so any retraction would mean our earlier review was wrong. Each blocking count was re-derived from the source at this commit by an independent pass that was instructed to try to refute it, and the branch was diffed against **both** its merge base and the current base-branch tip.

**Outcome: all three blocking counts still hold (B).** None was refuted; none was replaced. One new, non-blocking prose finding was added (a code comment describing a notice that does not exist). **Finding state: blocking 3 · non-blocking 6.**

CI conclusions worked from: **25 success, 0 failure**. No individual job is named or inferred here. The full suite was not run here; every statement below rests on reading the source at this commit.

### Has the base branch superseded any of this?

No. The base branch has moved a long way since the merge base, but not through any of the code these findings touch:

- `crates/rocm-dash-tui/src/ui/tabs/chat.rs` — **no change at all** between the merge base and the base-branch tip. The `draw_consent` ordering in count 2 is byte-identical on both sides.
- `persist_chat_endpoint` and the JSON/TOML config split in count 3 are unchanged on the base branch; the one-shot legacy-TOML migration already existed at the merge base and is not new work.
- The startup block in `event_loop` that count 1 is about is untouched on the base branch; the base-branch churn in `app/chat.rs` is unrelated signature work on `build_chat_agent` / `build_local_agent`.

So the answer to every count is the same as it was, for the same reasons.

---

## 🚫 Blocking (must fix before merge) — 3

### 1. (still holds) The new unit test in `app/mod.rs` does not test the change

Our claim under test, verbatim:

> `stale_startup_endpoint_uses_actionable_detect_offer` does not test the change. Its body calls only pre-existing reducers … and references no symbol added by this PR, so it compiles and passes verbatim with the production change reverted.

Re-derived and confirmed in every particular:

- The test body calls only `AppState::new`, `crate::llm::detected_llm_config`, `set_chat_config`, `set_detect_result`, and `apply_action(.., KeyAction::ChatDetectSave)`. It names **none** of `should_revalidate_stale_chat_url`, `stale_chat_url_replacement`, `normalize_base_url`, or the new `event_loop` block. All the reducers it uses are pre-existing — the diff adds call sites, not definitions. Revert the production hunks and this test still compiles and still passes.
- The one assertion not already covered by `detect_offer_lifecycle_accept_switches_chat` and `save_detect_offer_accepts_and_raises_persist_edge` is the `/detect save` transcript check, and it passes only because the test supplies `true` for `pre_consent`. `set_chat_config` maps `(Some(_), true) -> ChatConsent::Accepted`, and `set_detect_result` pushes that turn only `if accepted`. The sole production construction site hardcodes `chat_auto_consent: false`, so at real startup consent is `Pending` and the turn is never pushed — the assertion would be false against production's own wiring.
- No CLI flag, env var, or config key sets `chat_auto_consent`; the only other `false` literal is a test helper.

**Branch-level mutation note (standing focus a).** The *pure helpers* are genuinely branch-covered: `stale_revalidation_requires_unreachable_uncredentialed_persisted_url` flips each of the five conditions in `should_revalidate_stale_chat_url` one at a time, and `stale_chat_url_replacement_normalizes_equivalent_urls` covers the `localhost` branch of `normalize_base_url`. The gap is precisely the **wiring**: no test reaches the `event_loop` block that decides whether to probe, calls `detect_local_chat`, and routes the result into `set_detect_result`. Mutating that block — e.g. dropping the `if let Some(live)` so the offer is never raised — breaks nothing in the unit suite.

**What would resolve it:** make the test assert against `false` (production's value) and cover the actual wiring. `detect_local_chat_with_probe` already exists as an injectable-probe seam, or extract the `stale_offer` block from `event_loop` into a testable function and test that. Either way the test must fail when the wiring is removed.

*(For contrast, the new e2e scenario **is** load-bearing, re-verified here: with `chat_url` set, `should_detect_local_chat` returns `false`, so the outcome is `Configured`; `resolve_llm_config` returns the configured URL first and only gates the `DEFAULT_CHAT_BASE_URL` fallback on `probe_ok`. On revert the gate would render the stale port-9 URL while the `Then` step waits for the mock's assigned port. The scenario is the only thing currently proving the feature end to end, which is exactly why the unit-level gap matters.)*

### 2. (still holds) The startup path can surface the accept-and-use affordance before the privacy notice

Our claim under test, verbatim:

> `draw_consent` picks the offer panel ahead of `consent_gate_lines` whenever `chat_detect_offer` is `Some`, regardless of consent state.

Confirmed by reading the file, which the base branch has not touched since the merge base:

- `draw_consent` branches `if state.chat_detecting { … } else if let Some(offer) = state.chat_detect_offer.as_ref() { … } else { consent_gate_lines(…) }`. `chat_consent` is never consulted before the offer branch wins, and `"Your request only leaves this machine after you accept."` lives only in the `Pending` arm of `consent_gate_lines`.
- `draw` only calls `draw_consent` while consent is not `Accepted`, so this is exactly the pre-accept surface.
- At production startup, `set_chat_config(llm, false)` leaves consent `Pending`, and `set_detect_result` sets `chat_detect_offer` unconditionally when an offer is present. So a first-run user with a stale URL gets the offer panel as the first thing on the chat tab.
- From that panel, `y`/Enter and `s` go to `ChatDetectAccept` / `ChatDetectSave`, which set `chat_consent = Accepted` directly and focus the input — they never route through the notice-bearing gate.
- Pre-PR this was unreachable: the only other writer of `chat_detect_offer` is the async `ChatDetectResult` message from `request_detect`, itself reachable only via `d` at the gate (notice already on screen) or a post-accept `/detect`. The rendering precedence is pre-existing; **this PR is what makes it reachable**, which is why it is this PR's to resolve.

**What would resolve it:** include the privacy line in the offer panel whenever consent is not yet `Accepted` (or fall through to the gate and surface the offer beneath it), plus one assertion pinning it — `offer_state_shows_endpoint_and_choices` currently asserts only the panel's contents, and neither `chat.feature` scenario 2 nor the new scenario 8 covers this path.

### 3. (still holds) "Use & save" does not clear the stale value that caused the offer

Our claim under test, verbatim:

> `persist_chat_endpoint` writes `dirs::config_dir()/rocm-dash/config.toml`, but startup's `chat_url` comes from the unified `config.json` … So after saving, the stale `tui.chat_url` in `config.json` survives and the same offer reappears on every launch.

Traced end to end and confirmed:

- `persist_chat_endpoint` resolves its path via `rocm_dash_core::config::default_config_path()`, which is `dirs::config_dir()/rocm-dash/config.toml`, and writes TOML through the standalone `rocm_dash_core::config::Config` type.
- Startup's `chat_url` is `config.dashboard.tui.chat_url` from the unified config, whose path is `AppPaths::config_path()` → `config.json`.
- The full save dispatch (`chat_persist_dispatch` → the `event_loop` tick → `persist_chat_endpoint`) touches nothing on the unified side.
- The legacy-TOML migration is gated on `config.path().is_file() || !legacy.is_file()` — it returns early once `config.json` exists, which it will after any prior run. Its own doc says it "never clobbers an existing `config.json`."
- `DashboardConfig::with_chat_endpoint` is the unified-side helper and has exactly two references: its definition and one `#[test]`.

**So yes — the same offer reappears on every subsequent launch after the user "saved."** The PR's stated problem is fixed for the session only, which directly contradicts the call-site comment that accept/save/dismiss "work exactly as they do after an explicit `/detect` probe."

**What would resolve it:** persist through the unified config (`DashboardConfig::with_chat_endpoint` is the existing seam), or state the session-only scope explicitly in both the PR description and the offer text so the user is not told something was saved that was not.

---

## Non-blocking

- **(new) `app/mod.rs` — the comment above the probe clone states a fact the code does not have.** It reads "Clone so `probe_target` stays available for the stale-URL recovery **and its notice below**." There is no notice below: `probe_target` has exactly one use after the clone, the `stale_chat_url_replacement(&probe_target, live)` call. The nearest thing to a notice is the transcript turn `set_detect_result` pushes — and that is gated on consent being `Accepted`, which never holds on the production startup path (see count 1). Classifying per standing focus (c): this is **the codebase inviting the wrong conclusion**, and the same confusion is visible in the neighbouring "work exactly as they do after an explicit `/detect` probe" comment — both read as if startup and post-`/detect` were the same state, when the whole difference is that startup is pre-consent. That single mistaken premise is plausibly the common root of counts 1, 2 and 3. The cheap prevention: a one-line comment at the `set_detect_result` call site saying explicitly that at startup consent is `Pending`, so no transcript turn is pushed and the offer renders in place of the consent gate — plus dropping "and its notice below". Fixing the comment does not fix the behaviour, but it stops the next reader inheriting the premise.
- `app/chat.rs` — `stale_chat_url_replacement_none_when_nothing_live_found` passes for a trivial implementation (its input is `None`); it constrains nothing beyond `Option`'s type.
- `app/chat.rs` — asserting `Cow::Borrowed` in `normalize_base_url_strips_suffixes_without_allocating` pins an allocation strategy rather than behaviour; a correctness-preserving refactor (e.g. lowercasing before comparison) would break it.
- `app/chat.rs` — `normalize_base_url` does not canonicalize `[::1]`, `0.0.0.0`, or scheme/host case, so those spellings still produce a spurious "different server" offer. Worth a doc caveat if out of scope. (Related nit: the trailing-slash trim runs *before* the `/v1` strip, so an input like `http://h:8000//v1` normalizes to a value with a leftover trailing slash.)
- `app/chat.rs` — when the live endpoint has the same URL but a different served model, the freshly fetched config is discarded, so the stale model id is kept even though a verified one was in hand.
- `dash_steps.rs` — the new `Given` deserves a one-line rationale (port 9 as a canonically closed port; and that this writes the CLI's `config.json`, distinct from the dash TOML that `/detect save` writes), matching the sibling steps' commenting style and flagging the JSON/TOML split above.

---

## Positive signals

- The new pure helpers are cleanly separated and their unit tests are genuinely branch-level — each of the five conditions in `should_revalidate_stale_chat_url` is flipped individually, and the `localhost`/`127.0.0.1` equivalence is covered with a table. That is the standard the `event_loop` wiring should be held to as well.
- The e2e scenario is the one test here that would actually fail if the production change were reverted; keeping it is right.

## Note on the trust boundary

Nothing in the diff, the commit messages or the PR text contained instructions addressed to a reviewer. The internal-style ticket identifiers in the new comments match the repository's own documented convention in `AGENTS.md`, which explicitly permits them, so they are not flagged.

## Context (does not affect the finding)

This branch is currently conflicting with its base and needs a rebase regardless, and a second reviewer holds an independent change request, so these counts are not the only thing outstanding. Neither fact bears on whether the findings are correct.

PR #110, opened 2026-07-14
docs: permit Jira/EAI ticket IDs in branches, commits, and comments

rominf:
## Summary
- The OSS contribution guide (`AGENTS.md`) previously flagged any Jira/ticket-style identifier as a leak of internal tracking info.
- The team has decided bare Jira/EAI-style ticket IDs (e.g. `EAI-1234`) are fine to use in branch names, commit messages, and code/doc comments.
- Updates the leak-scan pattern and content-restriction guidance to stop treating those IDs as a violation, while still flagging actual links to internal tracking systems.

## Test plan
- [x] Docs-only change; no code/build gates apply (`cargo test`/`clippy`/`smoke_local.py` skipped as not relevant)
- [x] Re-ran the leak scan from the doc against this diff — only self-referential matches on the word "internal" in the policy text itself

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -56,7 +56,7 @@ Content restrictions for upstream surfaces:
 
 - do not include internal/proprietary names, aliases, URLs, hostnames, gateways, cluster names, or registry paths
 - do not include links to internal tracking systems or unrelated internal usernames
-- bare Jira/EAI-style ticket IDs (e.g. `EAI-1234`) are permitted in branch names, commit messages, and code/doc comments; do not flag them
+- bare Jira/EAI-style ticket IDs (e.g. `EAI-1234`) are permitted anywhere this rule applies; do not flag them
 - apply this rule to PR titles/bodies, issue text, comments, review replies, commit messages, branch names, code comments, fixtures, and logs
 
 Use neutral external framing (for example: "backend" or "gateway") rather than internal or vendor-specific ownership phrasing.
@@ -64,7 +64,7 @@ Use neutral external framing (for example: "backend" or "gateway") rather than i
 Leak scan before each upstream push/PR/comment batch:
 
 ```bash
-INTERNAL_KEYWORDS_PATTERN='internal|confidential|proprietary|private|confluence|\.corp|\.internal'
+INTERNAL_KEYWORDS_PATTERN='internal|confidential|proprietary|private|jira|confluence|\.corp|\.internal'
 git diff <upstream-base>..HEAD \
   | grep -inE "$INTERNAL_KEYWORDS_PATTERN" \
   && echo "REVIEW each hit" || echo "diff clean"
@@ -105,7 +105,7 @@ grep -rn "^<<<<<<<\|^=======\|^>>>>>>>" .
 For leak scans, use upstream base `origin/main` (or upstream default branch if different):
 
 ```bash
-INTERNAL_KEYWORDS_PATTERN='internal|confidential|proprietary|private|confluence|\.corp|\.internal'
+INTERNAL_KEYWORDS_PATTERN='internal|confidential|proprietary|private|jira|confluence|\.corp|\.internal'
 git diff origin/main..HEAD \
   | grep -inE "$INTERNAL_KEYWORDS_PATTERN" \
   && echo "REVIEW each hit" || echo "diff clean"
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -55,7 +55,8 @@ the full range on push (`verify-commits`), and a blocking CI gate
 Content restrictions for upstream surfaces:
 
 - do not include internal/proprietary names, aliases, URLs, hostnames, gateways, cluster names, or registry paths
-- do not include internal ticket identifiers, links to internal tracking systems, or unrelated internal usernames
+- do not include links to internal tracking systems or unrelated internal usernames
+- bare Jira/EAI-style ticket IDs (e.g. `EAI-1234`) are permitted in branch names, commit messages, and code/doc comments; do not flag them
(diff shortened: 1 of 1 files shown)


volen-silo:
🔴 Automated review · pr-review-watcher · 04b39f6a

## Summary
Docs-only policy change to `AGENTS.md` that permits bare Jira/EAI ticket IDs (e.g. `EAI-1234`) in upstream surfaces and removes `ticket-[0-9]+`/`jira` from the leak-scan pattern. Verdict: **Needs work** — the intent is sound but the permission's surface scope is narrower than the enforcement scope, leaving it ambiguous whether IDs are allowed in PR titles/issue text. Verified (grep-tested): both leak-scan copies (§2 L67, §4 L108) updated byte-identically; no other file still enforces the old prohibition; `EAI-1234` is correctly no longer flagged; the old `ticket-[0-9]+` never matched `EAI-1234` anyway (so the regex edit is largely cosmetic for that example — the real relaxation is prose); dropping `jira` means Jira links on non-`.internal` hosts (e.g. `jira.example.com`) are no longer caught. Blocking: 1 · Non-blocking: 3.

`AGENTS.md:59` vs `AGENTS.md:60` — **Permission scope contradicts the enforcement scope, defeating the PR's own "unambiguous permit" intent.** The removed line previously prohibited "internal ticket identifiers" across *all* surfaces the rule applies to (L60: PR titles/bodies, issue text, comments, review replies, commit messages, branch names, code comments, fixtures, logs — 8+ surfaces). The new permission (L59) only names 3 of them ("branch names, commit messages, and code/doc comments") and ends with "do not flag them." A reader cannot tell whether an `EAI-1234` in a **PR title, issue body, review reply, fixture, or log** is permitted or still disallowed — the very ambiguity this PR set out to remove. Fix: make L59's surface list match L60 exactly (e.g. "…are permitted anywhere this rule applies") or state the narrower scope is deliberate and say why titles/issue text are excluded.

## Non-blocking
- `AGENTS.md:67`,`AGENTS.md:108` — Removing `jira` unenforces the *retained* "no links to internal tracking systems" rule (L58) for Jira hosts without `.internal`/`.corp`/`confluence` in them: grep-verified that `https://jira.example.com/browse/EAI-1234` was caught before and is not now. The pattern is self-described as best-effort (L73) with mandated manual review, so it degrades a heuristic rather than a hard guarantee — but consider keeping a Jira-specific term, or note in the doc that Jira-link detection now relies on host conventions.
- `AGENTS.md:67`,`AGENTS.md:108` — The regex edit is effectively cosmetic for the `EAI-1234` example (old `ticket-[0-9]+` never matched it); the PR/Test-plan framing implies a functional scan fix when the behavioral change is the prose. Clarify in the description.
- `AGENTS.md:67`,`AGENTS.md:108` — Dropping `jira` also stops flagging plain prose mentions of the word "jira" (a product name, not a leak); harmless but an undocumented side effect of the pattern change.

fredespi (approved):

rominf:
Addressed both points from the automated review:

- **Blocking (scope mismatch):** Widened L59 to "permitted anywhere this rule applies" so it now matches L60's full enforcement scope (PR titles/bodies, issue text, comments, review replies, commit messages, branch names, code comments, fixtures, logs). Ticket IDs are already permitted in commit messages, which persist indefinitely in git history — that's a stronger form of exposure than a PR title or issue body, so there's no reason to keep those surfaces restricted.
- **Non-blocking (`jira` removal):** Restored `jira` in both leak-scan pattern copies (L67, L108). Only `ticket-[0-9]+` needed to go to stop flagging bare IDs like `EAI-1234` — dropping `jira` too was an unnecessary side effect that would've stopped catching actual Jira links (e.g. `jira.example.com/browse/...`), which are still supposed to be flagged under the retained "no links to internal tracking systems" rule. This also restores flagging of plain `jira` mentions, addressing your third point.
- **Non-blocking (cosmetic regex framing):** Confirmed — the `ticket-[0-9]+` removal was indeed cosmetic for the `EAI-1234` example (it never matched that format). The functional change is the prose permission in L59/L60, not the regex. Noting here for the record since it's a description clarification rather than a code change.

Pushed in 97a6513.

volen-silo:
🔴 Automated review · pr-review-watcher · 97a65135

## Summary
Docs/policy edit to `AGENTS.md` that relaxes the OSS leak policy to permit bare Jira/EAI-style ticket IDs (e.g. `EAI-1234`) on upstream surfaces while still barring links to internal tracking systems. Verdict: **Approve**. Verified: the prior scope-contradiction blocker is **resolved** — the permission now reads "permitted anywhere this rule applies" (`AGENTS.md:59`) and binds to the full surface enumeration on `AGENTS.md:60` (PR titles/bodies, issue text, comments, review replies, commit messages, branch names, code comments, fixtures, logs), so no surface is left ambiguous; the old "internal ticket identifiers" prohibition is fully deleted with no contradictory leftover; both leak-scan patterns (`:67`, `:108`) are byte-identical, correctly drop `ticket-[0-9]+`, and confirmed not to match bare `EAI-1234`. Blocking: 0 · Non-blocking: 2.

**None.**

## Non-blocking
- PR description — the "Test plan" claims the re-run leak scan produced "only self-referential matches on the word 'internal'", but the current diff also matches `jira`/`Jira` (the added permission line and the two added `INTERNAL_KEYWORDS_PATTERN` lines). The scan was evidently run before `jira` was restored; refresh the claim so the verification note is accurate.
- `AGENTS.md:67`,`:108` — the `jira` keyword only matches the literal substring "jira", so it won't catch Atlassian-hosted tracking links that lack it (e.g. `https://<org>.atlassian.net/browse/EAI-1234`), which the "links to internal tracking systems" rule (`:58`) intends to flag. Pre-existing heuristic gap, not introduced by this PR and mitigated by the documented "REVIEW each hit" manual triage — noted only for awareness.

PR #111, opened 2026-07-14
fix(install): actionable error when GPU family has no wheels

volen-silo:
## Problem

On a clean first run, `rocm install sdk` auto-detects the GPU, normalizes it to a TheRock
package family, and resolves a wheel/tarball index for that family. When the detected family
has **no published index** on the selected channel, the fetch returns HTTP 404 and the command
died with an opaque error:

```
failed to resolve TheRock release wheel runtime from candidate indexes:
  - https://repo.amd.com/rocm/whl/<family>: HTTP 404 while fetching ...
  - https://repo.amd.com/rocm/whl-multi-arch/<family>: HTTP 404 while fetching ...
```

Nothing told the user that `--family` was the escape hatch or which values are valid, so the
only recovery was to guess a working family.

## Fix

Turn the opaque failures into actionable guidance:

- Added `known_therock_families()` in `rocm-core` — the canonical set of families the CLI
  recognizes (the full output set of `normalize_therock_family`), guarded by a round-trip test
  so it can't drift from the normalizer.
- Added `family_resolution_hint()` — a pure, tested message builder that names the resolved
  family, lists the valid `--family` values, and suggests the alternate channel/format. Wording
  differs for auto-detected vs. user-supplied families.
- Wired the hint into the wheel-index failure (the reported 404 path), the tarball
  "no matching artifact" failure, and the "no supported family" bail.

### Deliberately not done

No auto-fallback to a nearby family: installing wheels built for a different architecture can
silently produce a broken runtime, which is worse than a clear stop on a first-run path.

## Testing

- New unit tests for the family list round-trip and the message builder (auto-detected vs.
  user-supplied, channel suggestions).
- `cargo test -p rocm-core -p rocm`, `cargo clippy --all-targets`, `cargo fmt --check` all clean.

The end-to-end 404 path depends on live index endpoints and real hardware, so it is covered here
by the message-builder unit tests and left to the manual setup checklist for on-hardware
confirmation.

### DIFF
--- a/apps/rocm/src/therock.rs
+++ b/apps/rocm/src/therock.rs
@@ -5,12 +5,13 @@
 use anyhow::{Context, Result, bail};
 use rocm_core::{
     AppPaths, ManagedToolConfig, RocmCliConfig, detect_host_gpu_diagnostics,
-    detect_host_therock_family, detect_managed_therock_family, ensure_uv_binary, managed_tools_dir,
-    normalize_runtime_path_for_host, normalize_runtime_path_for_storage,
-    normalize_runtime_path_text_for_host, normalize_runtime_path_text_for_storage,
-    normalize_therock_family, runtime_is_windows, runtime_os_name, runtime_path_for_windows_child,
-    runtime_path_list_split, runtime_python_executable_in_env, unix_time_millis, uv_command_env,
-    uv_pip_install_base, uv_venv_args, verify_rsa_pkcs1_sha256_signature,
+    detect_host_therock_family, detect_managed_therock_family, ensure_uv_binary,
+    known_therock_families, managed_tools_dir, normalize_runtime_path_for_host,
+    normalize_runtime_path_for_storage, normalize_runtime_path_text_for_host,
+    normalize_runtime_path_text_for_storage, normalize_therock_family, runtime_is_windows,
+    runtime_os_name, runtime_path_for_windows_child, runtime_path_list_split,
+    runtime_python_executable_in_env, unix_time_millis, uv_command_env, uv_pip_install_base,
+    uv_venv_args, verify_rsa_pkcs1_sha256_signature,
 };
 #[cfg(test)]
 use rocm_core::{generate_rsa_signing_keypair, sign_rsa_pkcs1_sha256_signature};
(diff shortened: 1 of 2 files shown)


volen-silo:
🔴 Automated review · pr-review-watcher · 06950a1c

## Summary
Turns opaque HTTP-404 / "no artifact" install failures into actionable guidance by adding `known_therock_families()` (rocm-core) and a tested `family_resolution_hint()` (therock.rs), wired into the wheel-index, tarball, and no-family-detected failure paths. Verdict: **Approve**. Verified: `cargo test -p rocm-core -p rocm` (169 + apps tests pass, incl. all 5 new tests), `cargo clippy -p rocm-core -p rocm --all-targets` and `cargo fmt --check` clean. Confirmed the load-bearing claims: `known_therock_families()` is the complete output set of `normalize_therock_family` (all 16 reachable outputs present, every listed entry round-trips to itself — incl. the `gfx90X-dgpu` vs `gfx90X-dcgpu` split); source classification (`host`/`managed-runtime` = auto vs `manifest`/`env` = user-supplied) matches every `source` value `resolve_family` can emit; the `--format tarball` suggestion is correctly suppressed on Windows (matching the existing tarball-Windows rejection); and `--channel`/`--format`/`--family` are all real clap flags. No internal ticket IDs (AGENTS.md §2 clean). Blocking: 0 · Non-blocking: 2.

**None.**

## Non-blocking
- crates/rocm-core/src/lib.rs:6060 — `known_therock_families_all_round_trip` proves listed→self but not completeness; a future new arm in `normalize_therock_family` could drift the user-facing list undetected. Consider a test asserting each normalizer output is contained in the list.
- apps/rocm/src/therock.rs:4145 — the 3 hint tests only exercise `format="wheel"`; the tarball branch (`--format wheel` suggestion), the Windows-wheel branch (no `--format` suggestion), and the `managed-runtime`/`env` sources are untested, so a regression in `alternate_format` (e.g. reintroducing `--format tarball` on Windows) would pass silently.

michaelroy-amd (approved):
Reviewed the install-family guidance and focused tests. The known-family list round-trips through normalization, source-specific hints cover all three failure paths, and Windows correctly omits the unsupported tarball suggestion. The earlier Windows Strix Halo failures were cold backend-install timing rather than this diff; a focused rerun of both failed scenarios passed at the same head: https://github.com/ROCm/rocm-cli/actions/runs/29445489844.

PR #112, opened 2026-07-14
feat(dash): mark demo and replay sessions as simulated

volen-silo:
## Problem

`rocm dash --demo` and `rocm dash --replay <file>` render synthetic/recorded telemetry through
the exact same UI as a live daemon, so nothing on screen (or in generated screenshots/casts)
distinguishes simulated data from a live system. Several indicators were also fabricated:

- A hardcoded `LIVE · last 60s`, a `live` sparkline label, and `Updates: Up to date` keyed only
  on being connected — which replay also sets.
- A launcher `✓ healthy` line derived purely from a process being *Running*, with no health probe.
- The one existing "demo data" banner was wired backwards: it only fired when *disconnected*, so
  it never appeared during an actual demo/replay session.
- The demo session was written to a fixed, world-readable path in the shared temp dir.

## Changes

Introduce a single explicit `AppState.simulated` signal (set for `--demo`, `--replay`, and the
asset generators) and make every honesty-relevant surface read it:

- **Persistent SIMULATED DATA marker** in the shared header, so it appears on every tab and in
  every generated asset (screenshots + casts).
- **Live indicators gated on observed state**: the header connection line, home's `LIVE · last 60s`
  / `live` sparkline label, and the `Up to date` tile no longer present simulated data as live.
- **Launcher health honesty**: a Running process is not a health signal — report `health: unknown`
  instead of a fabricated `healthy` (no probe exists today; an active one-shot probe is a proposed
  follow-up).
- **Secure demo session file**: written under the per-user data dir with an unpredictable name and
  `0o600` permissions, replacing the fixed world-readable temp path.
- **Corrected the Observe banner**: the old "demo data" text (which only fired when disconnected)
  now reads "no live data" and no longer collides with the new simulated marker.

## Testing

New regression tests cover the SIMULATED marker on every tab, its absence when live, home
indicator suppression, both asset generators, the corrected banner semantics, and the secure
demo-path (location + `0o600`/`0o700`). `cargo test --workspace --all-targets`,
`cargo clippy --workspace --all-targets`, and `cargo fmt --check` all clean.

On-terminal visual confirmation of the banner and generated assets belongs in the manual test
checklist.

### DIFF
--- a/apps/rocm/src/dash.rs
+++ b/apps/rocm/src/dash.rs
@@ -14,7 +14,7 @@
 //! The rest of `rocm` is synchronous; the async daemon/TUI run on a tokio
 //! runtime built here. The TUI lives entirely in the `rocm-dash-tui` crate.
 
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::time::Duration;
 
 use anyhow::{Context, Result};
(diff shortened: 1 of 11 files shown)


volen-silo:
🔴 Automated review · pr-review-watcher · c53e0ed5

## Summary
Adds an explicit `AppState.simulated` signal so `--demo`/`--replay`/asset-generator sessions carry a persistent SIMULATED DATA marker and stop presenting synthetic telemetry as live; also secures the demo session file (per-user dir, `0o600`) and corrects the mis-wired Observe banner and the fabricated launcher "healthy" line. Verdict: **Needs work** — the honesty goal is sound and well-tested, but the `simulated` gate is too broad in one place and the demo-dir creation reintroduces a TOCTOU window this repo already fixed elsewhere. Verified: `cargo test -p rocm-dash-tui -p rocm-dash-daemon -p rocm --all-targets` all pass (338 in `rocm`, 19/19 characterization incl. the new every-tab/absent-when-live tests, both example tests); confirmed against source that replay end/error genuinely drives `ConnState::Disconnected` (replay.rs:99-111 → app/mod.rs:1669) which the header then hides, and that the `DirBuilder::mode` pattern exists in server.rs:98 / agent.rs:1190. No EAI/ticket IDs in the diff (AGENTS.md §2 clean). Blocking: 2 · Non-blocking: 5.

**`crates/rocm-dash-tui/src/ui/mod.rs:296` — `simulated` permanently masks the real connection status.** The status text is hardcoded to `"simulated — not live"` for the whole session because `simulated` is never cleared. But during `--replay`, `replay::spawn` emits `Disconnected { reason: "end of recording" }` on completion and `Disconnected { reason: <error> }` on a failed/corrupt replay file (replay.rs:99-111), both wired into `state.conn` at app/mod.rs:1669. The user therefore never sees that the recording ended or that the file failed to load — a real information-loss regression this PR introduces. Fix: keep the SIMULATED DATA chip gated on `simulated` (it should stay permanent), but let the status-text branch still render `ConnState::Disconnected { reason }` (show the reason) instead of the blanket simulated label — i.e. only override the `Connected` case, not `Disconnected`.

**`apps/rocm/src/dash.rs:224-227` — TOCTOU/umask window when creating the demo dir.** `create_dir_all(&dir)` creates `{data_dir}/demo` at the process umask (typically world-readable `0o755`) and only *afterward* applies `set_permissions(0o700)`, whose result is discarded (`let _ =`). The dir is briefly world-readable at creation, and permanently so if the chmod fails. This repo already fixed this exact pattern twice — `server.rs:98` and `agent.rs:1190` both use `DirBuilder::new().recursive(true).mode(0o700).create(...)` with an explicit comment about "the brief window before set_permissions." Fix: use `DirBuilder` with `.mode(0o700)` here too (keeping the follow-up `set_permissions` for the pre-existing-dir case), matching the established pattern. Severity is bounded by the `0o600` file mode, but it reintroduces a documented in-house anti-pattern.

## Non-blocking
- `apps/rocm/src/dash.rs:216` / test name `demo_session_path_is_private_and_unpredictable` / PR body — "unpredictable name" is overstated; `session-{pid}-{millis}` is guessable. Confidentiality rests on the `0o700`/`0o600` perms, not name secrecy — reword to "unique" and note perms are the boundary.
- `crates/rocm-dash-tui/src/ui/tabs/observe.rs:36` — `telemetry_absent()` is unchanged, so at end-of-recording/backward-seek the amber "no live data" banner can appear alongside the SIMULATED chip (two honesty signals for different reasons).
- `crates/rocm-dash-tui/src/ui/tabs/home.rs:452` — the "tile shows conn-derived status" comment is now stale; the primary branch is `state.simulated`.
- `apps/rocm/src/dash.rs:233-241` — prune deletes all `*.ndjson` indiscriminately; two concurrent `rocm dash --demo` runs can delete each other's in-use session file. Low severity; consider skipping the file about to be written.
- No test for the launcher `health: unknown` change, the Updates-tile `unknown` branch, or the sparkline `sim` label (home.rs / launcher.rs) — behavior is correct by inspection but uncovered.

volen-silo:
### Review feedback addressed in c89fa3e

**Blocking**
- **Status masking (`ui/mod.rs`)** — the simulated label now replaces only the `Connected` case; `Disconnected { reason }` (end-of-recording, failed replay file) is rendered again, so playback status is no longer hidden. The permanent SIMULATED DATA chip is unchanged.
- **Demo-dir TOCTOU (`dash.rs`)** — the dir is now created with `DirBuilder::new().recursive(true).mode(0o700)` (no umask window), matching the `server.rs`/`agent.rs` pattern; the follow-up `set_permissions` remains for the pre-existing-dir case.

**Non-blocking**
- Prune now keeps the file about to be written and any session modified within the last hour, so concurrent `rocm dash --demo` runs can't delete each other's in-use file.
- Reworded "unpredictable" → "unique" (dir `0o700`/file `0o600` are the confidentiality boundary, not name secrecy).
- Fixed the stale Updates-tile comment.
- Added tests: launcher reports `health: unknown` (never a fabricated "healthy"), the Updates tile reads `unknown` when simulated, and the node-load sparkline is labeled `sim` (never `live`) when simulated.

**CodeQL "uncontrolled data used in path expression" (6 alerts)** — I believe these are false positives for this tool and plan to dismiss them (won't-fix) unless you'd prefer a code change:
- The tainted input is the operator's own `ROCM_CLI_DATA_DIR` / data directory (used identically across all of `AppPaths`) and the explicit `--replay <path>` argument. On a local, single-user CLI the operator already controls these paths and can write anywhere they have permission — no privilege or trust boundary is crossed by using them in a path expression.
- Two of the six alerts are in test code (`dash.rs:535,543`), which builds paths from a temp dir + PID.
- The genuine hardening here (private `0o700` dir / `0o600` file, no umask window) is already implemented above.

Happy to add explicit path validation if the team would rather silence these in code than dismiss them.

michaelroy-amd (approved):
Reviewed simulated-state propagation and the prior findings. Demo/replay state is explicit and runtime-only, live/health labels no longer misrepresent simulated sessions, disconnected reasons remain visible, and the demo directory/file are created with restrictive Unix modes without an umask window. Focused dashboard tests and CI are green.

PR #113, opened 2026-07-14
fix(lemonade): stop silently substituting valid Hugging Face model ids

rominf:
## Summary

`rocm serve Qwen/Qwen2.5-1.5B-Instruct` printed a deployment summary for `Qwen3-4B-Instruct-2507-GGUF`, while `rocm services` afterwards showed the requested `Qwen/Qwen2.5-1.5B-Instruct` — the two commands appeared to disagree about which model was running. In fact the requested model was silently swapped for an unrelated one, and the served `/v1/models` id (and the `model` field in chat completions) reflected the substitute, not the request.

## Root cause

`resolve_lemonade_model_ref` aliased any reference whose lowercased form merely **contained** the substring `qwen2.5-1.5b` (or `qwen3.5-0.8b`) to the bundled default model. The fully-qualified Hugging Face checkpoint id `Qwen/Qwen2.5-1.5B-Instruct` contains that substring, so it was rewritten to `Qwen3-4B-Instruct-2507-GGUF` with no notice on stdout, stderr, or in the logs. `rocm serve`'s summary shows the resolved id; `rocm services` reads back the stored raw request — hence the mismatch.

## Fix

- **Stop the silent substitution.** Drop the substring checks; only exact, recognized shorthand aliases (`qwen`, `qwen3-4b-instruct`, …) still resolve to the bundled GGUF. A syntactically valid `owner/repo` checkpoint now passes through unchanged, so the serve path either honours it or rejects it explicitly. `Qwen/Qwen2.5-1.5B-Instruct` is safetensors-only, so on Lemonade it now hits the existing "needs an explicit `owner/repo:variant`" message instead of serving a different model behind the user's back.
- **Report both identities.** When the requested reference resolves to a different canonical id, the deployment summary now shows `requested model` and `resolved model` rows (mirroring the plain-text serve output); a single `model` row is kept when they match, so the common case stays uncluttered.
- **README.** The quick-start used the exact substituted command, and the guidance claimed full HF ids are reliable cross-engine. Both are corrected: the quick-start uses the `qwen` alias (works on both engines), and the guidance now states the real split — Lemonade serves GGUF via `owner/repo:variant`, vLLM serves safetensors repos.

## Out of scope (follow-up)

The built-in catalog still lists this safetensors model as a llama.cpp/GGUF recipe with no real GGUF artifact — the removed substring hack was masking that. Making it genuinely Lemonade-servable (mapping it to a real GGUF build) is left as a separate change rather than silently substituting one here.

## Test plan

- New regression test: `resolve_lemonade_model_ref("Qwen/Qwen2.5-1.5B-Instruct")` returns it verbatim (this assertion fails before the change — the previous test codified the buggy substitution).
- New summary tests: matching request/resolved shows a single `model` row; divergent shows both `requested model` / `resolved model`.
- `cargo test -p rocm-engine-lemonade` (27 passed) and `cargo test -p rocm --bin rocm` (339 passed).
- `cargo fmt --all --check` and `cargo clippy --workspace --all-targets -- -D warnings` clean.

## Risk

Low. The resolver change only narrows an over-broad alias match; built-in aliases and the default-model path are unchanged. The summary change is additive and only alters output when requested/resolved differ.

Relates to EAI-7370.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,8 +203,13 @@ creates a separate managed runtime alongside it.
 Then serve a model:
 
 ```
-rocm serve Qwen/Qwen2.5-1.5B-Instruct
+rocm serve qwen
 ```
+
+`qwen` is a built-in alias for a small assistant model that serves out of the
+box. You can also serve any compatible Hugging Face model directly — see
+[Model serving](#model-serving) for the GGUF-vs-safetensors rule, since which
+form works depends on the engine your GPU selects.
 
 Use `rocm model` to see available model recipes and their GPU memory
 requirements.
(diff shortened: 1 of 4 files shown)


michaelroy-amd (approved):
Reviewed model resolution and deployment-summary behavior. Exact aliases still resolve to bundled defaults, valid Hugging Face model IDs now pass through unchanged, requested/resolved model reporting is consistent at call sites, and focused engine/summary tests plus CI are green.

PR #114, opened 2026-07-14
test(e2e): drive the dash TUI black-box over a pseudo-terminal

rominf:
## Summary

- Adds a pseudo-terminal driver (`portable-pty` + `vt100`) so the Cucumber suite can launch the real dash/chat raw-mode event loop, send keystrokes, resize the terminal, and assert on the current rendered screen rather than an escape-sequence transcript.
- Covers the interactive dashboard end to end: Home-to-ROCm navigation, help and command-palette discovery, managed-model visibility and loading status, live serving metrics, offline chat, and a managed local chat exchange using the registry-provided model and OS-assigned endpoint.
- Turns the local-endpoint privacy gate into passing black-box coverage and removes its former expected-failure entry.
- Removes the unused `ratatui` dependency from the `rocm` binary, eliminating the conflicting older dependency line that blocked the terminal emulator dependency.

## Why

The existing in-process render tests verify component behavior, but a piped `Command` never enters crossterm raw mode and therefore cannot prove that users can complete interactive terminal journeys. This adds the missing acceptance layer while retaining the in-process tests as the fast rendering layer.

## Design notes

- Piped and PTY launches share the same isolated config/data/cache environment; PTY scenarios additionally isolate `HOME` and `XDG_RUNTIME_DIR` so a host daemon cannot mask the scenario's embedded daemon and managed-service registry.
- The managed-service fixture uses plain on-disk JSON and an OS-assigned mock endpoint, keeping the scenarios black-box and free of GPU/network requirements.
- The metrics fixture serves deterministic vLLM-compatible counters and histograms; scenarios assert the rendered latency, queue, and KV-cache values through the real daemon scrape path.
- Waits poll visible screen state against deadlines, with the final screen included on failure. Teardown kills and reaps the child before joining the PTY reader.

## Scope / risk

- **Linux-only** for now (`@requires-os:linux`). The dependency supports ConPTY, but Windows is not a blocking contract until that path has dedicated stability coverage.
- Risk: low. The change is test infrastructure and scenarios plus one unused dependency removal; product behavior is unchanged.

## Test plan

- `cargo fmt --all --check`
- `cargo clippy -p e2e-cucumber --test e2e -- -D warnings`
- `cargo test -p e2e-cucumber --lib` — 23 passed
- Focused expanded PTY journeys — 6 scenarios / 39 steps passed, 0 xfail, 0 XPASS
- `cargo xtask e2e` — 17 passed, 3 existing expected failures, 0 XPASS, 0 unexpected failures
- Push hooks: commit signatures/sign-off, format, clippy, unit tests, license checks all passed

### DIFF
--- a/tests/e2e-cucumber/features/chat.feature
+++ b/tests/e2e-cucumber/features/chat.feature
@@ -10,8 +10,8 @@ Feature: Chat and endpoint detection
   # Driven through the interactive TUI (a pseudo-terminal): the privacy notice
   # lives on the chat consent gate, which only the real terminal path renders.
   # The model is registered so the CLI discovers it on its OS-assigned port and
-  # offers it — the notice must precede any request. (Was EAI-7222, previously
-  # an untestable-black-box gap.)
+  # offers it — the notice must precede any request. (Previously an
+  # untestable-black-box gap.)
   @id:chat-privacy-notice-accurate @requires-os:linux
   Scenario: 2 - The privacy notice is shown before using a local endpoint
     Given a model is being served locally
@@ -28,6 +28,7 @@ Feature: Chat and endpoint detection
     When the user accepts the local endpoint
     And the user sends a message to the managed model
     Then the managed model's response is displayed
+    And the mock received the typed prompt
     When the user quits interactive chat
     Then interactive chat exits successfully
 
--- a/tests/e2e-cucumber/features/chat.feature
+++ b/tests/e2e-cucumber/features/chat.feature
@@ -20,36 +20,47 @@ Feature: Chat and endpoint detection
     Then the local endpoint is shown for confirmation
     And the privacy notice is shown before any message is sent
 
+  @id:chat-managed-model-interactive @requires-os:linux
+  Scenario: 3 - Interactive chat uses a running managed model
+    Given a running managed model is available locally
+    When the user opens interactive chat
+    Then the local endpoint is shown for confirmation
+    When the user accepts the local endpoint
+    And the user sends a message to the managed model
+    Then the managed model's response is displayed
+    When the user quits interactive chat
+    Then interactive chat exits successfully
+
   @id:chat-endpoint-shown-in-services
-  Scenario: 3 - A served model's endpoint is shown in the services list
+  Scenario: 4 - A served model's endpoint is shown in the services list
     Given a model is being served
     And the model is registered with the CLI
     When the user lists running services
     Then the served model endpoint is listed
 
-  # Exercises the `rocm chat` CLI itself (one-shot `--prompt` via the local
-  # provider), not just the model endpoint over HTTP — so the command surface
-  # reports `rocm chat` as covered. Runs on mock (no GPU): the local provider
-  # resolves the planted managed-service record and talks to the mock server.
-  @id:chat-cli-oneshot-prompt
-  Scenario: 6 - The chat CLI answers a one-shot prompt against a local server
-    Given a model is being served
-    And the model is registered with the CLI
-    When the user sends a one-shot chat prompt through the CLI
-    Then the CLI prints the assistant's reply
-
   @id:chat-tool-definitions-accepted @requires-gpu
-  Scenario: 4 - Chat requests that include tool definitions are accepted
+  Scenario: 5 - Chat requests that include tool definitions are accepted
     Given a managed runtime is active
     And a model is served in the background
     When a chat request with tool definitions is sent
     Then the chat response is successful
 
   @id:chat-end-to-end-local-model @requires-gpu
-  Scenario: 5 - End-to-end chat through a locally served model
+  Scenario: 6 - End-to-end chat through a locally served model
     Given a managed runtime is active
     And a model is served in the background
     And the served model has been detected
     When the user sends a chat message
     Then the chat response is successful
     And the response contains a model-generated reply
+
+  # Exercises the `rocm chat` CLI itself (one-shot `--prompt` via the local
+  # provider), not just the model endpoint over HTTP — so the command surface
+  # reports `rocm chat` as covered. Runs on mock (no GPU): the local provider
+  # resolves the planted managed-service record and talks to the mock server.
+  @id:chat-cli-oneshot-prompt
+  Scenario: 7 - The chat CLI answers a one-shot prompt against a local server
+    Given a model is being served
+    And the model is registered with the CLI
+    When the user sends a one-shot chat prompt through the CLI
+    Then the CLI prints the assistant's reply
--- a/tests/e2e-cucumber/features/chat.feature
+++ b/tests/e2e-cucumber/features/chat.feature
@@ -19,6 +19,8 @@ Feature: Chat and endpoint detection
     When the user opens interactive chat
     Then the local endpoint is shown for confirmation
     And the privacy notice is shown before any message is sent
+    When the user quits interactive chat
+    Then interactive chat exits successfully
 
   @id:chat-managed-model-interactive @requires-os:linux
   Scenario: 3 - Interactive chat uses a running managed model
--- a/tests/e2e-cucumber/features/chat.feature
+++ b/tests/e2e-cucumber/features/chat.feature
@@ -7,11 +7,18 @@ Feature: Chat and endpoint detection
     When the user checks for running services
     Then the served model is listed
 
-  @id:chat-privacy-notice-accurate
-  Scenario: 2 - The privacy notice is accurate for local endpoints
+  # Driven through the interactive TUI (a pseudo-terminal): the privacy notice
+  # lives on the chat consent gate, which only the real terminal path renders.
+  # The model is registered so the CLI discovers it on its OS-assigned port and
+  # offers it — the notice must precede any request. (Was EAI-7222, previously
+  # an untestable-black-box gap.)
+  @id:chat-privacy-notice-accurate @requires-os:linux
+  Scenario: 2 - The privacy notice is shown before using a local endpoint
     Given a model is being served locally
-    When the user is offered the detected endpoint
-    Then the notice does not claim that requests leave the machine
+    And the model is registered with the CLI
+    When the user opens interactive chat
+    Then the local endpoint is shown for confirmation
+    And the privacy notice is shown before any message is sent
 
   @id:chat-endpoint-shown-in-services
   Scenario: 3 - A served model's endpoint is shown in the services list
--- a/tests/e2e-cucumber/features/dash.feature
+++ b/tests/e2e-cucumber/features/dash.feature
@@ -1,0 +1,24 @@
+Feature: Interactive dashboard
+
+  # These scenarios drive the real interactive TUI through a pseudo-terminal —
+  # the crossterm raw-mode event loop that a piped command can't reach. Linux
+  # only for now: portable-pty compiles on Windows via ConPTY, but that path is
+  # not yet promoted to a blocking contract (tracked as a follow-up).
+
+  @id:dash-opens-and-navigates @requires-os:linux
+  Scenario: 1 - A user opens the dashboard and navigates to ROCm setup
+    When the user opens the dashboard with demo data
+    Then the dashboard home view is displayed
+    When the user opens the ROCm view
+    Then ROCm setup actions are displayed
+    When the user quits the dashboard
+    Then the dashboard exits successfully
+
+  @id:dash-chat-offline-reply @requires-os:linux
+  Scenario: 2 - A user receives a response in interactive chat
+    Given interactive chat uses an offline assistant
+    When the user opens interactive chat
+    And the user sends a message about GPU health
+    Then the assistant's GPU status response is displayed
+    When the user quits the dashboard
+    Then the dashboard exits successfully
--- a/tests/e2e-cucumber/features/dash.feature
+++ b/tests/e2e-cucumber/features/dash.feature
@@ -20,8 +20,8 @@ Feature: Interactive dashboard
     When the user opens interactive chat
     And the user sends a message about GPU health
     Then the assistant's GPU status response is displayed
-    When the user quits the dashboard
-    Then the dashboard exits successfully
+    When the user quits interactive chat
+    Then interactive chat exits successfully
 
   @id:dash-loading-service-status @requires-os:linux
   Scenario: 3 - The dashboard reports a model that is still loading as loading
--- a/tests/e2e-cucumber/features/dash.feature
+++ b/tests/e2e-cucumber/features/dash.feature
@@ -22,3 +22,49 @@ Feature: Interactive dashboard
     Then the assistant's GPU status response is displayed
     When the user quits the dashboard
     Then the dashboard exits successfully
+
+  @id:dash-loading-service-status @requires-os:linux
+  Scenario: 3 - The dashboard reports a model that is still loading as loading
+    Given a managed model is still loading
+    When the user opens the dashboard
+    And the user opens the Observe view
+    Then the managed model is shown as loading rather than ready
+    When the user quits the dashboard
+    Then the dashboard exits successfully
+
+  @id:dash-managed-service-metrics @requires-os:linux
+  Scenario: 4 - Observe displays metrics from a managed model
+    Given a managed model exposes serving metrics
+    When the user opens the dashboard
+    And the user opens the Observe view
+    Then live serving metrics are displayed for the managed model
+    When the user quits the dashboard
+    Then the dashboard exits successfully
+
+  @id:dash-help-guidance @requires-os:linux
+  Scenario: 5 - A user can discover dashboard help and next-step guidance
+    When the user opens the dashboard with demo data
+    And the user opens dashboard help
+    Then navigation and next-step guidance are displayed
+    When the user closes dashboard help
+    And the user quits the dashboard
+    Then the dashboard exits successfully
+
+  @id:dash-command-palette-navigation @requires-os:linux
+  Scenario: 6 - A user navigates to Serving through the command palette
+    When the user opens the dashboard with demo data
+    And the user opens the command palette
+    Then dashboard destinations are displayed
+    When the user chooses Serving
+    Then Serving actions are displayed
+    When the user quits the dashboard
+    Then the dashboard exits successfully
+
+  @id:dash-managed-service-visible @requires-os:linux
+  Scenario: 7 - A managed model is visible in the dashboard
+    Given a running managed model is available locally
+    When the user opens the dashboard
+    And the user opens the Observe view
+    Then the managed model is displayed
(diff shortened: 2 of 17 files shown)


EugeneKSiloAi:
Should-fix
1. HOME/XDG isolation silently applies to all piped scenarios, contradicting the description.

The PR description says "PTY scenarios additionally isolate HOME and XDG_RUNTIME_DIR." But the implementation puts that isolation in the shared isolate_env():
e2e.rs
Lines 284-302
```
            let home = root.join("home");
            let runtime = root.join("runtime");
            std::fs::create_dir_all(&home).expect("failed to create isolated HOME");
            std::fs::create_dir_all(&runtime).expect("failed to create isolated runtime dir");
            env.push(("HOME", home.into_os_string()));
            env.push(("XDG_RUNTIME_DIR", runtime.into_os_string()));
```
Since isolate_cmd() (the piped std::process::Command path) now delegates to isolate_env(), every existing piped scenario also gets a fresh isolated HOME/XDG_RUNTIME_DIR — previously they inherited the real ones. That's a behavior change beyond the stated scope. Most at-risk are the @requires-gpu scenarios (chat-tool-definitions-accepted, chat-end-to-end-local-model), which likely weren't exercised in the local run (cargo xtask e2e — 17 passed looks like the mock tier), so the HOME change on those paths is unverified — e.g. anything resolving HOME-derived defaults (keyring on Linux, ~/.local, uv/pip fallbacks not covered by the explicit *_CACHE_DIR overrides).

Recommendation: either scope HOME/XDG_RUNTIME_DIR to the PTY spawn path only (they exist to stop a host daemon from masking the embedded one — a dashboard concern), or keep it shared but update the description and confirm the GPU scenarios still pass. Right now the code and the stated design disagree, which a reviewer can't silently accept.

2. Command-coverage isn't recorded for sessions that end without an explicit quit.

record_once only fires inside wait_for_exit:
tui_driver.rs
Lines 240-280
   ```
 pub async fn wait_for_exit(&mut self, timeout: Duration) -> Result<(), String> {
        ...
```
                    self.record_once(i32::try_from(status.exit_code()).unwrap_or(-1));
Scenario: 2 - The privacy notice is shown before using a local endpoint ends after privacy_notice_shown with no quit step, so its TuiSession is reaped by Drop (SIGKILL) and never records rocm chat coverage. Coverage isn't wrong overall (scenario 3 quits cleanly and records rocm chat), but if the command-surface report is meant to be complete per-scenario, consider a best-effort record_once in Drop too. Minor.

Nits (optional)
framed_screen mislabels geometry after resize. It hardcodes {COLS}x{ROWS} (80x24), so failure banners for Observe scenarios that called use_detail_size() (120x40) report the wrong dimensions:

tui_driver.rs
Lines 331-334
```
fn framed_screen(screen: &str) -> String {
    format!("--- last screen ({COLS}x{ROWS}) ---\n{screen}\n--- end screen ---")
}
```
Reading the size from the parser's current screen would make diagnostics accurate.

Post-exit drain window is a single 20ms poll. In wait_for_screen, once the child has exited it sleeps one POLL_INTERVAL then checks the screen once. For a large final frame the reader thread might not have drained fully in 20ms, producing a rare false "process exited before marker" failure. Low risk given the generous 20s budgets, but a short bounded drain loop would be more robust.

choose_serving ("jj" + Enter) and tab-index keys are order-coupled to the palette/tab layout. Inherent to TUI testing and acceptable, but worth a comment noting the assumed ordering so a future tab insertion doesn't produce a confusing failure.

**Comment by rominf** [comment]
Addressed in 3f646a7: HOME and XDG_RUNTIME_DIR isolation is now PTY-only, while piped/GPU scenarios retain their prior environment. PTY sessions now record command coverage during teardown, report the parser's current geometry, and use a bounded post-exit drain with a final race-safe screen check. I also documented the tab and palette ordering assumed by the key sequences.\n\nVerified with cargo fmt, clippy for the E2E test target, 23 harness unit tests, and focused privacy-notice and command-palette PTY scenarios.

File: tests/e2e-cucumber/tests/e2e/dash_steps.rs [hunk 1]
```diff
@@ -0,0 +1,396 @@
...
+        .unwrap_or_else(|e| panic!("failed to accept local endpoint: {e}"));
+}
+
+#[when("the user sends a message to the managed model")]
+async fn send_managed_model_message(world: &mut E2eWorld) {
+    let tui = session(world);
+    tui.wait_for_screen("No messages yet.", DEFAULT_TIMEOUT)
+        .await
+        .unwrap_or_else(|e| panic!("chat surface never became ready: {e}"));
+    tui.send("i")
```
volen-silo: **Blocking correctness bug.** By the time this step runs, the preceding `accept_local_endpoint` has sent `y`, which routes to `accept_chat_consent()` in the product code (`crates/rocm-dash-tui/src/app/mod.rs:1069-1073`) — that setter sets **both** `chat_consent = Accepted` **and** `chat_focused = true` (its doc: "Focuses the input so the user can type immediately").

With the input already focused, this `i` is no longer a focus gesture: the key router treats it as literal text (`app/mod.rs:3060-3067`), so the submitted prompt becomes `ihello from the terminal`. The corruption is invisible because the mock's `handle_chat` ignores the request body and always returns the canned reply (`src/mock_server.rs:228-239`), and `managed_model_response_displayed` only asserts on `mock response for testing` — so this scenario passes without ever sending the message it claims to.

Contrast the offline path (`send_gpu_message`): `--chat-mock` pre-accepts consent via `set_chat_config`, which does **not** touch `chat_focused`, so *its* leading `i` is correct. The managed path is not.

Fix: remove this `i` keystroke here (consent already focused the input). Consider also asserting the request actually carried the typed prompt (e.g. have the mock echo it back) so this class of bug can't hide behind the canned reply again.

tests/e2e-cucumber/tests/e2e/tui_driver.rs
```diff
@@ -0,0 +1,364 @@
...
+    }
+
+    /// The current visible screen as plain text (one row per line). `vt100` has
+    /// already resolved escape sequences and styling into cells, so this is
+    /// exactly what a user sees — color/attribute independent.
+    pub fn screen_text(&self) -> String {
+        self.screen_snapshot().0
+    }
+
+    fn screen_snapshot(&self) -> (String, (u16, u16)) {
```
volen-silo, line 200: **Reliability / diagnostics (non-blocking).** `screen_snapshot` does `self.parser.lock().map(...).unwrap_or_default()`. If the reader thread ever panics inside `parser.lock()` → `p.process(...)` (line 353), the mutex is poisoned; from then on every `screen_text()` silently returns an empty screen, so each `wait_for_screen` burns its full 20s deadline and reports a blank frame instead of the real fault. `Drop` also discards the reader `JoinHandle` result (lines 330-332), swallowing the panic.

A `vt100` parse panic is unlikely, but recovering the poisoned guard (`.unwrap_or_else(std::sync::PoisonError::into_inner)`) and/or surfacing a reader-thread panic would turn a 20s timeout-with-empty-screen into a direct diagnosis.

tests/e2e-cucumber/tests/e2e/tui_driver.rs
```diff
@@ -0,0 +1,364 @@
...
+// SPDX-License-Identifier: MIT
+
+//! Black-box driver for the interactive dash/chat TUI.
+//!
+//! The rest of the suite spawns `rocm` with piped stdin/stdout via
+//! `std::process::Command`. That can never exercise the interactive dashboard:
+//! the CLI only enters the crossterm raw-mode event loop when both stdin and
+//! stdout are a real terminal (`rocm_core::interactive_terminal`), and a pipe is
+//! not. So the dash was previously "untestable black-box" (e.g. the chat privacy
+//! notice, EAI-7222) and only had in-process render tests.
```
volen-silo: **Internal ticket reference (blocking for open-source).** `EAI-7222` is an internal tracker id and shouldn't ship in a repo going open source. The surrounding sentence already conveys the point — drop the id (e.g. "… was previously an untestable black-box gap").

tests/e2e-cucumber/features/chat.feature
```diff
@@ -7,42 +7,60 @@ Feature: Chat and endpoint detection
     When the user checks for running services
     Then the served model is listed
 
-  @id:chat-privacy-notice-accurate
-  Scenario: 2 - The privacy notice is accurate for local endpoints
+  # Driven through the interactive TUI (a pseudo-terminal): the privacy notice
+  # lives on the chat consent gate, which only the real terminal path renders.
+  # The model is registered so the CLI discovers it on its OS-assigned port and
+  # offers it — the notice must precede any request. (Was EAI-7222, previously
```
volen-silo: **Internal ticket reference (blocking for open-source).** Remove `EAI-7222` here; "previously an untestable black-box gap" already says what's needed without the internal id.

tests/e2e-cucumber/tests/e2e/dash_steps.rs
```diff
@@ -0,0 +1,396 @@
...
+async fn dashboard_exited(world: &mut E2eWorld) {
+    assert_tui_opened(world);
+}
+
+#[then("interactive chat exits successfully")]
+async fn interactive_chat_exited(world: &mut E2eWorld) {
+    assert_tui_opened(world);
+}
+
+// ── Then: chat privacy consent gate (real endpoint, EAI-7222) ──────
```
volen-silo: **Internal ticket reference (blocking for open-source).** Drop `EAI-7222` from this section header — e.g. `// ── Then: chat privacy consent gate (real endpoint) ──`.

tests/e2e-cucumber/tests/e2e.rs
```diff
@@ -272,16 +333,62 @@ impl E2eWorld {
     /// record is plain JSON matching the CLI's on-disk schema, not a typed import
     /// from the rocm-cli crates.
     pub fn register_mock_service(&self) {
+        self.register_mock_service_with(RegisterServiceOptions::default());
+    }
+
+    pub fn register_mock_service_with(&self, options: RegisterServiceOptions) {
```
volen-silo: **Maintainability (non-blocking).** This re-encodes essentially the same on-disk managed-service record that the public `mock_server::write_service_record` (`src/mock_server.rs:177-211`) already produces — same 20-odd fields; the only real differences are `startup_phase` / live PIDs here vs `supervisor_pid: 0` there. The two copies will drift.

Consider extending `write_service_record` to take the varying lifecycle fields (an options struct) and calling it from here, so the schema lives in one place — which is the stated "black-box, schema in one place" goal.

volen-silo (dismissed):
## Review: drive the dash TUI black-box over a pseudo-terminal

Overall this is a strong, well-documented addition. The PTY driver is a genuinely valuable acceptance layer, the isolation story (per-scenario HOME/XDG, OS-assigned mock ports, deterministic vt100 grid) is careful, and the dependency cleanup (dropping the dead ratatui 0.29 from the `rocm` binary) is verified safe: `cargo check --workspace --all-targets` passes with only ratatui 0.30 remaining. The removed privacy-notice xfail is correctly replaced by a real passing PTY scenario, with no orphaned step text left behind.

Requesting changes for one correctness defect that means a new scenario does not exercise the prompt it claims to send, plus a set of internal ticket references that should not ship in an open-source repo. A maintainability item and a reliability item are non-blocking. Specifics are in the inline comments.

### Blocking
1. `send_managed_model_message` sends `i` after consent already focused the input, so the submitted prompt is corrupted to `ihello from the terminal` and masked by the canned mock reply (inline on dash_steps.rs).
2. Internal ticket references (`EAI-7222`) remain in three added lines; this repo is going open source (inline on each).

### Non-blocking
3. The managed-service JSON record is duplicated between the World and the public `write_service_record` (inline).
4. A reader-thread panic poisons the vt100 mutex, turning later waits into blank-screen 20s timeouts that hide the real cause (inline).

tests/e2e-cucumber/tests/e2e/dash_steps.rs
rominf: Fixed in a2ce3ddb: removed the extra `i` because accepting consent already focuses the input. The scenario now also captures the mock request and asserts that the submitted user message is exactly `hello from the terminal`, so a corrupted prompt cannot be hidden by the canned reply.

tests/e2e-cucumber/tests/e2e/tui_driver.rs
rominf: Addressed in a2ce3ddb and 592b39df: parser locks recover poisoned guards, reader panics are captured and surfaced during waits (including the final drain boundary), and teardown preserves the diagnostic without double-panicking during an existing unwind.

tests/e2e-cucumber/tests/e2e/tui_driver.rs
rominf: Removed the ticket reference from the driver documentation in a2ce3ddb.

tests/e2e-cucumber/features/chat.feature
rominf: Removed the ticket reference from the feature comment in a2ce3ddb.

tests/e2e-cucumber/tests/e2e/dash_steps.rs
rominf: Removed the ticket reference from the section heading in a2ce3ddb.

tests/e2e-cucumber/tests/e2e.rs
rominf: Addressed in a2ce3ddb: the managed-service JSON schema now lives in `mock_server::write_service_record_with`, parameterized by `ServiceRecordOptions`. Both the Cucumber World and `rocm-demo-env` delegate to that shared writer while preserving their lifecycle/PID differences.

tests/e2e-cucumber/tests/e2e/dash_steps.rs
```diff
@@ -0,0 +1,396 @@
...
+//! screen, and clean exit — which the piped-`Command` steps structurally cannot.
+
+use cucumber::{given, then, when};
+use e2e_cucumber::mock_server::MockServer;
+
+use crate::e2e::tui_driver::{DEFAULT_TIMEOUT, TuiSession};
+use crate::{E2eWorld, RegisterServiceOptions};
+
+/// Borrow the scenario's active TUI session, or fail clearly if none was opened.
+const fn session(world: &mut E2eWorld) -> &mut TuiSession {
```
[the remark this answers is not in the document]
rominf: No code change needed: this compiles on the repository's pinned stable Rust toolchain, where `Option::expect` is const-evaluable. Both Linux and Windows build/test CI passed at the reviewed head.

tests/e2e-cucumber/features/dash.feature
```diff
@@ -0,0 +1,70 @@
...
+    Then the dashboard exits successfully
+
+  @id:dash-chat-offline-reply @requires-os:linux
+  Scenario: 2 - A user receives a response in interactive chat
+    Given interactive chat uses an offline assistant
+    When the user opens interactive chat
+    And the user sends a message about GPU health
+    Then the assistant's GPU status response is displayed
+    When the user quits the dashboard
+    Then the dashboard exits successfully
```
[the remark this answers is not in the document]
rominf: Fixed in 592b39df: the offline-chat scenario now uses the chat-specific quit and successful-exit steps.

tests/e2e-cucumber/features/chat.feature
```diff
@@ -7,42 +7,60 @@ Feature: Chat and endpoint detection
...
+  # an untestable-black-box gap.)
+  @id:chat-privacy-notice-accurate @requires-os:linux
+  Scenario: 2 - The privacy notice is shown before using a local endpoint
     Given a model is being served locally
-    When the user is offered the detected endpoint
-    Then the notice does not claim that requests leave the machine
+    And the model is registered with the CLI
+    When the user opens interactive chat
+    Then the local endpoint is shown for confirmation
+    And the privacy notice is shown before any message is sent
```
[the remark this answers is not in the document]
rominf: Fixed in f2fa1fe: the privacy-notice scenario now quits interactive chat explicitly and asserts a successful exit, so teardown records a clean command result rather than killing the session.

michaelroy-amd (changes requested):
## Review: PTY-backed TUI acceptance testing

The PTY + `vt100` approach is sound and worth adopting. Driving the compiled binary through a real pseudo-terminal is the correct acceptance layer for crossterm raw mode, and asserting on the current emulated screen is substantially stronger than matching escape-sequence transcripts.

The implementation is generally healthy: fixed terminal geometry, PTY-only HOME/XDG isolation, OS-assigned mock ports, deterministic metrics, exact prompt capture, bounded polling/draining, kill → reap → reader-join teardown, panic propagation, and consolidated service-record generation are all well handled. Removing the unused `ratatui` 0.29 dependency is also valid; the resulting tree contains only `ratatui` 0.30.2.

### Changes requested: isolate deterministic chat scenarios from host provider credentials

`tests/e2e-cucumber/tests/e2e/tui_driver.rs:115-129` inherits the full parent environment. That includes variables such as `OPENAI_API_KEY`, `ROCMDASH_CHAT_API_KEY`, `OPENAI_BASE_URL`, and `ANTHROPIC_API_KEY`. These are behavior inputs, not harmless process context: a configured chat key prevents local managed-service detection in the product startup path.

I reproduced the consequence at head `f2fa1fe`:

- normal workstation environment: 8 passed, 2 failed (`chat-privacy-notice-accurate` and `chat-managed-model-interactive`), both showing `Chat — no endpoint detected`;
- with provider credentials removed: 10 scenarios passed, 63 steps passed.

Please make scenarios that claim deterministic local/mock behavior hermetic by removing provider URL/key variables from the child environment. `portable_pty::CommandBuilder` supports `env_remove`. This should preferably be expressed as part of the scenario isolation contract, rather than requiring callers or CI to sanitize their shell manually. Keep it scoped if other scenarios intentionally test configured credentials.

Focused verification at this head:

- `cargo test -p e2e-cucumber --lib`: 27 passed
- `cargo clippy -p e2e-cucumber --test e2e -- -D warnings`: passed
- `cargo check --workspace --all-targets`: passed
- `cargo fmt --all --check`: passed
- hermetic Linux PTY run: 10 scenarios / 63 steps passed

Once the provider environment is isolated and the focused PTY scenarios pass without manually unsetting credentials, I think this is ready to approve.

michaelroy-amd (approved):
Approved after the provider-environment isolation fix in `53d939e`.

Verified the regression with host credentials injected: the two managed-chat scenarios now pass without manual shell cleanup. The complete Linux-tagged PTY set passes 10 scenarios / 63 steps; workspace tests, clippy, formatting, smoke, push hooks, commit verification, and current CodeQL checks are green.

volen-silo (approved):

PR #117, opened 2026-07-15
fix(serve): verify process identity and bound stop termination (EAI-7410)

michaelroy-amd:
## Summary

Fixes **EAI-7410**: the serve **stop** path signalled a PID read from a state file with no identity check and reported success without confirming the process actually exited.

Two defects, shared by both engine adapters and the CLI managed-service layer:

1. **PID reuse → wrong-process kill.** `stop` read a persisted PID and `SIGTERM`ed it (vLLM signalled the whole `/proc` descendant tree). Linux recycles PIDs, so a stale PID could belong to an unrelated process by stop time — killing an innocent process (and its subtree).
2. **Untruthful, unbounded termination.** Success was reported when `SIGTERM` was merely *delivered* (`ESRCH`/already-gone counted as success). No bounded wait, no confirmed exit, no `SIGKILL` escalation — so `stopped`/`graceful` were claimed even while the process was still alive or already gone.

## What changed

**New `rocm-core::proc_lifecycle`:**
- `ProcessIdentity` = `(pid, kernel start-time)` from `/proc/<pid>/stat` (field 22). A recycled PID has a different start-time, so the pair is a recycle-proof identity.
- `identity_state` → `Matches` / `Recycled` / `Indeterminate` / `Gone` (zombie-aware; refuses when a recorded identity can't be confirmed).
- `terminate_verified(id, scope, grace, force)` → verify identity, `SIGTERM`, poll for **actual exit** up to a bounded `grace`, escalate to `SIGKILL` only under `force`, and return a **truthful** `TerminationOutcome`. **Tree** scope snapshots every descendant and waits for **all** to exit, so a lingering GPU worker is never reported as a graceful stop; forced escalation only targets survivors it can positively re-verify.

**Wiring:**
- vLLM & lemonade persist `start_ticks`/`server_start_ticks` in engine state and verify identity in `stop_service` (vLLM `Tree`, lemonade `Single`); lemonade also refreshes the token at every state PID-swap.
- CLI `stop_internal_managed_service` verifies `ManagedServiceRecord` identities before signalling; the supervisor (launcher) and engine server carry their own tokens (`supervisor_start_ticks` / `engine_start_ticks`, the latter adopted in `refresh_from_engine_state`). A stop claims `"stopped"` only when every recorded process is confirmed gone; otherwise the prior status stands and the existing liveness refresh reconciles it.
- Removed the now-unused unverified `signal_process_tree`.

Behavior degrades gracefully: legacy state files without a token, and non-Linux platforms, fall back to documented best-effort (unchanged from prior behavior).

## Testing

- `cargo test --workspace` → **1459 passed, 0 failed** (`--test-threads=1`).
- `cargo clippy --workspace --all-targets -- -D warnings` clean; `cargo fmt --check` clean.
- New focused lifecycle tests: identity match/recycle/gone/indeterminate, refuse-to-signal on mismatch, graceful/timed-out/forced, **tree waits for descendants**, **forced tree kill of a SIGTERM-ignoring reparented descendant**, engine `identity_from_state` (incl. legacy/no-token), and managed-layer stop (refuse recycled PID, verified kill, distinct supervisor/engine PIDs).
- Reviewed across three rounds (reliability + security + OSS + dual fresh adversarial); all confirmed findings resolved.

## Follow-ups (out of scope — separate subsystems, same pattern)

- `apps/rocm/src/comfyui.rs` and the `rocmd` supervised-record stop path still use the store-then-kill pattern; recommend routing them through `terminate_verified` in a follow-up.

### DIFF
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -4478,6 +4478,9 @@ fn spawn_managed_engine_child(
     };
     record.supervisor_pid = child_pid;
     record.engine_pid = Some(child_pid);
+    // Capture the identity token while the child is alive, so a later stop
+    // verifies this exact process rather than a recycled PID.
+    record.supervisor_start_ticks = rocm_core::process_start_ticks(child_pid);
     record.status = "running".to_owned();
     record.write()?;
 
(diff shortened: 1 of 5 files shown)


volen-silo (commented):
## Review

Architecturally sound and it meets all three acceptance criteria. The design holds up well: recycle-proof identity via `/proc/<pid>` start-ticks, a conservative `Indeterminate` state that refuses to signal when identity can't be confirmed, per-descendant identity capture so a PID recycled mid-wait isn't mistaken for a survivor, and truthful outcome reporting. Test coverage hits stale state, identity mismatch, graceful, timed-out, and forced escalation.

### 🔴 Blocking — this is the `windows-build-and-test` failure

`crates/rocm-core/src/proc_lifecycle.rs:741`

```rust
#[cfg(test)]
mod tests {
    use super::*;
    use std::process::{Child, Command, Stdio};   // unused on Windows
```

`Child`/`Command`/`Stdio` are used only inside `#[cfg(unix)]` helpers (`spawn_ready`, `spawn`, `reap`) and the unix/linux tests. The four non-gated tests (`parse_start_ticks_*`, `outcome_truth_table`) don't touch them, so on Windows the import is unused and `-D warnings` fails the build. One-line fix:

```rust
    #[cfg(unix)]
    use std::process::{Child, Command, Stdio};
```

### 🟡 Non-blocking observations

1. **Windows stop latency.** `send_signal(Term)` is a documented no-op on Windows, but `terminate_verified` still runs the full `grace` wait before the forced-kill path. Since the managed layer hardcodes `force = true` (`apps/rocm/src/main.rs`, `terminate_recorded_service_pids`), every Windows managed stop now eats a guaranteed ~10s wait before the kill — a latency regression vs. the previous immediate `terminate_process`. Worth short-circuiting the first wait when `Term` is a no-op. (A *non-forced* Windows engine stop is worse: it always `TimedOut`s after the grace and never kills.)

2. **Compounding grace on the managed path.** `stop_internal_managed_service` calls the engine's own `stop_service` (waits up to `STOP_GRACE`) and then `terminate_recorded_service_pids` (another `MANAGED_STOP_GRACE`). The common case is fast — the second pass usually sees the engine PID already `AlreadyGone` — but a worst-case timeout stacks the two grace windows (~20s). Probably fine; flagging for awareness.

3. **Managed layer always forces.** `terminate_recorded_service_pids` passes `force = true` unconditionally, unlike the engine adapters which honor `request.force`. Likely intentional (a stop should actually stop), but it's undocumented at the call site — a one-line comment would help.

Only item 1's root cause (the unused import) blocks the merge; 1's latency note and 2–3 are follow-up-worthy at most.

rominf (dismissed):
Automated review: requesting changes for the Windows build failure.

Blocking: `crates/rocm-core/src/proc_lifecycle.rs:356` imports `Child`, `Command`, and `Stdio` unconditionally, but every consumer is compiled only on Unix/Linux. On Windows the import is unused and `-D warnings` fails deterministically. Please gate the import:

```rust
#[cfg(unix)]
use std::process::{Child, Command, Stdio};
```

This is the only failing code-relevant check; Linux build/test, clippy, CodeQL, and coverage are green. The branch is also behind `main` and needs a rebase.

Non-blocking:
1. The stopped-process count includes `TerminationOutcome::TimedOut`, so it can count a process that remains alive.
2. On Windows, the TERM operation is a no-op, but forced managed stops still wait the full grace period before killing, adding about ten seconds to every stop.
3. The engine stop grace and managed-process grace can stack in the worst case.
4. The managed layer hardcodes forced termination without documenting why it differs from adapters honoring `request.force`.
5. Adapter tests do not pin the intended `Tree` scope for vLLM versus `Single` for Lemonade.

clydehoang1 (approved):
Automated review:
This PR and I don’t see any major/blocking issues for the current changes.

What is verified locally:

cargo test -p rocm-core proc_lifecycle
cargo test -p rocm-engine-lemonade identity_from_state
cargo test -p rocm-engine-vllm identity_from_state
The process-identity + bounded termination changes look correct and materially safer (recycled PID protection, truthful stop outcomes, and engine-appropriate stop scope).

Residual risk is mainly integration-level behavior on full GPU/e2e paths, which should be covered by CI.

EugeneKSiloAi (approved):

PR #118, opened 2026-07-15
fix(examine): classify gfx1100/1101/1102 as discrete, not APU (EAI-7412)

michaelroy-amd:
### Problem
`gfx_is_apu_family` (crates/rocm-core/src/examine.rs) matched the whole `gfx110x` prefix, so the **discrete** RDNA3 parts **gfx1100 / gfx1101 / gfx1102** (Navi 31/32/33 → Radeon RX 7900/7800/7600) were tagged APUs alongside the real APU **gfx1103** (Phoenix) and the **gfx115x** (Strix) parts.

An old test (`gfx_apu_family_classification`) *protected* this wrong behavior, asserting `gfx_is_apu_family("gfx1100") == true` with a comment that it deliberately mirrored `examine.py`'s `gfx11[05]\d` regex bug.

### Impact
`is_apu` → `has_apu` / `has_discrete_amd` (`summarise_gpu_categories`) → gates `check_9_igpu_dgpu_collision` in diagnose.rs. On a real **gfx1103 APU + gfx1100 dGPU** box, both were tagged APU, so `has_discrete_amd` stayed `false` and the iGPU+dGPU collision fix was **silently suppressed** — for exactly the pairing it exists to help. A lone gfx1100 dGPU box also falsely reported "has an APU".

### Fix (test-first)
- **Correct the classification**: within `gfx110x`, only `gfx1103`+ are APUs; `gfx1100/1101/1102` are discrete. All `gfx115x` remain APUs. Trailing gcnArchName feature suffixes (e.g. `:sramecc+:xnack-`) are tolerated.
- **Replace the parity test** with the intended neighboring-family contract; add flag-derivation and lone-dGPU regression tests.
- **Correct the collision-fix note**: it now names the discrete target explicitly instead of the fragile "the higher-numbered gfx target is the dGPU" heuristic, which is *false* on RDNA3 (the APU can outrank its discrete neighbor). Added the previously-missing end-to-end `check_9` regression test.

Target → product mapping verified against LLVM AMDGPUUsage.

### Test plan
- `cargo test -p rocm-core` — 170 pass (RED→GREEN confirmed on the new contract test).
- `cargo test --workspace -- --test-threads=1` — full suite green.
- Reviewed via dual adversarial review; HIGH (backwards note) and MEDIUM (missing `check_9` test) findings fixed.

### Follow-ups (out of scope, noted)
- `gfx90c` (Renoir/Cezanne APU) is still unclassified — pre-existing gap, currently inert (`check_9` needs `has_apu`). Worth a separate ticket.
- External `rocm-doctor` `examine.py`/`diagnose.py` (not in this repo) should be updated in lockstep to drop the same bug; the surviving in-repo parity test only checks JSON key shape, not values, so nothing here regresses.

### DIFF
--- a/crates/rocm-core/src/diagnose.rs
+++ b/crates/rocm-core/src/diagnose.rs
@@ -874,6 +874,31 @@ fn check_9_igpu_dgpu_collision(e: &Examination, symptom: &str) -> Diagnosis {
     }
 
     let gfx_targets = amd_gfx_targets(e);
+    // Name the discrete GPU explicitly instead of guessing by gfx number: on
+    // RDNA3 the integrated APU (gfx1103 / gfx115x) can outrank its discrete
+    // neighbor (gfx1100/1101/1102), so "the higher-numbered target is the dGPU"
+    // is actively wrong for exactly the APU+dGPU pairing this check fires on.
+    let discrete_targets: Vec<String> = e
+        .gpus
+        .iter()
+        .filter(|g| g.is_amd && g.is_apu == Some(false) && !g.gfx_target.is_empty())
+        .map(|g| g.gfx_target.clone())
+        .collect();
+    let apu_targets: Vec<String> = e
+        .gpus
+        .iter()
+        .filter(|g| g.is_amd && g.is_apu == Some(true) && !g.gfx_target.is_empty())
+        .map(|g| g.gfx_target.clone())
+        .collect();
+    let note = if discrete_targets.is_empty() {
+        format!(
+            "Detected gfx targets: {gfx_targets:?}. Pin HIP_VISIBLE_DEVICES to the discrete GPU so the integrated APU is hidden."
+        )
+    } else {
+        format!(
+            "Detected gfx targets: {gfx_targets:?}. Discrete GPU(s): {discrete_targets:?}; integrated APU(s): {apu_targets:?}. Pin HIP_VISIBLE_DEVICES to the discrete GPU — do not assume the higher-numbered gfx target is the dGPU (on RDNA3 the APU can be higher)."
+        )
+    };
     let fix = if e.os_family == "windows" {
         Fix {
             summary: "Pin the HIP runtime to the discrete GPU with HIP_VISIBLE_DEVICES so the iGPU is hidden.".to_owned(),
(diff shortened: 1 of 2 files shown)


volen-silo (commented):
Reviewed the logic against the full probe→diagnose chain and ran the suite (`cargo test -p rocm-core` → **170 pass**, the 4 new tests green). This is a correct, cleanly-scoped, test-first fix. LGTM with two minor, non-blocking notes.

## ✅ Correct and well-wired

Traced the fix end-to-end and it does what the description says:

- `probe_gpus_rocminfo` (examine.rs:712) sets `gpu.is_apu = gfx_is_apu_family(gfx_target)`, overriding the earlier lspci guess — so on a real ROCm box the corrected classification is authoritative → `summarise_gpu_categories` → `has_apu`/`has_discrete_amd` → the `check_9` guard. The gfx1103-APU + gfx1100-dGPU box now correctly sets **both** flags, so the collision fix fires for exactly the pairing it exists for.
- `gfx_model_digit` handles the cases that matter: `gfx115x` → always APU, `gfx110x` → only `≥3`, trailing gcnArchName suffixes (`:sramecc+:xnack-`) tolerated, degenerate inputs (`gfx110`, ``) → `None`. The two prefixes are mutually exclusive so ordering is safe.
- The diagnose note rewrite is a real correctness fix, not cosmetic: naming the discrete target from `is_apu == Some(false)` avoids the "higher-numbered gfx is the dGPU" heuristic that is genuinely inverted on RDNA3 (gfx1103 APU > gfx1100 dGPU).
- Test coverage is thorough (boundary contract, flag derivation, lone-dGPU, and the end-to-end `check_9` regression). Good call replacing the bug-for-bug parity test with the intended contract, and the PR body flags the external `examine.py` lockstep + the surviving key-shape-only parity test — nice scope awareness. (The lspci-name fallback keys on marketing names, not gfx targets, so it never shared this prefix bug — scope is right.)

## 🟡 Minor: `digit >= 3` encodes a forward assumption about gfx110x silicon

Within `gfx110x`, `>= 3` classifies `gfx1104`–`gfx1109` as APUs too. Only `gfx1103` exists today, so this is correct now, but it bakes in "any future gfx110x ≥3 is an APU" — which would misclassify a hypothetical future discrete `gfx1104`. Fine to ship; worth a one-line comment on the `>= 3` so the assumption is explicit for the next person (the doc comment explains *why* 1100/1101/1102 are discrete but not why the ceiling is open-ended).

## 🟡 Minor: internal ticket ref vs. the going-open-source direction

The commit headline `…(EAI-7412)` and the new code comment `// End-to-end guard for EAI-7412:` (diagnose.rs:1478) embed an internal Jira ref. For context: none of the last 30 commit headlines on `main` carry a ticket ref, and neither `examine.rs` nor `diagnose.rs` currently contains one — so this would be the first in these files. Since the repo is heading open-source, consider dropping `EAI-7412` from the commit headline and rewording the comment to describe the guarantee without the ref (e.g. "End-to-end guard: gfx1103 APU + gfx1100 dGPU must trigger the collision fix"). The `expectations.toml` `bug =` field is the established structured place if a tracker link is needed. Non-blocking.

volen-silo (approved):

PR #121, opened 2026-07-16
Enforce GPU-required serving with a real device probe

volen-silo:
## Summary

Under the GPU-required policy, `rocm serve` (Lemonade) could proceed on a host with no usable AMD GPU, silently fall back to a CPU llama.cpp backend, assume device 0, and report the device as `rocm_gpu` regardless of what ran. This tightens enforcement so "GPU required" means a real GPU was found and used, and the reported backend/device matches execution.

### Enforcement
- **Real device probe** (`rocm-core`): `usable_amd_gpu_indices()` enumerates AMD GPU nodes from the KFD topology and applies the `HIP_VISIBLE_DEVICES` / `ROCR_VISIBLE_DEVICES` mask. `Some([])` is authoritative "no GPU"; `None` means unprobeable (don't block).
- **CLI pre-flight**: a GPU-required serve fails immediately when the probe reports no usable GPU — before preparing or launching any engine (no wasted engine download; actionable message instead of a late crash).
- **Engine gate** (Lemonade): the serve path re-checks before spawning; `auto` resolves to the first *present* device (no assumed device 0); an explicit `--gpu` index not present is rejected.
- **No CPU backend**: `cpu` dropped from backend selection; when only CPU is usable, serving fails with install guidance instead of degrading silently.
- **Truthful reporting**: healthcheck reports the backend and pinned ordinal that actually ran instead of a hardcoded `rocm_gpu`.
- Removed the CLI GPU-0 fallback when the GPU count is unknown.

### Tests
- Unit tests: mask parsing, auto/explicit/no-device resolution, CPU exclusion, device reporting.
- New E2E scenarios (`model_serving.feature`): no-GPU host refuses to serve (via a new `@requires-no-gpu` gate, runs on the **blocking** mock job), all-GPUs-masked refuses to serve, and a nonexistent `--gpu` index is rejected.
- The no-GPU fail-fast scenario was verified running live on a no-GPU host through the real binary. Masked / absent-index scenarios run on the Strix Halo / Instinct GPU runners.

## Notes / follow-ups
- **AC "reported device matches execution"** is unit-tested only: the healthcheck `device` field isn't surfaced by any user-facing CLI command, so a black-box E2E can't observe it yet. Surfacing backend/device in `rocm services list` would unlock an E2E — deliberately left as a separate change.
- **WSL:** the probe treats an unreadable KFD topology as "unknown" (won't block) when `/dev/kfd` exists; worth confirming on a WSL GPU host.

## Test plan
- [x] `cargo fmt`, `cargo clippy --workspace --all-targets`
- [x] `cargo test` for `rocm-core`, `rocm-engine-lemonade`, `rocm`, `e2e-cucumber`
- [x] E2E no-GPU scenario verified live
- [ ] Strix Halo / Instinct E2E (runs in CI)

### DIFF
--- a/crates/rocm-core/src/lib.rs
+++ b/crates/rocm-core/src/lib.rs
@@ -3550,16 +3550,26 @@ fn probe_usable_amd_gpu_indices() -> Option<Vec<u32>> {
 }
 
 /// Combine the KFD-topology and DRM-card AMD GPU counts into one "GPUs present"
-/// figure, taking the larger available count. Some hosts (e.g. Strix Halo APUs)
-/// enumerate the GPU only via DRM ip-discovery and report zero KFD GPU nodes, so
-/// relying on KFD alone would wrongly conclude there is no GPU and block serving.
+/// figure. KFD counts *compute* nodes and is authoritative for HIP ordinals, so
+/// it wins whenever it reports at least one GPU. DRM is used only as the
+/// zero-KFD fallback: some hosts (e.g. Strix Halo APUs) enumerate the GPU only
+/// via DRM ip-discovery and report zero KFD GPU nodes, so relying on KFD alone
+/// would wrongly conclude there is no GPU and block serving.
+///
+/// DRM must not *raise* a nonzero KFD count: a display/render-only AMD DRM card
+/// with no KFD compute node (e.g. KFD=1, DRM=2) would otherwise invent a usable
+/// HIP ordinal that passes `--gpu` validation but fails later inside HIP.
+///
 /// `None` only when NEITHER surface could be read (availability truly unknown).
 #[cfg(any(target_os = "linux", test))]
 fn combine_amd_gpu_counts(kfd: Option<usize>, drm: Option<usize>) -> Option<usize> {
-    match (kfd, drm) {
-        (Some(k), Some(d)) => Some(k.max(d)),
-        (Some(n), None) | (None, Some(n)) => Some(n),
-        (None, None) => None,
+    match kfd {
+        // KFD is compute-authoritative: prefer it whenever it sees a GPU.
+        Some(k) if k > 0 => Some(k),
+        // Zero KFD compute nodes: fall back to DRM for the APU shape.
+        Some(_) => Some(drm.unwrap_or(0)),
+        // KFD unreadable: use DRM if it could be read, else availability unknown.
+        None => drm,
     }
 }
 
@@ -8316,7 +8326,16 @@ last_installed_runtime_id = "therock-release"
     }
 
     #[test]
-    fn combine_amd_gpu_counts_takes_larger_and_needs_one_readable_surface() {
+    fn combine_amd_gpu_counts_prefers_compute_authoritative_kfd() {
+        // KFD is compute-authoritative: a nonzero KFD count wins, and DRM must not
+        // raise it. A display/render-only AMD DRM card (KFD=1, DRM=2) must NOT
+        // invent a second usable HIP ordinal, or an explicit `--gpu 1` would pass
+        // validation and then fail inside HIP.
+        assert_eq!(combine_amd_gpu_counts(Some(1), Some(2)), Some(1));
+        // Same principle with more DRM cards: still bounded by the KFD count.
+        assert_eq!(combine_amd_gpu_counts(Some(2), Some(8)), Some(2));
+        // KFD larger than DRM (e.g. multi-partition compute nodes): KFD still wins.
+        assert_eq!(combine_amd_gpu_counts(Some(3), Some(1)), Some(3));
         // Strix Halo shape: KFD reports 0 GPU nodes, DRM sees the iGPU → 1 present.
         assert_eq!(combine_amd_gpu_counts(Some(0), Some(1)), Some(1));
         // Discrete GPUs: both surfaces agree.
--- a/crates/rocm-core/src/lib.rs
+++ b/crates/rocm-core/src/lib.rs
@@ -3539,7 +3539,8 @@ pub fn has_usable_amd_gpu() -> bool {
 
 #[cfg(target_os = "linux")]
 fn probe_usable_amd_gpu_indices() -> Option<Vec<u32>> {
-    let present = linux_kfd_gpu_node_count()?;
+    let present =
+        combine_amd_gpu_counts(linux_kfd_gpu_node_count(), linux_drm_amdgpu_card_count())?;
     Some(usable_amd_gpu_indices_from(
         present,
         visibility_mask_from_env(),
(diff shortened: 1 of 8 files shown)


rominf (changes requested):
Automated review: requesting changes for one blocking issue.

The two new hardware scenarios are not scoped to Linux, but the enforcement they test is Linux-only. `probe_usable_amd_gpu_indices` returns `None` on non-Linux, and both the CLI preflight and Lemonade gate intentionally treat `None` as “unprobeable; do not block.” On the Strix Halo Windows runner, the masked-device and absent-index scenarios therefore proceed successfully and fail their expectation that serving is refused.

Please add `@requires-os:linux` to `serve-masked-devices-fail` and `serve-absent-gpu-index-rejected`. The existing OS-skip mechanism and Linux-only feature precedent make this the correct expression of the current support boundary; an xfail would misrepresent intentional behavior.

Non-blocking:
1. UUID values in `ROCR_VISIBLE_DEVICES` are parsed as an empty ordinal list, incorrectly turning a valid UUID mask into authoritative “no GPU.” Treat unparseable nonempty masks as unprobeable or document and reject the unsupported form explicitly.
2. CLI `--gpu` validation is mask-unaware while the engine gate is mask-aware, so the displayed plan can temporarily accept an index that the engine later rejects.
3. vLLM has no engine-side availability gate and now launches unpinned when the count probe fails but sysfs remains readable.

The separate Linux GPU failure reports zero unexpected failures and is unrelated expectation drift. Please also remove the generated-by-AI footer from the PR description.

michaelroy-amd:
> The two new hardware scenarios are not scoped to Linux.

Addressed in b26b8c5: both masked-device and absent-index scenarios now carry `@requires-os:linux`.

I also addressed the substantive non-blocking probe findings:
- unsupported nonempty visibility masks such as ROCr UUIDs now return an unprobeable result instead of being misclassified as authoritative zero GPUs;
- vLLM now performs the same engine-side GPU availability and explicit-index validation as Lemonade before spawning.

Rebased onto current `main`. Local verification:
- `cargo test --workspace --all-targets`: 1642 passed, 6 ignored
- `cargo clippy --workspace --all-targets -- -D warnings`: passed
- `python3 scripts/smoke_local.py`: passed
- `cargo xtask verify-commits --base origin/main --require-verified`: all 4 commits signed and signed off

The generated-by-AI footer was removed from the PR description. CI is running on head `b26b8c5`.

crates/rocm-core/src/lib.rs
```diff
@@ -3513,6 +3513,155 @@ fn detect_linux_kfd_gfx_target() -> Option<String> {
...
+
+/// Combine the KFD-topology and DRM-card AMD GPU counts into one "GPUs present"
+/// figure, taking the larger available count. Some hosts (e.g. Strix Halo APUs)
+/// enumerate the GPU only via DRM ip-discovery and report zero KFD GPU nodes, so
+/// relying on KFD alone would wrongly conclude there is no GPU and block serving.
+/// `None` only when NEITHER surface could be read (availability truly unknown).
+#[cfg(any(target_os = "linux", test))]
+fn combine_amd_gpu_counts(kfd: Option<usize>, drm: Option<usize>) -> Option<usize> {
+    match (kfd, drm) {
+        (Some(k), Some(d)) => Some(k.max(d)),
```
rominf: `max(kfd, drm)` can invent usable HIP ordinals when the host has an AMD DRM/display card that has no KFD compute node. For example, KFD=1 and DRM=2 produces `[0, 1]`, so an explicit `--gpu 1` passes both engine gates and fails later inside HIP instead of receiving the intended early actionable error. Please prefer the KFD count whenever it is nonzero, using DRM only for the documented `kfd == 0 && drm > 0` APU fallback, and add a regression test for `combine_amd_gpu_counts(Some(1), Some(2))`.

rominf (changes requested):
Requesting changes for one correctness gap in the new device probe. The implementation can treat a non-compute DRM card as a usable HIP ordinal, allowing an invalid explicit GPU selection through the early validation this PR is intended to provide. Please prefer a nonzero KFD compute-device count and reserve DRM for the documented zero-KFD fallback, with coverage for divergent nonzero counts.

crates/rocm-core/src/lib.rs
volen-silo: Fixed in cabf8a8. `combine_amd_gpu_counts` no longer takes `max(kfd, drm)`: KFD is compute-authoritative for HIP ordinals, so it now wins whenever it reports at least one GPU, and DRM is used only as the documented zero-KFD fallback for APU shapes (e.g. Strix Halo).

```rust
match kfd {
    Some(k) if k > 0 => Some(k),        // KFD compute-authoritative
    Some(_) => Some(drm.unwrap_or(0)),  // zero KFD -> DRM APU fallback
    None => drm,                        // KFD unreadable -> DRM, else unknown
}
```

So `KFD=1, DRM=2` now yields `Some(1)` (ordinals `[0]`), and an explicit `--gpu 1` on that host is rejected early with the actionable error instead of failing later inside HIP.

Added regression coverage for the divergent nonzero counts you asked about, alongside the existing APU/unknown-surface cases:
- `combine_amd_gpu_counts(Some(1), Some(2)) == Some(1)` (the reported case)
- `combine_amd_gpu_counts(Some(2), Some(8)) == Some(2)` (KFD < DRM, more display/render cards)
- `combine_amd_gpu_counts(Some(3), Some(1)) == Some(3)` (KFD > DRM, e.g. multi-partition)

`cargo fmt`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace --all-targets` all pass.

volen-silo:
Addressed the changes requested in the device-probe review.

**`cabf8a8` — `combine_amd_gpu_counts` now bounds the present-GPU count by the compute-authoritative KFD count.** Previously it took `max(kfd, drm)`, so a display/render-only AMD DRM card with no KFD compute node could inflate the count and let an explicit `--gpu N` pass the engine gates only to fail later inside HIP. KFD now wins whenever it reports at least one GPU; DRM is used only as the zero-KFD fallback for APU shapes. Added regression tests for the reported case (`Some(1), Some(2) -> Some(1)`) plus both divergent nonzero directions (KFD<DRM, KFD>DRM).

Verification (all green locally): `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace --all-targets` (0 failures; rocm-core 180 tests). Commit is signed (Verified) with a DCO sign-off.

rominf (commented):
The previously requested correctness fixes are present, and I found no remaining blocking issue in the current head (`cabf8a8`). In particular, the Linux-only hardware scenarios are scoped correctly, unsupported visibility masks are treated as unprobeable rather than authoritative zero devices, vLLM has the engine-side availability check, and a nonzero KFD compute-device count is authoritative over DRM.

Non-blocking follow-ups:

1. `crates/rocm-core/src/lib.rs:3516` says the probe applies `HIP_VISIBLE_DEVICES`, then `ROCR_VISIBLE_DEVICES`, while `visibility_mask_from_env()` uses only the first configured variable, preferring HIP. Rewording the public doc comment would make that limitation clearer.
2. `tests/e2e-cucumber/features/model_serving.feature:109` could also carry `@requires-os:linux`. The current no-GPU mock lane is Linux, so this does not affect present CI, but the underlying authoritative probe is Linux-only and a future non-Linux no-GPU lane would otherwise run an incompatible scenario.
3. Explicit CLI `--gpu` validation remains mask-unaware: `apps/rocm/src/main.rs:15207` validates against the unmasked `amd-smi` count, while the engine validates against the visibility-aware probe. This fails safely, but a masked index can appear in the serve plan before the engine rejects it.

Local targeted verification passed: rocm-core (180 tests), Lemonade (25), vLLM (31), E2E Cucumber (26), and `cargo clippy -p rocm-core --all-targets`.

All completed CI checks are green. The generic GPU E2E and Strix Halo Windows E2E jobs are still queued, so I am leaving this as a comment rather than approval until those hardware jobs complete.

rominf (approved):

PR #123, opened 2026-07-16
fix(driver): widen Rocky to 9.x and clarify Debian repo mapping

volen-silo:
## Summary
- Widen the driver-install matrix to accept Rocky Linux 9.4/9.6/9.7 (previously only 9.7), matching the RHEL 9.x set.
- Annotate the Debian install plan's `reason` so the Ubuntu codename it emits (12→jammy, 13→noble) reads as a deliberate, AMD-documented mapping rather than a misdetection.
- Add regression tests covering the above and locking in the existing (correct) Debian and EL behavior.

## Why / context
This addresses an audit that flagged the distro repository selection in `apps/rocm/src/main.rs`. Checked against AMD's official ROCm install docs, two of the three flagged items were **false positives** — the code already matches AMD:
- **Debian → Ubuntu suite is intentional.** AMD's [Debian native-install docs](https://rocm.docs.amd.com/projects/install-on-linux/en/latest/install/install-methods/package-manager/package-manager-debian.html) map Debian 12→jammy, 13→noble and serve them from the `.../ubuntu` graphics tree. Marking Debian unsupported or using `bookworm`/`trixie` would break the documented install.
- **EL point-release paths already match AMD's mixed convention** (`el/8`, `el/10` major; `el/9.x` point release), per the [RHEL docs](https://rocm.docs.amd.com/projects/install-on-linux/en/latest/install/install-methods/package-manager/package-manager-rhel.html).
- **Unsupported systems already fail without mutating package sources** — the fallthrough arm is non-mutating and command-less, and execution is gated on `supported && mutating`.

So this PR ships only the genuinely defensible residual: the Rocky 9.x consistency gap (a real inconsistency vs. RHEL) and a transparency note (the surprise that likely triggered the audit).

## Non-obvious decisions
- Debian's deb line and `codename` field are unchanged; only the human-readable `reason` gains an explanatory sentence. Behavior is identical.
- Rocky stays scoped to 9.x only (AMD documents Rocky Linux 9); 8.x/10.x remain unsupported, guarded by a test.

## Scope boundaries (non-goals)
- No Debian behavior change and no EL path-convention change — both are already correct.
- No new-distro support (e.g. AlmaLinux); the Red Hat compatibility boundary is deliberately not reopened.
- The SLES path's `zypper --no-gpg-checks` (a signature-verification bypass) is out of this Debian+EL scope and left for a follow-up.

## Test plan
- `cargo test -p rocm --bins driver_plan` → 11 passed (3 new: Rocky 9.4 flow, Rocky 8/10 unsupported, Debian Ubuntu-suite repo line + intentional note).
- `cargo fmt --check` and `cargo clippy -p rocm` clean.

## Risk
Low — additive matrix widening plus a message-only annotation, fully covered by unit tests; no change to command generation for existing supported distros.

### DIFF
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -2504,14 +2504,24 @@ fn build_driver_install_plan(
         ),
         ("debian", "12" | "13") => {
             let repo_codename = if version_id == "13" { "noble" } else { "jammy" };
-            apt_driver_plan(
+            let mut plan = apt_driver_plan(
                 os_id,
                 version_id,
                 repo_codename.to_owned(),
                 repo_version_expr,
                 dkms,
                 false,
-            )
+            );
+            // Debian deliberately reuses AMD's Ubuntu-suite repository: AMD's
+            // documented Debian install maps Debian 12 -> jammy and 13 -> noble
+            // and serves them from the .../ubuntu graphics tree. Surface that in
+            // the plan so the Ubuntu codename on a Debian host doesn't read as a
+            // misdetection.
+            plan.reason = format!(
+                "Debian intentionally uses AMD's Ubuntu-suite repository (codename {repo_codename}), per AMD's documented Debian install; the Ubuntu codename is deliberate, not a misdetection. {}",
+                plan.reason
+            );
+            plan
         }
         ("rhel", "10.1" | "10.0" | "9.7" | "9.6" | "9.4" | "8.10") => dnf_driver_plan(
             os_id,
@@ -2529,7 +2539,7 @@ fn build_driver_install_plan(
             dkms,
             DnfDriverDistro::Oracle,
         ),
-        ("rocky", "9.7") => dnf_driver_plan(
+        ("rocky", "9.4" | "9.6" | "9.7") => dnf_driver_plan(
             os_id,
             version_id,
             codename,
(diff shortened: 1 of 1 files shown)


michaelroy-amd:
Review findings for `93c000ddc01a66a62021e98bf183fa3a4d08a939`:

**No actionable code findings.**

I verified the changed driver-plan paths and tests:
- Rocky support remains explicitly limited to 9.4, 9.6, and 9.7.
- Rocky 9.x generates the expected `el/<point-release>` repository path and `.el9` package suffix.
- Rocky 8.10 and 10.0 remain unsupported, non-mutating, and command-free.
- Debian 12 -> `jammy` and Debian 13 -> `noble` match AMD's documented Debian repository mapping.
- Mutating driver plans continue to require explicit approval.

Local verification on the fetched PR head:
- `cargo test -p rocm --bin rocm driver_plan_ -- --nocapture`: 11 passed
- Three new focused regression tests: passed
- `cargo clippy -p rocm --all-targets -- -D warnings`: passed

Current GitHub state: the PR is mergeable, but not merge-ready because `E2E tests (Strix Halo, Windows)` is failing, `E2E tests (GPU)` remains queued, and required review is outstanding. The ordinary build, affected-tests, Windows build/test, Linux E2E, Ubuntu GPU E2E, CodeQL, hygiene, signatures/DCO, and clippy checks passed.

**Verdict:** code looks ready; resolve the remaining CI/review gates before merge.

rominf (approved):
Automated review: Approve.

The production changes are correct and well scoped:
- Rocky 9.4, 9.6, and 9.7 use the same `el/<point-release>/...-1.el9.noarch.rpm` path already used by the existing Rocky and RHEL 9.x flows; Rocky 8 and 10 remain unsupported.
- The Debian reason annotation is display-only. The actual codename and repository line are unchanged, and the documented Debian-to-Ubuntu-suite mapping is accurate.

All 11 driver-plan tests pass locally, and the new tests fail when the corresponding changes are reverted. I found no required blast-radius updates.

Non-blocking: Rocky 9.6 lacks its own positive case, although it shares the exact path covered by 9.4 and 9.7. Also, the point-release list models install-repository compatibility rather than a claim of identical GPU support; it will need to track future ROCm support-matrix updates.

The failing hardware E2E jobs report expectation drift and no unexpected failures; they do not exercise this distro-selection code.

PR #125, opened 2026-07-16
ci: cache rustc with sccache on compile-heavy jobs

michaelroy-amd:
## Problem

CI compilation dominates several required jobs, and the existing Rust cache does not preserve workspace crate outputs between runs.

The obvious backend for this is the GitHub Actions cache, but it does not fit: that cache is capped at 10 GB per repository with LRU eviction, and this repository already sits at ~10.6 GB. Adding rustc outputs there would evict the existing Rust and tool caches rather than add to them. The compile cache therefore goes to Azure blob storage, leaving the 10 GB Actions cache to the dependency caches that already rely on it.

## Change

- Install SHA-pinned `mozilla-actions/sccache-action` before Rust cache setup.
- Set `RUSTC_WRAPPER=sccache` in `build-and-test`, `windows-build-and-test`, and `clippy`, pointing sccache at an Azure blob container via `SCCACHE_AZURE_CONNECTION_STRING` (repository secret) and a per-job `SCCACHE_AZURE_KEY_PREFIX`.
- Restrict cache writes to `push` and `merge_group` (`SCCACHE_AZURE_RW_MODE`); pull requests read the cache but never populate it (see below).
- Probe the backend before any cargo step and degrade to a purely local cache if it is unreachable (see below).
- Preserve existing affected-path gates and `CARGO_INCREMENTAL=0` behavior.
- Report `sccache --show-stats` after each compile-heavy job.
- Leave coverage, E2E, signing, legal, and reporting jobs unchanged. The affected-crates `test` job is also deliberately left off sccache: on pull requests it compiles a crate subset, and a subset resolves Cargo features differently than the full workspace, so much of what it builds would get different keys than the compile-heavy jobs' entries and neither hit them nor be reusable by them.

`clippy` deliberately shares the `build-and-test` key prefix. The two jobs do not share most artifacts — clippy omits `--emit=link` for dependencies, which changes the cache key — but fully linked crates such as proc macros and build-script dependencies do hit across them, and the run below confirms it.

### Availability

sccache hard-fails at startup when the storage credential is missing or invalid, and v0.17.0 has no flag to soften that. Left alone, a rotated secret or a fork PR (which receives empty secrets) would red out three required checks for a reason unrelated to the change under test.

So a probe step starts the sccache server first. If the credential is absent or the backend is unreachable, it clears the Azure variables and lets sccache fall back to its always-available local disk cache; the build still runs, just without a shared cache. `RUSTC_WRAPPER` is set at job level and left alone by the probe on every path except the one where the sccache binary is missing altogether — there it must be cleared, because there is no wrapper to run. That distinction matters because `Swatinem/rust-cache` hashes `RUST*` environment variables into both its cache key and its restore key: toggling the wrapper moves those keys and forces a full cold dependency rebuild, so the paths that merely lose the *remote* backend deliberately leave it in place. The install step is `continue-on-error` and the probe checks that the binary exists, so a failed action release cannot break the build either.

Startup-time store validation is v0.17.0 behavior rather than a documented contract, so bumping the pinned version means re-verifying the four degradation paths; the probe comment says so in place.

### Cache write access

Any same-repo branch PR receives the credential, so an unrestricted cache would let anyone who can push a branch write objects that a later `main` build reads back. Keys are derived from the compiler, source, and flags, so a credential holder can compute the key a trusted build will look up and store an arbitrary artifact under it — code execution in that build. `SCCACHE_AZURE_RW_MODE` is therefore `READ_WRITE` only on `push` and `merge_group`, and `READ_ONLY` on pull requests. Pull requests keep the read-side speedup, which is where the measured gain comes from, and sccache skips the write half of its startup store check in read-only mode, so the availability probe above is unaffected.

Two related items are infrastructure follow-ups rather than workflow changes, and are not addressed here:

- **Retention.** The container is unbounded — `SCCACHE_CACHE_SIZE` governs only sccache's local disk cache, and cloud backends never evict. This needs a blob lifecycle rule (expire entries untouched for N days).
- **Credential scope.** `SCCACHE_AZURE_CONNECTION_STRING` currently carries an account key, which grants more than the one container. A container-scoped SAS would narrow that and needs no workflow change: sccache passes the string to OpenDAL, which accepts `SharedAccessSignature=` and prefers it over `AccountKey`.

## Verification

Two CI runs on this branch, cold then warm against the same Azure container:

| Job | Cold | Warm | Speedup | Warm hit rate |
|---|---:|---:|---:|---:|
| clippy | 208s | 106s | 1.96x | 99.90% (1041 hits / 1 miss) |
| Linux build/test | 588s | 126s | 4.67x | 100.00% (26 hits / 0 misses) |
| Windows build/test | 1299s | 418s | 3.11x | 100.00% (51 hits / 0 misses) |

Zero cache errors on every job in both runs, and `sccache --show-stats` reports the Azure backend (`azblob`) in each. The cold Linux run already showed a 38.62% hit rate because clippy had populated the shared prefix earlier in the same run, confirming the cross-job sharing described above.

Those runs predate the read-only restriction, so their warm pass both wrote and read. The speedups are read-side and carry over, but a pull request no longer caches artifacts unique to itself, and until this merges and `main` runs once there is nothing populated for pull requests to read.

The four degradation paths (missing credential, invalid credential, backend unreachable, sccache binary absent) were each exercised against the real v0.17.0 binary in both bash and PowerShell; all four compile successfully without a shared cache. After the probe changed, all four were re-exercised in both shells — this time driving each branch with a stub `sccache`, since branch selection rather than sccache's own behavior is what changed. Every path exits 0 and leaves the expected environment and step outputs, and the PowerShell run writes a plain UTF-8 environment file.

Also verified: workspace clippy with warnings denied, workspace build and smoke test, workflow YAML and action-pin checks, `xtask` workflow-contract tests, signed commits with DCO sign-off, OSS leak scan clean.

One unrelated flake appeared on the warm Windows run: `rocm_core::fix::tests::the_path_fix_finds_the_install_the_rest_of_the_cli_found`, which mutates the process-global `ROCM_PATH`. The Windows lane runs `cargo test` rather than nextest, so it shares one process across threads. The same test passed on the previous attempt with identical code, this PR changes no Rust source, and a re-run was green.

This rebuilds the intent of #71 from current `main`; it does not resolve the stale workflow branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -221,15 +221,17 @@ jobs:
         if: needs.changes.outputs.heavy == 'true'
         run: sudo apt-get update && sudo apt-get install -y pkg-config libcap-dev
 
-      - uses: actions-rust-lang/setup-rust-toolchain@166cdcfd11aee3cb47222f9ddb555ce30ddb9659 # v1.17.0
-        if: needs.changes.outputs.heavy == 'true'
-
       # Installs the sccache binary and wires it to the GitHub Actions cache
-      # backend. Runs AFTER setup-rust-toolchain so the wrapper is on PATH before
-      # the first cargo compile. Same `heavy` gate as the cargo steps below.
+      # backend. Runs BEFORE setup-rust-toolchain: RUSTC_WRAPPER=sccache is set at
+      # the job level, and setup-rust-toolchain's rust-cache runs `cargo metadata`
+      # (which honours RUSTC_WRAPPER) during setup — so sccache must already be on
+      # PATH by then. Same `heavy` gate as the cargo steps below.
       - name: Set up sccache
         if: needs.changes.outputs.heavy == 'true'
         uses: mozilla-actions/sccache-action@9e7fa8a12102821edf02ca5dbea1acd0f89a2696 # v0.0.10
+
+      - uses: actions-rust-lang/setup-rust-toolchain@166cdcfd11aee3cb47222f9ddb555ce30ddb9659 # v1.17.0
+        if: needs.changes.outputs.heavy == 'true'
 
       # Build the runtime binaries the smoke + acceptance steps below need (every
       # engine binary; smoke hard-fails on a missing one). Unit/integration tests
@@ -372,16 +374,18 @@ jobs:
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
+      # Installs sccache and wires it to the GitHub Actions cache. Runs BEFORE
+      # setup-rust-toolchain: RUSTC_WRAPPER=sccache is set at the job level and
+      # setup-rust-toolchain's rust-cache runs `cargo metadata` (which honours
+      # RUSTC_WRAPPER) during setup, so sccache must already be on PATH by then.
+      - name: Set up sccache
+        if: needs.changes.outputs.heavy == 'true'
+        uses: mozilla-actions/sccache-action@9e7fa8a12102821edf02ca5dbea1acd0f89a2696 # v0.0.10
+
       - uses: actions-rust-lang/setup-rust-toolchain@166cdcfd11aee3cb47222f9ddb555ce30ddb9659 # v1.17.0
         if: needs.changes.outputs.heavy == 'true'
         with:
           cache-on-failure: true
-
-      # Installs sccache and wires it to the GitHub Actions cache. Runs after
-      # setup-rust-toolchain so the wrapper is on PATH before the first compile.
-      - name: Set up sccache
-        if: needs.changes.outputs.heavy == 'true'
-        uses: mozilla-actions/sccache-action@9e7fa8a12102821edf02ca5dbea1acd0f89a2696 # v0.0.10
 
       - name: Build
         if: needs.changes.outputs.heavy == 'true'
@@ -495,14 +499,16 @@ jobs:
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
-      - uses: actions-rust-lang/setup-rust-toolchain@166cdcfd11aee3cb47222f9ddb555ce30ddb9659 # v1.17.0
-        if: needs.changes.outputs.rust == 'true'
-
-      # Installs sccache and wires it to the GitHub Actions cache. Runs after
-      # setup-rust-toolchain so the wrapper is on PATH before the first compile.
+      # Installs sccache and wires it to the GitHub Actions cache. Runs BEFORE
+      # setup-rust-toolchain: RUSTC_WRAPPER=sccache is set at the job level and
+      # setup-rust-toolchain's rust-cache runs `cargo metadata` (which honours
+      # RUSTC_WRAPPER) during setup, so sccache must already be on PATH by then.
       - name: Set up sccache
         if: needs.changes.outputs.rust == 'true'
         uses: mozilla-actions/sccache-action@9e7fa8a12102821edf02ca5dbea1acd0f89a2696 # v0.0.10
+
+      - uses: actions-rust-lang/setup-rust-toolchain@166cdcfd11aee3cb47222f9ddb555ce30ddb9659 # v1.17.0
+        if: needs.changes.outputs.rust == 'true'
 
       - name: Clippy
         if: needs.changes.outputs.rust == 'true'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -211,7 +211,22 @@ jobs:
     # thousands of small per-artifact entries on top of that, which under a full
     # cache are evicted before the next run reads them — leaving the cold-cache
     # penalty (a populating run compiles everything twice over) without the warm
-    # benefit. An external store has no such cap, so entries survive between runs.
+    # benefit. An external store has no such cap, so entries survive between runs
+    # — but nothing evicts them either: SCCACHE_CACHE_SIZE bounds only sccache's
+    # local disk cache, never a cloud backend. Retention is therefore the
+    # container's responsibility and has to come from a blob lifecycle rule
+    # (expire entries untouched for N days), not from anything in this workflow.
+    #
+    # Only trusted events populate the cache. Every branch a contributor can push
+    # receives the credential on its pull_request run, and a shared cache that
+    # such a run may write to is a code-execution path into a later main build:
+    # keys are derived from the compiler, source and flags, so anyone holding the
+    # credential can compute the key a trusted build will look up and store an
+    # arbitrary artifact under it. Limiting writes to push and merge_group leaves
+    # only main and the merge queue able to populate, while pull requests keep the
+    # full read-side speedup — which is where the measured gain comes from anyway.
+    # READ_ONLY also short-circuits the write half of sccache's startup store
+    # check, so the probe below behaves identically on pull requests.
     #
     # RUSTC_WRAPPER is set unconditionally rather than only when the backend is
     # reachable: rust-cache hashes every environment variable starting with
@@ -227,6 +242,7 @@ jobs:
       # Namespaced so the Linux and Windows lanes cannot collide, and so the
       # container stays legible if it is ever shared with another consumer.
       SCCACHE_AZURE_KEY_PREFIX: rocm-cli/build-and-test
+      SCCACHE_AZURE_RW_MODE: ${{ (github.event_name == 'push' || github.event_name == 'merge_group') && 'READ_WRITE' || 'READ_ONLY' }}
     # Don't spend per-OS build/test cycles unless the cheap lint gate is green.
     needs: [changes, clippy, prek]
     # A manual E2E dispatch skips this heavy job for a fast loop; the E2E jobs
@@ -271,6 +287,13 @@ jobs:
       # the rust-cache key stays put) that always starts, just without cross-run
       # reuse. Pull requests from forks get empty secrets and take the same path,
       # since sccache treats an empty value as unset.
+      #
+      # Bumping the pinned sccache version above means re-verifying the four
+      # degradation paths this handles (no binary, absent credential, invalid
+      # credential, unreachable backend). Validating the store at server startup
+      # is v0.17.0 behaviour, not a documented contract: a release that validated
+      # lazily instead would turn this step into a silent no-op and move the
+      # hard failure back onto the first cargo step, across three required checks.
       - name: Probe sccache cache backend
         id: sccache_probe
         if: needs.changes.outputs.heavy == 'true'
@@ -282,16 +305,19 @@ jobs:
           if ! command -v sccache >/dev/null; then
             echo "::warning::sccache is not installed; compiling without it."
             echo "RUSTC_WRAPPER=" >> "$GITHUB_ENV"
+            # Nothing will read the credential now; don't leave it in the job env.
+            disable
             exit 0
           fi
+          # sccache is the compiler wrapper from here on, remote backend or not,
+          # so there are statistics worth reporting either way.
+          echo "active=true" >> "$GITHUB_OUTPUT"
           if [ -z "${SCCACHE_AZURE_CONNECTION_STRING}" ]; then
             echo "::notice::No cache credential available (expected on forks); compiling without a shared cache."
             disable
             exit 0
           fi
-          if sccache --start-server; then
-            echo "enabled=true" >> "$GITHUB_OUTPUT"
-          else
+          if ! sccache --start-server; then
             echo "::warning::sccache could not reach the cache backend; compiling without a shared cache."
             # Leave no half-started server behind for the cargo steps to find.
             sccache --stop-server || true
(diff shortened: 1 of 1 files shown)


tomastola (approved):
**Verdict: Approve with comments.** The mechanism is correct — I verified the three non-obvious behaviors it rests on rather than taking them on trust, and all three hold. My substantive comments are about the *operational* side of introducing a shared, writable compile cache: who may write to it, and what bounds its growth. Neither blocks merge, but the first is worth resolving before this becomes load-bearing.

## What I verified (rather than assumed)

The design leans on three behaviors that are easy to get wrong. All three check out:

**1. The degradation mechanism actually works.** This is the crux: the probe blanks `SCCACHE_AZURE_*` via `$GITHUB_ENV`, but those same names are declared in the job-level `env:` block. There's a widely repeated claim that job-level `env:` shadows `$GITHUB_ENV` writes, which would make the entire fallback a no-op. That claim is wrong, and specifically it conflates job-level with step-level. From the runner source:

- `JobExtension.cs:243` — job-level `env:` is evaluated **once at job start** and written into `context.Global.EnvironmentVariables`.
- `FileCommandManager.cs:209` — `$GITHUB_ENV` writes update **that same dictionary**, overwriting it.
- `StepsRunner.cs:109-112` — each step seeds its env context from `Global.EnvironmentVariables`, i.e. post-`GITHUB_ENV`.
- `StepsRunner.cs:124-128` — only the **step-level** `env:` block overwrites on top.

So job-level `env:` is a seed, not a per-step re-application, and the probe's writes win. I also checked that no step-level `env:` in these jobs redeclares an `SCCACHE_*` or `RUSTC_*` key — the two "Acceptance install lifecycle" steps have step-level `env:`, but only for `E2E_INCLUDE_LIFECYCLE`/`E2E_ONLY_LIFECYCLE`, so nothing is shadowed. Correct.

**2. Empty values mean "unset" on both sides.** sccache v0.17.0 `config.rs:836`: `string_from_env_var` is `env::var(name).ok().filter(|s| !s.is_empty())`, and the Azure block (`config.rs:1069-1072`) only constructs a config when *both* the connection string and container are `Some`. Blanking them cleanly disables the remote backend. On the cargo side, an empty `RUSTC_WRAPPER` is documented as an explicit "no wrapper" that also overrides `build.rustc-wrapper` config. The PR body's claim is accurate, and it's accurate *for the pinned version*, which is the part that matters.

**3. The pin is real and doesn't smuggle in the GHA backend.** `fc920bf0` is the commit the annotated `v0.0.11` tag dereferences to. And since the whole rationale is avoiding the 10 GB Actions cache, I checked whether the action quietly enables it: at that revision it exports only `SCCACHE_PATH`, `ACTIONS_CACHE_SERVICE_V2`, `ACTIONS_RESULTS_URL`, `ACTIONS_RUNTIME_TOKEN` — **no `SCCACHE_GHA_ENABLED`** (grep count 0). Azure is genuinely the backend in use.

I also confirmed gate consistency mechanically: every cargo-invoking step in all three jobs carries exactly the same `if:` as that job's install+probe (`heavy` for the two build jobs, `rust` for clippy). There is no path where cargo runs with `RUSTC_WRAPPER=sccache` and no sccache.

The ordering rationale is the sharpest part of the change and it's correct: install and probe must precede `setup-rust-toolchain` because rust-cache runs `cargo metadata`, which honors `RUSTC_WRAPPER`. That's a real failure mode and it's easy to miss.

## Comments

### 1. PR branches get read-write access to the cache `main` reads from

This is the one I'd want addressed. Triggers are `pull_request` (not `pull_request_target`) with `permissions: contents: read`, so **fork PRs are safe** — empty secrets, and they take the local-cache path by design. The exposure is same-repo branch PRs: any contributor who can push a branch gets the secret and `READ_WRITE` on the shared prefix, and `build-and-test`/`clippy` PR runs write into the exact prefix that `main` builds later read from.

sccache keys include a hash of the compiler binary, source, and flags, which prevents *accidental* collisions — the in-diff comment is right that sharing between clippy and build-and-test can't poison either job. But it isn't a defense against someone holding the account key, who can compute the key a future trusted build will look up and store an arbitrary object there. That's code execution in a later `main` build.

Two mitigations, both cheap:

- Make PR runs read-only. sccache supports this natively (`config.rs:1075`, values `READ_ONLY`/`READ_WRITE`, defaulting to `READ_WRITE`):
  ```yaml
  SCCACHE_AZURE_RW_MODE: ${{ (github.event_name == 'push' || github.event_name == 'merge_group') && 'READ_WRITE' || 'READ_ONLY' }}
  ```
  PRs keep the full read speedup; only trusted events populate. Given the measured numbers come from *reading* a warm cache, this should cost close to nothing.
- Prefer a container-scoped SAS token over an account connection string. `SCCACHE_AZURE_CONNECTION_STRING` embeds the account key, which grants far more than the one container — and it's exposed to every step in three jobs, including third-party actions. They're SHA-pinned, which mitigates, but the blast radius is wider than it needs to be.

If the team's position is that push access is already trust enough, that's a legitimate call — but it should be a decision recorded in the comment block, not an implicit default.

### 2. Nothing bounds the container's growth

`SCCACHE_CACHE_SIZE` is parsed only under the `======= Local =======` section (`config.rs:1157-1159`) — it governs the local disk cache. The cloud backends don't evict. The PR argues for Azure because "an external store has no such cap," which is true and is exactly why this needs an answer: the container grows monotonically, thousands of small objects per run, forever. This needs an Azure blob lifecycle management rule (e.g. delete after N days of no access) and a note in the comment saying where that policy lives. Worth doing now, while the container is small.

### 3. The probe rests on an undocumented startup behavior

"sccache validates the store when its server starts" is an implementation detail of v0.17.0, not a documented contract. The version pin protects it today, which is good — but a future bump could move to lazy validation and silently turn the probe into a no-op, at which point the first cargo step hard-fails on three required checks. That's precisely the outage this PR exists to prevent. A line in the probe comment — *"bumping the pinned sccache version requires re-verifying the four degradation paths"* — would keep the next person from bumping it casually. The four-path verification is genuinely impressive; make it repeatable.

### 4. `test` is a compile job that didn't get sccache

The PR body enumerates what's intentionally excluded (coverage, E2E, signing, legal, reporting) but doesn't mention `test`, which runs `cargo nextest run` on affected crates (`ci.yml:424`) on the same Linux target. It would read the existing `build-and-test` prefix with likely-high hit rates. Excluding `coverage` is well-justified (`-C instrument-coverage` changes every key), but `test` looks like it was simply overlooked. Deliberate or not, it's worth a sentence.

### 5. Smaller items

- **The PR body overstates one invariant.** It says `RUSTC_WRAPPER` is "set unconditionally at job level and never toggled by the probe" — but the missing-binary branch does set `RUSTC_WRAPPER=`, and that branch moves the rust-cache key, which is the exact thing the comment says the design avoids. The behavior is right (with no binary, the wrapper must go), and the in-diff comment is more careful than the body. Just tighten the body's claim to "never toggled except when the binary is absent."
- **The stats-step comment doesn't match its condition.** It reads "Skipped when the probe left the wrapper off, where there are no statistics to report" — but on the no-credential/fork path the wrapper is *on* and local stats do exist; they're skipped anyway because `enabled` is only set on the remote-success path. Minor, but it means fork PRs get zero cache visibility. If that's intended, say so; if not, gate on the step having run at all.
- **The missing-binary branch skips `disable()`**, leaving `SCCACHE_AZURE_CONNECTION_STRING` populated for the rest of the job. Harmless (nothing reads it), but inconsistent with the other branch and needlessly keeps the secret in the environment.
- **The action already registers a `post: show_stats` step**, so stats print at job end regardless. Your explicit steps are still better placed — right after the compile, before later `cargo xtask` steps add noise — so I'd keep them, but a half-line saying why would prevent someone deleting them as redundant.
- **pwsh encoding.** `Out-File -Append` to `$GITHUB_ENV` is safe *only* because `shell: pwsh` (PS7 defaults to UTF-8). Under `shell: powershell` (5.1) it writes UTF-16LE and silently corrupts the env file. Adding `-Encoding utf8` costs nothing and makes the script robust to that edit.

## Things done well

Worth saying explicitly, since these are the parts most PRs skip: the cold/warm verification with per-job hit rates, all four degradation paths exercised against the real pinned binary in *both* shells, SHA-pinning the action *and* pinning `version` so the compiler wrapper can't float, `$PSNativeCommandUseErrorActionPreference = $false` to stop a future runner default from turning a failed probe into a failed step, and the `sccache --stop-server` cleanup so no half-started server is left for cargo to find. The Linux hit rates also empirically confirm the `CARGO_INCREMENTAL=0` assumption for those jobs — 99.9% would be impossible otherwise. And the Windows flake was diagnosed to a specific process-global test rather than waved off.

---

**Sources for the external behaviors I checked:** [Cargo environment variables reference](https://doc.rust-lang.org/cargo/reference/environment-variables.html) and [rust-lang/cargo#5985](https://github.com/rust-lang/cargo/pull/5985) (empty `RUSTC_WRAPPER`); [GitHub Docs — variables](https://docs.github.com/en/actions/reference/workflows-and-actions/variables) (env scoping — note the common community claim about job-level `env:` shadowing `$GITHUB_ENV` is what the runner source above contradicts); sccache `v0.17.0` `src/config.rs` and `Mozilla-Actions/sccache-action@fc920bf` `src/setup.ts`.

rominf:
Thanks — checking the runner's env precedence against `JobExtension`/`StepsRunner` rather than the widely-repeated claim is exactly the thing that would have been expensive to get wrong, since the whole fallback hangs on it.

Pushed `20cdbc20`. Flagging explicitly that this changes code you already approved (this repo does not dismiss approvals on push): it adds the read-only gate, four probe fixes, and comments; it does not touch the install/probe ordering or the key prefixes.

### 1. PR branches get read-write access

Adopted the read-only mitigation — `SCCACHE_AZURE_RW_MODE` is `READ_WRITE` only on `push`/`merge_group`, `READ_ONLY` everywhere else, with the reasoning recorded in the comment block rather than left implicit.

One thing worth adding, because it decides whether this is safe at all: I expected `READ_ONLY` to break the availability probe, since `--start-server` validates the store and a read-only credential can't complete a write check — which would have quietly dropped every PR onto the local-only path. It doesn't. `RemoteStorage::check()` reads `.sccache_check`, then returns early *before* the write probe when `rw_mode == ReadOnly` (`src/cache/cache.rs:277-282`). Read-only is validated read-side only, so the probe behaves identically on PRs.

On the SAS: agreed, and it needs no workflow change. sccache passes the string to `Azblob::from_connection_string`, and OpenDAL 0.55's parser accepts `SharedAccessSignature=` and explicitly prefers it over `AccountKey` (`core/src/raw/azure.rs:203`; there's a "prefers sas over key" case at `:382`). So `BlobEndpoint=…;SharedAccessSignature=…` is a drop-in rotation of the existing secret. That's an action on the storage account rather than something this PR can carry — it's recorded as a follow-up in the description, and I'm happy to open an issue if you'd rather it were tracked.

### 2. Nothing bounds the container's growth

Confirmed independently: `SCCACHE_CACHE_SIZE` is parsed into `disk_sz` and only ever reaches the disk config (`config.rs:1159` → `:1182`); no cloud backend evicts. The workflow comment and the description now say that retention has to come from a blob lifecycle rule and cannot come from anything in this workflow.

I deliberately did *not* write that such a rule exists — I can't verify that from here, and it needs whoever owns the storage account. Second ops follow-up, same offer to file it.

### 3. The probe rests on an undocumented startup behavior

Added to the probe comment: bumping the pinned version means re-verifying the four degradation paths, with the reason (startup-time validation is v0.17.0 behavior, not a contract; lazy validation would make the probe a silent no-op and move the hard failure onto three required checks).

### 4. `test` is a compile job that didn't get sccache

Overlooked rather than decided — thanks for catching it. Recorded as a deliberate exclusion now, and I chose not to wire it up: on PRs that job compiles a crate subset, and subset feature resolution differs from the full workspace, so much of what it builds gets different keys than the compile-heavy jobs' entries — it would neither hit them nor be reusable by them, while adding write traffic to a store nothing evicts (your #2). It already has its own `rust-cache` lane for the same reason. Noted as revisit-if-it-becomes-a-bottleneck.

### 5. Smaller items

- **Body overstated the invariant** — tightened; it now says the wrapper is left alone on every path *except* the missing-binary one, and why that path is different.
- **Stats comment vs. condition** — fixed the condition rather than the comment, since you're right that fork PRs were losing their only cache signal. The probe now emits `active=true` immediately after the binary check, and all three stats steps gate on that, so the local-only paths report too. The `enabled` output is gone.
- **Missing-binary branch skipped `disable()`** — fixed in both shells.
- **Action's own post `show_stats`** — confirmed at the pinned SHA (`action.yml` declares `post: dist/show_stats/index.js`); added a half-line saying the explicit steps are kept because the post step runs after later `cargo xtask` compiles have added noise.
- **pwsh encoding** — `-Encoding utf8` on every `Out-File`, with a note on why.

### Re-verification

The probe's branching changed, so the earlier four-path evidence no longer covered it. Re-ran all four in both bash and pwsh, driving each branch with a stub `sccache` (branch selection is what changed, not sccache's behavior):

| Path | exit | `GITHUB_ENV` | `GITHUB_OUTPUT` |
|---|---:|---|---|
| binary absent | 0 | `RUSTC_WRAPPER=`, both Azure vars blanked | — (stats skipped) |
| no credential (fork) | 0 | both Azure vars blanked | `active=true` |
| backend unreachable | 0 | both Azure vars blanked, server stopped | `active=true` |
| healthy backend | 0 | untouched | `active=true` |

Identical across both shells, and the pwsh env file is plain single-byte UTF-8. Also re-ran the `xtask` workflow-contract tests (131 pass) and re-validated the workflow YAML.

One caveat I'd rather state than leave implicit: the timing table was measured before the write restriction, so those warm runs both wrote and read. The speedups are read-side and carry over, but PRs no longer cache anything unique to themselves, and until this merges and `main` runs once there is nothing populated for a PR to read. I didn't re-measure, so I haven't relabeled the numbers as read-only results.

rominf:
@tomastola this is ready to merge from my side — all required checks are green — but your approval is on `d84be7ef` and the head is now `20cdbc20`, which I pushed after it. This repo doesn't dismiss approvals on push, so GitHub still shows it as approved; I'd rather not merge on that.

The delta is the read-only write gate plus four probe fixes, described in the comment above. Since the write gate is the security-relevant part, a second look at that specific change seems worth it rather than letting the stale approval carry it.

One unrelated note so it isn't mistaken for a regression: `E2E tests (Strix Halo, Ubuntu)` shows cancelled. It hit the lane's 35-minute cap (ran 35m41s, cancelled mid-test-step). It's an advisory lane rather than a required check, the same cancellation is hitting `main` pushes and several unrelated branches right now, and this PR changes no Rust source.

PR #126, opened 2026-07-16
test(e2e): scale serve-readiness VRAM floor to device total

fredespi:
## Summary

The E2E serve-readiness step waits for GPU memory to drain before each serve, but compared against a hardcoded 150 GiB floor (`MIN_FREE_VRAM_MIB`) sized for the 27B model on MI300X. A smaller card (e.g. Strix Halo, ~48 GiB unified VRAM) can never reach that floor, so `wait_for_free_vram` burned its full 2-minute deadline on every serve scenario — pure dead time.

Fix: probe the device's total VRAM and cap the floor at 90% of it (`required_free_vram_mib`). MI300X keeps its 150 GiB floor for the 27B model; smaller cards get a reachable floor that clears promptly once a killed serve drains.

## Changes
- Add `total_vram_mib()` / `vram_mib()` probe (amd-smi `TOTAL_VRAM` + rocm-smi total) alongside the existing free-VRAM read.
- Replace the hardcoded `MIN_FREE_VRAM_MIB` compare with `required_free_vram_mib(total) = min(150_000, total * 0.9)`.

## Test plan
- [x] Linux container gate green: clippy (`--workspace` + `-p e2e-cucumber --test e2e`) and tests under `RUSTFLAGS="-D warnings"`; e2e mock reconciliation 0 unexpected failures.
- [x] Scoped Strix-Ubuntu dispatch: serve-readiness step dropped from ~262s to 91s (the ~120s dead VRAM-drain wait is eliminated); MI300X floor unchanged.
- [ ] Full CI on this PR.

PR #129, opened 2026-07-17
fix(e2e): pre-warm lemonade engine on Strix-Windows; xfail its flaky serves (EAI-7455)

fredespi:
## Summary

Fixes the near-always-failing `E2E tests (Strix Halo, Windows)` lane. Two distinct causes:

1. **Per-scenario 4.6 GiB backend re-download.** Each lemonade serve reinstalled its llama.cpp/TheRock backend into the scenario's isolated data dir, overrunning the serve-readiness timeout. Native Windows serves route to lemonade (vLLM is skipped there), so only that platform hit it. **Fix:** pre-warm `rocm engines install lemonade` once per runner (into the persisted tree, with a retry for the daemon startup race) so the backend installs once, not per scenario.

2. **Residual flaky lemonade daemon.** After the download is fixed, the embedded lemonade server still intermittently refuses the backend-install/config RPC on native Windows ("Could not connect to Lemonade server"), worse under full-suite load. A harness-side retry could not reliably paper over this product-side race, so the six affected lemonade serve/chat scenarios are **xfail'd on `os=windows`** referencing **EAI-7455** — consistent with the existing `os=linux` lemonade xfails.

Also surfaces STDERR in the serve-step failure assert (the real serve error was previously hidden behind stdout-only output), which is what made the daemon flake diagnosable.

## Changes

- `.github/workflows/ci.yml` — Strix-Windows job pre-warms the lemonade engine once per runner, with a bounded retry gated on the install exit code (handles the daemon startup race).
- `tests/e2e-cucumber/expectations.toml` — xfail the 6 lemonade serve/chat scenarios on `os=windows, effective_engine=lemonade` → EAI-7455.
- `tests/e2e-cucumber/tests/e2e/serving_steps.rs` — serve-step failure asserts now include STDERR + rc.

## Test plan

- [x] Linux container gate green (clippy + workspace tests + e2e-cucumber lib, `-D warnings`).
- [x] Full Strix-Windows dispatch reconciles **11 xfail / 0 XPASS / 0 unexpected failures** (run #726).
- [x] Confirmed the 4.6 GiB backend downloads once in pre-warm, not per scenario.

## Notes

- EAI-7455 tracks the underlying lemonade daemon flake (product-side; likely the lemonade server, with rocm-cli readiness-gating as the actionable seam). The xfails are the interim handling so the flake stops masking real regressions on this non-blocking lane.
- The Strix-Windows E2E job is `continue-on-error` (non-blocking).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1194,6 +1194,40 @@ jobs:
             Write-Host "shared runtime already present at $env:E2E_SHARED_RUNTIMES_DIR - skipping pre-warm"
           }
 
+          # Pre-warm the lemonade engine into the shared runtime tree, with a RETRY.
+          # On native Windows serves route to lemonade, whose ~4.6 GiB llama.cpp/
+          # TheRock backend installs UNDER the runtime tree (lemonade self-manages
+          # its runtime); pre-warming it once here means scenarios find it already
+          # installed instead of each re-downloading it and overrunning the serve
+          # timeout. RETRY because `engines install lemonade` starts the embedded
+          # lemonade server then immediately fires config-set + backend-install RPCs
+          # before it is always listening, so the first install can fail ("Could not
+          # connect to Lemonade server") though a retry succeeds (EAI-7455). Gate on
+          # the install's own exit code — the marker dir is created before the backend
+          # RPC, so Test-Path is not proof of success.
+          $lemonadeReady = $false
+          for ($i=1; $i -le 4; $i++) {
+            New-Item -ItemType Directory -Force -Path "$prewarm\data","$prewarm\config","$prewarm\cache" | Out-Null
+            $env:ROCM_CLI_CONFIG_DIR = "$prewarm\config"
+            $env:ROCM_CLI_DATA_DIR   = "$prewarm\data"
+            $env:ROCM_CLI_CACHE_DIR  = "$prewarm\cache"
+            Write-Host "pre-warming shared lemonade engine (attempt $i/4)..."
+            & $env:ROCM_CLI_BINARY engines install lemonade
+            $installRc = $LASTEXITCODE
+            Remove-Item Env:\ROCM_CLI_CONFIG_DIR,Env:\ROCM_CLI_DATA_DIR,Env:\ROCM_CLI_CACHE_DIR -ErrorAction SilentlyContinue
+            if ($installRc -eq 0) {
+              Write-Host "shared lemonade engine pre-warmed into $prewarm (attempt $i)"
+              $lemonadeReady = $true
+              break
+            }
+            Write-Host "lemonade pre-warm attempt $i failed (rc=$installRc); killing stray lemonade and retrying..."
+            Get-Process lemond,llama-server,lemonade -ErrorAction SilentlyContinue | Stop-Process -Force -ErrorAction SilentlyContinue
+            Start-Sleep -Seconds 5
+          }
+          if (-not $lemonadeReady) {
+            Write-Host "lemonade pre-warm did not succeed after retries; scenarios will install their own"
+          }
+
           # Optional scenario-name filter for a scoped manual dispatch.
           $nameFilter = "${{ github.event.inputs.name_filter }}"
           if ($nameFilter) {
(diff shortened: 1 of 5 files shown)


rominf (changes requested):
Automated review: requesting changes. I found three blocking issues at the current head.

1. `tests/e2e-cucumber/tests/e2e/serving_steps.rs:130` unintentionally reverses #126 by restoring a fixed 150 GiB free-VRAM floor and deleting the device-scaled calculation. On a roughly 48 GiB Strix system that floor is unreachable, so each serve scenario can consume the full two-minute wait. Please retain the device-scaled implementation from `main`.

2. `.github/workflows/ci.yml` removes both `cargo xtask verify-pinned-keys` steps and the `docs/keys/**` heavy-change trigger added by #95. That leaves canonical-key changes without PR-time verification. This is unrelated to the E2E change and should be restored.

3. The new expectations use `xfail` for a nondeterministic failure, but the reconciler exits nonzero on every XPASS (`tests/e2e-cucumber/tests/e2e.rs:791-804`). In this PR run the flake did not occur, all six new cases XPASSed, and the Strix-Windows job went red despite zero unexpected failures. The new claim that occasional XPASS is harmless is therefore false. Please use a mechanism that tolerates both outcomes, such as a first-class flaky/allow-XPASS state, or skip these cases until the underlying issue is fixed.

Non-blocking cleanup:
- `serving_steps.rs:169` says `--json`, while the command and parser use CSV.
- The independent engine-prewarm comment appears in two jobs that do not contain that prewarm loop.
- `setup_large_gpu_model` still discards stderr, unlike the sibling setup paths updated here.

The separate GPU-lane failure contains only unrelated XPASS results and no unexpected failure; I did not treat it as caused by this PR.

fredespi:
Thanks — all three blocking issues addressed in the force-pushed `3a959ee` (rebased onto current main, rebuilt as a single clean commit so the stale-base clobber is gone).

1. **VRAM floor (#126):** fixed. The branch was built on a stale base, which is what reverted the device-scaled floor. Rebased onto current main; the diff no longer touches `wait_for_free_vram`/`required_free_vram_mib` — `main`'s device-scaled implementation is intact.

2. **verify-pinned-keys (#95):** fixed, same root cause. The rebase restores both `Pinned key consistency check` steps and the `docs/keys/**` heavy trigger; the branch diff no longer touches them.

3. **XPASS on a nondeterministic xfail:** you're right that plain xfail + exit-1-on-any-XPASS reddens the lane on a good run — that happened here. Added a first-class `flaky = true` expectation marker: a flaky xfail's XPASS is reported but non-fatal, while unexpected FAIL and non-flaky (stale) XPASS stay fatal. The six EAI-7455 lemonade/Windows entries use it. (Filed as EAI-7456; also unblocks the EAI-7333 vLLM XPASS drift on other lanes if adopted there.)

Non-blocking cleanup also done:
- `setup_large_gpu_model` now surfaces STDERR like the sibling paths.
- Removed the stale `--json` comment / misplaced pre-warm comments.

Diff is now 5 files, +198/-21 (harness marker + expectations + serve-step stderr + Windows pre-warm). Re-requesting review.

fredespi:
Closing — superseded by #142.

#142 ("stabilize GPU E2E and merge queue") independently landed the core of what this PR carried: the first-class `flaky` expectation marker that makes an intermittent xfail's XPASS non-fatal (the EAI-7456 idea).

The remaining unique piece here — the six native-Windows Lemonade scenarios marked flaky-xfail under EAI-7455 — is also already covered by #142's approach: those were added and then intentionally reverted there, because the failure was traced to a **borrowed org runner** (Lemonade's GitHub release download hitting an SSL `CURL code 35` after retries), not a permanent defect — to be re-added only if a bad runner recurs. Re-introducing them here would contradict that and would xfail scenarios that currently pass: a fresh full Strix-Windows run on current `main` (run 30342563245) is green — 27 scenarios, 0 unexpected failures, all six Lemonade scenarios passing with a single clean backend download.

EAI-7455 stays open in the backlog to track the flake if it resurfaces on the real Windows runner. Thanks @rominf for the reviews.

PR #130, opened 2026-07-17
test(e2e): cover Strix large model nightly

rominf:
## Summary

- Run the nightly large-model inference scenario with the platform-specific serving path: `Qwen/Qwen3.6-27B` through vLLM on MI300X and `unsloth/Qwen3.6-35B-A3B-GGUF:UD-Q4_K_XL` through Lemonade on Strix Halo.
- Add non-blocking Linux and Windows Strix Halo jobs to the nightly workflow, reusing each runner's established GPU preflight, Rust bootstrap, and persistent managed-runtime setup.
- Allow manual CI dispatches to opt both Strix platforms into `@nightly` scenarios and retain a scenario-name filter for targeted hardware validation.
- Keep the shared scenario platform-adaptive rather than duplicating it by operating system or serving engine.
- Pin the touched GitHub Actions references to immutable commit SHAs.

This keeps large model loads off the ordinary per-PR path while covering the hardware-verified Strix Halo checkpoint on both supported operating systems. Linux and Windows remain separate jobs because their process cleanup, GPU probing, toolchain setup, paths, and shells are platform-specific.

Risk: medium. The change affects non-blocking self-hosted GPU jobs and the existing nightly-only scenario; ordinary PR E2E continues to use the small models.

## Test plan

Local and pre-push checks:

- `cargo test -p e2e-cucumber --lib capability` (10 passed)
- `actionlint` on `nightly.yml` with the repository's custom self-hosted labels allowed
- repository pre-push hooks, including full `cargo clippy` and `cargo test`
- workflow YAML, commit signature/sign-off, license, conflict-marker, and OSS leak checks

Hardware evidence:

- Strix Halo Linux passed the scoped large-model serve and inference scenario: https://github.com/ROCm/rocm-cli/actions/runs/29605158243
- The existing Windows Strix E2E job is green on this PR, but it predates the Windows nightly opt-in and therefore does not verify the large model.
- MI300X reached the existing `Qwen/Qwen3.6-27B` scenario but timed out waiting for `/v1/models` after 2400 seconds: https://github.com/ROCm/rocm-cli/actions/runs/29605156216. The MI300X model selection and timeout are unchanged by the Windows follow-up.

The new Windows large-model path still requires a self-hosted Strix Halo nightly or scoped dispatch for hardware verification.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,12 +28,11 @@ on:
         options: [all, mock, app-dev-gpu, strix-ubuntu, strix-windows]
         # No tier input: the harness resolves pass/xfail/skip per scenario, so a
         # platform runs one job covering everything applicable to it.
-      # TEMP (probe): scenario-name regex forwarded to the cucumber harness
-      # (`cargo xtask e2e -- --name <regex>`) so a dispatch can run just a few
-      # scenarios. Empty = full suite. REMOVE after the #23 default-engine
-      # classification is validated on Strix.
+      # Scenario-name regex forwarded to the cucumber harness
+      # (`cargo xtask e2e -- --name <regex>`) so a dispatch can run only the
+      # selected scenarios. Empty = full suite.
       name_filter:
-        description: "TEMP scenario-name regex (cucumber --name); empty = full suite"
+        description: "Scenario-name regex (cucumber --name); empty = full suite"
         type: string
         default: ""
       # Opt a manual dispatch into the expensive `@nightly` scenarios (large-model
@@ -637,7 +636,7 @@ jobs:
       && (github.event_name != 'workflow_dispatch'
           || inputs.platform == 'all' || inputs.platform == 'mock')
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
 
       # Whether the real work runs this time. On push/PR/merge_group: only when
       # build-and-test succeeded and the change is heavy. On workflow_dispatch:
@@ -659,7 +658,7 @@ jobs:
         if: steps.gate.outputs.run == 'true'
         run: sudo apt-get update && sudo apt-get install -y pkg-config libcap-dev
 
-      - uses: actions-rust-lang/setup-rust-toolchain@v1
+      - uses: actions-rust-lang/setup-rust-toolchain@166cdcfd11aee3cb47222f9ddb555ce30ddb9659 # v1.17.0
         if: steps.gate.outputs.run == 'true'
 
       # The workspace test job selects only affected crates and runs the cucumber
@@ -718,7 +717,7 @@ jobs:
       # a single nightly scenario (e.g. the 27B serve) without the full nightly run.
       E2E_INCLUDE_NIGHTLY: "${{ inputs.include_nightly && '1' || '' }}"
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
 
       # Reclaim the GPU before running: a serve leaked by a killed/timed-out prior
       # run (its Drop teardown never executed) can keep an engine process spinning
@@ -785,7 +784,7 @@ jobs:
       # Swatinem/rust-cache otherwise tries to SAVE the large target dir to
       # GitHub's cache service in a post-step (slow/hangs) and its cleanup wipes
       # the local target.
-      - uses: actions-rust-lang/setup-rust-toolchain@v1
+      - uses: actions-rust-lang/setup-rust-toolchain@166cdcfd11aee3cb47222f9ddb555ce30ddb9659 # v1.17.0
         with:
           cache: false
 
@@ -856,12 +855,11 @@ jobs:
             echo "shared runtime already present at $E2E_SHARED_RUNTIMES_DIR — skipping pre-warm"
           fi
 
-          # TEMP: optional scenario-name filter for a scoped probe dispatch (same
-          # as the strix job) — lets a dispatch run just one scenario (e.g. the
-          # 27B nightly) instead of the whole suite. Empty = full suite.
+          # Optional scenario-name filter for a scoped dispatch — lets a manual
+          # run select only the large-model scenario instead of the whole suite.
           NAME_FILTER="${{ github.event.inputs.name_filter }}"
           if [ -n "$NAME_FILTER" ]; then
-            echo "TEMP name filter active: $NAME_FILTER"
+            echo "name filter active: $NAME_FILTER"
             cargo xtask e2e -- --name "$NAME_FILTER"
           else
             cargo xtask e2e
@@ -910,8 +908,11 @@ jobs:
       TMPDIR: /home/ubuntu/actions-runner/tmp
       PIP_CACHE_DIR: /home/ubuntu/actions-runner/pip-cache
       E2E_SERVE_TIMEOUT_SECS: "300"
+      # Match the MI300X dispatch path: opt into the platform-adaptive large-model
+      # scenario only when the manual include_nightly input is enabled.
+      E2E_INCLUDE_NIGHTLY: "${{ inputs.include_nightly && '1' || '' }}"
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
 
       - name: Prepare writable dirs on the nvme + reclaim GPU from stray E2E procs
         run: |
@@ -1011,10 +1012,10 @@ jobs:
             echo "shared runtime already present at $E2E_SHARED_RUNTIMES_DIR — skipping pre-warm"
           fi
 
-          # TEMP: optional scenario-name filter for a scoped probe dispatch.
+          # Optional scenario-name filter for a scoped manual dispatch.
           NAME_FILTER="${{ github.event.inputs.name_filter }}"
           if [ -n "$NAME_FILTER" ]; then
-            echo "TEMP name filter active: $NAME_FILTER"
+            echo "name filter active: $NAME_FILTER"
             cargo xtask e2e -- --name "$NAME_FILTER"
           else
             cargo xtask e2e
@@ -1050,7 +1051,7 @@ jobs:
       )
     continue-on-error: true
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
 
       # setup-rust-toolchain runs an internal bash script, which this Windows
       # runner lacks (bash: command not found). Bootstrap rustup with the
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -1050,7 +1051,7 @@ jobs:
       )
     continue-on-error: true
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
```
volen-silo, line 1058: BLOCKING: this Windows job has no `env:` block, so `E2E_INCLUDE_NIGHTLY` is never set on the `workflow_dispatch` path. `e2e-gpu-strix-ubuntu` and `e2e-gpu` both set `E2E_INCLUDE_NIGHTLY: "${{ inputs.include_nightly && '1' || '' }}"`, and this PR added that line to Ubuntu but not here. The README's documented `-f platform=strix-windows -f include_nightly=true` command therefore skips the `@nightly` large-model scenario silently. Add before `steps:`:

```yaml
    env:
      E2E_INCLUDE_NIGHTLY: "${{ inputs.include_nightly && '1' || '' }}"
```

tests/e2e-cucumber/tests/e2e/serving_steps.rs
```diff
@@ -346,24 +346,37 @@ async fn setup_lemonade_model(world: &mut E2eWorld) {
...
+    let (model, engine, ready_substr) =
+        if e2e_cucumber::capability::host_capability().effective_serve_engine == "lemonade" {
+            // ready_substr is the base name WITHOUT the quant. lemonade serves
+            // this explicit `owner/repo:variant` checkpoint under its verbatim
+            // ref, so the `/v1/models` id keeps the `UD-Q4_K_XL` tag — but matching
+            // only the quant-free base keeps the readiness check robust to any
+            // future id-normalization on the serve path (e.g. the quant rewriting
+            // lemonade applies to shorthand refs like `Qwen3-0.6B-GGUF`).
+            (
+                "unsloth/Qwen3.6-35B-A3B-GGUF:UD-Q4_K_XL",
```
volen-silo, line 365: Non-blocking: `:UD-Q4_K_XL` is not the catalog's recommended quant. `model_catalog.json` marks this model `"featured": true` with canonical `unsloth/Qwen3.6-35B-A3B-GGUF:Q4_K_M` (artifact_hint: "the `:Q4_K_M` variant selects the quant file"). `resolve_model_recipe` matches refs exactly, so this explicit `:UD-Q4_K_XL` ref resolves to no recipe and bypasses the recipe-driven path. Given the scenario titles itself "platform-recommended"/"featured", consider using the alias `qwen3.6-35b-a3b` (→ the featured `:Q4_K_M`) or softening the framing to "a large Lemonade GGUF model".

tests/e2e-cucumber/tests/e2e/serving_steps.rs
```diff
@@ -709,15 +722,29 @@ fn model_ids_match(resp_model: &str, expected: &str) -> bool {
 }
 
 fn normalize_model_id(id: &str) -> String {
-    let mut s = id.to_ascii_lowercase();
+    let mut s = id
+        .rsplit('/')
```
volen-silo: Non-blocking: comparing on `rsplit('/').next()` drops the org prefix, so two different orgs' repos with the same model-name suffix would spuriously match — a slight weakening of the anti-leaked-serve guard this fn exists for. Low real-world risk. Minor: the two `unwrap_or(id)` fallbacks are unreachable (`next()` on a split iterator is always `Some`).

tests/e2e-cucumber/README.md
```diff
@@ -115,6 +116,33 @@ The blocking mock job passes when every applicable scenario is pass-or-xfail wit
...
+
+# Strix Halo Linux / Lemonade / Qwen3.6-35B-A3B-GGUF (UD-Q4_K_XL)
+gh workflow run ci.yml --ref <ref> \
+  -f platform=strix-ubuntu \
+  -f include_nightly=true \
+  -f name_filter='large platform-recommended model'
+
+# Strix Halo Windows / Lemonade / Qwen3.6-35B-A3B-GGUF (UD-Q4_K_XL)
+gh workflow run ci.yml --ref <ref> \
+  -f platform=strix-windows \
```
volen-silo, line 142: This documented command will silently skip the large-model scenario on Windows until `e2e-gpu-strix-windows` in ci.yml sets `E2E_INCLUDE_NIGHTLY` (see the blocking comment on ci.yml). Also minor: the paragraph above says the nightly workflow "adds three non-blocking jobs" — the MI300X `e2e-gpu-nightly` job pre-existed; this PR adds two (Strix Ubuntu + Windows).

volen-silo (dismissed):
## Review: platform-adaptive large-model nightly (Strix Halo Linux + Windows)

Solid, well-documented change. The core refactor — making the `@nightly` large-model scenario engine-agnostic and selecting the serve engine from the probed host capability (vLLM on MI300X, Lemonade on Strix) — is correct: I traced `expectation.rs::resolve` and `capability.rs::effective_serve_engine` and the `@requires-engine:vllm` removal is safe (mock/no-GPU hosts still skip via `@requires-gpu`; both engine branches line up model/engine/ready-substring). SHA-pinning and the `TEMP`/`#23` probe-comment cleanup are complete and consistent, the drift-guard unit tests pass locally (`cargo test -p e2e-cucumber --lib capability` → 10/10), and the PR description is honest about the MI300X timeout.

One blocking issue: the on-demand CI dispatch path for **Strix Windows** never sets `E2E_INCLUDE_NIGHTLY`, so the command the README documents silently skips the large-model scenario. Details inline. A couple of non-blocking accuracy/fragility notes below.

### Blocking

**`e2e-gpu-strix-windows` on-demand nightly opt-in is missing.** The `e2e-gpu-strix-windows` job in `ci.yml` has no `env:` block, so `E2E_INCLUDE_NIGHTLY` is never set on the `workflow_dispatch` path. The sibling `e2e-gpu-strix-ubuntu` and `e2e-gpu` (MI300X) jobs both set `E2E_INCLUDE_NIGHTLY: "${{ inputs.include_nightly && '1' || '' }}"`, and this PR *added* that line to the Ubuntu job — but not Windows. Consequence: the exact command this PR's README documents —
`gh workflow run ci.yml -f platform=strix-windows -f include_nightly=true -f name_filter='large platform-recommended model'`
— runs the job, applies the name filter, and then **skips** the only matching scenario (`expectation.rs` skips `@nightly` when `E2E_INCLUDE_NIGHTLY` is unset), silently with no failure. This contradicts the PR's stated goal ("Allow manual CI dispatches to opt **both** Strix platforms into `@nightly` scenarios") and its own docs. The scheduled `nightly.yml` Windows job is unaffected (it hardcodes `E2E_INCLUDE_NIGHTLY: "1"`); this gap is scoped to the on-demand path only.
Fix (one line, mirroring the Ubuntu job) — add an `env:` block before `steps:`:
```yaml
    env:
      E2E_INCLUDE_NIGHTLY: "${{ inputs.include_nightly && '1' || '' }}"
```

### Non-blocking

- **"Platform-recommended"/"featured" framing overstates what's tested on Strix.** The scenario serves the hardcoded ref `unsloth/Qwen3.6-35B-A3B-GGUF:UD-Q4_K_XL`, but the catalog's `"featured": true` entry for this model is `:Q4_K_M` (`crates/rocm-core/src/model_catalog.json`, `artifact_hint`: "the `:Q4_K_M` variant selects the quant file"; alias `qwen3.6-35b-a3b`). `UD-Q4_K_XL` appears nowhere else in the repo, and `resolve_model_recipe` matches refs exactly (`normalize_model_ref` only trims/lowercases, does not canonicalize `:variant`), so this explicit ref resolves to **no** catalog recipe — it bypasses the recipe-driven path that "dogfooding the platform-recommended model" presumably means to cover. Either point the scenario at the alias `qwen3.6-35b-a3b` (exercises the true featured `:Q4_K_M` path), or soften the title/comment to "a large Lemonade GGUF model". Neither commit message justifies the `UD-Q4_K_XL` choice.

- **Lemonade readiness has no `@serve-timeout` escape hatch (latent fragility).** `isolate_cmd` raises `ROCM_CLI_VLLM_READY_TIMEOUT_SECS` from the `@serve-timeout:2400` tag, but there is no lemonade equivalent — `serve_direct_llama_server` (the direct-llama-server GGUF path this scenario now hits on Strix Linux) hard-codes a **2-minute** readiness cap (`engines/lemonade/src/lib.rs`, `Duration::from_mins(2)`) and kills the child on timeout, with no env override. Your reported Strix Linux run passed (pre-seeded weights load fast), so this isn't breaking today — but the `@serve-timeout:2400` tag gives a false sense of headroom on the lemonade path: a colder/slower ~22 GB load would fail-fast at 2 min regardless of the tag or the 90-min job cap. Pre-existing product code (outside this diff), flagged because this PR is what newly routes a large cold GGUF through it. Worth a follow-up to make that cap configurable, mirroring the vLLM one.

- **`normalize_model_id` now compares basename only.** Taking `rsplit('/').next()` before comparing drops the org prefix, so two different orgs' repos sharing a model-name suffix would spuriously match — a slight weakening of the anti-leaked-serve guard this function exists for. Low real-world risk (served models differ by base name too); noting for awareness. (The two `unwrap_or(id)` fallbacks after `rsplit`/`split` are unreachable — `next()` on a split iterator is always `Some` — harmless but dead.)

### Tradeoffs (no change requested)

- `timeout-minutes: 90` on the new Strix jobs must cover a cold release build + possible first-run SDK prewarm + up to 2400s model load + inference. The pre-existing MI300X nightly job already uses 90 and has timed out once under this pattern; the Strix jobs inherit the same profile. Jobs are `continue-on-error: true`, so a cap-kill only degrades diagnostic clarity (ambiguous runner-kill vs. a clean scenario failure), not merge-gating.
- The two new jobs duplicate ~200 lines of GPU-preflight / rust-bootstrap / prewarm logic already present in `ci.yml`'s Strix jobs (4 near-identical copies total). The PR justifies separate jobs (platform-specific cleanup, paths, shells), which is reasonable; a composite action could de-duplicate later.
- `nightly.yml`'s pre-existing `nightly` and `e2e-gpu-nightly` jobs still use unpinned `@v1`/`@v6` action tags, while this PR pins the refs it touched — an internal inconsistency, not introduced here. A follow-up could pin them for uniformity.

.github/workflows/ci.yml
rominf: Fixed in 9607d6b: the Windows Strix job now forwards the manual `include_nightly` input through `E2E_INCLUDE_NIGHTLY`, matching the Linux and MI300X dispatch paths.

tests/e2e-cucumber/tests/e2e/serving_steps.rs
rominf: Kept the hardware-verified `UD-Q4_K_XL` checkpoint and softened the scenario/docs wording from catalog-recommended or featured to platform-specific/representative in 9607d6b. This avoids changing the tested artifact without new hardware evidence.

tests/e2e-cucumber/tests/e2e/serving_steps.rs
rominf: Fixed in 9607d6b: normalization now preserves and compares the organization when both IDs provide one, while still accepting Lemonade artifact IDs that omit it. I also removed the unreachable fallbacks and added focused tests for same-org, missing-org, and cross-org cases.

tests/e2e-cucumber/README.md
rominf: Fixed in 9607d6b: the Windows dispatch now enables nightly scenarios, and the README now distinguishes the existing MI300X job from the two newly added Strix jobs.

EugeneKSiloAi (approved):

PR #132, opened 2026-07-20
fix(dash): correct chat and scrollbar scrolling

michaelroy-amd:
## Summary
- measure chat scrolling in wrapped terminal rows and preserve follow-tail state across keyboard, mouse, and resize transitions
- align scrollbar rendering and hit-testing with Ratatui thumb geometry, including pointer-offset dragging and tail-anchored dock inversion
- avoid deep transcript text copies while rendering and add rendered-boundary regressions

## Test plan
- `cargo test -p rocm-dash-tui`
- `cargo clippy -p rocm-dash-tui --all-targets -- -D warnings`
- `cargo test --workspace --all-targets`
- `git diff --check origin/main..HEAD`

## Notes
- Scope is limited to the dashboard TUI; engine behavior and GPU-required policy are unchanged.


rominf (approved):
Reviewed the scrolling geometry, follow-tail behavior, rendering boundaries, and regression coverage. The implementation is consistent with Ratatui's scrollbar geometry, and the focused tests and checks pass.

PR #134, opened 2026-07-21
ci: advisory hint for stale expected-failure markers (EAI-7478)

fredespi:
## Summary

A bug fix whose CI lane doesn't exercise the platform an xfail row is keyed to merges green, leaving the stale `expectations.toml` row to surface **later** as a mis-attributed XPASS on an unrelated PR (e.g. #94's vllm xfails resurfaced on #127's GPU lane). This closes that gap at authoring time with a lightweight advisory nudge — no change to the reconciler or the `heavy` filter.

## Changes

- **`scripts/xfail_expectations_hint.py`** — parses `tests/e2e-cucumber/expectations.toml`, collects the ticket ids a PR references (title, body, commit messages) in both the project tracker shape (`ABC-123`) and GitHub `#issue` shape, and warns via the step summary if any referenced ticket still has a live xfail row. **Advisory only: always exits 0.** Behavior is pinned by an embedded `--self-test` (the single source of truth for the rules).
- **`.github/workflows/ci.yml`** — wires it as a non-blocking, **non-required** `pull_request` job on `ubuntu-latest`, reusing the already-pinned checkout action. PR title/body are passed via `env:` (never interpolated into the shell).
- **`.github/pull_request_template.md`** — one ticket-neutral checkbox reminding authors to clear stale xfail rows for a fixed ticket.

## Test plan

- `python3 scripts/xfail_expectations_hint.py --self-test` — green (7 cases: tracker ref, `Fixes #123`, bare `#123`, no-xfail-row, nothing-tracked, URL-path-not-an-issue-ref, refs-from-commits).
- Verified against the real `expectations.toml`: `--body "Fixes EAI-7052"` reproduces the ticket's concrete case (warns on `serve-default-engine-inference` + `serve-default-engine-working-endpoint`); an unreferenced ticket is silent; exit code is always 0.
- `ruff check` clean; `ci.yml` parses; hawkeye license-header check passes.

Why match bare `#123` too: a bare `#123` sometimes means "PR 123", so this can occasionally nudge on an unrelated number — acceptable because the check is advisory. Favor recall: a stray nudge costs a glance; a missed stale xfail costs a mis-attributed red on an unrelated PR.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,7 @@
+<!--
+Copyright © Advanced Micro Devices, Inc., or its affiliates.
+
+SPDX-License-Identifier: MIT
+-->
+
+- [ ] If this PR fixes a bug, searched `tests/e2e-cucumber/expectations.toml` for the fixed ticket ID and removed/narrowed any now-stale xfail rows.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -176,7 +176,7 @@ jobs:
           python3 scripts/xfail_expectations_hint.py \
             --title "$PR_TITLE" \
             --body "$PR_BODY" \
-            --commits-file /tmp/pr-commits.txt
+            --commits-file /tmp/pr-commits.txt >> "$GITHUB_STEP_SUMMARY"
 
   build-and-test:
     runs-on: ubuntu-latest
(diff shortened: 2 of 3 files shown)


rominf (approved):
Reviewed the matching logic, workflow wiring, Actions input handling, and the CodeQL remediation. The current head removes the flagged file sink and all checks are green. Non-blocking follow-up: scripts/xfail_expectations_hint.py:224 reads expectations.toml without guarding a missing file, so a PR that renames or removes it would make this advisory job exit 1 despite the script's never-fail contract.

PR #136, opened 2026-07-21
test(e2e): run mockable serve chat scenarios on the no-GPU lane (EAI-7484)

fredespi:
## Summary

Move the two mockable serve-chat E2E scenarios off the scarce serial GPU runners onto the GitHub-hosted no-GPU lane, so they gate every PR at no coverage cost. Part of the E2E speedup effort ([EAI-7484]).

The chat scenarios *tool definitions accepted* (#4) and *end-to-end local model* (#5) only assert that a served model answers over HTTP — a tools-bearing request is accepted, and a non-empty reply comes back. Neither checks real generation, so neither needs a GPU.

## Changes

- **Capability-aware serve precondition**: `a model is served in the background` now does a real `rocm serve --managed` on a host with an AMD GPU (real-serve coverage preserved where hardware exists) and uses the in-process `MockServer` + a planted managed-service record on a no-GPU host.
- **Drop `@requires-gpu`** from chat scenarios 4 and 5 so they also run on the mock lane every PR. Real inference stays covered by the `@requires-gpu` `serve-*-inference` scenarios.
- **`a managed runtime is active`** becomes a no-op on a no-GPU host (a managed SDK can only install where a GPU family selects wheels; the mock serve needs no runtime). Only affects the two migrated scenarios.
- **xfail scoping**: the EAI-7423 lemonade-serve xfails for these two ids are scoped to `therock_family = "gfx*"`, so they apply only on a real GPU lemonade host and the scenarios expect-pass on the mock lane instead of XPASS-ing.
- **`serve-absent-gpu-index-rejected` (#13) stays on GPU** (comment updated): on a no-GPU host the GPU-required pre-flight refuses ("no usable AMD GPU") before the `--gpu` index is validated, so its index-specific rejection can only be observed with a real device present.

## Test plan

Linux container gate green:
- workspace + `e2e-cucumber` lib tests (25 passed)
- `cargo xtask e2e` mock lane: **4 xfail / 0 XPASS / 0 unexpected**; chat #4 and #5 now pass on the no-GPU lane.

Real-GPU behaviour (real serve on GPU hosts) is unchanged and will be exercised by the GPU CI lanes.

### DIFF
--- a/tests/e2e-cucumber/expectations.toml
+++ b/tests/e2e-cucumber/expectations.toml
@@ -126,9 +126,11 @@ serve_timeout_secs = 90
 # On a lemonade LINUX host this needs a live lemonade serve, which reaches ready
 # then shuts down immediately (EAI-7423). os=linux: passes on native Windows
 # lemonade (run 29354305288 XPASS'd it), consistent with the other lemonade
-# serve scenarios.
+# serve scenarios. therock_family=gfx* scopes this to a REAL GPU host: the
+# scenario now also runs on the no-GPU mock lane (backed by MockServer, not a
+# real lemonade serve), where EAI-7423 does not apply and it must expect-pass.
 [["chat-tool-definitions-accepted"]]
-when = { effective_engine = "lemonade", os = "linux" }
+when = { effective_engine = "lemonade", os = "linux", therock_family = "gfx*" }
 bug = "EAI-7423"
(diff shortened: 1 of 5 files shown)


rominf (approved):
Reviewed the mock/real lane split, expectation scoping, managed-service setup, and GPU-lane behavior. The mock lane now exercises the intended HTTP contract while preserving real inference coverage, and the focused checks pass.

PR #138, opened 2026-07-22
test(e2e): make XPASS non-fatal for known-flaky xfails

fredespi:
## Summary

The E2E reconciliation gate exits 1 on **any** XPASS. For an intermittent known bug (e.g. EAI-7333 vLLM readiness), a run where the bug doesn't reproduce passes its xfail scenario → counts as XPASS → the required check goes red with **zero real failures**. The `expectations.toml` comments already call these XPASSes "expected and harmless", yet the harness treated them as fatal.

This adds a per-condition `flaky = true` marker so a flaky xfail's XPASS is **reported but non-fatal**. Stale (non-flaky) XPASS and unexpected-FAIL stay fatal, so real regressions and genuinely-fixed bugs are still caught.

## Changes

- **`expectations.toml`**: new optional `flaky = true` key (documented in the grammar header); marked the two EAI-7333 vLLM entries (`serve-vllm-inference`, `serve-readiness-contract`) flaky, citing the runs where they flipped.
- **`src/expectation.rs`**: `flaky` field on `XfailEntry` + `Expectation::ExpectXfail` (`#[serde(default)]`, defaults false). Extracted the exit classification into a pure `reconcile()` → `Reconciliation { is_fatal() }` so the exit-code logic is unit-testable.
- **`tests/e2e.rs`**: exit decision now calls `reconcile()`; flaky XPASS printed as "(flaky, non-fatal)" and excluded from the exit-1 set.

Why not just flip the entries to expect-pass: EAI-7333 is genuinely intermittent (still open), so expect-pass would flip the failure to unexpected-FAIL on the next host/run where the bug resurfaces. The xfail is correct; only its XPASS should be non-fatal.

## Test plan

- [x] `cargo test -p e2e-cucumber --lib` — 18 passed (5 new: flaky parse/default, flaky-XPASS non-fatal, non-flaky-XPASS fatal, unexpected-FAIL always fatal, all-expected clean)
- [x] clippy clean
- [x] Full Linux container gate (clippy + workspace tests + mock E2E suite) green; reconciliation line printed `4 xfail, 0 XPASS (0 flaky, non-fatal), 0 unexpected failure(s)`

### DIFF
--- a/tests/e2e-cucumber/expectations.toml
+++ b/tests/e2e-cucumber/expectations.toml
@@ -20,6 +20,11 @@
 #   bug = "EAI-NNNN"                  ticket reference (shown in the report)
 #   reason = "..."                    human explanation (shown in the report)
 #   serve_timeout_secs = 90          optional: shorter serve wait for a known bug
+#   flaky = true                     optional: the bug is intermittent, so an XPASS
+#                                     (lucky run where it didn't reproduce) is
+#                                     reported but NON-FATAL. Unexpected-FAIL stays
+#                                     fatal. Use only for genuinely intermittent
+#                                     bugs, never to mask a fixed/stale entry.
 #
 # An empty `when = {}` means "always xfail" (platform-independent bug).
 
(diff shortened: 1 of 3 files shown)


fredespi:
Superseded by #142 (merged 2026-07-27), which landed an equivalent flaky-XPASS implementation: same `flaky` field on the xfail matrix, the same two EAI-7333 entries (`serve-vllm-inference`, `serve-readiness-contract`) marked flaky, and the same non-fatal-XPASS exit gate. This PR is now fully redundant. Closing rather than rebasing to a no-op.

PR #139, opened 2026-07-22
refactor(xtask): unify release packaging and install lifecycle E2E

rominf:
## Summary

Consolidates the release packaging and install-lifecycle verification onto a single cross-platform Rust harness, replacing four platform-specific shell/PowerShell scripts.

- **`cargo xtask package`** — a new pure-Rust command that builds the release distribution bundle (a `.tar.gz` on Unix, a `.zip` on Windows), each with a SHA-256 sidecar and, when a signing key is configured, a detached signature. It packages both the `rocm` and `rocmd` binaries and preserves the external artifact contract (bundle layout, checksum syntax, signature presence, installer compatibility) of the scripts it replaces.
- **Shared `paths` module** — extracts the workspace-path resolution (workspace root, active target dir, release binary dir, platform binary name) that had drifted across `e2e.rs`, `demos.rs`, and the old packaging scripts into one module with a unit-testable pure core.
- **`@lifecycle` E2E set** — replaces the standalone `acceptance-install-upgrade-tui-uninstall` scripts with an opt-in scenario set in the cucumber E2E suite. The scenarios drive `cargo xtask package`, run the real `install.sh` / `install.ps1`, and assert the install / upgrade / TUI / uninstall contract: checksum and detached-signature verification (including rejection of bad checksums, bad or missing signatures, and required-signature mode without a public key), first-install PATH setup, isolated-config smoke checks, and clean uninstall. All side effects are rooted under a per-scenario temp dir and mutated OS state is restored on teardown even if a step panics.
- **CI wiring** — points the CI, nightly, and release workflows at `cargo xtask package` and the `@lifecycle` set instead of the removed scripts, and widens CI change-detection so the heavy job triggers on `xtask/` and the lifecycle sources.

## Why

Packaging and install-acceptance logic was duplicated across separate Linux (shell) and Windows (PowerShell) scripts that had drifted apart and duplicated path-resolution logic already present in the xtask crate. A single cross-platform harness removes that drift, makes the packaging and lifecycle contract unit- and scenario-testable, and keeps the two platforms behaviorally in step.

## Design notes

- Signing inputs are read from the environment for parity with the scripts they replace, so existing release configuration keeps working.
- The `@lifecycle` set is gated behind `E2E_INCLUDE_LIFECYCLE` / `E2E_ONLY_LIFECYCLE` so the default `cargo xtask e2e` run stays fast; CI opts in explicitly.
- The `paths` module splits a pure, unit-testable core from filesystem lookups so the resolution logic can be tested without a built workspace.

## Test plan

Verified on Linux:

- `cargo test -p xtask` — 62 passed, including the new `package` tests (bundle layout, two-space lowercase-hex checksum syntax, stage-checksum-sign round trip, required-signature-without-key failure) and the `paths` unit tests.
- `cargo test -p e2e-cucumber --no-run` — the lifecycle steps and installer fixture compile cleanly.
- Run the Linux lifecycle scenarios locally with:
  `E2E_INCLUDE_LIFECYCLE=1 E2E_ONLY_LIFECYCLE=1 cargo xtask e2e`

## Deferred / not yet verified

- **Windows CI checks not verified locally.** The Windows lifecycle job (packaging + native-crypto signature verify + user-PATH persistence/restore + loopback-HTTP install + isolated smoke + uninstall) is wired into the workflow but was only exercised on Linux in this branch; it can only be validated on the GitHub Windows runner. This is why the PR is opened as a draft.

## Risk

Medium. Touches release/nightly/CI packaging paths and removes the previous acceptance scripts; the archive/layout contract is preserved and covered by unit tests, but the Windows path needs the CI runner to confirm.


juhovainio (approved):
lgtm!

PR #142, opened 2026-07-23
ci: stabilize GPU E2E and merge queue

rominf:
## Summary

- Make the real-GPU serve relaunch an actual retry: stop the stalled managed service — the whole engine process tree — before relaunching, so the second serve no longer competes with the first for device memory. Spend that relaunch only where it changes a result: a scenario the expectation matrix already declares a known bug keeps its shortened `serve_timeout_secs` and fails on the first attempt.
- Quote the stalled service's log tail and the device's free-VRAM state into the failure, so a serve that never publishes its model is diagnosable from CI output alone.
- Add first-class flaky-xfail semantics across the harness, `platform.json`, and consolidated report. Intermittent known bugs may reproduce or XPASS without reddening the lane, while deterministic XPASS remains fatal as stale metadata.
- Remove stale MI300X EAI-7052 expectations and mark the documented EAI-7333 readiness/inference races flaky.
- Add a checked-in advanced CodeQL workflow with `merge_group` support, replacing default setup so required CodeQL checks are produced in the merge queue.
- Increase the black-box dashboard wait budget from 20 to 30 seconds after the metrics scenario missed a healthy-but-slow frame by the old deadline.

## Root cause

PR and merge-group runs exposed three independent CI issues:

1. A managed vLLM launch can return successfully with status `starting` and never publish the model before the scenario deadline. The relaunch first added here could not rescue it: a vLLM still in engine init has not bound the serve port yet, so the port kill preceding each attempt cannot see it, and the relaunch therefore started while the first attempt still held its ~0.8-of-device memory reservation — dying on vLLM's own free-memory check. Both attempts were spent on a device the first attempt still owned; the exhausted two-minute VRAM-drain deadline before the second attempt is the signature. This failed `chat-end-to-end-local-model` on 3 of the last 17 MI300X runs, most recently dequeuing this PR from the merge queue.
2. The expectation matrix treated documented intermittent EAI-7333 passes as stale XPASS, repeatedly reddening otherwise healthy MI300X runs.
3. Branch protection required CodeQL contexts that GitHub default setup did not emit for `merge_group`; the queue timed out twice despite a green repository CI workflow.

Why a first serve attempt stalls at all remains unknown: the engine writes its reason to a service log in a per-scenario temp directory that no artifact captures, which is what the new log-tail reporting exists to surface.

The recovered dedicated Strix Halo Windows runner is used unchanged. The temporary organization-runner routing, wrapper-cleanup disablement, and Windows Lemonade EAI-7455 xfails have been removed.

## Risk and scope

- Risk: **medium**. Changes are limited to CI workflow triggers and E2E orchestration/reporting; production CLI behavior is unchanged.
- CodeQL remains enabled for actions, Python, and Rust through the checked-in advanced workflow.
- The permanent Windows Lemonade backend-download work remains outside this PR.

## Test plan

- [x] `cargo test -p e2e-cucumber --all-targets` — 38 passed
- [x] `cargo test -p e2e-report` — 34 passed
- [x] Advanced CodeQL passed for actions, Python, and Rust
- [x] Focused MI300X Qwen3.5 dispatch passed
- [x] Merge-group run emits and passes all required CodeQL checks (run 30255191412; that run was then dequeued by the GPU lane failure this PR fixes)
- [x] Dedicated Strix Halo Windows lane passes after runner recovery
- [x] MI300X lane green with the serve fix (run 30261532148): 0 unexpected failures, `chat-end-to-end-local-model` passes, and `chat-tool-definitions-accepted` now reaches its assertion and fails on the real tool-choice error instead of a serve timeout. The lane also drops from 25m09s to 8m22s, since a known-bug serve no longer pays for a second cold start.

The GPU serve path cannot be exercised without an AMD GPU, so CI is the authority for it; everything else above was run locally.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,37 @@
+name: CodeQL
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  merge_group:
+    branches: [main]
+    types: [checks_requested]
+  schedule:
+    - cron: "17 5 * * 1"
+
+permissions:
+  contents: read
+
+jobs:
+  analyze:
+    name: Analyze (${{ matrix.language }})
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+      packages: read
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [actions, python, rust]
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - uses: github/codeql-action/init@e4fba868fa4b1b91e1fdab776edc8cfbe6e9fb81 # v4.37.3
+        with:
+          languages: ${{ matrix.language }}
+
+      - uses: github/codeql-action/autobuild@e4fba868fa4b1b91e1fdab776edc8cfbe6e9fb81 # v4.37.3
+
+      - uses: github/codeql-action/analyze@e4fba868fa4b1b91e1fdab776edc8cfbe6e9fb81 # v4.37.3
--- a/crates/e2e-report/src/lib.rs
+++ b/crates/e2e-report/src/lib.rs
@@ -617,6 +617,9 @@ struct ManifestExpectation {
     bug: Option<String>,
     #[serde(default)]
     reason: Option<String>,
+    /// Non-deterministic known bug: an XPASS is expected and non-fatal.
+    #[serde(default)]
+    flaky: bool,
 }
 
 /// Read a platform's `platform.json` (sibling of `report.json`). Missing =
@@ -639,8 +642,10 @@ enum CellOutcome {
     Skip,
     /// Expected pass, but FAILED — a real regression.
     UnexpectedFail,
-    /// Expected xfail, but PASSED — stale entry (bug fixed here?).
+    /// Expected deterministic xfail, but PASSED — stale entry (bug fixed here?).
     Xpass,
+    /// Flaky xfail passed this run — expected for a non-deterministic known bug.
+    FlakyXpass,
     /// Expected skip, yet a result exists — harness/resolver disagreement.
     RanWhenNa,
     /// Expected to run (pass or xfail) but NO result was recorded — the scenario
@@ -655,12 +660,13 @@ impl CellOutcome {
     /// Reconcile one scenario's expectation against its actual result.
     /// `actual` is `Some(passed)` when the scenario ran, `None` when it did not
     /// appear in `report.json` (filtered out / skipped).
-    fn reconcile(expected: &str, actual: Option<bool>) -> Self {
+    fn reconcile(expected: &str, flaky: bool, actual: Option<bool>) -> Self {
         match (expected, actual) {
             ("pass", Some(true)) => Self::Pass,
             ("pass", Some(false)) => Self::UnexpectedFail,
             ("pass", None) => Self::Absent,
             ("xfail", Some(false)) => Self::Xfail,
+            ("xfail", Some(true)) if flaky => Self::FlakyXpass,
             ("xfail", Some(true)) => Self::Xpass,
             ("xfail", None) => Self::Absent,
             ("skip", None) => Self::Skip,
@@ -686,6 +692,7 @@ impl CellOutcome {
             Self::Skip => "n/a",
             Self::UnexpectedFail => "❌FAIL",
             Self::Xpass => "⚠️XPASS",
+            Self::FlakyXpass => "✅XPASS (flaky)",
             Self::RanWhenNa => "⚠️n/a-ran",
             Self::Absent => "⚠️no-result",
             Self::Missing => "·",
@@ -765,7 +772,8 @@ impl Grid {
                 if !ids.contains(&exp.id) {
                     ids.push(exp.id.clone());
                 }
-                let outcome = CellOutcome::reconcile(&exp.expected, actual.get(&exp.id).copied());
+                let outcome =
+                    CellOutcome::reconcile(&exp.expected, exp.flaky, actual.get(&exp.id).copied());
                 // A real result supersedes a defensive Missing on merge.
                 columns[col_idx]
                     .outcomes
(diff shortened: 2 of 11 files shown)


EugeneKSiloAi (approved):

clydehoang1 (approved):

volen-silo (dismissed):
Two substantive issues, plus small cleanup. The diagnosis and the shape of the fix are right; my concerns are about the retry silently degrading and one weakened assertion.

Verified locally: `cargo test -p e2e-report` (34 passed), `cargo clippy -p e2e-report -p e2e-cucumber --all-targets` clean. CodeQL actions are SHA-pinned with version comments per AGENTS.md §6, consistent with every other workflow.

## 1. The retry's load-bearing step can silently no-op

`tests/e2e-cucumber/tests/e2e/serving_steps.rs:417` → `tests/e2e-cucumber/tests/e2e.rs:378-410`

The whole premise is "stop the stalled service before relaunching, or the retry is just a second serve competing with the first". But `stop_managed_services` swallows every failure: `read_dir` error → `return`; read/parse/missing-`service_id` → `continue`; and the stop itself is `let _ = cmd.status();` with both streams to `Stdio::null()`.

That was fine as `Drop`-time cleanup. As the retry's load-bearing step it means: if the service record isn't on disk yet, or `rocm services stop` fails or can't escalate past the grace period, attempt 2 starts against a device attempt 1 still owns — exactly the bug being fixed — and nothing in the output says so. The failure then looks identical to a genuinely broken serve, which is the thing the log-tail work exists to prevent.

Suggestion: have `stop_scenario_services` return a status line (records found, per-service rc/stderr from `cmd.output()`) and push it into `diagnostics` alongside `device_state`.

## 2. `model_id.rs` — the wrong-organization guard is now disabled for any path-shaped id

`is_absolute_path` includes `slash_count > 1`, and that branch sets `org = None`. Since `model_ids_match` only compares orgs when *both* sides are `Some`, any id with two or more slashes skips the check entirely:

```
response: ".../models--another-owner--Qwen3-0.6B-GGUF/snapshots/x/Qwen3-0.6B-Q4_0.gguf"
expected: "unsloth/Qwen3-0.6B-GGUF"
=> model_ids_match == true
```

The organization is right there in the `models--<org>--<model>` segment and is discarded. This also contradicts the module doc at line 9 ("organizations must match when both ids include one"), and leaves `rejects_same_model_from_different_organization` unable to cover the path case.

Suggestion: recover the org from the `models--<org>--` segment, and add the wrong-org-in-path negative test.

## 3. Stale comments introduced by this change

`tests/e2e-cucumber/tests/e2e/tui_driver.rs:70` and `:222` both still say "a 20s timeout"; the constant is now 30.

## Non-blocking

- `src/serve_log.rs:43` — `read_to_string` fails whole-hog on invalid UTF-8. vLLM logs carry progress bars and ANSI, so a single bad byte turns the tail into `<failed to read ...: stream did not contain valid UTF-8>` in precisely the failure this exists for. `fs::read` + `from_utf8_lossy` is a one-liner.
- `crates/e2e-report/src/lib.rs:2303` — `C::reconcile("xfail", None)` was *changed* to `("xfail", true, None)` rather than adding the flaky case, so the non-flaky xfail/no-result path is no longer asserted.
- Worst-case serve budget doubles: 2 × (≤120s VRAM drain + 300s readiness) ≈ 14 min per GPU-serve scenario, times two such scenarios, against the Strix Halo Ubuntu lane's 35-minute job cap (currently 16m38s). Headroom exists, but a double stall now hits the *job* timeout, which kills the run with no report — losing exactly the diagnostics this PR adds. Worth a cap on total retry wall-clock.
- `assert!(rc == 0)` at `serving_steps.rs:392` exits before the retry, so a launch that fails on vLLM's own free-memory check gets no second attempt and no service-log tail — arguably the case that most wants one.

## Tradeoff worth naming

`flaky = true` makes a scenario effectively unasserted for as long as the entry lives, and nothing will ever prompt its removal (`xfail_expectations_hint.py` only fires when a PR references the ticket). It is the right tool for a genuine intermittent hardware race — just noting the entries are now permanently invisible.

## Nice

The `serve_log.rs` error-path design (every unreadable case explains itself rather than failing) with a test per case, and the `device_state` string threading a "why" into the failure instead of a bare timeout.

rominf:
Thanks — all three blocking points and both non-blocking code items are fixed, and both design questions are implemented too.

**1. The retry's stop can silently no-op** — `5c0e42c`. `stop_managed_services` now reports, per service, what it found and how the stop went: a missing/unreadable/unparsable record, or `rc` plus the last stderr line from `cmd.output()`, or a stop that couldn't be spawned. The retry appends that as a `--- STOP ---` block next to `device_state`, so "attempt 2 ran against a device attempt 1 still owned" is stated rather than inferred. Teardown still discards the status.

**2. Wrong-organization guard disabled for path-shaped ids** — `739a92b`. The organization is recovered from the `models--<org>--<model>` segment instead of discarded, so the guard stays live for cache paths and the module doc holds again. Your example is now `rejects_wrong_organization_named_by_a_cache_path` — I confirmed it fails against the previous code and passes after. Also added the matching-org path case, and `models--<model>` with no organization, which stays `None` rather than inventing one.

**3. Stale comments** — `83724a4`. Both now say 30s.

**Non-blocking**

- `serve_log.rs` — `15b3c99`: `fs::read` + `from_utf8_lossy`, with a test that a log ending in a truncated multi-byte sequence still yields its readable tail.
- `e2e-report` reconcile — `a2782c3`: the non-flaky `("xfail", false, None)` case is restored alongside the flaky one.
- **Retry wall-clock** — `5c0e42c`, bounded per *run* rather than per scenario. A per-scenario cap still multiplies by however many scenarios stall, which is the double-stall case that would blow the 35-minute job. The relaunch is now drawn from a run-wide budget (`E2E_SERVE_RELAUNCH_BUDGET`, default 1), so the worst case the retry adds is one cold start (~8 min at the lanes' 300s readiness) for the whole run instead of one per stalling scenario, and a denied relaunch says so in the failure. Default 1 matches the observed failure — a single scenario stalling in a run; a run that stalls twice is worth more as a report than as a job killed before it writes one.
- **`assert!(rc == 0)` before the retry** — `5c0e42c`. A non-zero exit now takes the same diagnostics → stop → retry path, so a launch rejected by vLLM's own free-memory check gets both a second attempt and a service-log tail. Cost worth naming: a genuinely broken invocation is retried too, doubling its time to failure — bounded by the same run budget.

**On flaky permanence** — agreed, and it's why the flag is per-entry rather than a lane-wide switch. What should retire these entries is the EAI-7333 fix landing, not an automated nudge: the hint script can't observe an intermittent bug that has stopped reproducing. I'd rather that stay explicit than have entries silently expire into red.

Local: `cargo test -p e2e-cucumber --all-targets` 42 passed, `cargo test -p e2e-report` 34 passed, `cargo clippy --workspace --all-targets -- -D warnings` clean. The GPU serve path itself is only exercisable on the lanes — CI is running on the pushed commits.

juhovainio (approved):

PR #149, opened 2026-07-29
fix(diagnose): correct fix-9's auto flag and test the rocm-doctor contract

volen-silo:
## What

`skills/rocm-doctor/` is a published Agent Skill, and this repo is its source
of truth. The skill is a thin driver: the probe, the closed 15-mode catalog and
the fixes all ship inside the `rocm` binary, so it is versioned with the binary
and lives beside it. The `amd/skills` catalog federates skills from the repos
that own them, which is how the published copy stays in step.

What the skill owns is a *contract*: which fix-ids exist, which the CLI applies
itself, which machines each is for, what a diagnosis carries, and where a report
goes when nothing matched. Nothing tested that seam. `diagnose.feature`
deliberately asserts only the *shape* of a diagnosis so it stays
host-independent, so a renamed fix-id or a 16th failure mode would merge green
and silently break the published skill.

## Changes

**Contract feature** — `rocm_doctor_skill.feature` (6 scenarios) parses
`reference.md`'s catalog table, its auto-applicable prose, the fields and
thresholds it names, the verdicts it enumerates and the trackers it lists, then
diffs each against what the CLI reports. Query-only: no GPU, no serve, no
download, no mutation, so it runs on the blocking mock lane with no capability
tags. Reading a document as test data is a deliberate, narrow exception to the
suite's black-box rule — nothing is imported from the codebase, and the document
is the thing under test.

**rocm-core unit tests** — exit codes `1`, `4` and `5` were asserted nowhere and
no test reached a runner that writes to disk. The two that looked like coverage
did not: the dry-run test picks `fix-2`, whose Linux runner takes no
`FixOptions` and never prompts, and the e2e preview scenario picks print-only
`fix-1`, which returns before any runner.

**Two CLI corrections** — `check_9_igpu_dgpu_collision` reported
`auto_applicable: false` on Linux while `fix::RECIPES` says `true`, so an agent
was told fix-9 was print-only while `rocm fix` would apply it. And
`route_when_no_match` matched on lemonade / ollama / lm-studio, which
`Examination::probe` can never report — the CLI advertised routing it cannot do.

**Skill corrections** — the skill gated on "is `matched` empty?", which
`DiagnoseReport::has_match` exists to prevent; it promised that every
auto-applicable fix honours `--dry-run` and confirms before mutating, which is
false for `fix-2-unset-override` on Linux (its runner reports and stops); the
installer was labelled "Linux / macOS" when `install.sh` is Linux x86_64 only;
and the nightly-only install claim was stale.

**Licence config** — `skills/rocm-doctor/**` is excluded: `SKILL.md` must open
with YAML frontmatter for the skill loader, which pushes any header out of
hawkeye's detection window, and catalog skills carry no headers.

## Scope

Claims proven elsewhere are not restated. Two earlier scenarios were dropped for
that reason. `skills/**` joins the `heavy` paths filter, since `reference.md` is
now a fixture.

## Verification

- `cargo xtask e2e` — 81 scenarios, 79 pass; the 2 failures are the expected
  WSL2 `diagnose` xfails (0 unexpected)
- `cargo test --workspace --all-targets --exclude e2e-cucumber` — clean
- `cargo clippy --locked --workspace --all-targets` and
  `--locked -p e2e-cucumber --test e2e` — clean
- `hawkeye check` and all `prek` hooks — clean, and verified *positively*:
  stripping `README.md`'s or `rocm-cli-assistant`'s header still exits 1, so the
  new exclusion is not swallowing the check
- Both new assertions verified to fail when they should, not just to pass

## Known coverage gap

The fix-9 `auto_applicable` correction is user-observable — `render_report_text`
adds a `flags:` line when the flag is set, and `Fix` derives `Serialize`, so the
JSON flips too. `AGENTS.md` asks for a Gherkin scenario when that happens, and
there is none. This is a gap I could not close rather than one I overlooked:
`check_9_igpu_dgpu_collision` only fires on an `Examination` carrying an APU
**and** a discrete AMD GPU, the suite drives the real binary against the real
host, no `--from-examination`-style input exists to hand it a synthetic one, and
no CI lane has both. It is covered by a `rocm-core` regression test that runs
`diagnose` for both OS families and asserts every emitted fix agrees with the
catalog.

Not verified locally: scenario 6 rejecting a *matching* symptom. This host is
WSL2, where the catalog is out of scope and never runs, so `has_match` is false
either way. That case bites on the no-GPU Linux lane.

Unrelated pre-existing flake seen while verifying: `dash-gen-tps-*` assert
against a wall-clock validity window and fail intermittently on full runs — on
`main` as well as here, and never in isolation. Not touched by this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,6 +122,11 @@ jobs:
               - '**/*.ps1'
               - '**/*.feature'
               - 'tests/e2e-cucumber/**'
+              # skills/rocm-doctor/reference.md is an e2e FIXTURE: the
+              # rocm_doctor_skill.feature scenarios parse its catalog table and
+              # compare it to `rocm fix`. A skill-only edit must therefore run
+              # the E2E job, which gates on `heavy`.
+              - 'skills/**'
               - 'install*'
               # Pinned-key consistency compares docs/keys/* against the installers,
               # so a canonical-key-only change must trigger the heavy job that runs
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -184,21 +184,27 @@ Required consistency points:
 on two counts, and both change how you edit it:
 
 - **This repo is its source of truth.** The skill is a thin driver over the
-  `rocm` binary, so it is versioned with the binary and lives here. The
-  [`amd/skills`](https://github.com/amd/skills) catalog federates it: a nightly
-  job vendors the folder from `main` and opens a pull request there. So edit it
-  here, and never edit the catalog's copy — the next federation run overwrites
-  it. Federation does not carry the skill's `evals/` folder, which is why the
-  catalog keeps a dataset of its own.
+  `rocm` binary, so it is versioned with the binary and lives here.
+  [`amd/skills`](https://github.com/amd/skills) is still in Phase-1
+  incubation for this skill: it carries no automated federation for
+  `rocm-doctor` yet — `.github/federation.json` there only declares
+  `AMD-AGI/TraceLens` as a source, and this skill instead sits under
+  `staging/rocm-doctor`, outside any job's coverage. Until federation picks it
+  up, the rocm-cli team hand-syncs `staging/rocm-doctor` from this folder
+  whenever it changes materially. So edit it here, and never edit the
+  `amd/skills` copy directly — the next hand-sync overwrites it.
 - **`reference.md` is an e2e fixture.** `tests/e2e-cucumber/features/rocm_doctor_skill.feature`
   parses its closed-catalog table and compares it to what `rocm fix` reports.
-  The catalog is authoritative in `crates/rocm-core/src/fix.rs` — adding,
-  renaming, or re-scoping a failure mode means changing the CLI **first**, then
-  the two docs. That feature is what catches you if you forget.
+  The failure catalog itself is authoritative in `crates/rocm-core/src/fix.rs`
+  (the `RECIPES` list) and `crates/rocm-core/src/diagnose.rs` (each mode's
+  checker and OS scoping) — adding, renaming, or re-scoping a failure mode
+  means changing the CLI **first**, then the two docs. That feature is what
+  catches you if you forget.
 
 The folder is excluded from `licenserc.toml`: `SKILL.md` must open with YAML
-frontmatter for the skill loader, and the catalog's skills carry no headers.
-The licence is stated in `skill-card.md` instead.
+frontmatter for the skill loader, and skills published this way carry no
+license headers of their own. The licence is stated in `skill-card.md`
+instead.
 
 ## 8) Verification Matrix For This Repo
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -168,6 +168,7 @@ When changing assistant-adjacent behavior, keep consistency with:
 
 - `docs/llm-tool-use.md`
 - `skills/rocm-cli-assistant/SKILL.md`
+- `skills/rocm-doctor/SKILL.md` and `skills/rocm-doctor/reference.md`
 
 Required consistency points:
 
@@ -175,6 +176,29 @@ Required consistency points:
 - mutating actions require approval flow
 - avoid invented shell/package-manager commands in assistant behavior paths
 - preserve built-in assistant constraints and no-CPU-fallback policy
+
+### `skills/rocm-doctor/` — published from here, and a test fixture
+
+`skills/rocm-cli-assistant/SKILL.md` is compiled into the binary
+(`include_str!` in `apps/rocm/src/main.rs`). `skills/rocm-doctor/` is different
+on two counts, and both change how you edit it:
+
+- **This repo is its source of truth.** The skill is a thin driver over the
+  `rocm` binary, so it is versioned with the binary and lives here. The
+  [`amd/skills`](https://github.com/amd/skills) catalog federates it: a nightly
+  job vendors the folder from `main` and opens a pull request there. So edit it
+  here, and never edit the catalog's copy — the next federation run overwrites
+  it. Federation does not carry the skill's `evals/` folder, which is why the
+  catalog keeps a dataset of its own.
+- **`reference.md` is an e2e fixture.** `tests/e2e-cucumber/features/rocm_doctor_skill.feature`
+  parses its closed-catalog table and compares it to what `rocm fix` reports.
+  The catalog is authoritative in `crates/rocm-core/src/fix.rs` — adding,
+  renaming, or re-scoping a failure mode means changing the CLI **first**, then
+  the two docs. That feature is what catches you if you forget.
+
+The folder is excluded from `licenserc.toml`: `SKILL.md` must open with YAML
+frontmatter for the skill loader, and the catalog's skills carry no headers.
+The licence is stated in `skill-card.md` instead.
 
 ## 8) Verification Matrix For This Repo
 
--- a/crates/rocm-core/src/diagnose.rs
+++ b/crates/rocm-core/src/diagnose.rs
@@ -2368,6 +2368,73 @@ mod tests {
     /// `Examination::probe` sets in `examine.rs`. It is not derived from that
     /// code, so adding a fifth framework there will not fail this test — see
     /// the note on [`route_when_no_match`] for the three places to edit.
+    /// The gate the rocm-doctor skill tells an agent to read, and the reason it
+    /// is not "is `matched` empty?".
+    ///
+    /// This cannot be asserted from the e2e suite: `diagnose` scores several
+    /// checkers from host state alone, with no symptom keyword involved, so on a
+    /// runner that happens to have (say) `amdgpu` blacklisted the catalog
+    /// explains the host no matter what symptom is passed. Constructing the
+    /// `Examination` is the only way to hold the premise still.
+    #[test]
+    fn sub_threshold_causes_leave_has_match_false_and_route_upstream() {
+        // Missing both groups scores 45 -- worth listing, not enough to
+        // establish. `matched` is NOT empty here, which is the whole point: an
+        // agent gating on emptiness would propose this fix for a host where
+        // nothing was established, and never route the user anywhere.
+        let mut e = linux_base();
+        e.in_render_group = Some(false);
+        e.in_video_group = Some(false);
+        let report = diagnose(&e, "the office printer keeps jamming on page three");
+
+        assert!(
+            !report.matched.is_empty(),
+            "this test is pointless unless something sub-threshold was listed"
+        );
+        assert!(
+            report
+                .matched
+                .iter()
+                .all(|d| d.score < report.min_score_for_match),
+            "expected every cause below {}, got {:?}",
+            report.min_score_for_match,
+            report
+                .matched
+                .iter()
+                .map(|d| (&d.id, d.score))
+                .collect::<Vec<_>>()
+        );
+        assert!(
+            !report.has_match,
+            "nothing cleared the threshold, so has_match must be false -- \
+             skills/rocm-doctor/ tells an agent to route upstream on exactly this"
+        );
+        assert!(
+            report.route_when_no_match.url.starts_with("http"),
+            "the skill's rule is to hand over this tracker when nothing was \
+             established, so it must name somewhere to go: {:?}",
+            report.route_when_no_match
+        );
+    }
+
+    /// The other half: the flag has to discriminate, or asserting it is free.
+    #[test]
+    fn an_established_cause_sets_has_match() {
+        let mut e = linux_base();
+        e.in_render_group = Some(false);
+        e.in_video_group = Some(false);
+        let report = diagnose(&e, "cannot open /dev/kfd: permission denied");
+        assert!(
+            report.has_match,
+            "a symptom that matches the catalog must set has_match: {:?}",
+            report
+                .matched
+                .iter()
+                .map(|d| (&d.id, d.score))
+                .collect::<Vec<_>>()
+        );
+    }
+
     #[test]
     fn routing_targets_cover_every_framework_the_probe_reports() {
         let mut targets = Vec::new();
--- a/crates/rocm-core/src/diagnose.rs
+++ b/crates/rocm-core/src/diagnose.rs
@@ -973,7 +973,14 @@ fn check_9_igpu_dgpu_collision(e: &Examination, symptom: &str) -> Diagnosis {
             fix_id: "fix-9-igpu-dgpu".to_owned(),
             auto_applicable: true,
             verify: "powershell -NoProfile -Command \"$env:HIP_VISIBLE_DEVICES=1; python -c \\\"import torch; print(torch.cuda.device_count())\\\"\"".to_owned(),
-            notes: vec![note],
+            notes: vec![
+                note,
+                "auto-applicable here means `rocm fix fix-9-igpu-dgpu` has a runner \
+                 for it — but that runner only pins HIP_VISIBLE_DEVICES when you pass \
+                 --device-index N. Without it, `rocm fix` just prints the query that \
+                 identifies which index is the discrete GPU."
+                    .to_owned(),
+            ],
             ..Fix::default()
         }
     } else {
@@ -993,7 +1000,14 @@ fn check_9_igpu_dgpu_collision(e: &Examination, symptom: &str) -> Diagnosis {
             // flag to decide whether to offer to run the fix or just print it.
             auto_applicable: true,
             verify: "HIP_VISIBLE_DEVICES=1 python -c \"import torch; print(torch.cuda.device_count())\"".to_owned(),
-            notes: vec![note],
+            notes: vec![
+                note,
+                "auto-applicable here means `rocm fix fix-9-igpu-dgpu` has a runner \
+                 for it — but that runner only pins HIP_VISIBLE_DEVICES when you pass \
+                 --device-index N. Without it, `rocm fix` just prints the query that \
+                 identifies which index is the discrete GPU."
+                    .to_owned(),
+            ],
             ..Fix::default()
         }
     };
--- a/crates/rocm-core/src/diagnose.rs
+++ b/crates/rocm-core/src/diagnose.rs
@@ -106,14 +106,17 @@ impl DiagnoseReport {
 }
 
 /// Upstream tracker for a framework key.
+/// Upstream tracker for a routing target.
+///
+/// Only the targets [`route_when_no_match`] can actually produce are listed.
+/// Arms for lemonade / ollama / lm-studio / amdgpu-install were unreachable —
+/// the host probe never reports those frameworks — so they described a
+/// capability the CLI does not have. Routing an app that merely appears in the
+/// symptom text is the caller's job, not the probe's.
 fn upstream_tracker(target: &str) -> &'static str {
     match target {
         "pytorch" => "https://github.com/pytorch/pytorch/issues  (tag with rocm label)",
         "llama-cpp" => "https://github.com/ggml-org/llama.cpp/issues",
-        "lemonade" => "https://github.com/lemonade-sdk/lemonade/issues",
-        "ollama" => "https://github.com/ollama/ollama/issues",
-        "lm-studio" => "https://lmstudio.ai/docs/app  (use in-app support; no public repo)",
-        "amdgpu-install" => "https://repo.radeon.com  (raise via your AMD support contact)",
         _ => "https://github.com/ROCm/ROCm/issues",
     }
 }
@@ -985,7 +988,11 @@ fn check_9_igpu_dgpu_collision(e: &Examination, symptom: &str) -> Diagnosis {
                 "# Persist in your shell rc or your launch script.".to_owned(),
             ],
             fix_id: "fix-9-igpu-dgpu".to_owned(),
-            auto_applicable: false,
+            // `rocm fix fix-9-igpu-dgpu --device-index N` really does apply this
+            // on Linux (see `fix::run_hip_visible_devices_linux`), so the
+            // diagnosis must not tell an agent otherwise — it branches on this
+            // flag to decide whether to offer to run the fix or just print it.
+            auto_applicable: true,
             verify: "HIP_VISIBLE_DEVICES=1 python -c \"import torch; print(torch.cuda.device_count())\"".to_owned(),
             notes: vec![note],
             ..Fix::default()
@@ -1907,13 +1914,17 @@ fn catalog_covers(e: &Examination) -> bool {
         .any(|(_, applicable)| applicable.contains(&family))
 }
 
+/// Where to send a user when nothing in the catalog matched.
+///
+/// Keyed off the *host-detected* framework, which `Examination::probe` only
+/// ever sets to `pytorch`, `llama-cpp`, `unknown` or `skipped` — so those two
+/// named arms plus the ROCm-core default are the whole reachable set. If a probe
+/// for another framework is added, add its arm here and in [`upstream_tracker`];
+/// `routing_targets_cover_every_framework_the_probe_reports` will flag it.
 fn route_when_no_match(e: &Examination) -> Route {
     let target = match e.framework.as_str() {
         "pytorch" => "pytorch",
         "llama-cpp" => "llama-cpp",
-        "lemonade" => "lemonade",
-        "ollama" => "ollama",
-        "lm-studio" => "lm-studio",
         _ => "rocm-core",
     };
     Route {
@@ -2340,6 +2351,102 @@ mod tests {
         assert_eq!(top.id, "fix-1-arch");
         // 50 (keyword) + 55 (missing arch), clamped to 100.
         assert_eq!(top.score, 100);
+    }
+
+    /// Routing must stay defined for every framework the probe can report, and
+    /// must not claim targets it can never reach.
+    ///
+    /// `Examination::probe` sets `framework` to exactly these four values (see
+    /// `examine.rs`). The catalog docs previously advertised lemonade / ollama /
+    /// lm-studio routing that no probe could ever trigger; this pins the
+    /// reachable set so a re-added arm has to come with a probe that reaches it.
+    #[test]
+    fn routing_targets_cover_every_framework_the_probe_reports() {
+        let mut targets = Vec::new();
+        for framework in ["skipped", "pytorch", "llama-cpp", "unknown"] {
+            let e = Examination {
+                framework: framework.to_owned(),
+                ..Examination::default()
+            };
+            let route = route_when_no_match(&e);
+            assert!(
+                route.url.starts_with("http"),
+                "{framework}: routed to a non-URL {:?}",
+                route.url
+            );
+            targets.push(route.target);
+        }
+        targets.sort_unstable();
+        targets.dedup();
+        assert_eq!(
+            targets,
+            vec!["llama-cpp", "pytorch", "rocm-core"],
+            "the reachable routing targets changed; update the docs in \
+             skills/rocm-doctor/reference.md (and the amd/skills copy) to match"
+        );
+    }
+
+    /// Every diagnosis must agree with the fix catalog about whether the CLI
+    /// can apply the fix itself.
+    ///
+    /// An agent following the rocm-doctor skill branches on `auto_applicable`
+    /// from `diagnose --json` to decide whether to offer to run `rocm fix` or
+    /// merely print the plan, while `fix::apply` dispatches on `RECIPES`. When
+    /// the two disagree the CLI contradicts itself: fix-9 advertised
+    /// `auto_applicable: false` on Linux and then applied itself anyway.
+    ///
+    /// Both OS families are exercised because the checkers build their `Fix`
+    /// per-OS and only one branch is taken per run — the Linux branch was the
+    /// stale one, and a Linux-only test would not have seen it.
+    #[test]
+    fn every_diagnosis_agrees_with_the_fix_catalog_on_auto_applicability() {
+        for os in ["linux", "windows"] {
+            let mut e = Examination {
+                os_family: os.to_owned(),
+                ..Examination::default()
+            };
+            // Trip several checkers at once so this covers more of the catalog
+            // than a single fix.
+            e.in_render_group = Some(false);
+            e.in_video_group = Some(false);
+            e.has_apu = true;
+            e.has_discrete_amd = true;
+            e.gpus = vec![
+                Gpu {
+                    gfx_target: "gfx1103".to_owned(),
+                    is_amd: true,
+                    is_apu: Some(true),
+                    ..Gpu::default()
+                },
+                Gpu {
+                    gfx_target: "gfx1100".to_owned(),
+                    is_amd: true,
+                    is_apu: Some(false),
+                    ..Gpu::default()
+                },
+            ];
+
+            let report = diagnose(&e, "torch crashes with a segfault");
+            assert!(
+                !report.matched.is_empty(),
+                "{os}: expected at least one diagnosis to compare against the catalog"
+            );
+            for d in &report.matched {
+                let fix = d
+                    .fix
+                    .as_ref()
+                    .unwrap_or_else(|| panic!("{os}/{}: matched with no fix", d.id));
+                let catalog = crate::fix::auto_applicable_for(&fix.fix_id).unwrap_or_else(|| {
+                    panic!("{os}/{}: emitted a fix-id not in the catalog", fix.fix_id)
+                });
+                assert_eq!(
+                    fix.auto_applicable, catalog,
+                    "{os}/{}: diagnose reports auto_applicable={}, but the fix catalog \
+                     (what `rocm fix` actually does) says {catalog}",
+                    fix.fix_id, fix.auto_applicable
+                );
+            }
+        }
     }
 
     #[test]
--- a/crates/rocm-core/src/diagnose.rs
+++ b/crates/rocm-core/src/diagnose.rs
@@ -1918,9 +1918,13 @@ fn catalog_covers(e: &Examination) -> bool {
 ///
 /// Keyed off the *host-detected* framework, which `Examination::probe` only
 /// ever sets to `pytorch`, `llama-cpp`, `unknown` or `skipped` — so those two
-/// named arms plus the ROCm-core default are the whole reachable set. If a probe
-/// for another framework is added, add its arm here and in [`upstream_tracker`];
-/// `routing_targets_cover_every_framework_the_probe_reports` will flag it.
+/// named arms plus the ROCm-core default are the whole reachable set.
+///
+/// Adding a framework to `examine.rs`'s probe means adding its arm here, its
+/// tracker in [`upstream_tracker`], and its name to the hand-maintained list in
+/// `routing_targets_cover_every_framework_the_probe_reports`. That test reads
+/// its own list rather than deriving one from `examine.rs`, so it cannot notice
+/// a new framework on its own — all three edits are manual.
 fn route_when_no_match(e: &Examination) -> Route {
     let target = match e.framework.as_str() {
         "pytorch" => "pytorch",
@@ -2356,10 +2360,14 @@ mod tests {
     /// Routing must stay defined for every framework the probe can report, and
     /// must not claim targets it can never reach.
     ///
-    /// `Examination::probe` sets `framework` to exactly these four values (see
-    /// `examine.rs`). The catalog docs previously advertised lemonade / ollama /
-    /// lm-studio routing that no probe could ever trigger; this pins the
-    /// reachable set so a re-added arm has to come with a probe that reaches it.
+    /// The catalog docs previously advertised lemonade / ollama / lm-studio
+    /// routing that no probe could ever trigger; this pins the reachable set so
+    /// a re-added arm has to come with a probe that reaches it.
+    ///
+    /// The framework list below is **hand-maintained** to match what
+    /// `Examination::probe` sets in `examine.rs`. It is not derived from that
+    /// code, so adding a fifth framework there will not fail this test — see
+    /// the note on [`route_when_no_match`] for the three places to edit.
     #[test]
     fn routing_targets_cover_every_framework_the_probe_reports() {
         let mut targets = Vec::new();
--- a/crates/rocm-core/src/diagnose.rs
+++ b/crates/rocm-core/src/diagnose.rs
@@ -105,7 +105,6 @@ impl DiagnoseReport {
     }
 }
 
-/// Upstream tracker for a framework key.
 /// Upstream tracker for a routing target.
 ///
 /// Only the targets [`route_when_no_match`] can actually produce are listed.
--- a/crates/rocm-core/src/fix.rs
+++ b/crates/rocm-core/src/fix.rs
@@ -702,6 +702,18 @@ fn looks_like_a_diagnosis_position(value: &str) -> bool {
     !digits.is_empty() && digits.chars().all(|c| c.is_ascii_digit())
 }
 
+/// Whether the CLI will apply `fix_id` itself, or `None` if it isn't a known
+/// fix. `RECIPES` is the authority: `apply()` dispatches on it, so this is the
+/// value any other surface describing a fix has to agree with.
+///
+/// Test-only: the one production consumer is `apply()`, which reads the recipe
+/// directly. This exists so `diagnose`'s tests can assert the two surfaces
+/// agree without exposing `RECIPES`.
+#[cfg(test)]
+pub(crate) fn auto_applicable_for(fix_id: &str) -> Option<bool> {
+    find_recipe(fix_id).map(|r| r.auto_applicable)
+}
+
 /// List every fix-id (id, kind, OS scope, title).
 #[must_use]
 pub fn list_recipes() -> String {
@@ -813,13 +825,30 @@ pub fn apply(fix_id: &str, opts: &FixOptions) -> i32 {
 // Consent / environment helpers
 // ---------------------------------------------------------------------------
 
-fn confirm(prompt: &str, assume_yes: bool) -> bool {
+/// The half of the consent decision that needs no I/O: `Some(verdict)` when the
+/// answer is already determined, `None` when the user must actually be asked.
+///
+/// Split out from [`confirm`] so the two rules that matter — `--yes` approves,
+/// and a non-interactive shell *refuses* rather than silently proceeding — are
+/// testable without a controlled stdin. A test that drove `confirm` directly
+/// would block on `read_line` whenever the suite happened to run with a terminal
+/// attached.
+const fn consent_without_prompt(assume_yes: bool, stdin_is_terminal: bool) -> Option<bool> {
     if assume_yes {
-        return true;
+        return Some(true);
     }
-    if !std::io::stdin().is_terminal() {
-        fail!("Non-interactive shell and --yes not passed; refusing to apply.");
-        return false;
+    if !stdin_is_terminal {
+        return Some(false);
+    }
+    None
+}
+
+fn confirm(prompt: &str, assume_yes: bool) -> bool {
+    if let Some(verdict) = consent_without_prompt(assume_yes, std::io::stdin().is_terminal()) {
+        if !verdict {
+            fail!("Non-interactive shell and --yes not passed; refusing to apply.");
+        }
+        return verdict;
     }
     print!("{prompt} [y/N]: ");
     let _ = std::io::stdout().flush();
@@ -1184,8 +1213,29 @@ fn run_hip_visible_devices_linux(opts: &FixOptions) -> i32 {
         fail!("Could not determine your home directory.");
         return 3;
     };
+    pin_device_in_rc_file(&rc_file, idx, opts, || {
+        confirm(&format!("Append to {}?", rc_file.display()), opts.yes)
+    })
+}
+
+/// The part of fix-9 that actually touches the user's machine: decide whether
+/// the rc file already pins a device, honour `--dry-run`, ask for consent, and
+/// append.
+///
+/// Taking the rc path and the consent decision as parameters keeps this — the
+/// one auto-fix path on Linux that really mutates a file — testable end to end
+/// against a temp file, with no `$HOME` mutation and no dependency on whether
+/// stdin happens to be a terminal. Exit codes are the contract documented in
+/// `skills/rocm-doctor/reference.md`: 0 applied/dry-run/already-set, 4 write
+/// failed, 5 declined.
+fn pin_device_in_rc_file(
+    rc_file: &Path,
+    idx: i64,
+    opts: &FixOptions,
+    consent: impl FnOnce() -> bool,
+) -> i32 {
     let export_line = format!("export HIP_VISIBLE_DEVICES={idx}");
-    if let Ok(existing) = std::fs::read_to_string(&rc_file)
+    if let Ok(existing) = std::fs::read_to_string(rc_file)
         && existing.contains("HIP_VISIBLE_DEVICES=")
     {
         println!(
@@ -1200,11 +1250,11 @@ fn run_hip_visible_devices_linux(opts: &FixOptions) -> i32 {
         println!("(dry-run; not executed)");
         return 0;
     }
-    if !confirm(&format!("Append to {}?", rc_file.display()), opts.yes) {
+    if !consent() {
         return 5;
     }
     if let Err(exc) = append_line(
-        &rc_file,
+        rc_file,
         "# Added by rocm examine (fix-9-igpu-dgpu)",
         &export_line,
     ) {
@@ -1607,5 +1657,170 @@ mod tests {
         // behaviour that a script could start depending on.
         assert_eq!(apply("#1", &FixOptions::default()), 2);
         assert_eq!(apply("bogus", &FixOptions::default()), 2);
+    }
+
+    // ── Consent and mutation contract ──────────────────────────────
+    //
+    // `skills/rocm-doctor/reference.md` documents six exit codes, and the skill
+    // tells an agent it may run an auto-fix because the CLI "refuses on a
+    // non-interactive shell without --yes" and "confirms before mutating".
+    // Before these tests, codes 1, 4 and 5 were asserted nowhere in the repo and
+    // no test ever reached a runner that writes to disk: the two that look like
+    // they cover it don't — `dry_run_never_mutates_...` picks fix-2, whose Linux
+    // runner takes no FixOptions and never prompts, and diagnose.feature's
+    // preview scenario picks print-only fix-1, which returns before any runner.
+    //
+    // fix-9 is the auto-fix used here because its Linux runner is the one that
+    // genuinely appends to a file, and `pin_device_in_rc_file` lets that run
+    // against a temp path with an injected consent verdict.
+
+    /// A unique scratch path under the workspace's test-artifact dir, following
+    /// the convention in `lib.rs`'s tests (no `tempfile` dev-dependency here).
+    fn scratch_dir(name: &str) -> PathBuf {
+        let dir = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+            .join("..")
+            .join("..")
+            .join(".rocm-work")
+            .join("tests")
+            .join("core")
+            .join(format!(
+                "fix-{name}-{}-{:?}",
+                std::process::id(),
+                std::thread::current().id()
+            ));
+        std::fs::create_dir_all(&dir).expect("failed to create scratch dir");
+        dir
+    }
+
+    fn pinning_opts() -> FixOptions {
+        FixOptions {
+            device_index: Some(1),
+            ..FixOptions::default()
+        }
+    }
+
+    #[test]
+    fn yes_flag_approves_without_prompting() {
+        assert_eq!(consent_without_prompt(true, false), Some(true));
+        assert_eq!(consent_without_prompt(true, true), Some(true));
+    }
+
+    #[test]
+    fn non_interactive_shell_refuses_instead_of_proceeding() {
+        // The rule the skill relies on: without --yes and with nothing to prompt,
+        // the answer is a definite NO, never an implicit yes.
+        assert_eq!(consent_without_prompt(false, false), Some(false));
+    }
+
+    #[test]
+    fn interactive_shell_without_yes_must_actually_ask() {
+        assert_eq!(consent_without_prompt(false, true), None);
+    }
+
+    #[test]
+    fn declining_consent_returns_5_and_writes_nothing() {
+        let rc = scratch_dir("declined").join(".bashrc");
+        std::fs::write(&rc, "# existing\n").expect("seed rc file");
+
+        let code = pin_device_in_rc_file(&rc, 1, &pinning_opts(), || false);
+
+        assert_eq!(code, 5, "a declined fix must exit 5");
+        assert_eq!(
+            std::fs::read_to_string(&rc).expect("read rc"),
+            "# existing\n",
+            "a declined fix must not touch the file"
+        );
+        std::fs::remove_dir_all(rc.parent().expect("parent")).ok();
+    }
+
+    #[test]
+    fn dry_run_returns_0_and_writes_nothing_even_with_consent() {
+        let rc = scratch_dir("dryrun").join(".bashrc");
+        std::fs::write(&rc, "# existing\n").expect("seed rc file");
+        let opts = FixOptions {
+            dry_run: true,
+            ..pinning_opts()
+        };
+
+        // Consent would be granted; --dry-run must short-circuit before it.
+        let code = pin_device_in_rc_file(&rc, 1, &opts, || {
+            panic!("dry-run must not reach the consent gate")
+        });
+
+        assert_eq!(code, 0);
+        assert_eq!(
+            std::fs::read_to_string(&rc).expect("read rc"),
+            "# existing\n",
+            "a dry-run must not touch the file"
+        );
+        std::fs::remove_dir_all(rc.parent().expect("parent")).ok();
+    }
+
+    #[test]
+    fn granting_consent_appends_the_export_and_returns_0() {
+        let rc = scratch_dir("granted").join(".bashrc");
+        std::fs::write(&rc, "# existing\n").expect("seed rc file");
+
+        let code = pin_device_in_rc_file(&rc, 3, &pinning_opts(), || true);
+
+        assert_eq!(code, 0);
+        let body = std::fs::read_to_string(&rc).expect("read rc");
+        assert!(
+            body.starts_with("# existing\n"),
+            "existing rc content must be preserved:\n{body}"
+        );
+        assert!(
+            body.contains("export HIP_VISIBLE_DEVICES=3"),
+            "expected the export line to be appended:\n{body}"
+        );
+        std::fs::remove_dir_all(rc.parent().expect("parent")).ok();
+    }
+
+    #[test]
+    fn an_rc_file_that_already_pins_a_device_is_left_alone() {
+        let rc = scratch_dir("already-pinned").join(".bashrc");
+        let original = "export HIP_VISIBLE_DEVICES=0\n";
+        std::fs::write(&rc, original).expect("seed rc file");
+
+        let code = pin_device_in_rc_file(&rc, 1, &pinning_opts(), || {
+            panic!("an already-pinned rc file must not reach the consent gate")
+        });
+
+        assert_eq!(code, 0, "already-pinned is success, not failure");
+        assert_eq!(
+            std::fs::read_to_string(&rc).expect("read rc"),
+            original,
+            "must not append a second, conflicting pin"
+        );
+        std::fs::remove_dir_all(rc.parent().expect("parent")).ok();
+    }
+
+    #[test]
+    fn a_failed_write_returns_4_rather_than_reporting_success() {
+        // A directory where the rc file should be: the append fails at open().
+        // This is the only way the documented "attempted but the command
+        // failed" code is reachable for this fix.
+        let dir = scratch_dir("write-fails");
+        let rc = dir.join(".bashrc");
+        std::fs::create_dir_all(&rc).expect("create dir in place of rc file");
+
+        let code = pin_device_in_rc_file(&rc, 1, &pinning_opts(), || true);
+
+        assert_eq!(code, 4, "a failed write must exit 4, not 0");
+        std::fs::remove_dir_all(&dir).ok();
+    }
+
+    #[test]
+    fn exit_code_1_is_unreachable_by_construction() {
+        // apply() returns 1 only for an auto_applicable recipe with no runner.
+        // `auto_applicable_recipes_have_a_runner` keeps that impossible; assert
+        // the reachability argument here so reference.md's row 1 is accounted
+        // for rather than silently untested.
+        assert!(
+            RECIPES
+                .iter()
+                .all(|r| !r.auto_applicable || r.runner.is_some()),
+            "an auto-applicable recipe without a runner would make exit 1 reachable"
+        );
     }
 }
--- a/crates/rocm-core/src/fix.rs
+++ b/crates/rocm-core/src/fix.rs
@@ -986,15 +986,26 @@ fn run_unset_override_linux() -> i32 {
     let Some(home) = home_dir() else {
         return 0;
     };
-    let candidates = [
+    report_persistent_override(&[
         home.join(".bashrc"),
         home.join(".bash_profile"),
         home.join(".zshrc"),
         home.join(".profile"),
         home.join(".config").join("fish").join("config.fish"),
-    ];
-    let hits: Vec<PathBuf> = candidates
-        .into_iter()
+    ])
+}
+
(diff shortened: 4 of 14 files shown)


rominf (commented):
Reviewed at `987dd9a2`. Checked out locally: `fix::tests` 18/18 pass, all 22 CI checks green, and I re-verified the description's load-bearing claims (exit codes 1/4/5 were indeed unasserted before; the 15-mode catalog, OS scopes and auto-flags in `reference.md` match `RECIPES` exactly).

The refactor is the good part — splitting `consent_without_prompt` out of `confirm` and taking the rc path plus an injected consent verdict in `pin_device_in_rc_file` makes the mutation path genuinely testable without touching `$HOME` or stdin, and it's behaviour-preserving. The comments explaining *why* the two tests that looked like coverage weren't are worth keeping.

Three things I'd want resolved before this merges, then some smaller ones.

### Dependent on an unmerged upstream PR, but not draft

The premise is that `skills/rocm-doctor/` is byte-verbatim with amd/skills#87 — but #87 is still open and can still change in review. Merging first can land a mirror that never matches what upstream finally accepts, and the drift job fails open so nothing will say so. Suggest draft until #87 lands, then mirror the merged files.

That also sequences the next point for free.

### The mirror ships documentation that's already known to be wrong

You flag it in the PR body, so this is agreement rather than a new find — but it interacts badly with the framing. The point of the change is that the documented contract becomes executable and can't silently break; a reader will reasonably assume the whole file is under test. It isn't: the new scenarios cover the catalog, and `SKILL.md`'s install instructions are untested and currently false (`install.sh:310-319` is Linux x86_64 only and hard-fails otherwise, against "Linux / macOS"; the nightly-only claim is likewise stale).

Fixing it upstream is the right layer. Landing #87 with the correction first and mirroring the corrected file avoids importing a known-wrong doc.

### Scenario 2's pinned set duplicates an existing unit test, and its rationale doesn't hold

`skill_steps.rs:264-266` justifies the hard-coded `AUTO_APPLICABLE` with "both files name these four ids in prose". `SKILL.md` contains no fix-ids at all — `grep 'fix-[0-9]' skills/rocm-doctor/SKILL.md` is empty. Only `reference.md` names them (line 43), and the assertion doesn't read that prose either: it compares the CLI to a Rust array, which is exactly what `exactly_the_four_known_fixes_are_auto` (`fix.rs:1018`) already asserts, same ids, same order.

So it's a duplicate by the PR's own "claims already proven elsewhere are not restated" standard. Either drop it, or make it do the thing the comment describes and parse the prose line out of `reference.md`.

### Smaller

- `world.model_name` is reused to carry a symptom string (`skill_steps.rs:184`) and a fix-id (`:196`). The PR already adds a dedicated `skill_reference` field; these deserve the same rather than borrowing an unrelated name.
- `pinning_without_a_device_index_is_not_applicable` returns early on non-Linux, so it's a vacuous green on the Windows lane. `#[cfg(target_os = "linux")]` states that instead of passing silently.
- Both parsers `panic!` on an unrecognised row (`skill_steps.rs:96`, `:123`). Loud failure is defensible, but it means a cosmetic change to `rocm fix` output surfaces as a panic rather than an assertion diff naming the two sets.
- Scratch-dir cleanup is `remove_dir_all(...).ok()` after the assertions, so a failing test leaves debris under `.rocm-work/`. Minor, and the no-`tempfile` convention is a fair reason.
- `!skills/rocm-cli-assistant/*.md` is non-recursive, and a new top-level directory with tracked markdown goes unchecked silently. AGENTS.md documents the hazard, which is the right mitigation — worth being sure that note is the first thing anyone editing `licenserc.toml` sees.
- The drift job's `if: github.event_name == 'pull_request'` skips the merge queue. Harmless while it's advisory, but it means the last state before merge is never compared.

On the licence change specifically: I checked it positively rather than trusting the green check, since a passing `hawkeye check` only proves nothing is failing. Stripping README.md's header makes hawkeye flag it, so `!/*.md` anchoring works and the mirror is genuinely exempt — and no currently-tracked markdown lost coverage in the blanket-to-enumeration swap.

licenserc.toml
```diff
@@ -2,13 +2,18 @@
...
-## copyright header.
+## Note on *.md: hawkeye excludes markdown by default, so each scope we DO want
+## header-checked is force-included with a `!` negation below. A negation cannot
+## be narrowed afterwards — a positive exclude never wins against it, and a
+## pattern with no slash (`!*.md`) matches at every depth — so the scopes are
+## enumerated and anchored instead of using one blanket `!**/*.md`. Adding a new
+## top-level directory that holds tracked markdown means adding it to that list;
+## otherwise its headers go unchecked.
+##
+## `skills/rocm-doctor/` is deliberately absent: it is a byte-verbatim mirror of
```
rominf: **This reverts main's HawkEye v7 migration, and after rebase the mechanism this comment describes won't exist.**

`main` merged #300 (`9979df89`, "migrate HawkEye to v7"), which deleted `licenserc-md-header.toml` and moved to the v7 schema (`[header] text`, `[files]`, `[[rules]] style_out = "xml"`). This file is still v6 (`inlineHeader`, `additionalHeaders`, `[mapping.MARKDOWN_STYLE]`) and resurrects the deleted file.

`git merge-tree origin/main pr/149` reports exactly one conflict, and it's this file — everything else auto-merges. CI pins `HAWKEYE_VERSION: v7.0.0`, so taking the PR side breaks the blocking `License header check` job.

The subtler problem is that the *purpose* stops being solvable the way this comment describes. Main's v7 config has no `!**/*.md` negation at all — it just lists `**/*.md` under `[files] includes` and header-checks every tracked markdown file. So after rebase there's no "enumerated scopes" list to omit `skills/rocm-doctor` from, and the mirrored files would be *required* to carry AMD headers — which is precisely what this exemption exists to prevent, and what the frontmatter constraint makes impossible.

Fix: rebase and re-express it in v7 terms — add `"skills/rocm-doctor/**"` to `[files] excludes` — then drop the resurrected `licenserc-md-header.toml` and this comment block. (The revert also flips the docs URL back from `fast/hawkeye` to `korandoru/hawkeye`.)

.github/workflows/ci.yml
```diff
@@ -174,6 +179,61 @@ jobs:
...
+  # a required check.
+  skill-mirror-drift:
+    name: rocm-doctor mirror drift (advisory)
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request'
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+      - name: Compare skills/rocm-doctor against amd/skills
+        env:
+          UPSTREAM: https://raw.githubusercontent.com/amd/skills/main/skills/rocm-doctor
```
rominf: **This URL 404s, so the drift job can never fire.**

Checked against the live repo: `amd/skills` `contents/skills` holds eight skills and none is `rocm-doctor`. The files are at **`staging/rocm-doctor/`**:

```
GET repos/amd/skills/contents/staging/rocm-doctor  → 200 (SKILL.md, reference.md, skill-card.md, evals)
GET repos/amd/skills/contents/skills/rocm-doctor   → 404
```

So every fetch fails, the job takes the `missing` branch, prints "not published yet", and passes green. The one check whose entire purpose is catching mirror rot is structurally incapable of firing — and because it fails open by design, nobody will ever notice it isn't working. (It's currently masking finding on the SKILL.md divergence, which it would otherwise have caught.)

Two fixes: point `UPSTREAM` at `.../amd/skills/main/staging/rocm-doctor`, and make the job distinguish "all three files 404" — which is suspicious and should be reported loudly in the summary — from a transient single-file fetch error. Fail-open is right for a flaky network; it's wrong for "the path is gone".

skills/rocm-doctor/SKILL.md
```diff
@@ -0,0 +1,139 @@
...
+
+Given a "ROCm / PyTorch / llama.cpp isn't working on my AMD GPU" complaint,
+identify which **known misconfiguration** is the cause and either fix it (with
+consent) or hand back the exact next step.
+
+This skill does **not** probe or reason on its own. The `rocm` CLI owns the
+probe, the closed failure-mode catalog, and the fixes; the skill just drives it
+and relays the results. The catalog is a **closed list** — if the symptom
+doesn't match a known mode, route the user upstream instead of guessing.
+
```
rominf, line 29: **This file is no longer a byte-verbatim mirror — and what's missing is a safety instruction.**

I diffed all three files against `amd/skills@main:staging/rocm-doctor/`. `reference.md` and `skill-card.md` are identical. `SKILL.md` is not. Upstream has, and this copy lacks:

- an entire **`## Scope gate — check before anything else`** section (~35 lines): read the symptom and answer "is this an AMD GPU on native Linux or Windows?" first; if not, decline, give **no** troubleshooting — "no commands to run, no driver or CUDA advice, no diagnostic checklist… not even generic GPU suggestions" — and do not run `rocm examine`/`diagnose`/`fix`
- a **`## Prerequisites`** section
- an extra `## Rules` bullet: "Never run the workflow — or offer *any* troubleshooting, generic GPU fixes included — for a non-AMD GPU or a WSL2 setup"
- the workflow's "Only start here once the Scope gate passes" preamble
- a markdown link `[reference.md](reference.md)` where this copy has bare backticks

The Scope gate is a substantive behavioral instruction, not formatting — an agent driving the mirrored copy will happily start diagnosing an NVIDIA box.

Since the "byte-verbatim mirror" premise is restated in `AGENTS.md:180-190`, `licenserc.toml:13-16`, `tests/e2e-cucumber/README.md:218-222`, and this feature's own header comment, it needs to be true. Re-mirror all three from `staging/rocm-doctor/`.

crates/rocm-core/src/fix.rs
```diff
@@ -1197,4 +1247,180 @@ mod tests {
...
+        std::fs::create_dir_all(&rc).expect("create dir in place of rc file");
+
+        let code = pin_device_in_rc_file(&rc, 1, &pinning_opts(), || true);
+
+        assert_eq!(code, 4, "a failed write must exit 4, not 0");
+        std::fs::remove_dir_all(&dir).ok();
+    }
+
+    #[test]
+    fn pinning_without_a_device_index_is_not_applicable() {
```
rominf: **This asserts 3 where `main` deliberately returns 0 — and the files auto-merge, so you get both tests.**

`main` changed the missing-`--device-index` branch to `return 0` (`crates/rocm-core/src/fix.rs:864-871`):

```rust
let Some(idx) = opts.device_index else {
    // Print-only path: without an index the fix only prints the query that
    // helps the user identify N, so it succeeds like any other preview.
    println!("Run `rocminfo | grep -E 'Agent |Marketing|gfx'` …");
    return 0;
};
```

guarded by its own regression test `fix_9_without_device_index_is_print_only_and_returns_zero`, whose comment says explicitly "must return 0 -- not the environment/OS code 3."

The nasty part: `git merge-tree` merges `fix.rs` **cleanly**. So the merged tree contains both `fix_9_without_device_index_is_print_only_and_returns_zero` (asserting 0) and this test (asserting 3) for the same call path. That's a guaranteed red `Test (affected crates)` after rebase, with no conflict marker to warn you it's coming.

Simplest fix is to drop this test — main's already covers the path.

There's a doc consequence too: `skills/rocm-doctor/reference.md:39` still documents exit **3** for "missing/negative `--device-index`", which is now wrong for the *missing* case. That row needs correcting upstream to say *negative* only. And note the new feature deliberately tests no exit codes — so the one place the mirrored document is demonstrably stale is exactly where the "the contract is now executable" claim has a hole.

skills/rocm-doctor/SKILL.md
```diff
@@ -0,0 +1,139 @@
...
+     `notes`, and the `needs_sudo` / `needs_reboot` / `needs_relogin` /
+     `auto_applicable` flags). `score >= 75` = high confidence; `50–74` = likely
+     (confirm one more piece of evidence with the user first).
+   - `out_of_scope` — when set (e.g. WSL2), do **not** diagnose. First, if the
+     user's symptom clearly names an app that ships its own runtime (Lemonade,
+     Ollama, LM Studio), route them to that app's tracker (see
+     [Framework routing](#framework-routing)) — those trackers apply regardless
+     of platform. Otherwise relay the `out_of_scope` message and stop (see
+     [Out of scope](#out-of-scope)).
+   - `route_when_no_match` — when `matched` is empty, hand the user this
```
rominf: **The skill tells the agent to gate on `matched` being empty, which `rocm-core` documents as the wrong gate.**

Step 1 here says "`route_when_no_match` — when `matched` is empty, hand the user this upstream tracker". `DiagnoseReport::has_match` exists specifically because that's wrong (`crates/rocm-core/src/diagnose.rs:67-81`):

> Serialized because a JSON consumer cannot otherwise tell a real cause from a weak signal, and the obvious substitute — "is `matched` empty?" — is wrong. Several checkers open with a nonzero base score for a situation that is merely *potentially* relevant (being in a container, having an APU alongside a discrete GPU), so a perfectly healthy host produces a non-empty `matched` full of sub-threshold entries. **A caller gating on emptiness proposes a fix for a machine with nothing wrong, and never routes the user to `route_when_no_match`.**

That's a precise description of what an agent following this instruction will do. And `reference.md` never mentions `has_match` at all — it's absent from both mirrored files (`grep -c has_match` → 0 on each).

Worth noting the new feature *cannot* catch this class: the field-presence assertion is one-directional (`documented ⊆ report`), so a field the CLI emits and the doc omits is invisible to it. That's the exact class of drift the feature was written to catch.

Two fixes: upstream, add `has_match` to `reference.md`'s field list and change step 1 to branch on it; here, add a reverse assertion that every top-level key of `diagnose --json` is documented (or an explicit allow-list of intentionally-undocumented keys).

skills/rocm-doctor/reference.md
```diff
@@ -0,0 +1,91 @@
...
+Linux-only: fix-3, -4, -5, -7, -10, -11, -12. Windows-only: fix-13, -14, -15.
+Cross-platform: fix-1, -2, -6, -8, -9.
+
+## Framework routing
+
+`rocm diagnose` diagnoses frameworks built against the **system** ROCm/HIP:
+
+- **PyTorch**, **llama.cpp** — in scope.
+
+Apps that ship their own runtime are routed upstream (via `route_when_no_match`):
```
rominf: **This PR deletes routing arms the mirrored doc still advertises, and the check is one-directional.**

`reference.md:79-83` says:

> Apps that ship their own runtime are routed upstream (via `route_when_no_match`): **Lemonade** → …/lemonade/issues, **Ollama** → …/ollama/issues, **LM Studio** → in-app support

This PR's own change to `crates/rocm-core/src/diagnose.rs` removes the `lemonade`, `ollama`, `lm-studio` and `amdgpu-install` arms from both `upstream_tracker` and `route_when_no_match`. So the blessed fixture documents three targets that are now unreachable.

`assert_route_is_documented` (`skill_steps.rs:573-582`) only checks the CLI's actual target is **among** the documented set — never the converse — so three documented-but-dead targets pass silently. The PR body acknowledges the doc is wrong and ships it as the authoritative fixture anyway.

Two other things about this deletion:

- It's a CLI behavior change inside a `test(e2e):` PR. `AGENTS.md` §11 wants one logical change per PR, and the title doesn't advertise this.
- `Examination` derives `Deserialize`, and #250 adds `doctor --from-examination <path>`, which can supply an arbitrary `framework` string. That makes the deleted arms reachable again from user-supplied input — so "unreachable today" may not survive #250.

Fix the routing section upstream first, then mirror; and assert the documented target set **equals** the reachable set.

crates/rocm-core/src/diagnose.rs
```diff
@@ -1307,13 +1314,17 @@ fn run_all_checks(e: &Examination, symptom: &str) -> Vec<Diagnosis> {
     results
 }
 
+/// Where to send a user when nothing in the catalog matched.
+///
+/// Keyed off the *host-detected* framework, which `Examination::probe` only
+/// ever sets to `pytorch`, `llama-cpp`, `unknown` or `skipped` — so those two
+/// named arms plus the ROCm-core default are the whole reachable set. If a probe
+/// for another framework is added, add its arm here and in [`upstream_tracker`];
+/// `routing_targets_cover_every_framework_the_probe_reports` will flag it.
```
rominf: **This comment promises a guarantee the test doesn't provide.**

The comment says "If a probe for another framework is added, add its arm here and in `upstream_tracker`; `routing_targets_cover_every_framework_the_probe_reports` will flag it."

That test (`:1643`) iterates a **hard-coded** list:

```rust
for framework in ["skipped", "pytorch", "llama-cpp", "unknown"] {
```

It never reads `examine.rs`. Adding a fifth framework to `probe_framework` — say `"vllm"` — changes nothing about this test; it stays green, and the new framework silently falls to the `_ => "rocm-core"` arm.

The comment is worse than no comment, because it'll cause a future author to skip the manual update they actually still need to do.

Either derive the list from a single exported constant or enum shared with `examine.rs`'s `probe_framework` (which would make the guarantee real), or reword the comment to say the list is hand-maintained and must be updated alongside.

tests/e2e-cucumber/tests/e2e/skill_steps.rs
```diff
@@ -0,0 +1,619 @@
...
+            panic!(
+                "`rocm fix` offers {id}, which skills/rocm-doctor/reference.md does not document"
+            )
+        });
+        assert_eq!(
+            documented.auto, offered.auto,
+            "{id}: reference.md says auto-applicable={}, `rocm fix` says {}",
+            documented.auto, offered.auto
+        );
+    }
```
rominf, line 508: **"both files name these four ids in prose" isn't true, and this duplicates an existing unit test.**

`skills/rocm-doctor/SKILL.md` contains **zero** `fix-N` ids — `grep -c 'fix-[0-9]'` returns 0. Only `reference.md:43` names them ("Only four fixes are auto-applicable — `fix-2-unset-override`, …"), and this assertion doesn't read that prose anyway: it compares the CLI against the `AUTO_APPLICABLE` Rust array at :38.

So the stated justification for pinning the array doesn't hold. And `crates/rocm-core/src/fix.rs:1141` (`exactly_the_four_known_fixes_are_auto`) already asserts the same four ids in the same order, from the same source.

It also cuts against the rule this PR itself adds at `tests/e2e-cucumber/README.md:238`: *"If a scenario in this feature does not read `reference.md`, it is in the wrong file."*

Either drop the pinned array (the unit test covers it), or parse the "Only four fixes are auto-applicable — …" line out of `reference.md:43` so the assertion is genuinely doc-derived — which is what the feature is for.

tests/e2e-cucumber/features/rocm_doctor_skill.feature
```diff
@@ -0,0 +1,81 @@
...
+    Given the ROCm Doctor skill as it is published
+    When an agent inspects the machine for tooling
+    Then the inspection succeeds whatever it finds
+    And its verdict is one the skill documents
+
+  # The skill's standing rule is "never invent a fix — if nothing matched, route
+  # the user upstream". That rule is only followable if the address the CLI hands
+  # over is the one the skill's own routing table gives, so this is the last
+  # claim in the document that the binary can be held to.
+  @id:skill-escalation-target-matches-cli
```
rominf, line 90: **Scenario 6's `Given` is inert — the assertion passes identically for a matched symptom.**

`diagnose()` populates `route_when_no_match` **unconditionally**, regardless of whether anything matched. This `Given` only sets a symptom string; no assertion downstream checks `has_match`, or `matched` emptiness, or that the catalog actually failed to explain the report.

So "A report the catalog cannot explain is routed where the skill says" is really testing "the route field is well-formed and documented". Swapping `UNMATCHED_SYMPTOM` for a known symptom would not change the outcome, which means the scenario's stated behavior is never exercised.

Asserting `has_match == false` for the unmatched symptom before checking the route would make it real (skipping when `out_of_scope` is set — that's the only host-dependent branch). Otherwise, fold the route check into scenario 4 and drop the misleading `Given`.

Two small ones while I'm here:

- **`world.model_name` is being repurposed to carry the symptom** (`skill_steps.rs:400`, `:405`, `:419`). That field is the shared world's model identifier; overloading it will collide silently if a future scenario needs both. This PR already shows the right pattern by adding a dedicated `skill_reference` field in the same change (`tests/e2e-cucumber/tests/e2e.rs:88`) — a `skill_symptom: Option<String>` alongside it is the fix. I raised this on the previous revision and it's still here.
- **The `actions/checkout` pin on the new job** (`.github/workflows/ci.yml:194`, `df4cb1c…` v6.0.3) matched its siblings when this branch was cut, but `main` has since bumped every other job in the file to `3d3c42e5…` v7.0.1. Just something the rebase needs to pick up, not an authoring mistake.

rominf (changes requested):
I reviewed this at `04fe5aa5`. The idea is good — making a published skill's documented contract executable, rather than letting it rot silently, is exactly the right instinct, and the doc-as-fixture parsing in `skill_steps.rs` is nicely done. The ten new `rocm-core` unit tests are real coverage, and I confirmed the scenarios actually run (plain untagged, `skill_steps.rs` is registered in `e2e.rs`).

But the branch is 38 commits behind `main` and that has caused real damage, plus the mirror premise no longer holds. My recommendation is: **rebase, re-mirror all three files from upstream, and fix upstream first for the three places the doc is now wrong** — then this is close.

The blocking set:

1. `licenserc.toml` reverts main's HawkEye v7 migration (#300). It's the *only* merge conflict, and after rebase the exemption mechanism it edits no longer exists, so the mirrored `.md` files would be header-required and the blocking check would fail.
2. The drift job fetches a path that 404s. The one check whose job is catching mirror rot can never fire — and it fails open, so nobody will notice.
3. The mirror is not byte-verbatim any more. Upstream's `SKILL.md` has gained a **Scope gate** section (a substantive safety instruction: decline non-AMD/WSL2 setups and give no troubleshooting) that this copy doesn't have.
4. A new unit test asserts exit 3 where `main` deliberately changed the branch to return 0 — and `fix.rs` auto-merges *without conflict*, so the merged tree gets two tests asserting different values for the same call. Guaranteed red after rebase, with no conflict marker to warn you.

Then five smaller ones about the tests not checking what they claim, and two nits.

**Also worth flagging on the PR body:** `amd/skills#87` merged on 2026-08-13, so "Depends on amd/skills#87, which is still open… the upstream files 404" and the draft rationale are both stale — and the PR is not in draft. The dependency is discharged; the body should say so.

**Sequencing thought:** #250 renames `rocm diagnose` → `rocm doctor` and changes `render_report_text`'s signature. This PR promotes the pre-consolidation surface to a *tested, published contract* — and the new feature only checks CLI-against-doc, never doc-against-CLI-surface, so it won't flag the mismatch. I'd land #250 first, then rebase this with the mirror re-fetched after upstream is corrected.

**CI:** `E2E tests (Strix Halo, Windows)` is red, but the branch is 38 commits stale so that result doesn't mean much until rebase.

Method note: parts of this were read by parallel helper passes that ran degraded (sequential, no fan-out). Everything above I verified myself — including diffing the mirror against upstream, running `git merge-tree`, and querying the upstream repo's contents API.

licenserc.toml
volen-silo: Redone on v7. `licenserc.toml` now keeps main's `[header]` / `[files]` / `[[rules]]` schema and just adds `"skills/rocm-doctor/**"` to `[files] excludes`; the resurrected `licenserc-md-header.toml` and the v6 comment block are gone, and the docs URL stays `fast/hawkeye`.

Verified positively rather than trusting the green check, since a passing `hawkeye check` only proves nothing is failing. Stripping `README.md`'s header makes it exit 1, and so does stripping `skills/rocm-cli-assistant/SKILL.md` — so the exclusion is scoped to the one folder and nothing else lost coverage. `9eb0eabc`

.github/workflows/ci.yml
volen-silo: Job deleted rather than repointed. The premise changed underneath it: `amd/skills` now federates — `.github/federation.json` names a source repo per skill and a nightly job vendors the folder from `main` — and its README's architecture diagram already places `rocm-doctor` under a ROCm product repo. So this repo is the source of truth and the catalog is the consumer, which makes a hand-maintained mirror plus a drift check the thing federation replaces.

You were right that it could never fire, and right about the fail-open masking it. Both stop mattering once nothing is being mirrored inward.

skills/rocm-doctor/SKILL.md
volen-silo: Re-synced all three from `staging/rocm-doctor/`, so the Scope gate, Prerequisites, the extra Rules bullet, the workflow preamble and the `[reference.md](reference.md)` link are all in.

With the direction inverted (this repo is now the source of truth), the "byte-verbatim mirror" claim is gone from `AGENTS.md`, `licenserc.toml`, `tests/e2e-cucumber/README.md` and the feature header — it is no longer something that has to be kept true by hand.

crates/rocm-core/src/fix.rs
volen-silo: Dropped, and confirmed the way you predicted: the merge was clean and `cargo test -p rocm-core --lib fix::` went red with `left: 0, right: 3` before I removed it. Main's `fix_9_without_device_index_is_print_only_and_returns_zero` covers the path.

`reference.md`'s exit-code row 3 now reads "OS mismatch, negative `--device-index`" — the missing case is a print-only plan, which row 0 already covers. Fixed here rather than upstream, since this repo owns the file now.

skills/rocm-doctor/SKILL.md
volen-silo: Fixed both halves in `9eb0eabc`. `reference.md` now documents `has_match` with the reason emptiness is the wrong substitute, and `SKILL.md` step 1 gates on it — the `route_when_no_match` bullet says "when `has_match` is false" instead of "when `matched` is empty".

Added the reverse assertion too. The field check is now an equality, not `documented ⊆ emitted`. It needed no allow-list: `rocm diagnose --json` emits exactly six top-level keys and `has_match` was the only one the document omitted.

Checked it is not vacuous — removing `has_match` from `reference.md` fails scenario 4 with `diagnose --json emits ["has_match"], which skills/rocm-doctor/reference.md never names`. (First attempt at that negative test passed, because the `route_when_no_match` bullet still mentioned the field and the parser picked it up there; it fails once both mentions go.)

skills/rocm-doctor/reference.md
volen-silo: Both parts done.

The routing section is corrected: Lemonade / Ollama / LM Studio are now attributed to the skill's own hand-off table, with the reason stated — the probe reports only `pytorch`, `llama-cpp`, `unknown` or `skipped`, so `route_when_no_match` has no arm for them. `SKILL.md`'s hedge ("only when the host probe detects that app") is gone for the same reason.

On the one-directional check: `reference.md` now lists the CLI's own targets separately, and `assert_route_is_documented` reads that list instead of the union — so a documented-but-dead CLI target can no longer pass by matching an entry that belongs to the skill's table. Worth being straight about the limit: a black-box e2e sees one route per run and cannot enumerate the CLI's reachable set, so this is "the CLI is held to its own list", not full set equality. The equality is pinned in `routing_targets_cover_every_framework_the_probe_reports`.

Your #250 point stands and I have not addressed it — the arms become reachable again if an arbitrary `framework` can be supplied. Flagging rather than guessing at it here.

crates/rocm-core/src/diagnose.rs
volen-silo: Reworded — the comment now says the list is hand-maintained and names the three places that must be edited together (the arm here, the tracker in `upstream_tracker`, the list in the test), and the test's own doc comment says it reads its own list rather than deriving one from `examine.rs`, so it cannot notice a new framework.

Took the reword over the shared-constant option: making the guarantee real means exporting the probe's framework set as public surface, which is a larger change than this PR should carry. Happy to do it as a follow-up if you would rather have the real guarantee.

tests/e2e-cucumber/features/rocm_doctor_skill.feature
volen-silo: All three.

The `Given` is no longer inert — scenario 6 asserts `has_match` is false before checking the route. I first guarded it with "skip when `out_of_scope` is set", then removed the guard: on WSL2 the catalog is not run at all, so `has_match` is false there too and the claim holds by a different path. Guarding it would have made the check vacuous on exactly this host.

Being straight about what I could verify: I confirmed the assertion is live (forcing it to expect `true` fails with `left: Some(false)`), but I could not produce the converse locally — this host is WSL2, so even a known symptom yields `has_match: false`. A matching symptom only exists on a host the catalog covers, so the no-GPU Linux lane is where it bites.

`skill_symptom: Option<String>` added alongside `skill_reference`; `model_name` is no longer overloaded. Sorry for missing that the first time you raised it.

The `actions/checkout` pin is moot — that job is gone.

tests/e2e-cucumber/tests/e2e/skill_steps.rs
volen-silo: Took your second option — it now parses the prose. `documented_auto_prose` reads the "Only four fixes are auto-applicable — …" sentence out of `reference.md` (bounded by the em dashes, so the `rocm fix fix-2-unset-override` example later in the same paragraph stays out) and compares that set to what `rocm fix` reports. The `AUTO_APPLICABLE` Rust array is gone.

Dropping the scenario outright would have taken the per-id auto-flag doc↔CLI comparison with it, which is the drift this feature exists to catch and which `exactly_the_four_known_fixes_are_auto` does not cover — it compares the CLI to a Rust array, never to the document. So the scenario stays and now earns its place by the rule you quoted: it reads `reference.md`.

Side effect worth having: the table and the prose are now checked against each other, so a rename applied to the `yes` cells and the CLI together still fails if the prose was missed.

tests/e2e-cucumber/features/rocm_doctor_skill.feature
```diff
@@ -0,0 +1,81 @@
+Feature: The ROCm Doctor skill's instructions match the CLI they drive
```
rominf, line 1: This new feature file has no entry in `FEATURE_KEYS` (`tests/e2e-cucumber/tests/feature_naming.rs:23-42`), so `feature_files_and_declared_keys_agree` fails — confirmed locally:

```
rocm_doctor_skill.feature has no key in FEATURE_KEYS — add one so its scenarios are indexed and its ids are qualified like every other feature
```

This runs in the plain `cargo test` set (no GPU/binary needed) and is what's failing `Test (affected crates)`, `windows-build-and-test`, and `E2E tests` in CI.

Fix: add `("rocm_doctor_skill.feature", "skill")` to `FEATURE_KEYS` — the six `@id:skill-*` tags already used here match that key — then rename each `Scenario: N - ...` to `Scenario: skill-0N - ...` (01-06) to satisfy `scenario_names_are_indexed_sequentially_per_feature`.

rominf (commented):
Reviewed at `9eb0eabc`.

Went through all 9 threads from my previous pass and checked each fix against the actual current code rather than taking the replies at face value — all genuinely hold:
- `licenserc.toml` is back on the v7 schema with `skills/rocm-doctor/**` excluded, and the resurrected v6 file is gone.
- The dead `amd/skills` drift-check job is deleted outright (the repo is now the source of truth, so there's nothing to drift-check).
- `SKILL.md`/`reference.md` are re-synced: the Scope gate, Prerequisites, and the `has_match` gating are all present, in both directions (the reverse field-coverage assertion actually catches a doc omission when I mentally removed `has_match` from it).
- The conflicting device-index test in `fix.rs` is dropped in favor of main's, and `reference.md`'s exit-3 row is corrected to "negative only."
- The dead Lemonade/Ollama/LM Studio routing arms are gone from `diagnose.rs`, and the routing table now lives in the skill's own hand-off, checked against the CLI's real reachable set.
- The `diagnose.rs` comment is honest now about the framework list being hand-maintained rather than promising a guarantee the test doesn't provide.
- `skill_steps.rs` parses the "auto-applicable" prose out of `reference.md` instead of pinning a redundant array, and scenario 6's `has_match` assertion is live (verified via the WSL no-catalog path) with a dedicated `skill_symptom` field instead of overloading `model_name`.

I resolved all 9 threads.

One new thing turned up on a fresh pass over the current diff: the new `rocm_doctor_skill.feature` has no entry in `FEATURE_KEYS` (`tests/e2e-cucumber/tests/feature_naming.rs`), so `cargo test -p e2e-cucumber --test feature_naming` fails — reproduced locally, `feature_files_and_declared_keys_agree` panics with "rocm_doctor_skill.feature has no key in FEATURE_KEYS — add one so its scenarios are indexed and its ids are qualified like every other feature." That's what's turning `Test (affected crates)`, `windows-build-and-test`, and `E2E tests` red right now. Left as an inline comment below.

Everything else checked out: `cargo check`/`cargo build` on the touched crates, the full `rocm-core` unit suite (48/48), clippy with `-D warnings`, and the real `rocm_doctor_skill.feature` run against a built `rocm` binary (all 6 scenarios pass) — which also confirms the fix-9 `auto_applicable` correction is a genuine, previously-unnoticed bug fix backed by a real regression test, not just a doc tweak.

rominf (changes requested):
Marking the verdict properly — my review just above should have been request-changes rather than a plain comment (my error, not a change of opinion). Nothing new here.

The blocker is the one inline comment on `tests/e2e-cucumber/features/rocm_doctor_skill.feature`: the file has no `FEATURE_KEYS` entry, so `feature_files_and_declared_keys_agree` fails and `Test (affected crates)`, `windows-build-and-test` and `E2E tests` are red. Everything from the previous round is resolved and verified.

tests/e2e-cucumber/features/rocm_doctor_skill.feature
volen-silo: Fixed in `e6fe813d`. Added `("rocm_doctor_skill.feature", "skill")` to `FEATURE_KEYS` and renamed the six scenarios to `skill-01`..`skill-06` — indexing the file is what makes the sibling naming tests start applying to it. The `@id:` tags already carried the `skill-` prefix, so `every_scenario_has_a_feature_qualified_id` needed nothing.

`cargo test -p e2e-cucumber --test feature_naming` is 4/4, and the full `cargo test --workspace --all-targets` is green — no `--exclude e2e-cucumber` this time, which is exactly how I missed it.

**One correction, though: `E2E tests` was red for a second, independent reason, and would still be red after this fix alone.**

Scenario 6 asserted `has_match == false`. I added that last round in response to your inert-`Given` comment, and the premise does not survive a real host. `diagnose` scores several checkers from host state alone, with no symptom keyword involved — the GitHub-hosted Linux runner ships `/etc/modprobe.d/blacklist-radeon-instinct.conf` with `amdgpu` unloaded, which is `fix-5-amdgpu-load` at score **90** whatever symptom it is given:

```
assertion `left == right` failed: ... the symptom must not have matched:
  left: Some(true)
 right: Some(false)
```

It passed on your host and on the WSL2 lane for the same reason: WSL2 short-circuits the catalog as out-of-scope, so `has_match` is false there by a different path. That made it look host-independent when it was not. I did say last round that I could not verify the converse locally — that turned out to be the case that mattered.

So the claim moved to where the host can be held still, in `crates/rocm-core/src/diagnose.rs`:

- `sub_threshold_causes_leave_has_match_false_and_route_upstream` — builds an `Examination` missing both groups (score 45), then asserts `matched` is **non-empty**, every cause is below `min_score_for_match`, `has_match` is false, and the route still names somewhere to go. That is precisely the shape that makes gating on emptiness wrong, and no probed host can be relied on to produce it.
- `an_established_cause_sets_has_match` — a matching symptom flips the flag, so the assertion is not free.

What is left in scenario 6 is the half only this feature can check: whatever target the CLI hands back is one the published document names. It is renamed to say that, rather than to claim a premise it does not establish, and `README.md` now explains why no scenario asserts match-or-not in either direction.

Full suite locally: 81 scenarios, 79 pass, the 2 failures are the declared WSL2 `diagnose` xfails (0 unexpected).

skills/rocm-doctor/reference.md
```diff
@@ -0,0 +1,105 @@
...
+| 2 | usage error (incl. unknown fix-id) |
+| 3 | not applicable on this host (OS mismatch, negative `--device-index`) — nothing changed |
+| 4 | attempted but the command failed |
+| 5 | user declined at the prompt |
+
+Only four fixes are auto-applicable — `fix-2-unset-override`,
+`fix-4-render-group`, `fix-6-path`, `fix-9-igpu-dgpu` — and the rest print their
+plan for the user to run. Pass the **full** id (`rocm fix fix-2-unset-override`,
+not `rocm fix fix-2`; a short id returns exit 2, unknown fix-id). Auto fixes
+print the exact command, honor `--dry-run`, refuse on a non-interactive shell
```
rominf: **`fix-2-unset-override` does not honor `--dry-run` or `--yes` on Linux, so this paragraph is not true for it.**

`run_unset_override` (`crates/rocm-core/src/fix.rs:655`) splits by platform. The Windows arm takes `opts` and does exactly what this paragraph describes — `if opts.dry_run { ... } else if confirm(..., opts.yes) { ... }` (`fix.rs:709`). The Linux arm is `run_unset_override_linux()` (`fix.rs:663`), which takes no `FixOptions` at all: it prints where the override is set, says "This command does NOT edit your shell rc files for you; that's your dotfiles", and returns 0. It never mutates, so `--dry-run` is a no-op, `--yes` is never read, and there is no prompt to refuse.

On Linux, then, fix-2 is print-only — but this paragraph, the catalog row at line 61 (`| fix-2-unset-override | both | ... | yes |`), and `SKILL.md:142` ("Only the four auto-applicable fixes prompt and mutate. The other 11 are **print-only**") all tell an agent otherwise. An agent following this doc will report a `--dry-run` preview the user never got.

Two ways out: scope the claim per platform (fix-2 is auto on Windows, print-only on Linux), or redefine the Auto-fix column as "has a runner" rather than "prompts and mutates" and move the dry-run/confirm promise to the fixes that keep it.

Worth pinning either way. `auto_applicable_recipes_have_a_runner` only checks that an auto recipe *has* a runner — fix-2 does — so nothing currently fails when the runner turns out to be a no-op on one platform. That is exactly the doc/CLI drift this PR exists to catch.

crates/rocm-core/src/diagnose.rs
```diff
@@ -957,7 +960,11 @@ fn check_9_igpu_dgpu_collision(e: &Examination, symptom: &str) -> Diagnosis {
                 "# Persist in your shell rc or your launch script.".to_owned(),
             ],
             fix_id: "fix-9-igpu-dgpu".to_owned(),
-            auto_applicable: false,
+            // `rocm fix fix-9-igpu-dgpu --device-index N` really does apply this
+            // on Linux (see `fix::run_hip_visible_devices_linux`), so the
+            // diagnosis must not tell an agent otherwise — it branches on this
+            // flag to decide whether to offer to run the fix or just print it.
+            auto_applicable: true,
```
rominf, line 1001: **This flip is user-observable, and there is no scenario for it.**

`render_report_text` pushes `"rocm fix can run it"` onto the flags line when `fix.auto_applicable` (`diagnose.rs:1442`), and `Fix` derives `Serialize`, so the JSON flips too. A fix-9 diagnosis that previously printed no `flags:` line now prints one.

`AGENTS.md` is explicit that a unit test does not discharge this:

> If the change alters what a user of the CLI can observe — command output, exit codes, files or paths the CLI creates, or which runtime/engine it selects — then that behavior must be covered by a Gherkin scenario in `tests/e2e-cucumber/features/` [...] a unit test asserting the internal helper does NOT discharge this; it proves the function, not the behavior

Nothing under `tests/e2e-cucumber/features/` mentions `fix-9` or the auto-applicable flag today, so the old and new values are equally green.

I recognise that provoking a real iGPU+dGPU collision on a CI host may not be feasible. If it isn't, that is worth stating in the PR description rather than leaving the gap silent — a reader can't otherwise tell an uncoverable case from an overlooked one.

crates/rocm-core/src/diagnose.rs
```diff
@@ -106,14 +106,17 @@ impl DiagnoseReport {
 }
 
 /// Upstream tracker for a framework key.
+/// Upstream tracker for a routing target.
```
rominf, line 108: **Duplicated doc line.** Line 108 above still reads "Upstream tracker for a framework key." and this line adds "Upstream tracker for a routing target." The new one is the accurate summary now that the framework-keyed arms are gone; the old line should go with them.

rominf (changes requested):
Re-reviewed at `e6fe813d`. Both things I flagged last round are fixed, and I owe you a correction on one of them.

**Correction: I attributed all three red checks to the missing `FEATURE_KEYS` entry. That was wrong.** Only `Test (affected crates)` and `windows-build-and-test` came from that. `E2E tests` failed for the separate reason you found — scenario 6's `has_match == false` assertion. And that assertion was there because I asked for it in my previous review, on the reasoning that `out_of_scope` was the only host-dependent branch. It isn't: the catalog scores host state, so `fix-5-amdgpu-load` reaches 90 on the hosted Linux runner whatever symptom it is handed. Your diagnosis is right and mine was wrong. Sorry for the round trip.

Confirmed fixed at this head:

- `rocm_doctor_skill.feature` is in `FEATURE_KEYS` and the six scenarios are indexed `skill-01`..`skill-06`, so the three sibling naming tests now actually run against it instead of skipping it vacuously.
- The `has_match` claim moved into `rocm-core` against a constructed `Examination`. `sub_threshold_causes_leave_has_match_false_and_route_upstream` plus the companion that flips the flag is the right shape — it pins the thing you actually mean without depending on what the runner happens to look like.

All checks are green on `e6fe813d` as of now, including the `E2E tests` and `windows-build-and-test` lanes.

Three things still open, inline below. The first matters most: the skill doc makes a promise the CLI does not keep on Linux, and the whole point of this PR is that the doc and the CLI agree.

crates/rocm-core/src/diagnose.rs
volen-silo: Dropped in `cfe1bb21`. The new summary line was added without removing the old one when the framework-keyed arms went.

crates/rocm-core/src/diagnose.rs
volen-silo: Confirmed, and it is not coverable at e2e — so it is now stated in the PR description rather than left silent, per your suggestion.

`check_9_igpu_dgpu_collision` only fires on an `Examination` carrying an APU *and* a discrete AMD GPU. The suite drives the real binary against the real host, and I checked there is no way to hand it a synthetic one: no `--from-examination`-style input exists on any branch, and no CI lane has both an APU and a dGPU. So the flip is covered by the `rocm-core` regression test that runs `diagnose` for both OS families and asserts every emitted fix agrees with the catalog, and the PR now says plainly that the e2e half is uncoverable and why.

skills/rocm-doctor/reference.md
volen-silo: You are right, and this one stung a little — my own commit message for the unit-test commit says "the dry-run test picks `fix-2`, whose Linux runner takes no `FixOptions` and never prompts". I knew the fact and still did not notice the doc claimed the opposite.

Verified before changing anything: `run_unset_override` splits on `runtime_is_windows()`; the Windows arm takes `opts` and does the dry-run/confirm dance, `run_unset_override_linux()` takes no options and returns 0 after reporting.

Took your second option, in `cfe1bb21`. Making `auto_applicable` OS-aware would be the root-cause fix, but it is a flat `bool` on a `const` recipe, so it ripples into the `rocm fix` listing, the diagnosis JSON, `exactly_the_four_known_fixes_are_auto` and the catalog row — a CLI behaviour change that belongs in its own PR, not bolted onto this one. Happy to open it if you would rather the flag mean "prompts and mutates".

So the flag keeps its meaning and the docs stop overpromising. `reference.md` now defines the Auto-fix column as what `rocm fix` actually reports — the CLI has a runner and will carry it out itself, explicitly *not* a promise that it mutates — and the dry-run/refuse/confirm contract moves to prose naming fix-2-on-Linux as the exception. `SKILL.md:142` gets the same correction, since it made the same claim.

On pinning it — you are right that `auto_applicable_recipes_have_a_runner` cannot catch this, since fix-2 does have a runner. My first attempt at a test was worthless and I want to flag that rather than bury it: I asserted `apply("fix-2-unset-override", ...)` exits 0 with no `--yes`, then checked it discriminated by routing fix-2 through the Windows arm — **it still passed**, because `ps_env_scope` finds no PowerShell on Linux, returns empty, and the Windows arm exits 0 before ever reaching `confirm`. An exit code cannot express this claim.

What landed instead asserts the files. The rc-file scan is split into `report_persistent_override(candidates)`, taking its paths the way this PR already made `pin_device_in_rc_file` take its rc path — deriving them from `$HOME` would mean mutating a process-global in a parallel suite, which is a real hazard here (`the_path_fix_finds_the_install_the_rest_of_the_cli_found` sets `ROCM_PATH` and flaked on the Windows lane for exactly that reason). The test seeds a scratch `.bashrc` carrying the override plus a clean `.profile`, and asserts both are byte-identical afterwards — not an exit code, since a runner that started stripping the line would still return 0. Confirmed it fails when the runner is made to mutate:

```
assertion `left == right` failed: fix-2 on Linux must leave the user's dotfiles byte-identical
  left: "export PATH=$PATH:/x\n"
 right: "export HSA_OVERRIDE_GFX_VERSION=10.3.0\nexport PATH=$PATH:/x\n"
```

AGENTS.md
```diff
@@ -176,6 +177,29 @@ Required consistency points:
...
+
+`skills/rocm-cli-assistant/SKILL.md` is compiled into the binary
+(`include_str!` in `apps/rocm/src/main.rs`). `skills/rocm-doctor/` is different
+on two counts, and both change how you edit it:
+
+- **This repo is its source of truth.** The skill is a thin driver over the
+  `rocm` binary, so it is versioned with the binary and lives here. The
+  [`amd/skills`](https://github.com/amd/skills) catalog federates it: a nightly
+  job vendors the folder from `main` and opens a pull request there. So edit it
+  here, and never edit the catalog's copy — the next federation run overwrites
```
rominf: This instruction points at a sync that doesn't happen, and I'd treat it as blocking because of what was removed on the strength of it.

I checked the catalog directly rather than going from the PR text. `amd/skills/.github/federation.json` declares exactly one source — `AMD-AGI/TraceLens`, skill `tracelens-analysis-orchestrator`. There is no `ROCm/rocm-cli` entry. And `rocm-doctor` isn't under `amd/skills/skills/` at all; it's in `amd/skills/staging/rocm-doctor/`, which `federate-skills.yml` never touches, since it vendors only the sources declared in `federation.json`. The two copies have already drifted apart.

So "never edit the catalog's copy — the next federation run overwrites it" tells a future agent to rely on something that won't run. That matters more than a stale sentence: last round the amd/skills drift-check job was deleted outright, on the reasoning that both copies stop mattering once federation replaces the mirror. Federation doesn't currently replace it, so this PR removes the only detector and substitutes a mechanism that isn't wired up. Right now nothing keeps the published copy honest.

Two ways out, either fine by me: land the `federation.json` entry and promote `staging/rocm-doctor` first, and reference that from here; or say plainly that the catalog copy is hand-synced from staging today, name who syncs it, and keep a drift detector until federation is live. The same claim appears in `licenserc.toml`, the e2e README, and the feature header, so they'd all move together.

Separately, line 191 says federation doesn't carry the skill's `evals/` folder — but `skills/rocm-doctor/` in this PR has no `evals/` at all. That's phantom detail in a file loaded on every task; either drop the sentence or land the folder.

crates/rocm-core/src/diagnose.rs
```diff
@@ -957,7 +959,11 @@ fn check_9_igpu_dgpu_collision(e: &Examination, symptom: &str) -> Diagnosis {
                 "# Persist in your shell rc or your launch script.".to_owned(),
             ],
             fix_id: "fix-9-igpu-dgpu".to_owned(),
-            auto_applicable: false,
+            // `rocm fix fix-9-igpu-dgpu --device-index N` really does apply this
+            // on Linux (see `fix::run_hip_visible_devices_linux`), so the
+            // diagnosis must not tell an agent otherwise — it branches on this
+            // flag to decide whether to offer to run the fix or just print it.
+            auto_applicable: true,
             verify: "HIP_VISIBLE_DEVICES=1 python -c \"import torch; print(torch.cuda.device_count())\"".to_owned(),
```
rominf, line 1002: This is the fix-2 problem again, in the PR that promotes the fix.

Flipping this to `true` makes `render_report_text` emit "rocm fix can run it" and "apply with: `rocm fix fix-9-igpu-dgpu`", and flips the JSON through the `Serialize` derive. But `run_hip_visible_devices_linux` short-circuits when `opts.device_index` is `None`: it prints the `rocminfo` identification query, says to re-run with `--device-index N`, and returns 0. The Windows arm does the same. No prompt, no mutation, and nothing for `--dry-run` to preview.

The comment just above shows you know this — `--device-index N` really does apply it. The problem is that nothing an agent reads says so. SKILL.md step 3 tells the agent to run exactly `rocm fix <fix-id>`, and both SKILL.md and reference.md name fix-2 as the one exception among the four that doesn't mutate everywhere. The `--device-index` requirement lives only in `RECIPES[fix-9].notes`, which a `diagnose --json` consumer never reads.

Net effect: an agent is newly told the CLI will apply fix-9, runs the documented command, gets exit 0, and reports a change that never happened. That's the same shape as the fix-2 overpromise from the last round.

Extend both exception paragraphs to cover fix-9, and ideally surface the requirement in the diagnosis's own notes so a JSON consumer can see it.

skills/rocm-doctor/reference.md
```diff
@@ -0,0 +1,116 @@
...
+`fix-2-unset-override` mutates on Windows only.** Its Linux runner reports where
+the override is set and which rc files carry it, then stops — it will not edit
+your dotfiles. So on Linux it never prompts, `--dry-run` has nothing to preview,
+and `--yes` is never read. Do not tell a Linux user a dry run previewed a change
+that was never going to happen.
+
+## Closed catalog (15 failure modes)
+
+| id | OS | Failure mode | Typical signal | Auto-fix |
+| --- | --- | --- | --- | --- |
```
rominf, line 77: This line is a second statement of the OS scope, and nothing reads it.

`assert_same_os_scope` compares the CLI against `parse_reference_catalog`, which reads the table's OS column only. So re-scoping a fix in the CLI and updating the table leaves this summary line stale and every scenario still green — while the feature header promises that "a re-scoped OS … cannot merge here and silently break the published skill".

This is the same double-statement problem you fixed for the auto set with `documented_auto_prose`; it just wasn't applied to the OS claim. Parsing this line the same way would close it.

Two more in the same family, since they share a root cause. The counts are unparsed — `## Closed catalog (15 failure modes)` above, and SKILL.md's "The other 11 are print-only" — so adding a sixteenth recipe plus a table row leaves all six scenarios green with both numbers stale, which contradicts the header's "or a 16th failure mode" clause. And SKILL.md is never opened by any step: `reference_md_path()` is the only path constructed, yet the Given is worded "the skill as it is published" and the e2e README says this feature is the only thing catching the document drifting. SKILL.md independently restates the auto four, the routing table and the per-fix field list, none of it compared to anything.

Either extend the parsers to cover these, or narrow the header and README claims to what they actually check. I'd rather have a smaller promise than a wider one that isn't kept.

.github/workflows/ci.yml
```diff
@@ -122,6 +122,11 @@ jobs:
               - '**/*.ps1'
               - '**/*.feature'
               - 'tests/e2e-cucumber/**'
+              # skills/rocm-doctor/reference.md is an e2e FIXTURE: the
+              # rocm_doctor_skill.feature scenarios parse its catalog table and
+              # compare it to `rocm fix`. A skill-only edit must therefore run
+              # the E2E job, which gates on `heavy`.
+              - 'skills/**'
```
rominf, line 129: Anchoring here because the file I mean isn't in this diff. `skills/**` was added to ci.yml's `heavy` filter — good — but `.github/workflows/e2e-selfhosted.yml` has its own copy that opens with "Same `heavy` set as ci.yml: anything that can affect the E2E suite" and then repeats the list verbatim. It didn't get `skills/**`.

So that comment is now false in the file asserting it, and `rocm_doctor_skill.feature` — which those lanes also run — won't be re-exercised on a skills-only PR. There's no parity check in `workflow_contract.rs` to catch it. The lanes are advisory so the blast radius is small, but the comment actively misleads the next person to edit it. Mirror the entry, or name the intentional difference.

crates/rocm-core/src/diagnose.rs
```diff
@@ -1661,6 +1675,173 @@ mod tests {
...
+        let report = diagnose(&e, "the office printer keeps jamming on page three");
+
+        assert!(
+            !report.matched.is_empty(),
+            "this test is pointless unless something sub-threshold was listed"
+        );
+        assert!(
+            report
+                .matched
+                .iter()
```
rominf, line 1919: Scope point, and it's about what lands on main rather than about the code.

This PR carries a real CLI behaviour change — fix-9's auto flag, visible in both the text report and the JSON, plus the removal of four `upstream_tracker` and three `route_when_no_match` arms. It's cleanly separated into `15e4be38`, which is the right instinct. But this repo squash-merges, so the commit that lands on main takes the PR title, and that title is `test(e2e): …`. The conventional-commit type on main will say `test` for a change that alters CLI output.

Either retitle to reflect the behaviour change and describe the test work in the body, or split `15e4be38` out so both commits type honestly. I raised the scope point last round and it was answered by describing the change in the PR body, which helps a reviewer but doesn't reach the commit log.

Related and already handled reasonably: AGENTS.md §3 wants a Gherkin scenario for user-observable change with no stated escape hatch, and the gap is now disclosed in the PR body. I'm fine with that as a documented deviation — just want it called that rather than treated as compliance.

AGENTS.md
```diff
@@ -176,6 +177,29 @@ Required consistency points:
...
+  it. Federation does not carry the skill's `evals/` folder, which is why the
+  catalog keeps a dataset of its own.
+- **`reference.md` is an e2e fixture.** `tests/e2e-cucumber/features/rocm_doctor_skill.feature`
+  parses its closed-catalog table and compares it to what `rocm fix` reports.
+  The catalog is authoritative in `crates/rocm-core/src/fix.rs` — adding,
```
rominf: "Catalog" does three jobs in this subsection and I had to re-read it twice. Line 188 and 190 mean the external `amd/skills` repo; 194 and this line mean the 15-mode failure catalog; then 200 goes back to the external one. Reading "the catalog is authoritative in `fix.rs`" directly after "never edit the catalog's copy" requires inferring which referent is meant at each hop, and the two are unrelated. This file is loaded on every task, so it's the highest-traffic place in the repo for that kind of ambiguity — worth reserving the word for one meaning and calling the other the failure-mode catalog.

Small accuracy point in the same line: it's only half right. `RECIPES` lives in `fix.rs`, but the failure-mode checkers and their OS scoping live in `diagnose.rs` — which is where "adding or re-scoping a failure mode" actually starts, and it's the file the fix-9 change above touches. Naming both would save the next person a wrong turn.

tests/e2e-cucumber/tests/e2e/skill_steps.rs
```diff
@@ -0,0 +1,702 @@
...
+            .get(field.as_str())
+            .and_then(serde_json::Value::as_i64)
+            .unwrap_or_else(|| panic!("diagnose JSON has no numeric '{field}'"));
+        assert_eq!(
+            actual, *expected,
+            "{field} is {actual}, but reference.md states {expected} — and the skill's \
+             workflow text reasons about that number when it grades a cause"
+        );
+    }
+}
```
rominf, line 618: This comment describes a branch that can't be taken.

It guards the `if let Some(expected) = known` on the grounds that PyTorch and llama.cpp are named in scope without URLs, so for those there's nothing to compare. But `documented_cli_routes` sets `inside = true` only at the sentinel line and parses just the bullets after it — the PyTorch/llama.cpp bullet sits above the sentinel and is never parsed at all. All three bullets that are parsed carry `https://` URLs, so `route_url` returns `Some` for every entry and `known` is always `Some(_)`.

So the `None` arm is unreachable, and the comment implies the URL comparison is sometimes skipped when it never is. Same class as the `diagnose.rs` comment we corrected last round. Either drop the arm and unwrap with a message naming the missing bullet, or reword to say every documented CLI route currently carries a URL.

rominf (changes requested):
Everything from my last two rounds holds — I re-verified each against the code rather than the replies. licenserc is back on the v7 schema, `has_match` is documented and gated with a real two-way field equality rather than an allow-list, the exit-3 row is right, the unreachable routing arms really are unreachable (I traced every producer of `Examination.framework`), the auto set is parsed from prose, `skill_symptom` is a dedicated field that something actually consumes, and `FEATURE_KEYS` is wired. The `report_persistent_override` and `pin_device_in_rc_file` extractions are behaviour-preserving, and the fix-2 dotfile byte-identity test is the strongest thing in the PR.

The blocker is that the premise the whole design rests on isn't true yet, and I only found it by going and looking at the catalog rather than reading the PR text. Details inline, but in short: `amd/skills` does not federate this repo, `rocm-doctor` sits in its `staging/` directory which federation never touches, and the two copies have already diverged — while the drift-check that would have caught that was deleted last round on the grounds that federation replaces it.

Second, fix-9 repeats the fix-2 overpromise we corrected in the last round, in the same PR that promoted it.

The rest are smaller: several contract restatements that the parsers never read (including SKILL.md, which no step opens at all), a workflow filter that no longer matches the invariant it claims, the commit type, and two comment-accuracy points.

Nothing here is about the quality of the test scaffolding, which is good. It's about the guarantees being claimed for it being wider than what it checks.

tests/e2e-cucumber/features/rocm_doctor_skill.feature
volen-silo: Fixed: `FEATURE_KEYS` now has `("rocm_doctor_skill.feature", "skill")` (tests/e2e-cucumber/tests/feature_naming.rs:38) and the six scenarios are named `skill-01`…`skill-06`. `feature_files_and_declared_keys_agree` and `scenario_names_are_indexed_sequentially_per_feature` both pass.

skills/rocm-doctor/reference.md
volen-silo: Fixed: reworded to "Two exceptions," fix-2 unchanged and a new fix-9 bullet added in both `reference.md` and `SKILL.md`, and the four-fixes/print-only claim is scoped per-platform rather than blanket. `auto_applicable_recipes_have_a_runner` still only checks a runner exists, but the doc no longer overpromises what it does.

crates/rocm-core/src/diagnose.rs
volen-silo: Fixed: the stale duplicate is gone, only "Upstream tracker for a routing target." remains at diagnose.rs:108.

crates/rocm-core/src/diagnose.rs
volen-silo: Confirmed not locally reachable: `check_9_igpu_dgpu_collision` only fires when `Examination` carries both an APU and a discrete AMD GPU, the suite drives the real binary against the real host, there's no `--from-examination`-style seam to hand it a synthetic host, and no CI lane has that hardware combination. Disclosed in the PR body under "Known coverage gap" per the AGENTS.md escape hatch, rather than left silent. It's covered at the unit level instead: a `rocm-core` regression test runs `diagnose` for both OS families and asserts every emitted fix's `auto_applicable` agrees with the `fix::RECIPES` catalog, which is exactly the field that flipped.

crates/rocm-core/src/diagnose.rs
volen-silo: Fixed: extended both exception paragraphs to cover fix-9 ("Two exceptions" in reference.md, same in SKILL.md step 3), naming the `--device-index N` requirement explicitly. Also surfaced it in `check_9_igpu_dgpu_collision`'s own notes (diagnose.rs), so a `diagnose --json` consumer sees the caveat too, not just SKILL.md/reference.md.

tests/e2e-cucumber/tests/e2e/skill_steps.rs
volen-silo: Fixed: dropped the `if let Some(expected)` guard and replaced it with an `unwrap_or_else` that panics naming the missing bullet, plus a comment explaining `documented_cli_routes` only ever returns `Some` in practice (every parsed `route_when_no_match` bullet carries a URL). Fails loudly instead of silently skipping if that ever stops being true.

.github/workflows/ci.yml
volen-silo: Fixed: added `'skills/**'` to e2e-selfhosted.yml's `heavy` filter, right after `'tests/e2e-cucumber/**'`, matching ci.yml's placement and its own "Same heavy set as ci.yml" comment.

AGENTS.md
volen-silo: Fixed: reworded so "catalog" only ever refers to the external amd/skills catalog in this subsection, and the failure-mode catalog is spelled out as such. Also corrected the attribution — it now names both `fix.rs`'s `RECIPES` and `diagnose.rs`'s checkers/OS-scoping as authoritative, since diagnose.rs is where a re-scoped failure mode actually starts (and where the fix-9 change above lives).

AGENTS.md
volen-silo: Agreed, and took the second path rather than landing a federation.json entry from this PR. AGENTS.md's federation bullet now says plainly that amd/skills is still Phase-1 incubation for this skill — no automated sync, `federation.json` only declares TraceLens, and rocm-doctor sits in `staging/rocm-doctor` outside any job's coverage — and that the rocm-cli team hand-syncs staging/rocm-doctor from here whenever it changes materially, rather than describing a job that doesn't run. Also dropped the phantom `evals/` sentence — this skill has no `evals/` folder. I did not restore a drift-check job or touch amd/skills from this PR; left open for you to re-verify the wording matches what you found live.

volen-silo: One thing I did not touch, flagging rather than silently leaving inconsistent: `licenserc.toml`'s comment still says the folder is "vendored verbatim into the amd/skills catalog" — same stale claim, out of this PR's assigned scope, so it's still there for a follow-up.

skills/rocm-doctor/reference.md
volen-silo: Took the narrowing path rather than extending the parsers, on all three points: the feature header now says explicitly that the guarantee covers the table's cells and the auto-applicable prose line — what skill_steps.rs actually parses — and that the OS-scope summary line, the "(N failure modes)" heading count, and SKILL.md's "the other M are print-only" line are free-standing prose that restates those facts and is not parsed, so it can go stale even while the table stays correct. Same caveat added to tests/e2e-cucumber/README.md's description of this feature. Did not extend the parsers to cover SKILL.md, and did not change what skill_steps.rs checks — left open for you to confirm the narrowed claim matches what's actually verified.

crates/rocm-core/src/diagnose.rs
volen-silo: Retitled to `fix(diagnose): correct fix-9's auto flag and test the rocm-doctor contract` so the type that lands on main via squash-merge is `fix`, not `test`, matching the real CLI behavior change. Kept it as one PR rather than splitting `15e4be38` out, since the test commit exists specifically to pin the fix and catch this class of drift going forward. No ticket IDs in the title.

volen-silo:
Pushed a round addressing this review pass:

- fix-9's "one exception" doc gap: extended both reference.md and SKILL.md to name fix-9 alongside fix-2, and surfaced the `--device-index N` requirement in `check_9_igpu_dgpu_collision`'s own diagnosis notes so a JSON consumer sees it too, not just the docs.
- Reworded the dead `None` arm in `skill_steps.rs`'s `assert_route_is_documented` into a loud panic, since `documented_cli_routes` only ever returns `Some` in practice.
- Added `'skills/**'` to `e2e-selfhosted.yml`'s `heavy` filter so it matches `ci.yml`.
- Reworded AGENTS.md so "catalog" isn't overloaded between the external amd/skills catalog and the failure-mode catalog, and attributed the failure catalog's checkers/OS-scoping to `diagnose.rs` alongside `fix.rs`.
- Rewrote the federation description to match what's actually true today: no automated sync for rocm-doctor, `staging/rocm-doctor` is outside any job's coverage, and the rocm-cli team hand-syncs it during Phase-1 incubation. Left that thread open for you to re-verify the wording.
- Narrowed the feature header and README claims to what `skill_steps.rs` actually parses (the catalog table and auto-applicable prose), and called out that the OS-scope summary line and the failure-mode/print-only counts are unparsed prose that can drift. Left that thread open too, since it's the same kind of judgment call.
- Retitled the PR so the squash-merge commit type is `fix`, matching the real CLI behavior change (fix-9's auto flag, the removed routing arms), rather than `test`.

Also confirmed the four already-fixed threads from the previous round still hold (feature naming, the fix-2 exception paragraph, the duplicated doc-comment line, and the fix-9 flag-flip coverage gap, which is disclosed in the PR body under "Known coverage gap").

Verification: `cargo test --workspace --all-targets --exclude e2e-cucumber` clean, `cargo clippy --workspace --all-targets --exclude e2e-cucumber -- -D warnings` and the e2e-cucumber crate's clippy both clean, `cargo fmt --all --check` clean, `prek` hooks clean (license-headers hook skipped locally — `hawkeye` isn't installed in this environment, it normally runs in CI's dedicated job). Ran the full `rocm_doctor_skill.feature` (6/6 scenarios) and the broader `diagnose.feature` — all pass except the 2 pre-existing WSL2-environment `diagnose` xfails, unrelated to this round's changes.

PR #152, opened 2026-07-30
fix(examine): encode GC version minor/revision as hex digits

volen-silo:
- [x] If this PR fixes a bug, searched `tests/e2e-cucumber/expectations.toml` for the fixed ticket ID and removed/narrowed any now-stale xfail rows. (No EAI-7346 rows exist.)

## Problem

An LLVM gfx target is `gfx` + major in **decimal** + minor and revision as **single hex digits** — `gfx90a` is GC 9.0.**10** (`10 == 0xa`), `gfx1250` is GC 12.5.0. `gfx_target_from_gc_version` concatenated all three components as decimal, which agrees with hex only while every component is below 10.

Checked against every gfx target this repo already knows — gfx900/906/908/90a/942/950, gfx1030–1036, gfx1100–1103, gfx1150–1153, gfx1200/1201/1250 — and none contradicts the hex convention.

The concrete victim is **gfx90a** (MI210/MI250/MI250X), on the *primary* detection path (`detect_linux_gfx_target` tries KFD first, `ip_discovery` second — both feed the same helper):

1. KFD packs gfx90a as `90010` (major·10000 + minor·100 + step).
2. `parse_linux_kfd_gfx_target` decodes 9 / 0 / 10 — correctly.
3. The helper produced `"gfx9010"` instead of `"gfx90a"`.
4. `normalize_therock_family("gfx9010")` misses every specific arm and falls through to the loose `starts_with("gfx90")` one → **`"gfx90X-dcgpu"`** instead of `"gfx90a"`.
5. Wrong TheRock family → wrong runtime wheel; and gfx90a drops out of `VLLM_PREFERRED_THEROCK_FAMILIES`, so engine selection changes too.

## Approach

- Format minor and revision as hex; `major` stays decimal (`gfx1030`, `gfx1250` need both digits).
- A minor or revision that cannot be a single hex digit yields `None` rather than a fabricated target, so detection tries the next source and otherwise reports the target as unknown.
- The 5/6-digit branch of `parse_linux_kfd_gfx_target` no longer falls back to `gfx{digits}`. That value is KFD's *packed version*, not a target name, and passing it through was how out-of-range decodes still reached the loose family arm. The 3/4-digit branch keeps `gfx{digits}`, which is correct there.

Sub-10 components are unchanged, so every existing expectation holds.

## Behavior change

An unknown target is recoverable with `--family`, but this does turn a previously-succeeding (and wrong-wheel) `rocm install sdk` on an affected host into a hard error from `resolve_family`, which lists the recognized families. That is the intended direction — failing loudly beats silently installing the wrong runtime — but it is a visible change.

## What this does not close

Not `Fixes EAI-7346`, deliberately. The reported MI455X symptom (gfx1250 read as gfx1210) does not appear to follow from this code: GC 12.5.0 packs to `120500` and decodes to `gfx1250` both before and after. Under the old encoding `gfx1210` was also reachable from GC 1.2.10 and 1.21.0 (packed `10210` / `12100`), which this change maps to `gfx12a` and `None` — but neither is a plausible GC IP version.

I could not obtain the MI455X sysfs values, so that is reasoning about the code, not an observation of the hardware. Closing EAI-7346 needs `gfx_target_version` and `ip_discovery/die/*/GC/*/{major,minor,revision}` from the machine. If the cause is the GC IP version differing from the LLVM target, the fix is an override consulted **before or inside** `detect_linux_sysfs_gfx_target` — the existing PCI/marketing table cannot serve, since it is only reached when sysfs returns `None` and, on Linux, only under WSL.

Related gap, not addressed here: `normalize_therock_family("gfx1250")` returns `None`, so even a correctly detected gfx1250 maps to no family. Naming MI455X's TheRock family would be a guess, and guessing reintroduces the silent-wrong-family class this PR fixes.

## Testing

`cargo fmt --all -- --check` and `cargo clippy --workspace --all-targets -- -D warnings` clean. `cargo test --workspace --no-fail-fast` passes except two pre-existing `rocm-core` `proc_lifecycle::tests::tree_*` failures, which fail on WSL2 for process-tree signalling reasons unrelated to this change (untouched file).

New coverage: the `gfx90a` regression at the helper, through `parse_linux_kfd_gfx_target("90010")`, through an `ip_discovery` fixture, and end to end via `normalize_therock_family` — plus out-of-range rejection. Every new assertion except the `major == 0` case fails against the old implementation.

**Not run locally:** no GPU on this host, so `rocm examine` was not exercised against real gfx90a hardware; the KFD and `ip_discovery` paths are covered by fixtures only.

Refs EAI-7346.

### DIFF
--- a/crates/rocm-core/src/lib.rs
+++ b/crates/rocm-core/src/lib.rs
@@ -3841,13 +3841,23 @@ fn detect_ip_discovery_gc_target(ip_discovery_dir: &Path) -> Option<String> {
 }
 
 #[cfg(any(target_os = "linux", test))]
-/// Build the LLVM gfx target for a GC (Graphics Core) IP version.
+/// Render gfx target *components* as an LLVM target token.
 ///
-/// The target is `gfx` + major in decimal + minor and revision as **single hex
-/// digits** — `gfx90a` is GC 9.0.**10** (`10 == 0xa`), and `gfx1250` is GC
-/// 12.5.0. Concatenating the components as decimal agrees with hex only while
-/// every component is below 10, so it silently produced `gfx9010` for gfx90a
-/// hardware (MI210/MI250), which then normalized to the wrong TheRock family.
+/// The token is `gfx` + major in decimal + minor and revision as **single hex
+/// digits** — the `a` in `gfx90a` is revision `10`. Concatenating the components
+/// as decimal agrees with hex only while every component is below 10, so it
+/// silently produced `gfx9010` for gfx90a hardware (MI210/MI250), which then
+/// normalized to the wrong TheRock family.
+///
+/// **The caller owns whether its numbers are target components at all.** KFD's
+/// `gfx_target_version` packs exactly this triple, so decoding it and calling
+/// here is sound. A GC (Graphics Core) IP version is a *different* quantity that
+/// merely coincides with the target on many parts: it does not on the GC 9.4.x
+/// line, where GC 9.4.0/9.4.1/9.4.2/9.4.3 are gfx906/gfx908/gfx90a/gfx942. So
+/// [`detect_ip_discovery_gc_target`] can still yield a wrong-but-plausible token
+/// for those parts — pre-existing, unchanged by the hex encoding, and not
+/// something this function can detect, since the components it receives are
+/// well-formed either way.
 ///
 /// A minor or revision that cannot be a single hex digit does not describe any
 /// gfx target, so it yields `None` rather than a fabricated token: the caller
@@ -6995,14 +7005,24 @@ Class Name:                Display
             Some("gfx1103".to_owned())
         );
 
-        // The same defect reached this fallback path, not just the KFD one.
+        // KNOWN WRONG, and pinned so the gap stays visible: on the GC 9.4.x line
+        // the GC IP version is not the LLVM target. Aldebaran (MI200/MI250)
+        // reports GC 9.4.2 here but is gfx90a, so this path yields MI300's target
+        // instead and resolves to the wrong TheRock family — the right wheel for
+        // this host is `gfx90a`. Pre-existing: the decimal encoding produced
+        // `gfx942` for 9/4/2 too, so the hex fix neither caused nor closes it.
+        // Fixing it needs a GC-IP-version → target table for GC 9.4.x
+        // (9.4.0/9.4.1/9.4.3 are gfx906/gfx908/gfx942), not a different encoding.
         fs::write(gc.join("major"), "9")?;
-        fs::write(gc.join("minor"), "0")?;
-        fs::write(gc.join("revision"), "10")?;
+        fs::write(gc.join("minor"), "4")?;
+        fs::write(gc.join("revision"), "2")?;
+        let aldebaran = detect_ip_discovery_gc_target(&root.join("ip_discovery"));
+        assert_eq!(aldebaran, Some("gfx942".to_owned()));
         assert_eq!(
-            detect_ip_discovery_gc_target(&root.join("ip_discovery")),
-            Some("gfx90a".to_owned())
+            normalize_therock_family(&aldebaran.expect("target")),
+            Some("gfx94X-dcgpu".to_owned())
         );
+
         fs::remove_dir_all(root).ok();
         Ok(())
     }
--- a/crates/rocm-core/src/lib.rs
+++ b/crates/rocm-core/src/lib.rs
@@ -3747,15 +3747,21 @@ fn parse_linux_kfd_gfx_target(value: &str) -> Option<String> {
     }
     match digits.len() {
         3 | 4 => Some(format!("gfx{digits}")),
+        // A 5/6-digit value is KFD's *packed* version (major·10000 + minor·100 +
+        // step), not a target name, so there is no `gfx{digits}` fallback here.
+        // Fabricating one from a version that failed to decode fed
+        // `gfx90010`-shaped tokens into `normalize_therock_family`, where the
+        // loose `gfx90` arm mapped them to a plausible-looking but wrong family.
+        // Yielding `None` instead lets detection try the next KFD node and then
+        // `ip_discovery`, and otherwise report the target as unknown — which is
+        // recoverable with `--family`, where a wrong family silently installs
+        // the wrong runtime wheel.
         5 | 6 => {
             let raw: u32 = digits.parse().ok()?;
             let major = raw / 10_000;
             let minor = (raw / 100) % 100;
             let revision = raw % 100;
-            if let Some(token) = gfx_target_from_gc_version(major, minor, revision) {
-                return Some(token);
-            }
-            Some(format!("gfx{digits}"))
+            gfx_target_from_gc_version(major, minor, revision)
         }
         _ => None,
     }
@@ -3835,11 +3841,26 @@ fn detect_ip_discovery_gc_target(ip_discovery_dir: &Path) -> Option<String> {
 }
 
 #[cfg(any(target_os = "linux", test))]
+/// Build the LLVM gfx target for a GC (Graphics Core) IP version.
+///
+/// The target is `gfx` + major in decimal + minor and revision as **single hex
+/// digits** — `gfx90a` is GC 9.0.**10** (`10 == 0xa`), and `gfx1250` is GC
+/// 12.5.0. Concatenating the components as decimal agrees with hex only while
+/// every component is below 10, so it silently produced `gfx9010` for gfx90a
+/// hardware (MI210/MI250), which then normalized to the wrong TheRock family.
+///
+/// A minor or revision that cannot be a single hex digit does not describe any
+/// gfx target, so it yields `None` rather than a fabricated token: the caller
+/// tries the next detection source and otherwise reports the target as unknown,
+/// which is recoverable with `--family`, where a fabricated one silently
+/// installs the wrong runtime wheel. `major` is only checked for zero — it is
+/// printed in decimal and has no single-digit bound (`gfx1030`, `gfx1250`), so
+/// an implausible major still concatenates into a token.
 fn gfx_target_from_gc_version(major: u32, minor: u32, revision: u32) -> Option<String> {
-    if major == 0 {
+    if major == 0 || minor > 0xf || revision > 0xf {
         return None;
     }
-    Some(format!("gfx{major}{minor}{revision}"))
+    Some(format!("gfx{major}{minor:x}{revision:x}"))
 }
 
 #[derive(Debug, Clone, Copy, Serialize, Deserialize, Eq, PartialEq)]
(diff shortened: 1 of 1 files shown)


rominf:
## Review

Approve — no blocking issues. The fix lands in the shared `gfx_target_from_gc_version` helper, so one change covers both the KFD and `ip_discovery` detection paths; I swept for a second decimal-concatenation decoder elsewhere and found none (the Windows/WSL path resolves via PCI-ID and marketing-name tables). Regression coverage at four levels for one bug, and the "what this does not close" section is the kind of disclosure that makes a fix trustworthy.

Verified locally: `cargo fmt --all -- --check` clean; `cargo clippy --workspace --all-targets -- -D warnings` clean; `cargo test -p rocm-core --lib` → 201 passed, 2 failed, both the pre-existing WSL2 `proc_lifecycle` process-tree failures in an untouched file. I also confirmed the claim that every new assertion except the `major == 0` case fails against the old implementation — it holds exactly as stated.

### 1. The stated justification for removing the `gfx{digits}` fallback describes a path the old code could not take

This is the one I'd ask you to address before merge. It's comment-only, but it's the reason given for removing code, and it goes into public history.

The new block comment above the `5 | 6 =>` arm of `parse_linux_kfd_gfx_target` says the fallback "fed `gfx90010`-shaped tokens into `normalize_therock_family`, where the loose `gfx90` arm mapped them to a plausible-looking but wrong family."

In the pre-change code the `Some(format!("gfx{digits}"))` fallback was reached only when `gfx_target_from_gc_version` returned `None`, and the old helper returned `None` only for `major == 0`. With `digits.len()` of 5 or 6 that requires a leading-zero string like `"009010"` — so for every realistic KFD `gfx_target_version` the fallback was effectively unreachable. The mechanism that actually produced the wrong family was the decimal concatenation in the helper: `(9,0,10)` → `"gfx9010"` → loose `starts_with("gfx90")` arm → `gfx90X-dcgpu`. Never `gfx90010`.

The test comment on `parse_linux_kfd_gfx_target("121600")` is wrong in a second way: under the old code that returned `Some("gfx12160")` from the concat, not `gfx{digits}`, and `normalize_therock_family("gfx12160")` matched no arm and already returned `None` — so it never normalized to a wrong family either. The assertion itself is still worth keeping (it pins the new bound end-to-end); only the comment needs correcting.

Worth noting this also slightly overstates the change: the fallback removal is inert except for leading-zero inputs. The real behavior change — previously-succeeding hosts now getting `None` — comes from the new `minor > 0xf || revision > 0xf` bound in the helper. Same correction applies to the commit body and the PR description sentence "passing it through was how out-of-range decodes still reached the loose family arm."

### 2. `gfx94c` is synthetic but reads as real

In `gc_version_encodes_components_above_nine_as_hex_digits`, `gfx94c` is not a shipped part and appears nowhere else in the tree — fine for exercising a second hex digit beyond `a` on a pure formatting function, but the adjacent `gfx90a` case is annotated as real MI210/MI250 hardware, so the unannotated one reads as equally real. A one-liner (`// not a shipped part — chosen only to exercise a hex digit other than 'a'`) avoids a reader mining tests for supported targets and picking up a nonexistent one.

### 3. Doc comment placement

The new `///` block sits below `#[cfg(any(target_os = "linux", test))]`. Legal, but the file has 12 instances of `///` immediately preceding `#[cfg(...)]` and, after this PR, exactly one of the reverse.

### 4. Two untested mechanisms this PR makes load-bearing (follow-up)

The safety argument for returning `None` rests on two behaviors nothing exercises:

- **Fall-through on `None`** — the KFD node loop and DRM card loop `continue` past a node that now yields `None`. Before this PR a malformed node almost always yielded something, so the `continue` was rarely the deciding factor; now it is. No test builds two nodes (one out-of-range, one valid) and asserts the valid one wins.
- **The new hard-error path** — `resolve_family`'s `bail!` is untested. I read the message and it is genuinely actionable (names `--family`, lists `known_therock_families()`, appends `detect_host_gpu_diagnostics()`), but nothing pins that.

Both look correct on inspection; non-blocking given how strong the coverage of the actual defect is.

### 5. Diagnostics don't surface the raw GC triple on failure (follow-up)

When detection now returns `None`, `detect_host_gpu_diagnostics` re-derives the target and prints `detected_gfx_target: <unknown>` — the `major`/`minor`/`revision` (or the packed `gfx_target_version`) that caused the rejection appear nowhere. Pre-existing shape, but this PR is what makes hitting that path expected, and per your own scope note the missing MI455X sysfs values are exactly what's needed next. Echoing the raw values would let affected users self-report that data.

### 6. `normalize_therock_family` still has no arm for `gfx90c` or `gfx1250` (pre-existing)

Now that the encoder emits real hex-bearing names, `gfx90c` (Renoir/Cezanne APUs, GC 9.0.12) resolves correctly and then falls through the loose `starts_with("gfx90")` arm to `gfx90X-dcgpu` — a datacenter family for an APU. Not a regression (old `"gfx9012"` hit the same arm), just newly visible. The loose catch-all is the same mechanism behind this bug and will keep producing plausible-but-wrong families for any gfx9 target without a specific arm. Correctly not bundled here — worth its own issue proposing explicit arms or a known-target table.

### Tradeoffs, surfaced rather than contested

- **`None` → hard error.** I traced the user-visible result: clean and actionable, not a panic. Right direction, and the PR flags it rather than burying it — but it is a user-visible regression in "it used to work" terms for affected hosts, so worth being a conscious maintainer call.
- **Bounding `minor`/`revision` but not `major`** is coherent per the LLVM convention and the doc comment says so. A typed `GcVersion` with `TryFrom`, or validation against a known-targets table so a syntactically-valid-but-nonexistent triple like `(12,5,1)` → `gfx1251` isn't reported as "detected" only to dead-end at family resolution, would be a stronger contract. Correctly deferred.
- **Scoping out the MI455X symptom** — confirmed independently that `gfx_target_from_gc_version(12, 5, 0)` yields `gfx1250` both before and after, so this is a genuine no-op there. Anyone tracking that work shouldn't read this merging as closing it.

### Release notes

Two user-visible changes worth calling out:

- On a host whose GC version decodes to an out-of-range `minor`/`revision`, `rocm install sdk` now fails with `unable to resolve a supported TheRock GPU family for this host` instead of silently installing a wrong-family runtime wheel. Recovery is `--family <FAMILY>`.
- gfx90a hosts (MI210 / MI250 / MI250X) on the KFD or `ip_discovery` paths now resolve to family `gfx90a` instead of `gfx90X-dcgpu`. This re-admits gfx90a to `VLLM_PREFERRED_THEROCK_FAMILIES`, so **engine selection changes** on those machines — relevant to anyone who has been working around the old behavior with an explicit `--family` or env override.

Nothing required outside the diff: no changelog in this repo, no `expectations.toml` update, no stale xfail rows, and no fixture under `tests/e2e-cucumber/**` encodes a fabricated token.

rominf:
## Review

The core fix is correct and well-evidenced. KFD's `gfx_target_version` really does pack the *LLVM target* (`90010` for gfx90a), so hex-decoding minor/revision is the right call, and the `gfx9010` → `gfx90X-dcgpu` mismapping you describe is real. I also traced the removed `gfx{digits}` fallback and found no regression: `detect_linux_kfd_gfx_target` (`crates/rocm-core/src/lib.rs:3565`) continues to the next node on `None`, `detect_linux_sysfs_gfx_target` (`:3556`) still chains to `ip_discovery`, and every sub-10 target (gfx900/906/908/1030/1100–1103/1150–1153/1200/1201/1250) is byte-identical under both encodings.

The requested changes below are to the **new doc comment and the new `ip_discovery` test**, not to the fix.

---

### 1. `crates/rocm-core/src/lib.rs:3846-3858` — the documented invariant does not hold for CDNA / Vega20

The doc comment states as a general law that the LLVM gfx target is `gfx` + major (decimal) + minor and revision (hex) *from the GC IP version*. That holds for RDNA, but on the CDNA and Vega20 line the GC IP version is not the LLVM target:

| GPU | GC IP version | This code emits | Real LLVM target | `normalize_therock_family` |
|---|---|---|---|---|
| Vega20 (MI50/60) | 9.4.0 | `gfx940` | `gfx906` | `gfx94X-dcgpu` ❌ |
| Arcturus (MI100) | 9.4.1 | `gfx941` | `gfx908` | `gfx94X-dcgpu` ❌ |
| Aldebaran (MI200/MI250) | **9.4.2** | `gfx942` | **`gfx90a`** | `gfx94X-dcgpu` ❌ |
| Aqua Vanjaram (MI300X) | 9.4.3 | `gfx943` | `gfx942` | `gfx94X-dcgpu` ✔ (coincidentally) |

Failure scenario: a host with `amdgpu` loaded but no readable KFD topology falls through to `detect_linux_drm_ip_discovery_gfx_target`. On an MI250 that reads GC 9.4.2, produces `gfx942`, and normalizes to `gfx94X-dcgpu` — the wrong TheRock wheel, and gfx90a drops out of `VLLM_PREFERRED_THEROCK_FAMILIES` (`:1303`), so engine selection shifts too. That is the same silent-wrong-family class this PR sets out to remove, still live on the secondary path.

**This is pre-existing, not a regression from this PR** — the old decimal code produced `gfx942` for 9/4/2 as well. What is new is that the comment now documents the false premise as fact. Suggest scoping it: the KFD `gfx_target_version` packs the LLVM target; the GC IP version read from `ip_discovery` does not always equal it, with a known divergence across GC 9.4.x.

Sourcing: this table is derived from **kernel source** (`drivers/gpu/drm/amd/amdgpu/amdgpu_discovery.c` and the `gfx_v9_4_*` IP blocks), **not from observed hardware sysfs dumps**. If you have access to an MI200/MI300 box, a real `ip_discovery/die/*/GC/*/{major,minor,revision}` read would settle it directly.

### 2. `crates/rocm-core/src/lib.rs:6998-7005` — the new `ip_discovery` fixture asserts a case that cannot occur

The fixture writes `major=9, minor=0, revision=10` and asserts `gfx90a`, under the comment "The same defect reached this fallback path, not just the KFD one." Real gfx90a hardware reports `9/4/2` there, so this exercises the shared helper a second time rather than the `ip_discovery` path's actual data — the helper is already covered at `:6919-6941`.

Failure scenario: a future change breaks GC 9.4.x handling on the `ip_discovery` path and this test stays green, because no test feeds it a value that line of hardware actually produces.

Options: drop the fixture as redundant, or switch it to `9/4/2` and assert the current (wrong) output with an explicit `// KNOWN WRONG: GC IP version != LLVM target on GC 9.4.x` marker, so the gap is visible rather than papered over.

### 3. The "What this does not close" reasoning — worth re-checking, given #1

The PR argues the reported MI455X symptom (gfx1250 read as gfx1210) cannot follow from this code, because GC 12.5.0 packs to `120500` → `gfx1250` under both encodings. That reasoning covers the KFD path only. If MI455X's GC IP version were 12.1.0 while its LLVM target is gfx1250 — the same CDNA-style divergence as #1 — then `detect_ip_discovery_gc_target` emits `gfx1210` under both the old and the new encoding, matching the reported symptom exactly.

To be clear, **this is a hypothesis, not a claim**: I could not obtain MI455X GC IP version data, so I have not verified that its GC version diverges from its target. It seems worth re-checking before the description tells readers the symptom is unrelated, since the mechanism now has a plausible candidate.

### 4. Minor, pre-existing and already acknowledged

`normalize_therock_family("gfx1250")` returns `None` (`:2828`), so even perfect detection of gfx1250 resolves to no family. Noting it only so it isn't lost — declining to guess the family here is the right call.

---

### Verified clean

- `cargo test -p rocm-core` for the `gfx` / `gc_version` / `kfd` filters: 13 / 3 / 4 passed.
- `cargo clippy --locked -p rocm-core --all-targets -- -D warnings`: clean (re-run after touching the source, since post-`--` args are not fingerprinted).
- `cargo fmt --check`: clean.
- `extract_first_gfx_token` handles the trailing hex letter correctly, and nothing elsewhere assumes gfx tokens are all-numeric after the `gfx` prefix.

### Worth calling out

Declining to claim the fix closes the reported issue, rather than asserting a fix that was never reproduced on hardware, is the right instinct — as is the "Behavior change" section stating plainly that affected hosts now hard-fail instead of silently installing the wrong wheel. Returning `None` over a fabricated token is also the right layer decision: it keeps the failure recoverable with `--family` instead of silent.

<sub>Automated review pass; findings verified against source where noted, with the unverified parts flagged as such.</sub>

juhovainio (approved):
@volen-silo approving this but please check Roman's latest comments before merging

volen-silo:
Thanks — verified all four against the code before acting. 1 and 2 held up and are fixed in `9eb2484`; 3 and 4 need no code change, reasoning below.

**1 — the invariant does not hold for CDNA/Vega20.** Correct, and the doc comment was the wrong place to assert it. `gfx_target_from_gc_version` is fed by two callers carrying *different* quantities: `parse_linux_kfd_gfx_target` passes the decoded KFD packed target, `detect_ip_discovery_gc_target` passes the raw GC IP version. Those coincide on RDNA and not on GC 9.4.x. The comment now states which caller owns the guarantee and names the divergence, instead of recording the false premise as fact.

Your table also has hardware confirmation, which I think settles the sourcing caveat for the MI300 row: the MI300X E2E lane in #150 reports `gfx943`. The real target is gfx942, and gfx943 is exactly GC 9.4.3 rendered by this path — so the divergence is observable in CI, not only in kernel source. It also implies that lane reaches `ip_discovery` rather than KFD, since KFD packs `90402` and would have decoded to gfx942.

I left the mapping itself alone — pre-existing, as you say, and closing it needs a GC-IP-version to target table for the 9.4.x line, which is a different change from the encoding.

**2 — the fixture asserts a case that cannot occur.** Agreed. Took your second option: it now writes 9/4/2, the value real Aldebaran reports, pins the wrong-but-plausible `gfx942`, and asserts the `gfx94X-dcgpu` family it resolves to, under a `KNOWN WRONG` marker recording what a real fix would take. Confirmed it is load-bearing by patching in a 9.4.2 → gfx90a mapping: the test fails, so a future fix trips the marker rather than passing silently.

**3 — the "what this does not close" reasoning.** Fair, though the description already stops short of a claim: it says the change "does not appear to close" the symptom, that this is reasoning about the code and not an observation of the hardware, and it names a GC-IP-version-differing-from-target override as the fix if that turns out to be the cause. Given #1 is now confirmed on real MI300 silicon, that candidate is stronger than when it was written. Happy to sharpen the wording if it still reads as too settled — either way it needs the MI455X sysfs values.

**4 — `normalize_therock_family("gfx1250")` returns `None`.** Confirmed, no arm matches. Left alone, for the reason you give.

One unverified aside, since it bears on how narrowly #1 should be scoped: Navi10 may be GC 10.1.10 against target gfx1010, which would make 9.4.x not the only divergent line. Family resolution is unaffected either way — both encodings land on `gfx101X-dgpu` — and I have no hardware to check it, so the new comment is worded as "coincides on many parts" rather than naming 9.4.x as the sole exception.

PR #153, opened 2026-07-30
fix(serve): stop hardcoding vLLM GPU memory utilization (EAI-7353)

rominf:
## Problem

Serving a 0.5B model on a large-VRAM machine reserved roughly 200 GB of VRAM.

The vLLM adapter always passed `--gpu-memory-utilization 0.80`, and vLLM reserves that fraction of **total** device VRAM for its KV cache regardless of how small the model is. There was no way for a user to override it.

## Changes

- **Stop overriding vLLM's default.** When the user does not ask for a value, no `--gpu-memory-utilization` argument is passed at all, so vLLM's own default applies. Chosen over hardcoding vLLM's current 0.9 so the value cannot drift out of sync with upstream.
- **Add `rocm serve --gpu-memory-utilization <FRACTION>`** so the value can be set explicitly. The flag is vLLM-scoped and follows the existing `--tool-call-parser` precedent: same warn-and-ignore behavior on other engines, routed through `EngineRecipeHint.required_flags`, so the engine protocol is unchanged.
- **Validate `0 < value <= 1`** with an error that names the flag.
- **Extract a pure `vllm_serve_args` helper** so the launch argv is unit-testable.
- **Update `docs/vllm.md`**: drop the now-stale "0.80 by default for display/WSL headroom" rationale, document the flag, and state that it is a fraction of total — not free — VRAM.
- **Update `docs/testing.md`**: point the vLLM TheRock acceptance runbook at the renamed test so the documented command runs.

## Why an engine-scoped flag

An engine-agnostic knob was considered and rejected: llama.cpp/lemonade has no fractional-VRAM-reservation concept, and multi-engine wrappers in the wider ecosystem (Xinference, LocalAI, RamaLama, KubeAI, Jan, llama-swap) all decline to unify this for the same reason. `gpu-memory-utilization` is also the near-universal spelling (vLLM, TGI's equivalent, Ray Serve, KServe, KubeAI, NIM).

## Behavior change

On machines where the old 0.80 was providing display or WSL headroom, the effective reservation now rises to vLLM's default. Passing `--gpu-memory-utilization 0.8` restores the previous behavior.

Related: the e2e free-VRAM floor in `tests/e2e-cucumber/tests/e2e/serving_steps.rs` caps at 90% of total, which now sits exactly at vLLM's default instead of having slack. That is a tight tolerance on small cards and may deserve a deliberate margin in a follow-up.

## Test plan

- 7 new unit tests plus one rewritten (`resolve_model_surfaces_conservative_vram_default` → `resolve_model_omits_gpu_memory_utilization_default`) across `engines/vllm` and `apps/rocm`: flag absent (no argument emitted), flag present (value reaches argv), and invalid input (`0`, `1.5`, `-0.2`, `abc`, `NaN`, `inf`, empty).
- Fails-before verified by temporarily restoring the hardcoded flag/value pair.
- No `expectations.toml` rows reference this ticket, so none needed narrowing.

Risk: low for the code path; medium for users who relied on the implicit 0.80 headroom, which the docs now cover.

### DIFF
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -21986,10 +21986,25 @@ install therock";
 
     #[test]
     fn gpu_memory_utilization_override_leaves_non_vllm_engines_untouched() {
+        // The override is vLLM-specific: other engines are never rewritten, with
+        // or without a recipe of their own.
         assert!(
             engine_recipe_with_gpu_memory_utilization_override("lemonade", None, Some(0.5))
                 .is_none()
         );
+        let existing = EngineRecipeHint {
+            contract_version: ENGINE_RECIPE_CONTRACT_VERSION.to_owned(),
+            engine: "lemonade".to_owned(),
+            required_flags: vec!["--some-flag".to_owned()],
+            ..EngineRecipeHint::default()
+        };
+        let hint = engine_recipe_with_gpu_memory_utilization_override(
+            "lemonade",
+            Some(existing.clone()),
+            Some(0.5),
+        )
+        .unwrap();
+        assert_eq!(hint.required_flags, existing.required_flags);
     }
 
     #[test]
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -4240,7 +4240,7 @@ fn parse_gpu_memory_utilization(value: Option<&str>) -> Result<Option<f64>> {
     if !parsed.is_finite() || parsed <= 0.0 || parsed > 1.0 {
         bail!(
             "--gpu-memory-utilization must be greater than 0 and at most 1 (a fraction of \
-             the GPU's TOTAL VRAM); got `{trimmed}`"
+             the GPU's TOTAL VRAM, e.g. 0.5); got `{trimmed}`"
         );
     }
     Ok(Some(parsed))
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -336,6 +336,12 @@ rocm serve qwen2.5-7b-instruct --verbose --device gpu_required")]
         /// and implies `--enable-auto-tool-choice`. Applies to vLLM only.
         #[arg(long, value_name = "NAME")]
         tool_call_parser: Option<String>,
+        /// Fraction of each GPU's TOTAL VRAM that vLLM may claim for weights and
+        /// KV cache (`0.0 < value <= 1.0`). Omitted by default, so vLLM applies
+        /// its own default. Lower it when a small model should not reserve most
+        /// of a large card. Applies to vLLM only.
+        #[arg(long, value_name = "FRACTION", allow_hyphen_values = true)]
+        gpu_memory_utilization: Option<String>,
         /// API key that clients must present to a public (non-loopback) endpoint.
         /// When binding a public interface and this is omitted, a strong key is
         /// generated automatically. Prefer the `ROCM_SERVE_API_KEY` environment
@@ -1623,6 +1629,7 @@ fn dispatch(cli: Cli) -> Result<()> {
             no_smoke_test,
             allow_public_bind,
             tool_call_parser,
+            gpu_memory_utilization,
             api_key,
         }) => serve(ServeArgs {
             model,
@@ -1639,6 +1646,7 @@ fn dispatch(cli: Cli) -> Result<()> {
             no_smoke_test,
             allow_public_bind,
             tool_call_parser,
+            gpu_memory_utilization,
             api_key,
         }),
         Some(Command::Comfyui { command }) => comfyui(command),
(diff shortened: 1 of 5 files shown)


juhovainio:
## Review: `fix(serve): stop hardcoding vLLM GPU memory utilization` (EAI-7353)

**Verification performed:** checked out the branch locally, confirmed the diff scope, ran the full affected test suites (`rocm`, `rocm-engine-vllm`), and ran `clippy` on both crates. Everything compiles, all 8 new/changed tests pass, clippy is clean.

### What it does
Removes rocm-cli's hardcoded `--gpu-memory-utilization 0.80` for vLLM (which was reserving that fraction of *total* VRAM regardless of model size). Now vLLM's own default applies unless the user passes the new `rocm serve --gpu-memory-utilization <FRACTION>`, threaded through the existing `EngineRecipeHint.required_flags` channel (same mechanism `--tool-call-parser` already uses), so no engine protocol change is needed.

### Code quality
- Clean, minimal diff that reuses an existing extension point rather than inventing a new one.
- `vllm_serve_args` extraction into a pure, unit-testable function is a solid improvement — argv construction was previously only exercisable via a live process spawn.
- Validation (`parse_gpu_memory_utilization`) happens before anything launches, giving a fast, actionable error instead of a buried vLLM argparse failure.
- The note-plumbing (`gpu_memory_utilization_note` stored without a `note:` prefix, added by each rendering path) is a little subtle but I confirmed both call sites (plan path and `serve_summary::render_summary`) apply the prefix exactly once — no duplication.

### Issues / things worth a second look
- **`allow_hyphen_values = true`** on the new flag is a real, self-acknowledged tradeoff: `--gpu-memory-utilization` with no value will swallow the *next* argument (e.g. `--gpu-memory-utilization --verbose` eats `--verbose` as the value, then fails with an error naming the flag but showing `--verbose` as the bad input). Documented in a comment — worth an explicit nod that this is accepted, not an oversight.
- **Behavior change is real and user-facing**: deployments relying on the old 0.80 ceiling now reserve whatever vLLM's own (typically higher) default is. Well documented in both the PR description and `docs/vllm.md`, with the exact flag to restore old behavior — good practice for a breaking-ish change.
- **Self-identified follow-up risk** (not a blocker): the e2e free-VRAM floor (`MAX_FREE_VRAM_FLOOR_MIB`) now sits right at vLLM's default utilization instead of having slack. Worth a tracking ticket if one doesn't exist.
- Minor: `format!("{value}")` for a whole-number fraction like `1.0` emits `"1"` rather than `"1.0"` — cosmetic only, Python's argparse float type accepts both.

### Test coverage
Thorough — 7 new tests + 1 rename, covering: flag absent (no default injected), explicit value reaching argv, replacement of a recipe-authored value while preserving other flags, non-vLLM engines left untouched, the full invalid-input matrix (`0`, `1.5`, `-0.2`, `abc`, empty, `NaN`, `inf`), and that the "ignored for this engine" note reaches the interactive summary path (not just the plan path).

### Security
No concerns — local CLI argument validation and `Command::arg` construction, no shell injection surface.

### Verdict
Solid, well-tested, well-documented change. No blocking issues found. The `allow_hyphen_values` edge case and the e2e VRAM-floor tolerance are worth author awareness but shouldn't hold up merging.

juhovainio (approved):

rominf:
Status update: this PR is rebased onto current `main` and 20 of 21 checks are green, including `windows-build-and-test`, which was the original failure here. That failure was the loopback-HTTP install download flake in `lifecycle-windows-http-install`; rebasing onto #174 resolved it, no code change was needed.

The one remaining red check, `E2E tests (GPU)`, is not a code failure — the self-hosted GPU runners are unhealthy. The job is assigned, runs zero steps, and dies in about one second with a runner-level error:

```
System.ArgumentNullException: Value cannot be null. (Parameter 'configuredSettings')
   at GitHub.Runner.Sdk.ArgUtil.NotNull(Object value, String name)
   at GitHub.Runner.Common.ConfigurationStore.GetSettings()
   at GitHub.Runner.Worker.JobRunner.RunAsync(...)
```

Evidence that this is environmental and repo-wide rather than specific to this PR:

- Both GPU runners are affected — `github-runner-0` and `github-runner-1` — with the identical error.
- It reproduces on other open PRs and on a merge-group run, not just this branch.
- Last successful `E2E tests (GPU)` was 2026-08-05 17:18 UTC; every attempt today fails this way.
- I re-ran the failed job twice; it fails identically each time, so it is a persistent runner-configuration problem, not a flake that re-running clears.

`E2E tests (GPU)` is a required check and the merge queue is configured ALLGREEN, so this blocks this PR and will eject anything enqueued until the runners are repaired. Someone with access to the runner hosts will need to re-configure or re-register them.

Not merging until that is fixed and the check is genuinely green.

PR #165, opened 2026-08-03
fix(install): stop leaking temp files when an atomic write fails

### LINKED ISSUE #158: [Issue]: Interrupted downloads leave orphaned partial files that are never cleaned up
### Problem Description

Failed downloads leave large partial files behind, and nothing ever removes them. On a machine that is already low on disk this compounds: each retry leaks another partial, so the disk fills faster the more you retry.

There are two separate defects.

**1. `write_file_atomically` leaks its temp file on write failure**

The temp file is named `path.tmp-<unix_ms>`, and the error paths return without removing it. Because the name embeds a millisecond timestamp, every failed attempt leaves a *distinct* orphan rather than overwriting the previous one.

- `apps/rocm/src/therock.rs:2187-2202`
- same defect in `apps/rocmd/src/lib.rs:1064-1074`

This is on the SDK tarball path (`apps/rocm/src/therock.rs:1062`), where the orphan is bounded only by free space.

**2. `download_file_to_path` writes straight to the final path**

`crates/rocm-core/src/lib.rs:115-133` does `File::create(destination)` then `io::copy` — no temp file, no rename, no cleanup on error. An interrupted transfer therefore leaves a truncated file *at the destination path*, where it is indistinguishable from a complete download.

That corrupt file then becomes a permanent cache entry: `engines/lemonade/src/lib.rs:979` accepts any existing file as valid (`if archive.is_file() { "Using cached ..." }`) with no size or hash check, so extraction fails on every subsequent run. `--reinstall` clears the runtime directory (lines 986-990) but does not re-download the archive, so there is no supported way out short of deleting the file by hand.

Related: the SDK tarball is read fully into memory before being written (`read_to_end` into a `Vec<u8>` at `apps/rocm/src/therock.rs:2119-2123`, passed to `write_file_atomically` at 2078-2088), so peak cost is the whole tarball in RAM plus a full copy on disk before the rename.

### Steps to Reproduce

Both reproduced with tests run against this tree.

Temp-file leak — fill a small filesystem, then call `write_file_atomically` twice:

```
attempt 0: ... No space left on device (os error 28)
  leftovers: [("artifact.tar.tmp-1785755921807", 67108864)]
attempt 1: ... No space left on device (os error 28)
  leftovers: [("artifact.tar.tmp-1785755921807", 67108864),
              ("artifact.tar.tmp-1785755921875", 0)]
```

Two attempts, two distinct orphans, the first holding all the space that was left.

Truncated download — serve `Content-Length: 1048576`, send 16 bytes, hang up:

```
download error: failed to write .../lemonade.tar.gz: response body closed before all bytes were read
left behind 16 bytes at .../lemonade.tar.gz
```

### Suggested Fix

- Download to a temp path and rename only on success; remove the temp on every error path.
- Stream to disk instead of buffering the whole body in memory.
- Validate a cached archive (size against `Content-Length`, or a checksum) before trusting it, so a truncated file is re-fetched rather than reused.

Minor: `with_extension("tmp-...")` mangles multi-dot names — `artifact.tar.gz` becomes `artifact.tar.tmp-<ms>`. Harmless for the rename itself, but worth knowing for any cleanup glob.

### Additional Information

Found while investigating unbounded disk growth. Related issues to follow for retention/GC and free-space preflight.

rominf:
## Summary

`write_file_atomically` leaves its scratch file behind when a write or rename
fails. The name embeds a millisecond timestamp, so each retry leaks a *distinct*
orphan instead of reusing one — on a full disk, retrying makes things worse.

- Remove the temp file on every failure path, in both copies of the helper
  (`apps/rocm/src/therock.rs`, `apps/rocmd/src/lib.rs`).
- Build the temp name from the whole file name, so a multi-extension artifact
  keeps its extensions: `sdk.tar.gz` yields `sdk.tar.gz.tmp-<id>` where
  `with_extension` dropped the `.gz`.

## Root cause

Reproduced by filling a small filesystem and calling the helper twice:

```
attempt 0: ... No space left on device (os error 28)
  leftovers: [("artifact.tar.tmp-1785755921807", 67108864)]
attempt 1: ... No space left on device (os error 28)
  leftovers: [("artifact.tar.tmp-1785755921807", 67108864),
              ("artifact.tar.tmp-1785755921875", 0)]
```

Two attempts, two orphans, the first holding all the space that was left.

## Scope

This PR originally also rewrote the download path. #198 has since landed
`download_file_streaming`, which covers that ground more thoroughly (sibling
`.part` file, resume, length and digest cross-checks), so those changes are
dropped in favour of it. What remains is the one defect #198 did not touch.

Note this does **not** close #158. Both in-call cleanup gaps are now fixed, but
a `.part` file orphaned by a hard crash or Ctrl-C still survives: the name is
keyed on the process id, so a later run never removes it, and nothing sweeps
for them. Leaving #158 open for that.

## Test plan

- `write_file_atomically_cleans_up_temp_when_the_rename_fails` (both crates) —
  portable, runs in CI. Fails before, passes after.
- `write_file_atomically_temp_name_preserves_multi_dot_file_names`,
  `temp_sibling_path_preserves_multi_dot_file_names`.
- `write_file_atomically_cleans_up_temp_on_write_failure` (both crates) —
  the original `/dev/shm` ENOSPC reproduction, `#[ignore]`d because it fills a
  shared tmpfs. Run with `cargo test -- --ignored`.

Risk: low. Two self-contained helpers; behaviour on the success path is
unchanged apart from the temp file's name.

Refs #158

### DIFF
--- a/.github/workflows/e2e-selfhosted.yml
+++ b/.github/workflows/e2e-selfhosted.yml
@@ -275,11 +275,12 @@ jobs:
           prewarm="$RUNNER_WORKSPACE/e2e-prewarm"
           export E2E_SHARED_RUNTIMES_DIR="$prewarm/data/runtimes"
 
-          # Build the rocm binary ONCE and reuse it for both the pre-warm and the
-          # suite (via ROCM_CLI_BINARY) so xtask does not rebuild. Honors
+          # Build the rocm and rocmd binaries ONCE and reuse them for both the
+          # pre-warm and suite so xtask does not rebuild. Honors
           # CARGO_TARGET_DIR set above.
-          cargo build --release -p rocm
+          cargo build --release -p rocm -p rocmd
           export ROCM_CLI_BINARY="$CARGO_TARGET_DIR/release/rocm"
+          export ROCM_CLI_ROCMD_BINARY="$CARGO_TARGET_DIR/release/rocmd"
 
           # Pre-warm the shared runtime ONCE, SERIALLY, before the suite — never
           # lazily inside a concurrent scenario (two multi-GiB installs racing the
@@ -442,10 +443,10 @@ jobs:
           prewarm="$RUNNER_WORKSPACE/e2e-prewarm"
           export E2E_SHARED_RUNTIMES_DIR="$prewarm/data/runtimes"
 
-          # Build the rocm binary once; reuse for pre-warm + suite so xtask doesn't
-          # rebuild.
-          cargo build --release -p rocm
+          # Build the rocm and rocmd binaries once; reuse them for pre-warm + suite.
+          cargo build --release -p rocm -p rocmd
           export ROCM_CLI_BINARY="$CARGO_TARGET_DIR/release/rocm"
+          export ROCM_CLI_ROCMD_BINARY="$CARGO_TARGET_DIR/release/rocmd"
 
           # Pre-warm once, serially, in place (no mv/symlink). Skipped once the tree
           # is populated (persists across runs on RUNNER_WORKSPACE).
@@ -593,13 +594,14 @@ jobs:
           $prewarm = "$env:RUNNER_WORKSPACE\e2e-prewarm"
           $env:E2E_SHARED_RUNTIMES_DIR = "$prewarm\data\runtimes"
 
-          # Build the rocm binary once; reuse for pre-warm + suite so xtask doesn't
-          # rebuild. This job does not set CARGO_TARGET_DIR, so the binary lands in
+          # Build the rocm and rocmd binaries once; reuse them for pre-warm + suite.
+          # This job does not set CARGO_TARGET_DIR, so the binaries land in
           # the default target\release (fall back to it when the env var is unset).
-          cargo build --release -p rocm
+          cargo build --release -p rocm -p rocmd
           if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
           $targetDir = if ($env:CARGO_TARGET_DIR) { $env:CARGO_TARGET_DIR } else { "target" }
           $env:ROCM_CLI_BINARY = "$targetDir\release\rocm.exe"
+          $env:ROCM_CLI_ROCMD_BINARY = "$targetDir\release\rocmd.exe"
 
           # Pre-warm once, in place (no move/symlink). Skipped once the tree is
           # populated (persists across runs on RUNNER_WORKSPACE).
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -384,9 +384,10 @@ jobs:
           prewarm="$RUNNER_WORKSPACE/e2e-prewarm"
           export E2E_SHARED_RUNTIMES_DIR="$prewarm/data/runtimes"
 
-          # Build once; reuse for pre-warm + suite so xtask doesn't rebuild.
-          cargo build --release -p rocm
+          # Build both binaries once; reuse them for pre-warm + suite.
+          cargo build --release -p rocm -p rocmd
           export ROCM_CLI_BINARY="$CARGO_TARGET_DIR/release/rocm"
+          export ROCM_CLI_ROCMD_BINARY="$CARGO_TARGET_DIR/release/rocmd"
 
           # Pre-warm the shared runtime once/serially before the suite; skipped
           # once populated (persists across runs). Install in place (no mv) so the
@@ -490,8 +491,9 @@ jobs:
           prewarm="$RUNNER_WORKSPACE/e2e-prewarm"
           export E2E_SHARED_RUNTIMES_DIR="$prewarm/data/runtimes"
 
-          cargo build --release -p rocm
+          cargo build --release -p rocm -p rocmd
           export ROCM_CLI_BINARY="$CARGO_TARGET_DIR/release/rocm"
+          export ROCM_CLI_ROCMD_BINARY="$CARGO_TARGET_DIR/release/rocmd"
 
           if [ ! -d "$E2E_SHARED_RUNTIMES_DIR/registry" ]; then
             echo "pre-warming shared runtime (first run on this runner)…"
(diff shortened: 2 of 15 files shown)


juhovainio (dismissed):
lgtm

rominf:
## Merge-queue ejection: root cause and fix

This PR was ejected from the merge queue when the blocking `E2E tests` job failed on `dash-managed-service-metrics` (merge-group run 31102672143, job 92621417125):

```
FAIL: 'dash-managed-service-metrics' was expected to pass on this host but FAILED — a regression.
Step panicked: TTFT metrics did not appear: timed out after 30s waiting for "50ms".
```

### Root cause

The captured screen in that log shows the dashboard still on **Home** — the `4` keypress that should have switched to Observe never took effect, so the step waited 30s for metrics on a view it never reached.

The two scenarios that open Observe launch the TUI and send the tab key with no assertion in between, unlike the demo-data journeys which assert the home view first. A key written into the pseudo-terminal before the dashboard is reading input can be consumed by whatever holds the terminal at that moment, and nothing retries it. The step then fails much later, in an assertion about a view the dashboard never left — which is exactly the failure signature above.

This is a latent synchronisation defect in the E2E step, not in this PR's production code: nothing in the download/atomic-write change is reachable from `rocm dash`. It surfaced here rather than in this PR's own checks because those ran against the older base.

### Fix

Send the tab key repeatedly until the Observe chip is actually marked active, so the step depends on the dashboard having acted on the key rather than on it having been ready when the key was written. The helper is restricted to idempotent keys (a tab jump, not a toggle).

### Reproduction

Deterministic, by pointing the harness at a wrapper that drains the terminal before exec'ing the real binary — i.e. a dashboard that is not yet reading when the key arrives:

```sh
cat > /tmp/wrapper/rocm <<'SH'
#!/bin/sh
stty raw -echo min 0 time 20; dd bs=1 count=16 of=/dev/null 2>/dev/null; stty sane
exec /path/to/target/release/rocm "$@"
SH
ROCM_CLI_BINARY=/tmp/wrapper/rocm cargo xtask e2e -- -n "Observe displays metrics"
```

Before: `1 scenario (1 failed)` with the CI symptom verbatim. After: `1 scenario (1 passed)`. The sibling scenario (`dash-loading-service-status`), which uses the same step, also passes under the wrapper.

Verified in addition: full `cargo xtask e2e` matches the base commit exactly (the two `diagnose-*` failures are pre-existing on this host, confirmed on `ec2bcb32` itself), `cargo fmt --all --check`, both clippy invocations, `cargo xtask manifest --check`, and `cargo test --workspace --all-targets` (only two pre-existing `proc_lifecycle` failures local to this host).

Note: the guard is exercised by these two scenarios on every run, but the key-loss condition itself is only reproducible with the wrapper above, so there is no automated test that would fail without it.

rominf:
## Re-scoped — please re-review, the existing approval predates this

Heads up that this PR now does substantially less than when it was approved. The
approval at 8c1ec0ce was for code that is largely no longer here, and this repo
does not dismiss stale reviews on push, so I have re-requested review rather than
let it merge on a review of code that no longer exists.

**Dropped: the download half.** #198 (c6b5bb54) landed `download_file_streaming`
on `main`, which implements the same idea more completely than this PR did —
sibling `.part` file, rename on success, plus resume via `Range`, `Content-Range`
offset validation, SHA-256 and `Content-Length` cross-checks, a free-space
preflight, and retry classification. Keeping this PR's version would have meant a
second, weaker download path, so it is gone in favour of main's.

**Kept: the `write_file_atomically` temp leak**, which #198 did not touch. Both
copies (`apps/rocm/src/therock.rs`, `apps/rocmd/src/lib.rs`) still returned early
on a write or rename failure without removing the scratch file, and the name
embeds a timestamp so each retry leaked a *distinct* orphan. #198 stopped routing
the SDK tarball through the `therock.rs` copy, but three callers remain, so both
copies are fixed here.

**Split out: the e2e TUI flake fix.** The Observe-tab keystroke retry was
unrelated scope creep and is moving to its own PR.

**`Fixes #158` is now `Refs #158`.** The in-call cleanup gaps are closed, but the
issue's actual complaint — orphans that are never cleaned up — survives. A `.part`
file left by a hard crash or Ctrl-C is keyed on the process id, so the next run's
cleanup targets a different path and never removes it, and nothing in the tree
sweeps for them. #158 should stay open; I have not touched it.

michaelroy-amd (commented):
- **Review focus:** Please review atomic publication semantics across Unix/Windows and the single-generation metadata cache; then confirm the E2E binary contract proves the user-visible cleanup path.
- **Problem / impact:** Failed atomic writes leaked unique temp files, so retries—especially under ENOSPC—consumed more space. Timestamp collisions could share a scratch path, Windows replacement could damage an existing destination, and separately published body/signature/metadata files could expose mixed cache generations.
- **State:** [#165](https://github.com/ROCm/rocm-cli/pull/165), head `32a861ec301aae944af926cba466cf85506356e6`; `OPEN`, non-draft, `MERGEABLE`, `REVIEW_REQUIRED` (`BLOCKED`); 15 files, +1,724/-175.
  1. `apps/rocm/src/therock.rs`: collision-safe temp reservation, platform publication, cleanup, and body/signature/metadata consolidation into one cache entry.
  2. `apps/rocmd/src/lib.rs`: matching atomic-write behavior for daemon-side marker/state publication.
  3. Windows-specific `windows-sys` replacement path versus Unix rename semantics: preserving the old destination when publication fails.
  4. `tests/e2e-cucumber/features/artifact_prefetch.feature` and `artifact_steps.rs`: failed marker publication leaves no temp and does not alter an occupied destination.
  5. `tests/e2e-cucumber/src/lib.rs`, lifecycle steps, `xtask/src/e2e.rs`: explicit `ROCM_CLI_ROCMD_BINARY` and the requirement that `rocm`/`rocmd` come from one directory.
  6. Self-hosted/nightly workflows plus `xtask/src/workflow_contract.rs`: every prebuilt hardware lane builds and exports both binaries.
- **Invariants to verify:** Temp creation uses exclusive reservation and bounded collision retries; every write/publish error removes only its own temp; a failed replacement preserves the previous destination; successful publication is atomic on both supported OSes; body, signature bytes, and metadata always come from one cache generation, including legacy-cache migration and 304 revalidation; lifecycle packaging cannot mix binaries from different directories.
- **Verification evidence:** **Required:** all required checks green in the same hosted categories listed above, including signature+DCO and CodeQL. The hosted E2E run covers the new artifact-prefetch scenario. **Advisory:** Strix Ubuntu and the self-hosted consolidated report pass. Red: [GPU](https://github.com/ROCm/rocm-cli/actions/runs/32063255588/job/95492259558) has one unexpected `serve-hf-checkpoint-inference` failure from the missing Lemonade runtime directory; [Strix Windows](https://github.com/ROCm/rocm-cli/actions/runs/32063255588/job/95492259638) has one unexpected `serve-hf-checkpoint-inference` failure because `http://127.0.0.1:11435/v1/models` did not serve `Qwen3-0.6B` within 600 seconds.
- **Recommended sequence:** atomic helper and OS publication paths → cache-generation model → failure/race tests → Gherkin scenario → harness/workflow binary propagation.

michaelroy-amd (approved):
Approved at `32a861ec301aae944af926cba466cf85506356e6`. The focused review above covers the changed contracts and invariants. All required checks pass; the current self-hosted advisory failures are documented there and are not merge-required.

PR #166, opened 2026-08-03
fix(install): delete the SDK archive once it is extracted

### LINKED ISSUE #162: [Issue]: SDK tarball is never deleted after extraction
### Problem Description

The SDK tarball is kept forever after it has been extracted, so each tarball install costs the compressed archive *plus* the unpacked tree — roughly double, for no benefit.

`install_tarball_runtime` downloads to `cache_dir/therock/<file_name>` (`apps/rocm/src/therock.rs:1029`), extracts it (`apps/rocm/src/therock.rs:1061-1062`), writes the manifest, and returns. Nothing removes the archive on that path or anywhere else. Because the runtime key includes the version, every installed version leaves its own archive behind.

This looks like an oversight rather than an intentional cache: the same download-extract-install pattern in `ensure_uv_binary` *does* clean up after itself, at `crates/rocm-core/src/uv.rs:179`. The manifest even records `tarball_file_name`, but nothing consumes it for cleanup.

### Steps to Reproduce

1. Install the SDK with the tarball format.
2. The extracted runtime appears under the data directory, and the full archive also remains under `cache_dir/therock/`.
3. Install another version. Both archives are still there.

### Suggested Fix

Delete the archive once extraction succeeds, mirroring `crates/rocm-core/src/uv.rs:179`.

If keeping it is intentional — to make a reinstall of the same version cheap — then it should be bounded and documented, and the file needs an integrity check before reuse, since a truncated archive is currently indistinguishable from a complete one (see the companion issue on partial downloads).

### Additional Information

Found while investigating unbounded disk growth.

rominf:
## Summary

Delete the downloaded SDK archive once it has been extracted.

## Root cause

A tarball SDK install downloaded the archive into the cache directory, extracted it, and left the archive there permanently. Only the extracted tree is used afterwards, so every installed version cost roughly twice the disk it needed to. Because the runtime key includes the version, each installed version kept its own archive.

This looks like an oversight rather than a deliberate cache: `ensure_uv_binary` already removes its own download after unpacking, on exactly the same download-extract-install pattern. The manifest even records `tarball_file_name`, but nothing ever consumed it for cleanup.

## Technical decisions

Removal is best-effort. The install has already succeeded by the time the archive is redundant, so a cleanup failure is reported rather than raised — failing an otherwise successful install because a file could not be unlinked would be worse than leaving the file.

The work is in a small `extract_tarball_and_discard_archive` helper so the behaviour is unit-testable without performing a real multi-gigabyte download.

## Tests

`extracting_the_sdk_archive_removes_it` builds a small tar fixture, extracts it through the helper, and asserts both that the contents landed and that the archive is gone. It skips cleanly if `tar` is unavailable.

Fixes #162

### DIFF
--- a/apps/rocm/src/therock.rs
+++ b/apps/rocm/src/therock.rs
@@ -1060,7 +1060,7 @@ fn install_tarball_runtime(
     }
 
     download_file(&artifact.url, &cache_path)?;
-    extract_tarball(&cache_path, &install_root)?;
+    extract_tarball_and_discard_archive(&cache_path, &install_root)?;
 
     let manifest = InstalledRuntimeManifest {
         runtime_key: runtime_key.clone(),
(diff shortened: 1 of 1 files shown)


juhovainio (approved):
lgtm

Issue #168, opened 2026-08-03
[Issue]: Process-tree tests count a zombie as running, failing on hosts that reap slowly


rominf:
### Problem Description

Two process-lifecycle tests fail on some hosts because they check whether a process is still alive with a bare PID check, which cannot distinguish a live process from a terminated-but-unreaped one (a zombie).

- `proc_lifecycle::tests::tree_stop_waits_for_descendants`
- `proc_lifecycle::tests::tree_forced_kill_reaches_sigterm_ignoring_descendant`

Both assert on `crate::process_is_running(grandchild)`, which is `kill(pid, 0)`. That call succeeds for a zombie, so a descendant that was terminated correctly still reads as running until the host gets round to reaping it. Whether the test passes therefore depends on reaping latency, which varies by environment — it passes on the CI runners and fails reliably under WSL2.

**The product code is correct; only the tests are wrong.** `identity_state` deliberately treats a zombie as gone via `process_has_exited` (`crates/rocm-core/src/proc_lifecycle.rs:325-332`, matching state `Z`), and `terminate_verified` uses that logic throughout. The tests bypass it and use the weaker check instead.

### Steps to Reproduce

Run `cargo test -p rocm-core proc_lifecycle` on a host that reaps orphans slowly (WSL2 reproduces it every time):

```
thread 'proc_lifecycle::tests::tree_stop_waits_for_descendants' panicked at
crates/rocm-core/src/proc_lifecycle.rs:546:9:
Tree stop must terminate the descendant, not just the root
```

Instrumenting the same scenario shows the termination worked and only the assertion is wrong:

```
PROBE root=67962 grandchild=67963 tree=[67962, 67963]
PROBE after_term grandchild_running=true
PROBE grandchild_proc_state=Z
PROBE grandchild_identity_state=Gone
```

The tree was enumerated correctly, both processes were signalled, the descendant is a zombie (`Z`), and the product's own `identity_state` correctly reports `Gone`. Only the bare PID check disagrees.

### Impact

Beyond the two failing tests, this blocks the `pre-push` hook (which runs the test suite) for anyone developing on an affected host, so pushes require `--no-verify` — which also skips the signing and sign-off checks that hook enforces.

### Suggested Fix

Assert through the same semantics the product uses: capture a `ProcessIdentity` for the descendant before the stop and check `is_exited(identity_state(&grandchild))` afterwards. A regression test pinning the zombie semantics would stop the weaker check from creeping back in.

PR #170, opened 2026-08-03
fix(uv): colocate the uv cache with the ROCm CLI data dir

### LINKED ISSUE #160: [Issue]: uv cache is not colocated with the data dir, so environments are full copies instead of hardlinks
### Problem Description

ROCm CLI never tells `uv` where to put its cache, so the cache stays at `$HOME/.cache/uv` while managed environments are created under the ROCm CLI data directory. When those two paths are on different filesystems, `uv` cannot hardlink and silently falls back to copying every file — so each environment carries its own full copy of the ROCm SDK and the torch stack instead of sharing one cached copy.

This is not an exotic configuration. It happens whenever `$HOME/.cache` is a separate mount from the data directory — a split `/home`, a container overlay, or a user who moved the data directory to a larger disk with `ROCM_CLI_DATA_DIR` (`crates/rocm-core/src/lib.rs:810`) or `--prefix`. The workaround for running out of space therefore *multiplies* the space used.

`UV_CACHE_DIR` and `--link-mode` are not set anywhere in `apps/` or `engines/`; the only place either appears is the end-to-end test harness (`tests/e2e-cucumber/tests/e2e.rs:214`).

### Steps to Reproduce

Measured with uv 0.9.30 on a host where `$HOME/.cache` is a separate ext4 mount from the root overlay, installing a small package and comparing inodes.

Environment on the same filesystem as the cache — true hardlink:

```
inode=1740514 links=2 dev=830  .../venv/lib/python3.13/site-packages/six.py
inode=1740514 links=2 dev=830  .../.cache/uv/archive-v0/.../six.py
```

Environment on a different filesystem — full copy:

```
warning: Failed to hardlink files; falling back to full copy. This may lead to
         degraded performance. If the cache and target directories are on
         different filesystems, hardlinking may not be supported.

inode=1740556 links=1 dev=66   .../venv/lib/python3.13/site-packages/six.py
inode=1740514 links=2 dev=830  .../.cache/uv/archive-v0/.../six.py
```

Distinct inode, link count 1, plain regular file — a copy, not a reflink or symlink. Setting `UV_LINK_MODE=hardlink` explicitly does not change this: the request is downgraded to a copy and the command still exits 0.

`uv`'s own default is `hardlink` on Linux and Windows and `clone` on macOS, so this only degrades when the filesystems differ.

### Why the warning does not save us

`uv` does warn, but two of the three install paths discard it:

- `apps/rocm/src/therock.rs:2635-2639` inherits stderr, so the warning is visible.
- `apps/rocm/src/comfyui.rs:1422-1423` pipes stderr into a log file.
- `engines/vllm/src/lib.rs:940-943` captures output with `.output()` and drops stderr entirely when the command succeeds.

### Suggested Fix

Point `uv` at a cache inside the ROCm CLI data directory — for example alongside the existing helpers in `crates/rocm-core/src/runtime.rs` — and set it in `uv_command_env()` (`crates/rocm-core/src/uv.rs:63`) so every `uv` invocation inherits it. The cache then moves with the data directory and hardlinking keeps working wherever the user puts it, and the cache also comes under the reach of `rocm uninstall`, which today does not touch `$HOME/.cache/uv`.

Tradeoff worth a maintainer opinion: a ROCm-CLI-owned cache is no longer shared with other `uv` projects on the machine, so unrelated work loses some cache reuse. The alternative — leaving the cache shared and only warning — keeps that sharing but leaves the duplication in place.

### Additional Information

The mechanism above is verified by inode evidence. The size penalty at realistic wheel sizes was deliberately not measured (it would require downloading the full torch and ROCm SDK stack), so the magnitude is inferred from package sizes rather than observed.

rominf:
## Summary

Point `uv` at a cache inside the ROCm CLI data directory, so the cache sits on the same filesystem as the environments it populates and hardlinking keeps working.

## Root cause

`uv_command_env()` set only `UV_HTTP_TIMEOUT`, never `UV_CACHE_DIR`, so the cache stayed at `$HOME/.cache/uv` while managed environments were created under the data directory. When those are on different filesystems, `uv` cannot hardlink and falls back to copying every file — each environment then carries its own full copy of the ROCm SDK and the torch stack.

This is not an exotic setup. It happens with a split `/home`, a container overlay, or a user who moved the data directory to a larger disk with `ROCM_CLI_DATA_DIR` — so the workaround for running out of space *multiplied* the space used.

Measured with uv 0.9.30 on a host where `$HOME/.cache` is a separate mount, installing a small package and comparing inodes:

```
same filesystem  -> inode=1740514 links=2 dev=830   (hardlink into the cache)
different fs     -> inode=1740556 links=1 dev=66    (full copy)
```

A distinct inode with link count 1 — a copy, not a reflink or symlink. Setting `UV_LINK_MODE=hardlink` explicitly does not help: the request is downgraded and the command still exits 0.

`uv` does print a fallback warning, but two of the three install paths discard it — ComfyUI pipes stderr to a log file, and the vLLM install captures output with `.output()` and drops stderr on success.

## Technical decisions

**Keyed off `data_dir`, not `cache_dir`.** A cache would naturally live in the cache directory, but `ROCM_CLI_CACHE_DIR` can point at a *different* filesystem from the environments, which would reintroduce the exact bug. Colocation with the environments is the property that matters here.

**The escape hatch is `ROCM_CLI_UV_CACHE_DIR`**, matching the sibling knobs in the same module (`ROCM_CLI_UV_BINARY`, `ROCM_CLI_UV_VERSION`, `ROCM_CLI_UV_TIMEOUT_SECS`). A bare `UV_CACHE_DIR` exported for unrelated Python work would otherwise silently opt a user out of the fix with no signal. An ambient `UV_CACHE_DIR` is still honored — the e2e harness sets one to share a cache across scenarios — but it ranks below the namespaced variable, and `UvCacheSource` lets callers tell managed / override / inherited apart.

**A one-shot notice, not a migration.** When a pre-colocation `$HOME/.cache/uv` is still on disk, the CLI prints one notice and records that it did, following `maybe_migrate_legacy_dashboard_config`. Nothing is migrated or deleted: the cache is content-addressed and may be shared with unrelated `uv` projects, so removing it is the user's call.

Threading `&AppPaths` through `uv_command_env` touched the call sites in `therock.rs`, `comfyui.rs`, and the vLLM engine; all of them already had `paths` in scope, so no new `AppPaths::discover()` calls were introduced.

## Non-goal: `--prefix` installs — #187

**This colocates with the data directory only.** `--prefix` relocates `install_root` for a single SDK install and leaves `paths.data_dir` — and therefore the uv cache — untouched. A `--prefix` on a different filesystem from the data directory still copies rather than hardlinks, and that case is *not* fixed here.

It is deliberately out of scope because the obvious fix does not work. Threading the target root through and using `<install_root>/uv-cache` breaks every wheel install: `uv` materializes `UV_CACHE_DIR` on *every* invocation including `uv venv`, and `uv venv` refuses to create an environment in a directory that already exists. Verified:

```
$ UV_CACHE_DIR=$P/fresh/uv-cache uv venv --python 3.12 $P/fresh
Creating virtual environment at: /tmp/.../fresh
error: Failed to create virtual environment
  Caused by: A directory already exists at: /tmp/.../fresh

$ ls $P/fresh
uv-cache          <-- uv created its cache there, which then blocked venv creation
```

Since `install_root` *is* the venv directory, the cache cannot live inside it. (This is also why the existing `<install-root>/pip-cache` convention in `docs/testing.md` works for pip but not for uv: pip creates its cache during downloads, after the venv exists.) Choosing a different scheme is a design decision for maintainers, so it is tracked in **#187** with the evidence and four options.

In the meantime the gap is documented in `docs/manual-testing.md`, pinned by `uv_cache_does_not_follow_a_prefix_install_root`, and `ROCM_CLI_UV_CACHE_DIR` is a manual workaround.

The related pre-existing `Downloads/cache:` message — which reports `<install_root>/pip-cache` although nothing sets `PIP_CACHE_DIR` — is folded into #187, since whatever scheme resolves `--prefix` decides what that message should say.

## Consequences worth noting

- `rocm uninstall` now covers the uv cache as a side effect, since it removes the data directory. That reclaims more space but makes uninstall slower. The `--keep-data` / `--keep-cache` help text has been corrected to say so: `--keep-cache` does *not* cover the uv cache.
- `MANIFEST.md` now discloses the wheel cache and its size, not just the uv binary.
- The cache is no longer shared with other `uv` projects on the machine. Users who want the old sharing can set `ROCM_CLI_UV_CACHE_DIR`.
- Existing installs re-download once.

## Tests

Cache resolution is a pure function over the two override variables, exercised for unset, empty, whitespace-only, trimmed, inherited, and namespaced-override inputs, plus precedence between the two. The relocated-data-dir test goes through `AppPaths::with_managed_root` rather than two hand-built `AppPaths`. `uv_cache_does_not_follow_a_prefix_install_root` pins the #187 gap so it cannot regress silently in either direction.

They exercise the pure helper rather than mutating the process environment, because the crate denies `unsafe_code` and `std::env::set_var` is `unsafe` in this edition. That also keeps them parallel-safe.

## CI note

`cargo test --workspace --all-targets --no-fail-fast` on this branch shows two failures in `proc_lifecycle` that are unrelated to it — see #168, fixed by #169. That file is not touched by this PR, and those are the only two failures in the workspace. `cargo fmt --all --check` and `cargo clippy --locked --workspace --all-targets -- -D warnings` are clean.

Fixes #160

### DIFF
--- a/MANIFEST.md
+++ b/MANIFEST.md
@@ -672,6 +672,15 @@ official uv GitHub releases at `https://github.com/astral-sh/uv/releases/`.
 The binary is cached in the rocm-cli managed data directory and reused for
 subsequent operations. The version may be pinned via `ROCM_CLI_UV_VERSION`.
 
+`uv`'s own content-addressed package cache is also kept in the managed data
+directory (at `<data-dir>/uv-cache`), so that it shares a filesystem with the
+environments `uv` populates and packages can be hardlinked into them instead of
+copied. This cache holds every wheel `uv` downloads — the ROCm SDK and the
+torch stack included — so it is typically the largest directory rocm-cli
+manages, on the order of several GB per SDK version installed. It is removed by
+`rocm uninstall` unless `--keep-data` is passed, and its location can be
+overridden with `ROCM_CLI_UV_CACHE_DIR`.
+
 ### Lemonade Embeddable Runtime
 
 When `rocm engines install lemonade` is run, the CLI downloads a prebuilt
--- a/apps/rocm/src/comfyui.rs
+++ b/apps/rocm/src/comfyui.rs
@@ -348,6 +348,7 @@ pub(crate) fn install(
         let uv = ensure_uv_binary(paths)
             .context("failed to acquire uv binary for ComfyUI dependency install")?;
         run_uv_logged_command(
+            paths,
             &uv,
             uv_install_args(&runtime.python, &packages),
             Some(&runtime_env),
@@ -1400,6 +1401,7 @@ fn uv_install_args(venv_python: &Path, packages: &[String]) -> Vec<String> {
 }
 
 fn run_uv_logged_command(
+    paths: &AppPaths,
     uv: &Path,
     args: Vec<String>,
     runtime_env: Option<&ComfyUiRuntimeEnvironment>,
@@ -1421,7 +1423,7 @@ fn run_uv_logged_command(
         .stdin(Stdio::null())
         .stdout(Stdio::piped())
         .stderr(Stdio::piped());
-    for (key, value) in uv_command_env() {
+    for (key, value) in uv_command_env(paths) {
         command.env(key, value);
     }
     if let Some(runtime_env) = runtime_env {
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -42,7 +42,7 @@ use rocm_core::{
     read_tcp_stream_to_string, resolve_builtin_model_recipe, resolve_model_recipe,
     runtime_install_root_is_protected, runtime_path_is_same_or_inside,
     runtime_python_activation_hint, runtime_python_env_bin_dir, runtime_python_executable_in_env,
-    shell_command_for_host, write_all_tcp_stream,
+    shell_command_for_host, uv_cache_source, write_all_tcp_stream,
 };
 use rocm_engine_protocol::{
     DEFAULT_LOG_TAIL_LINES, DetectRequest, DetectResponse, DevicePolicy,
@@ -435,10 +435,12 @@ rocm logs --search error timeout")]
         /// Keep saved settings.
         #[arg(long)]
         keep_config: bool,
-        /// Keep app data such as logs, services, and engines.
+        /// Keep app data such as logs, services, engines, and the uv package cache
+        /// (often the largest directory rocm-cli manages).
         #[arg(long)]
         keep_data: bool,
-        /// Keep caches.
+        /// Keep caches under the cache directory. Does not cover the uv package cache,
+        /// which lives under the data directory; use --keep-data for that.
         #[arg(long)]
         keep_cache: bool,
         /// Allow removing development binaries inside the current build tree.
@@ -938,6 +940,7 @@ fn main() -> Result<()> {
         .and_then(|paths| logging::init(&paths));
 
     maybe_migrate_legacy_dashboard_config();
+    maybe_notice_legacy_uv_cache();
 
     let raw_args: Vec<String> = std::env::args().skip(1).collect();
     if raw_args.is_empty() {
@@ -966,6 +969,54 @@ fn main() -> Result<()> {
     }
 
     dispatch(Cli::parse())
+}
+
+/// Legacy `uv` cache location, used before the cache was colocated with the managed
+/// data directory. Kept relative so the check works on every platform's home dir.
+const LEGACY_UV_CACHE_RELATIVE: [&str; 2] = [".cache", "uv"];
+
+/// One-shot notice that a pre-colocation `uv` cache is still occupying space at the
+/// default `uv` location. Nothing is migrated or deleted: the cache is
+/// content-addressed and may be shared with unrelated `uv` projects on the machine, so
+/// removing it is the user's call. Silent when the managed cache does not exist yet
+/// (nothing has moved) or when an override is in effect.
+fn maybe_notice_legacy_uv_cache() {
+    let Ok(paths) = AppPaths::discover() else {
+        return;
+    };
+    let cache = uv_cache_source(&paths);
+    if cache.is_override() {
+        return;
+    }
+    // Only worth mentioning once the managed cache is actually in use; otherwise the
+    // legacy directory is simply the cache still being used by other tools.
+    if !cache.path().is_dir() {
+        return;
+    }
+    let Some(home) = std::env::var_os("HOME")
+        .or_else(|| std::env::var_os("USERPROFILE"))
+        .map(PathBuf::from)
+    else {
+        return;
+    };
+    let legacy = LEGACY_UV_CACHE_RELATIVE
+        .iter()
+        .fold(home, |dir, part| dir.join(part));
+    if !legacy.is_dir() {
+        return;
+    }
+    // One-shot: a standing reminder on every invocation would be noise, and the user may
+    // reasonably decide to keep the legacy cache for other uv projects.
+    let marker = paths.data_dir.join(".legacy-uv-cache-notice");
+    if marker.exists() {
+        return;
+    }
+    eprintln!(
+        "rocm: the uv cache now lives at {}; the previous cache at {} is no longer used by rocm-cli and can be removed if no other uv project needs it",
+        cache.path().display(),
+        legacy.display()
+    );
+    let _ = fs::write(&marker, b"");
 }
 
 /// One-shot, best-effort migration of a legacy rocm-dash `config.toml` into the
--- a/apps/rocm/src/therock.rs
+++ b/apps/rocm/src/therock.rs
@@ -14,7 +14,9 @@ use rocm_core::{
     uv_venv_args, verify_rsa_pkcs1_sha256_signature,
 };
 #[cfg(test)]
-use rocm_core::{generate_rsa_signing_keypair, sign_rsa_pkcs1_sha256_signature};
+use rocm_core::{
+    generate_rsa_signing_keypair, managed_uv_cache_dir, sign_rsa_pkcs1_sha256_signature,
+};
 use serde::{Deserialize, Serialize};
 use std::cmp::Ordering;
 use std::fmt::Write as _;
@@ -3606,12 +3608,32 @@ mod tests {
     }
 
     #[test]
-    fn managed_uv_cache_defaults_inside_generated_runtime_folder() {
+    fn managed_uv_cache_sits_under_the_data_dir_for_generated_runtime_folders() {
         let (_root, paths) = test_paths("managed-uv-cache");
         let runtime_key = "release-wheel-gfx120x-all-7-14-0";
         let install_root = managed_runtime_root(&paths, "wheel", runtime_key);
-        // uv caches live beside the venv; verify the wheel root path structure
         assert!(install_root.starts_with(&paths.data_dir));
+        // Without --prefix the generated runtime folder is itself under the data dir, so
+        // the uv cache shares a filesystem with the environment it populates.
+        assert!(managed_uv_cache_dir(&paths.data_dir).starts_with(&paths.data_dir));
+    }
+
+    #[test]
+    fn uv_cache_does_not_follow_a_prefix_install_root() {
+        // Documents a known gap rather than an intended behavior: `--prefix` relocates
+        // install_root only, while the uv cache stays keyed off the data dir. When the two
+        // land on different filesystems uv falls back to copying. Tracked separately; see
+        // the `--prefix` non-goal on the PR that introduced the colocation.
+        let (_root, paths) = test_paths("prefix-uv-cache");
+        let prefix_root = PathBuf::from("/mnt/elsewhere/envs/my-env");
+        let cache = managed_uv_cache_dir(&paths.data_dir);
+
+        assert!(
+            !cache.starts_with(&prefix_root),
+            "cache {} unexpectedly followed the --prefix root",
+            cache.display()
+        );
+        assert!(cache.starts_with(&paths.data_dir));
     }
 
     #[test]
--- a/apps/rocm/src/therock.rs
+++ b/apps/rocm/src/therock.rs
@@ -913,7 +913,7 @@ fn install_wheel_runtime(
         "Creating Python environment at {}.",
         install_root.display()
     ));
-    ensure_uv_venv(&uv, &python_launcher.executable, &install_root)?;
+    ensure_uv_venv(paths, &uv, &python_launcher.executable, &install_root)?;
     let env_python = venv_python_path(&install_root);
 
     progress_line(format!(
@@ -928,6 +928,7 @@ fn install_wheel_runtime(
     }
     install_args.extend(therock_pip_package_specs(&resolution.package_versions));
     run_uv_progress_command(
+        paths,
         &uv,
         install_args
             .iter()
@@ -2326,7 +2327,12 @@ fn extract_tarball_and_discard_archive(archive_path: &Path, target_dir: &Path) -
     Ok(())
 }
 
-fn ensure_uv_venv(uv: &Path, python_launcher: &Path, install_root: &Path) -> Result<()> {
+fn ensure_uv_venv(
+    paths: &AppPaths,
+    uv: &Path,
+    python_launcher: &Path,
+    install_root: &Path,
+) -> Result<()> {
     let env_python = venv_python_path(install_root);
     if env_python.is_file() {
         if run_command(
@@ -2353,7 +2359,7 @@ fn ensure_uv_venv(uv: &Path, python_launcher: &Path, install_root: &Path) -> Res
             .map(String::as_str)
             .collect::<Vec<_>>()
             .as_slice(),
-        &uv_command_env(),
+        &uv_command_env(paths),
         "create managed TheRock runtime virtual environment",
     )?;
     if !env_python.is_file() {
@@ -2738,10 +2744,15 @@ fn run_command_with_env(
     bail!("{context_text}: {detail}")
 }
 
-fn run_uv_progress_command(uv: &Path, args: &[&str], context_text: &str) -> Result<()> {
+fn run_uv_progress_command(
+    paths: &AppPaths,
+    uv: &Path,
+    args: &[&str],
+    context_text: &str,
+) -> Result<()> {
     let mut command = Command::new(uv);
     command.args(args);
-    for (key, value) in &uv_command_env() {
+    for (key, value) in &uv_command_env(paths) {
         command.env(key, value);
     }
     let status = command
@@ -2845,7 +2856,7 @@ fn ensure_managed_python(paths: &AppPaths) -> Result<PythonLauncher> {
     progress_line(format!("Installing Python {version} via uv..."));
     let status = Command::new(&uv)
         .args(["python", "install", &version])
-        .envs(uv_command_env())
+        .envs(uv_command_env(paths))
         .stdin(Stdio::null())
         .stdout(Stdio::inherit())
         .stderr(Stdio::inherit())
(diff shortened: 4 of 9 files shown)


rominf:
The `windows-build-and-test` failure here is the known `lifecycle-windows-http-install` flake tracked in #173, not this change.

Failure signature on every attempt, unchanged:

```
rocm-cli installer: failed to download
http://127.0.0.1:<port>/rocm-cli-windows-amd64.zip:
An error occurred while sending the request.
FAIL: 'lifecycle-windows-http-install' was expected to pass on this host but FAILED
```

That is `install.ps1` fetching from the test's local HTTP server, before any `rocm` code executes. This branch changes only Rust — `uv.rs`, `runtime.rs`, `therock.rs`, `comfyui.rs`, and the vLLM engine — and touches neither the installer nor packaging, so it cannot reach that code path.

The same scenario failed and then passed on re-run, with no code change, on #166, #169 and #172 today. The per-PR failure count rose through the afternoon (1, then 2, then 3, then 5 here), which looks like a worsening runner or network condition rather than anything branch-specific.

I have re-run this job five times and stopped there rather than keep consuming runner time. Could a maintainer re-run it, or advise if the Windows lane is known to be unhealthy right now? Everything else on this PR is green, and `fmt`, `clippy -D warnings`, and the test suite pass locally.

volen-silo (dismissed):
Reviewed at `e10c649`. The core mechanism is right and the execution is careful — root cause identified at the correct layer, measured evidence rather than assertion, every uv spawn site found and threaded (I checked all six independently), and a test seam that works around `#![deny(unsafe_code)]` making `set_var` unusable. The PR body also self-discloses its known gaps, which made this much faster to review. Two blocking items, one of them a two-minute edit.

## Blocking

### 1. `--prefix` installs are not covered — and that's the mainline flow

`crates/rocm-core/src/uv.rs:78` derives the cache from `paths.data_dir` unconditionally. But at `apps/rocm/src/therock.rs:812`:

```rust
let install_root = prefix.unwrap_or_else(|| managed_runtime_root(paths, "wheel", &runtime_key));
```

With `--prefix`, `install_root` is a raw user path and `data_dir` is untouched. The two are never cross-checked, so the same-filesystem property this PR exists to establish does not hold for prefixed installs.

This isn't an edge case:

- The built-in assistant's system prompt (`apps/rocm/src/main.rs:8439`) instructs the model to *always* obtain an install folder and pass `--prefix`, and forbids proposing an install command without one. Every assistant-driven SDK install goes down this path.
- It cascades to ComfyUI — `apps/rocm/src/comfyui.rs:666` roots the app at `install_root.join("apps").join(APP_ID)`, so the uv install at `comfyui.rs:1426` inherits the prefixed root. Two of the three uv install paths are affected.
- One combination is a genuine regression. `ROCM_CLI_DATA_DIR=/mnt/big` plus `--prefix $HOME/envs`: before this PR the cache was `~/.cache/uv`, same filesystem as `$HOME/envs`, hardlinks worked. After, the cache is `/mnt/big/uv-cache` and uv falls back to copying. The two knobs are independent and nothing validates one against the other.

There's also an existing convention pointing the other way. `docs/testing.md:161-164` specifies a localized cache inside the selected ROCm runtime folder at `<install-root>/pip-cache`, **"including both generated managed folders and explicit `--prefix` folders"** — and `docs/manual-testing.md:41-46` says the same in user terms. The documented intent is colocate-with-install-root, which covers `--prefix`; this PR colocates with `data_dir`, which doesn't. Either follow that convention or say why it's being changed.

The fix looks cheap, which is why I'm flagging it rather than deferring. The target root is already in scope at four of the five call sites: `ensure_uv_venv` (`therock.rs:2218`) takes `install_root`; `run_uv_progress_command`'s only caller (`therock.rs:922`) has it as a local; `run_uv_logged_command`'s only caller (`comfyui.rs:350`) has `runtime.python`; `install_vllm_with_uv` (`engines/vllm/src/lib.rs:926`) takes `python: &Path`. A `uv_command_env_for_target(paths, target_root)` is a signature change at four sites. `ensure_managed_python` (`therock.rs:2747`, `:2760`) genuinely doesn't fit — it installs an interpreter into uv's own global store — and should stay keyed off `data_dir`.

If you'd rather land the improvement now and handle `--prefix` separately, that's a reasonable call — but then it needs to be an explicit non-goal in the description with a follow-up issue, not silently implied as covered. Which brings me to:

### 2. The description and commit message state something factually wrong about `--prefix`

PR body and commit `e10c649` both say:

> a user who moved the data directory to a larger disk with `ROCM_CLI_DATA_DIR` or `--prefix`

`--prefix` does not move the data directory. It relocates `install_root` for a single SDK install and leaves `paths.data_dir` — and therefore the uv cache — exactly where it was. Listing it among the fixed scenarios tells reviewers a case is handled when it isn't, and it's precisely the sentence that would stop the next reader from noticing finding 1. Free to correct in the body; the commit message needs an amend or a follow-up note.

## Non-blocking

- **`apps/rocm/src/main.rs:438-442` — the uninstall flag help is now misleading.** `--keep-data` says "Keep app data such as logs, services, and engines" and `--keep-cache` says "Keep caches." `build_uninstall_plan` buckets strictly by `paths.data_dir` / `paths.cache_dir`, so after this PR `--keep-cache` has zero effect on the uv cache while `--keep-data` silently preserves what will often be the largest directory the product manages. A one-line amendment to `--keep-data` is close to required. (The broader data-vs-cache taxonomy issue pre-dates this PR — `managed_pip_cache_dir` was already under the data root — so re-bucketing is a separate concern.)
- **`crates/rocm-core/src/uv.rs:78-81` — the behavior the PR actually adds is untested.** All three new tests call the private `uv_command_env_with_inherited_cache` with a pre-resolved `Option<&OsStr>`, so the env read and the `.filter(|v| !v.is_empty())` are never exercised in either direction. Avoiding `set_var` is the right call, but the empty-string boundary is a pure function and could be tested as one.
- **`crates/rocm-core/src/uv.rs:440-448`** — `command_env_cache_dir_follows_a_relocated_data_dir` is near-tautological. `managed_uv_cache_dir` is an unbranched `root.join("uv-cache")`, so building two `AppPaths` with different `data_dir`s and asserting the results differ only fails if the implementation ignores `paths` entirely, which the preceding test already covers. It doesn't touch `AppPaths::with_managed_root` or `ROCM_CLI_DATA_DIR`, so the name overstates it.
- **Stale `$HOME/.cache/uv`** — you flag this as known. Worth noting there's house precedent for the light version: `maybe_migrate_legacy_dashboard_config` (`main.rs:924-939`) does one-shot legacy-location detection and prints a one-line notice. A notice would fit; a full cache migration seems neither needed nor advisable.
- **`crates/rocm-core/src/uv.rs:65` — the escape hatch piggybacks on a third-party variable, against local convention.** Every neighbouring override in this same module is namespaced: `ROCM_CLI_UV_BINARY`, `ROCM_CLI_UV_VERSION`, `ROCM_CLI_UV_TIMEOUT_SECS`. `UV_CACHE_DIR` is uv's generic variable, so a developer who has it exported for unrelated Python work silently opts out of the fix with no signal — indistinguishable from an accident. A `ROCM_CLI_UV_CACHE_DIR` (keeping `UV_CACHE_DIR` inheritance for the e2e harness) would match the convention and let the CLI report that an override is active.
- **`MANIFEST.md:666-672`** — the uv section discloses that the uv *binary* is cached in the managed data directory but not the multi-GB wheel cache that now lives there. For a disclosure document that materially understates the footprint.

## While you're here (pre-existing, but this PR is the natural place)

- **`apps/rocm/src/main.rs:9411`, `:9423`** render `Downloads/cache: <value>` from `active_runtime_pip_cache_dir` / `setup_runtime_pip_cache_dir`, which fall back to `<install_root>/pip-cache`. But nothing anywhere sets `PIP_CACHE_DIR` or `--cache-dir`, and every manifest constructor writes `pip_cache_dir: None`. So the CLI tells the user downloads stay in their chosen ROCm folder while they actually go to the uv cache. Already wrong before this PR — but this is the PR that decides where downloads live, and fixing finding 1 by keying off `install_root` would make the existing message true.
- **`apps/rocm/src/therock.rs:3409`** — `managed_uv_cache_defaults_inside_generated_runtime_folder` only asserts `install_root.starts_with(&paths.data_dir)` and never touches `managed_uv_cache_dir` or `uv_command_env`. It predates this PR, but its name now claims coverage of exactly what changed here. Repurposing it into a `--prefix` divergence test would have caught finding 1.

## Nit

`crates/rocm-core/src/uv.rs:79` — the inherited-value check doesn't trim, while both siblings in the same file do (`uv_version` at `:225`, `env_secs` at `:332`). `UV_CACHE_DIR="   "` is treated as a deliberate override.

## On the `data_dir`-over-`cache_dir` question you raised

Your reasoning holds. `ROCM_CLI_CACHE_DIR` can point at a different filesystem from the environments, which would reintroduce the bug, and `AppPaths::with_managed_root` already derives `cache_dir` from `data_dir` by default. The right axis to colocate on is the one the environments live on — which is also the argument for `install_root` in finding 1.

rominf:
Thanks — this was a genuinely useful review; finding 1 turned out to be more interesting than either of us expected.

**Note on what changed since your review at `e10c649`:** I force-pushed. `e10c649` was amended (message only, content identical) to fix the `--prefix` claim, and is now `9298e2a`. All new work is in a separate commit `35931cb` on top, so the delta since your review is exactly that one commit.

---

### 1. `--prefix` installs are not covered — scoped out, with evidence: #187

I tried the cheap fix first, and **it does not work — it breaks every wheel install.** `uv` materializes `UV_CACHE_DIR` on *every* invocation, including `uv venv`, and `uv venv` refuses to create an environment in a directory that already exists. Since `install_root` *is* the venv directory (`venv_python_path(&install_root)`), `<install_root>/uv-cache` is self-defeating:

```
$ UV_CACHE_DIR=$P/fresh/uv-cache uv venv --python 3.12 $P/fresh
Using CPython 3.12.3 interpreter at: /usr/bin/python3.12
Creating virtual environment at: /tmp/.../fresh
error: Failed to create virtual environment
  Caused by: A directory already exists at: /tmp/.../fresh

$ ls $P/fresh
uv-cache        <-- uv created its own cache dir, which then blocked venv creation
```

That also explains why the `docs/testing.md` convention holds for pip but not uv: pip creates its cache during downloads, i.e. after the venv exists, whereas uv both creates the venv and does the downloading. The documented convention is not transferable to uv as-is.

So this isn't a four-call-site signature change — it needs a cache-location scheme chosen first, and the candidates all have real costs (sibling-of-install-root pollutes the user's directory or `$HOME`; a split cache that switches location after the venv exists is fragile against reordering; validate-and-warn instead of relocate; or accept the copy). That's a maintainer call, not one to make inside this PR.

Opened **#187** with the reproduction, the four options, and your `Downloads/cache:` point folded in — you're right that whichever scheme lands is what decides whether that message becomes true, so it belongs with the decision rather than here.

In this PR the gap is now explicit rather than implied: a non-goal section in the description, a `--prefix` caveat in `docs/manual-testing.md`, `ROCM_CLI_UV_CACHE_DIR` as a manual workaround, and `uv_cache_does_not_follow_a_prefix_install_root` pinning it so it can't drift silently in either direction.

I did confirm your regression combo (`ROCM_CLI_DATA_DIR=/mnt/big` + `--prefix $HOME/envs`) is real: `uv_command_env` never sees `install_root`, so the cache lands on the data-dir filesystem regardless of the prefix.

### 2. Factually wrong `--prefix` claim in body and commit — fixed

You were right, and it was exactly the sentence that would have stopped the next reader noticing finding 1. Removed from the PR body and from the commit message (amended: `e10c649` → `9298e2a`). The commit now states the `--prefix` limitation explicitly instead of listing it as fixed.

### Non-blocking

- **`--keep-data` / `--keep-cache` help** (`35931cb`) — both corrected. `--keep-data` now names the uv cache and calls out that it is often the largest directory the CLI manages; `--keep-cache` now says it does *not* cover the uv cache and points at `--keep-data`. Agreed the data-vs-cache re-bucketing is a separate concern; not touched.
- **Untested env read / empty-string boundary** (`35931cb`) — resolution is now a pure function over both variables, tested for unset, empty, whitespace-only, trimmed, inherited, and namespaced-override, plus precedence between the two. Still no `set_var`.
- **Near-tautological relocated-data-dir test** (`35931cb`) — replaced. It now goes through `AppPaths::with_managed_root`, which is the path `ROCM_CLI_DATA_DIR` actually takes, and asserts the resolved cache lands under the relocated root rather than merely differing.
- **Stale `$HOME/.cache/uv`** (`35931cb`) — took the `maybe_migrate_legacy_dashboard_config` precedent: one-shot detection, one-line notice, a marker so it doesn't nag, and nothing migrated or deleted. It stays silent when an override is in effect, so the e2e harness never sees it.
- **Un-namespaced escape hatch** (`35931cb`) — agreed, and your "indistinguishable from an accident" framing is the right one. Added `ROCM_CLI_UV_CACHE_DIR`, ranked above an ambient `UV_CACHE_DIR` (still honored for the harness). A `UvCacheSource` enum distinguishes managed / override / inherited so the CLI can report that an override is active.
- **`MANIFEST.md`** (`35931cb`) — now discloses the wheel cache, its location, its rough magnitude, that `rocm uninstall` removes it, and the override.

### While you're here

- **`Downloads/cache:` message** — folded into #187 rather than fixed here, since it can't be made true without deciding the `--prefix` scheme.
- **`managed_uv_cache_defaults_inside_generated_runtime_folder`** (`35931cb`) — renamed to say what it checks, and joined by `uv_cache_does_not_follow_a_prefix_install_root`, which is the divergence test you described. It asserts the current (gap) behavior deliberately, with a comment saying so and pointing at the tracking issue.

### Nit

- **Trim** (`35931cb`) — fixed, for both variables. `OsStr` has no `trim`, so it trims the lossy view and keeps the value when it survives, so a path is never silently rewritten.

---

Verification, all actually run on this branch: `cargo fmt --all --check` clean; `cargo clippy --locked --workspace --all-targets -- -D warnings` clean (sources touched first so the cache couldn't hide lints); `cargo test --workspace --all-targets --no-fail-fast` — the only two failures in the whole workspace are the known `proc_lifecycle` ones from #168, in a file this PR doesn't touch.

Leaving all threads for you to resolve.

rominf:
**CI status on `35931cb`** — every hosted check is green: `build-and-test`, `windows-build-and-test`, `clippy`, `Test (affected crates)`, `E2E tests`, `Coverage`, `CodeQL` (actions/python/rust), `prek`, `License header check`, `Commit signatures + sign-off`, `Third-party notices current`, `Lint (PowerShell)`.

Two self-hosted checks I can't drive from here:

- **`E2E tests (GPU)` — FAILURE, infrastructure, not this branch.** The job starts and completes in ~1 second with **zero steps executed** on `github-runner-0` (`self-hosted, linux, amd-gpu`). It fails identically on every recent `ci.yml` run in the repo, not just this PR:

  ```
  run=31104090115  E2E tests (GPU)  failure  dur=1s  steps=0
  run=31103889746  E2E tests (GPU)  failure  dur=0s  steps=0
  run=31103688037  E2E tests (GPU)  failure  dur=1s  steps=0
  run=31103619164  E2E tests (GPU)  failure  dur=1s  steps=0   <- this PR
  run=31103315258  E2E tests (GPU)  failure  dur=0s  steps=0
  ```

  `gh run rerun --failed` is refused while the run is in progress. Could a maintainer re-run it once the GPU runner is healthy?

- **`E2E tests (Strix Halo, Ubuntu` / `Windows)` — still queued** after ~40 minutes, waiting on the shared native runners. They do complete successfully on other runs in the same window, so this looks like queue contention rather than a problem with the branch.

Flagging rather than leaving a bare red X. Happy to re-push or re-run if anything here turns out to be real.

volen-silo (approved):

PR #175, opened 2026-08-03
fix(uninstall): report the caches it does not remove

### LINKED ISSUE #163: [Issue]: rocm uninstall leaves the uv cache and model weights behind
### Problem Description

`rocm uninstall` presents itself as removing what ROCm CLI put on the machine, but it leaves behind the two largest things ROCm CLI causes to be downloaded: the `uv` package cache and the model cache. A user who uninstalls to reclaim space can be left with tens of gigabytes still gone.

`build_uninstall_plan` (`apps/rocm/src/main.rs:14877`) removes binaries plus exactly three directories — the config, data, and cache dirs (`apps/rocm/src/main.rs:14908-14924`). Neither shared cache is inside any of them:

- **`uv` cache.** `uv_command_env()` (`crates/rocm-core/src/uv.rs:63-68`) sets only `UV_HTTP_TIMEOUT`, never `UV_CACHE_DIR`, so every `uv` invocation writes to `uv`'s default location under the user's home cache. (The companion issue on hardlink dedup proposes moving this inside the data directory, which would also bring it under uninstall's reach.)
- **Model cache.** Nothing in the product sets `HF_HOME` — the only occurrences are a dash demo fixture, the end-to-end harness, and a standalone script. `hf_cache_roots_from` (`engines/lemonade/src/lib.rs:1975-1994`) only *reads* the environment for discovery and otherwise falls back to the shared default, so weights land outside anything ROCm CLI manages, reports, or removes.

The per-runtime pip cache is inside the runtime root and is surfaced by `examine`, so it is the one cache that uninstall does cover today.

### Steps to Reproduce

1. Install the SDK and serve a model, so both caches are populated.
2. Run `rocm uninstall`.
3. The `uv` cache and the downloaded model weights are still on disk, and nothing in the output mentions them.

### Suggested Fix

At minimum, **report** them. The uninstall review already has a "Please review:" section — naming these paths and their sizes, and stating plainly that they are not being removed, would stop the command from implying a clean slate it does not deliver.

Beyond reporting, the two caches deserve different treatment:

- The `uv` cache is shared with any other `uv` project on the machine and ROCm CLI never claimed it, so deleting it by default would slow down unrelated work. Removing it should be opt-in — or it should be relocated inside the data directory, after which it is covered for free.
- Model weights are large, slow to re-download, sometimes gated behind a token, and may include files the user fetched themselves. These should be reported with their location, not deleted.

### Additional Information

Found while investigating unbounded disk growth.

rominf:
## Summary

`rocm uninstall` now names the caches it leaves behind, with their paths and why. Nothing new is deleted.

```
Please review:
  - the uv package cache is shared with other uv projects on this computer and is not removed: /home/<user>/.cache/uv
  - downloaded model files are slow to fetch again and may include your own, so they are not removed: /home/<user>/.cache/huggingface/hub
```

## Root cause

`build_uninstall_plan` removes binaries plus the config, data, and cache directories. The two largest things an install causes to appear on disk are in neither:

- The `uv` package cache — `uv_command_env()` never set `UV_CACHE_DIR`, so `uv` used its own default outside the ROCm CLI directories.
- Downloaded model files — nothing in the product sets `HF_HOME`, so weights land in the shared default. `hf_cache_roots_from` only *reads* the environment for discovery.

A user who uninstalls to reclaim space can be left with tens of gigabytes still gone, with nothing in the output to suggest it.

## Technical decisions

**Report, don't delete.** The `uv` cache is shared with every other `uv` project on the machine and ROCm CLI never claimed it; deleting it would slow down unrelated work. Model files are large, slow to re-fetch, sometimes gated behind a token, and may be the user's own. Naming them is what closes the honesty gap; deleting them is a separate decision.

**Locations are resolved at run time** from `UV_CACHE_DIR`, `HF_HOME`, and `HUGGINGFACE_HUB_CACHE`, falling back to the standard defaults — rather than assuming a fixed path.

**Anything already inside a directory being removed is skipped**, so a note is never printed for a cache that uninstall does in fact delete. This matters because #170 moves the `uv` cache under the data directory: once that lands, the `uv` note correctly stops appearing without further changes here, and it keeps working for users who point `UV_CACHE_DIR` elsewhere. The two PRs are independent and can land in either order.

**Missing paths are not mentioned**, so the output stays quiet on a machine that never populated them.

## Tests

The filtering logic is split into a pure `shared_cache_notes_for`, so the three tests do not depend on the caller's real environment:

- a cache inside a removed directory is not reported (the note would be false)
- a cache outside everything being removed is reported, with its path
- a missing path is not mentioned

## CI note

`cargo test --workspace` shows two unrelated failures in `proc_lifecycle` — see #168, fixed by #169. Everything else passes; fmt and `clippy -D warnings` are clean.

Fixes #163

### DIFF
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -15283,6 +15283,10 @@ fn build_uninstall_plan(paths: &AppPaths, options: &UninstallOptions) -> Result<
         }
     }
 
+    for note in shared_cache_notes(paths) {
+        plan.warnings.push(note);
+    }
+
     let managed_services = load_managed_services(paths).unwrap_or_default();
     if !managed_services.is_empty() {
         plan.warnings.push(format!(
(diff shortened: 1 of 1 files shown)


juhovainio (approved):
lgtm @rominf , please check that we have e2e coverage for this as well

PR #184, opened 2026-08-06
feat(release): distribute rocm-cli as a Python wheel

michaelroy-amd:
## Status: not for immediate merge

Opening this for review now. We will merge once we feel the quality bar is there, pending any other changes that come out of review. Note that the publish path is dormant by design (see the gate below), so merging this does not publish anything.

## What this adds

A second distribution channel: `pip install rocm-cli` installs the prebuilt `rocm` and `rocmd` binaries, alongside the existing `install.sh` / `install.ps1` bundles.

- **`scripts/build_wheel.py`** (new) packages binaries the release job has *already* built and signed. Nothing is rebuilt, so the wheel payload stays byte-identical to the copy inside the signed archive for the same tag.
- **`.github/workflows/release.yml`** builds and smoke-tests a wheel in each existing build job and adds a `publish-pypi` job using PyPI Trusted Publishing.
- **`apps/rocm/src/main.rs`** stops `rocm uninstall` from deleting binaries a Python package manager owns.
- Docs for the new channel and its (different) trust root.

## Design decisions worth a reviewer's attention

**Binaries ship in the wheel's `.data/scripts/` directory, not as console-script shims.** This is load-bearing, not a style choice. `std::env::current_exe()` has to resolve to the real binary: the CLI re-execs itself to launch managed engine services (`managed_service_launcher_path`) and locates `rocmd` as a sibling file (`sibling_binary_path`). A Python shim would make `current_exe()` the interpreter and break both. Both smoke tests assert the installed files are native executables by magic bytes (`7f454c46` / `MZ`), so a future refactor that introduced a shim fails CI rather than shipping.

**Wheels are written to `dist/wheels/`, a subdirectory.** `release_readiness.py::validate_exact_dist_assets` iterates `dist` for files only, so the strict release-asset gate is untouched and `release_readiness.py` needed no changes. The Windows upload glob was also changed from `Get-ChildItem dist -Include … -Recurse` to `Get-ChildItem dist\* -File -Include …`; with `-Recurse` it matched `dist\wheels\*.whl.sha256` and would have pushed wheel sidecars into the GitHub release asset set.

**Wheels only, no source distribution**, and only `manylinux_2_17_x86_64` and `win_amd64`. On any other platform pip fails with "no matching distribution" instead of attempting a Rust build, which matches the supported-platform rule in AGENTS.md §6. The existing manylinux2014 container build already satisfies `manylinux_2_17`: the released Linux binaries link only allowlisted shared objects and need at most `GLIBC_2.16`.

**Tag to PEP 440 mapping is exhaustive-or-error** and cross-checked against `[workspace.package] version`, so a tag can never publish a wheel whose binaries report a different version. `v0.1.0-experimental.1` maps to `0.1.0a1`. Because PyPI never allows a version to be re-uploaded, `-alpha.N` and `-experimental.N` collapse onto the same `aN` form and only one spelling may be used per serial; this is documented in `docs/release-trust.md`.

**Publication is dormant.** `publish-pypi` runs only when the repository variable `ROCM_CLI_PUBLISH_PYPI` is `1` or `true` (an allowlist, so `FALSE`/`no`/`off` stay dormant), mirroring the existing `ROCM_CLI_REQUIRE_PRODUCTION_TRUST` pattern. Until then CI builds, smoke-tests, and retains the wheels as workflow artifacts and uploads nothing. The GitHub release asset set is unchanged either way; the pipeline does gain steps that can fail, which the docs say explicitly rather than claiming nothing changed.

**The uninstall fix is a real bug, reproduced first.** `rocm uninstall` globbed the running executable's directory and deleted every rocm-named file. Inside a pip environment:

```
3 item(s) would be removed:
  - config: …/.rocm
  - binary: …/venv/bin/rocm
  - binary: …/venv/bin/rocmd
```

That deletes files pip owns and leaves the wheel `RECORD` dangling while `pip list` still reports the package installed. A new tier in `build_uninstall_plan` detects a Python-managed layout (`pyvenv.cfg`, or a `*.dist-info/RECORD` that claims the script) and skips binary removal, naming the pip / pipx / `uv tool` command instead. Probe failures resolve conservatively: a tree that exists but cannot be read is treated as managed rather than deleted, because a false positive only leaves files in place while a false negative corrupts another package manager's state.

## Trust model, stated plainly

PyPI cannot carry the detached `.sig` sidecar the installers verify, so this channel has a *different* trust root: Trusted Publishing (OIDC, no long-lived upload token) plus PEP 740 attestations. `docs/release-trust.md` is explicit that pip does not verify attestations at install time and offers no flag to require them, so for a plain `pip install` they are an auditable provenance record rather than an install gate. It also states that the cross-job checksum catches transit corruption but cannot detect a compromised builder that wrote a consistent wheel and sidecar. The GitHub release assets remain the signature-bearing artifacts.

## Verification performed

Local, all passing:

- `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace --all-targets`
- `cargo test -p rocm --bin rocm is_python_managed` (6 tests), plus mutation probes: stubbing the detector to `false` fails 5 of 6, and restoring the old `RECORD` parser fails the quoted-row test
- `python scripts/build_wheel.py --self-test`, `release_readiness.py --self-test`, `build_single_exe_release.py self-test`, `scripts/smoke_local.py`, `cargo xtask verify-pinned-keys`, both WSL script self-tests, `ruff`
- End to end: built a wheel from the published `v0.1.0-experimental.1` Linux binaries, installed it into a fresh venv, and confirmed `rocm --version`, `rocmd --version`, `rocm examine`, and mode `0755`. Also confirmed `pipx install` and `uv tool install` place both executables
- `release.yml` parses; every non-pwsh `run` block passes `bash -n`; all six pwsh blocks parse under a real Windows PowerShell parser; the new PE magic check was executed there and correctly rejected a `#!python` shim; all ten `uses:` refs are 40-hex SHAs with a `# vX.Y.Z` comment
- Leak scan over the diff is clean; commit is signed and signed-off

Not runnable here, so stated rather than claimed:

- GitHub Actions runtime: the manylinux container build, the Windows job, artifact transfer, and the publish gate itself
- A native Windows runner (PowerShell parsing and behavior checks are not a substitute)
- Real PyPI or TestPyPI publication
- GPU hardware lanes (not reachable from this change)

## Owner prerequisites before the gate can be opened

1. Access to the existing (currently empty) `AMD` organization on PyPI. `rocm-cli` is unregistered on both PyPI and TestPyPI; `rocm` itself is squatted by an unrelated placeholder, which is why the distribution name is `rocm-cli`.
2. A pending Trusted Publisher for `ROCm/rocm-cli`, workflow `release.yml`, environment `pypi`. The pending-publisher flow creates the project on first OIDC publish, so the name is claimed without ever minting an API token.
3. A TestPyPI rehearsal via `workflow_dispatch` (the dispatch input defaults to `testpypi` precisely so the irreversible action is always opt-in).
4. Only then set `ROCM_CLI_PUBLISH_PYPI`.

## Review notes

This branch already went through six independent adversarial review passes plus a cross-review round; 19 findings were applied and 1 was rejected on spec grounds (a claim that `License-File` must be `licenses/LICENSE.TXT`; the metadata spec defines that path as relative to the project root, and the wheel then stores it at `.dist-info/licenses/<path>`, so the bare filename is correct). Fresh eyes are still very welcome, particularly on the workflow semantics that can only be proven by running them.

---

- [x] If this PR fixes a bug, searched `tests/e2e-cucumber/expectations.toml` for the fixed ticket ID and removed/narrowed any now-stale xfail rows. — no xfail rows relate to this change; the one existing e2e assertion on uninstall output (`skipping running executable on Windows`, `lifecycle_steps.rs`) still holds, because an installer-managed Windows binary is not detected as Python-managed and still reaches that branch.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,6 +8,14 @@ on:
       tag:
         description: "Release tag (e.g. v0.1.0)"
         required: true
+      pypi_repository:
+        description: "PyPI index for the publish job (defaults to a TestPyPI rehearsal)"
+        required: false
+        default: testpypi
+        type: choice
+        options:
+          - testpypi
+          - pypi
 
 permissions:
   contents: write
@@ -144,6 +152,62 @@ jobs:
             --asset "${DIST}.tar.gz" \
             --asset rocm-cli-linux-amd64.tar.gz
 
+      - name: Build Python wheel
+        run: |
+          # Reuse the exact target/release binaries that were just packaged and
+          # signed — the wheel payload must stay byte-identical to the binaries
+          # inside the signed archive, so nothing is rebuilt here.
+          python scripts/build_wheel.py \
+            --bin-dir target/release \
+            --platform linux-amd64 \
+            --tag "${{ steps.version.outputs.value }}" \
+            --out-dir dist/wheels
+
+      - name: Smoke test the Python wheel
+        run: |
+          shopt -s nullglob
+          wheels=(dist/wheels/*.whl)
+          if [ "${#wheels[@]}" -ne 1 ]; then
+            echo "expected exactly one linux wheel, found ${#wheels[@]}" >&2
+            exit 1
+          fi
+          venv="${RUNNER_TEMP}/wheel-smoke"
+          rm -rf "${venv}"
+          python -m venv "${venv}"
+          "${venv}/bin/python" -m pip install --no-index --disable-pip-version-check "${wheels[0]}"
+          # The wheel ships the real ELF binaries in its .data/scripts
+          # directory rather than console-script shims, so these must be the
+          # native executables themselves and must report their own version.
+          for binary in rocm rocmd; do
+            magic="$(head -c 4 "${venv}/bin/${binary}" | od -An -tx1 | tr -d ' \n')"
+            if [ "${magic}" != "7f454c46" ]; then
+              echo "${binary} is not an ELF executable (magic ${magic}); a shim would break current_exe()" >&2
+              exit 1
+            fi
+          done
+          # Assert the installed binaries are the version the wheel claims, so
+          # a stale target/release tree (for example from a restored build
+          # cache) cannot ship binaries that disagree with the release tag.
+          wheel_version="$(basename "${wheels[0]}")"
+          wheel_version="${wheel_version#rocm_cli-}"
+          wheel_version="${wheel_version%%-py3-none-*}"
+          release="$(printf '%s' "${wheel_version}" | sed -E 's/^([0-9]+\.[0-9]+\.[0-9]+).*/\1/')"
+          for binary in rocm rocmd; do
+            reported="$("${venv}/bin/${binary}" --version)"
+            if [ "${reported}" != "${binary} ${release}" ]; then
+              echo "${binary} reports '${reported}', expected '${binary} ${release}' from wheel ${wheel_version}" >&2
+              exit 1
+            fi
+          done
+          rm -rf "${venv}"
+
+      - name: Upload Python wheel
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: wheel-linux-amd64
+          path: dist/wheels/
+          if-no-files-found: error
+
       - name: Create release
         env:
           GH_TOKEN: ${{ github.token }}
@@ -222,8 +286,86 @@ jobs:
           GH_TOKEN: ${{ github.token }}
         run: |
           $version = "${{ needs.release.outputs.version }}"
-          $assets = Get-ChildItem dist -File -Include *.zip,*.sha256,*.sig -Recurse | ForEach-Object { $_.FullName }
+          # `dist\*` plus -Include, never -Recurse: with -Recurse this glob also
+          # matches the wheel checksum sidecars under dist\wheels, which are not
+          # GitHub release assets. Without a path wildcard, -Include silently
+          # matches nothing at all.
+          $assets = Get-ChildItem dist\* -File -Include *.zip,*.sha256,*.sig | ForEach-Object { $_.FullName }
           gh release upload $version @assets --clobber
+
+      # The wheel steps run after the release upload so the release asset set
+      # is produced from a tree that has no wheels in it yet. The upload glob
+      # above is non-recursive, so this ordering is defence in depth rather
+      # than the only thing keeping the asset set correct.
+      - name: Build Python wheel
+        shell: pwsh
+        run: |
+          $ErrorActionPreference = "Stop"
+          # Reuse the exact target\release binaries that were just packaged and
+          # signed — the wheel payload must stay byte-identical to the binaries
+          # inside the signed archive, so nothing is rebuilt here.
+          python .\scripts\build_wheel.py `
+            --bin-dir target\release `
+            --platform windows-amd64 `
+            --tag "${{ needs.release.outputs.version }}" `
+            --out-dir dist\wheels
+
+      - name: Smoke test the Python wheel
+        shell: pwsh
+        run: |
+          $ErrorActionPreference = "Stop"
+          function Invoke-Checked {
+            param([string]$Exe, [string[]]$Arguments)
+            & $Exe @Arguments
+            if ($LASTEXITCODE -ne 0) {
+              throw "$Exe $($Arguments -join ' ') failed with exit code $LASTEXITCODE"
+            }
+          }
+          $wheels = @(Get-ChildItem dist\wheels -File -Filter *.whl)
+          if ($wheels.Count -ne 1) {
+            throw "expected exactly one windows wheel, found $($wheels.Count)"
+          }
+          $venv = Join-Path $env:RUNNER_TEMP "wheel-smoke"
+          if (Test-Path -LiteralPath $venv) {
+            Remove-Item -Recurse -Force -LiteralPath $venv
+          }
+          Invoke-Checked python @("-m", "venv", $venv)
+          Invoke-Checked "$venv\Scripts\python.exe" @("-m", "pip", "install", "--no-index", "--disable-pip-version-check", $wheels[0].FullName)
+          # The wheel ships the real PE binaries in its .data/scripts directory
+          # rather than console-script shims, so these must be the native
+          # executables themselves and must report their own version.
+          foreach ($binary in @("rocm.exe", "rocmd.exe")) {
+            $path = Join-Path "$venv\Scripts" $binary
+            $magic = [System.IO.File]::ReadAllBytes($path)[0..1]
+            if ($magic[0] -ne 0x4D -or $magic[1] -ne 0x5A) {
+              throw "$binary is not a PE executable; a shim would break current_exe()"
+            }
+          }
+          # Assert the installed binaries are the version the wheel claims, so
+          # a stale target\release tree cannot ship binaries that disagree with
+          # the release tag.
+          $wheelVersion = $wheels[0].Name -replace '^rocm_cli-', '' -replace '-py3-none-.*$', ''
+          if ($wheelVersion -notmatch '^(?<release>\d+\.\d+\.\d+)') {
+            throw "cannot read a release segment out of wheel version '$wheelVersion'"
+          }
+          $release = $Matches['release']
+          foreach ($binary in @("rocm", "rocmd")) {
+            $reported = (& "$venv\Scripts\$binary.exe" "--version") -join ""
+            if ($LASTEXITCODE -ne 0) {
+              throw "$binary --version failed with exit code $LASTEXITCODE"
+            }
+            if ($reported.Trim() -ne "$binary $release") {
+              throw "$binary reports '$reported', expected '$binary $release' from wheel $wheelVersion"
+            }
+          }
+          Remove-Item -Recurse -Force -LiteralPath $venv
+
+      - name: Upload Python wheel
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: wheel-windows-amd64
+          path: dist/wheels/
+          if-no-files-found: error
 
   publish-release:
     name: Publish release
@@ -240,3 +382,110 @@ jobs:
           VERSION="${{ needs.release.outputs.version }}"
           RELEASE_ID="$(gh release view "${VERSION}" --json databaseId -q .databaseId)"
           gh api --method PATCH "repos/${GITHUB_REPOSITORY}/releases/${RELEASE_ID}" -f draft=false
+
+  publish-pypi:
+    name: Publish wheels to PyPI
+    runs-on: ubuntu-latest
+    needs:
+      - release
+      - windows-release
+      # Sequenced after publish-release so wheels never reach an index before
+      # the GitHub release they correspond to is out of draft.
+      - publish-release
+    # PyPI publication gate — dormant until the owner sets the repository
+    # variable ROCM_CLI_PUBLISH_PYPI to "1" or "true", mirroring the
+    # ROCM_CLI_REQUIRE_PRODUCTION_TRUST gate in the env block above. It stays
+    # unset until the project owns the PyPI `rocm-cli` project and has
+    # registered this workflow as its Trusted Publisher. The gate is an
+    # allowlist rather than a not-falsy test so that any other value —
+    # including "FALSE", "no", or "off" — leaves the job dormant. While it is
+    # dormant this job is skipped and nothing reaches any index; the wheels are
+    # still built, smoke-tested, and retained as workflow artifacts, and the
+    # GitHub release asset set is unchanged.
+    if: vars.ROCM_CLI_PUBLISH_PYPI == '1' || vars.ROCM_CLI_PUBLISH_PYPI == 'true'
+    environment: pypi
+    permissions:
+      # Trusted Publishing mints a short-lived OIDC token instead of using a
+      # stored API token, so this job needs id-token: write and no upload
+      # secrets. Declaring permissions here also keeps the workflow-level
+      # `contents: write` from applying to a job that writes nothing back.
+      contents: read
+      id-token: write
+    steps:
+      - name: Resolve publication target
+        id: target
+        env:
+          # Empty on tag pushes, which always mean the real index. A manual
+          # dispatch defaults to testpypi so that the irreversible action —
+          # a real upload burns that version forever — is always opt-in.
+          REQUESTED: ${{ github.event.inputs.pypi_repository }}
+        run: |
+          TARGET="${REQUESTED:-pypi}"
+          case "${TARGET}" in
+            pypi|testpypi) ;;
+            *)
+              echo "unknown pypi_repository input: ${TARGET}" >&2
+              exit 1
+              ;;
+          esac
+          echo "value=${TARGET}" >> "$GITHUB_OUTPUT"
+
+      - name: Download built wheels
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          pattern: wheel-*
+          path: dist/wheels
+          merge-multiple: true
+
+      - name: Verify wheel checksums and stage the upload directory
+        run: |
+          shopt -s nullglob
+          wheels=(dist/wheels/*.whl)
+          if [ "${#wheels[@]}" -ne 2 ]; then
+            echo "expected 2 wheels (linux + windows), found ${#wheels[@]}" >&2
+            exit 1
+          fi
+
+          # Re-verify each wheel against the sidecar written beside it on the
+          # builder runner. The sidecar travels in the same artifact, so this
+          # catches a file corrupted or truncated in artifact storage transit,
+          # not a compromised builder that wrote both files consistently.
+          for wheel in "${wheels[@]}"; do
+            sidecar="${wheel}.sha256"
+            if [ ! -f "${sidecar}" ]; then
+              echo "missing checksum sidecar: ${sidecar}" >&2
+              exit 1
+            fi
+            expected="$(awk 'NR == 1 { print $1 }' "${sidecar}")"
+            actual="$(sha256sum "${wheel}" | awk '{ print $1 }')"
+            if [ -z "${expected}" ] || [ "${expected}" != "${actual}" ]; then
+              echo "checksum mismatch for ${wheel}: expected '${expected}', got '${actual}'" >&2
+              exit 1
+            fi
+            echo "verified $(basename "${wheel}") ${actual}"
+          done
+
+          # The publish action uploads every file in packages-dir, so stage the
+          # wheels on their own — the .sha256 sidecars are not distributions.
+          mkdir -p dist/pypi
+          cp "${wheels[@]}" dist/pypi/
+
+      # Exactly one of the two steps below runs. A rehearsal resolves to
+      # testpypi and therefore cannot reach the real index, and any other
+      # value already failed in "Resolve publication target". Neither step
+      # passes credentials (Trusted Publishing only), and both leave PEP 740
+      # attestations at the action's enabled default.
+      - name: Publish to TestPyPI (rehearsal)
+        if: steps.target.outputs.value == 'testpypi'
+        uses: pypa/gh-action-pypi-publish@dc37677b2e1c63e2034f94d8a5b11f265b73ba33 # v1.14.2
+        with:
+          packages-dir: dist/pypi
+          repository-url: https://test.pypi.org/legacy/
+          print-hash: true
+
+      - name: Publish to PyPI
+        if: steps.target.outputs.value == 'pypi'
+        uses: pypa/gh-action-pypi-publish@dc37677b2e1c63e2034f94d8a5b11f265b73ba33 # v1.14.2
+        with:
+          packages-dir: dist/pypi
+          print-hash: true
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -53,6 +53,7 @@ use rocm_engine_protocol::{
 };
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
+use std::borrow::Cow;
 use std::collections::{BTreeMap, VecDeque};
 use std::ffi::OsString;
 use std::fmt::Write as _;
@@ -14888,6 +14889,11 @@ fn build_uninstall_plan(paths: &AppPaths, options: &UninstallOptions) -> Result<
                 "binary removal skipped because {} looks like a cargo target build; pass --force-dev-binaries to remove sibling debug/release binaries",
                 current_exe.display()
             ));
+        } else if is_python_managed_layout(&current_exe) {
+            plan.skipped.push(format!(
+                "binary removal skipped because {} was installed by a Python package manager; remove it with the matching command instead: `pip uninstall rocm-cli`, `pipx uninstall rocm-cli`, or `uv tool uninstall rocm-cli`",
+                current_exe.display()
+            ));
         } else {
             for path in collect_installed_binary_candidates(&current_exe)? {
                 if rocm_core::runtime_is_windows() && path == current_exe {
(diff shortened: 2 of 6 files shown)


volen-silo (changes requested):
Reviewed at `3f27beb`. The wheel packaging itself holds up well — spec conformance, the `.data/scripts/` rationale, the dormant publish gate, OIDC scoping, action pinning, and the `dist/wheels/` subdirectory trick all check out, and the `-Recurse` → `dist\*` glob change is a genuine catch. Two blocking items below, both in the supporting fix and its coverage rather than in the packaging.

## Blocking

### 1. `is_python_managed_layout` misses `pip install --user` on Windows — the delete path stays live

`apps/rocm/src/main.rs:15133` (`python_site_packages_dirs`), reached from `build_uninstall_plan`.

The probe only considers `<env_root>/Lib/site-packages` and `<env_root>/lib/python*/site-packages`. CPython's `nt_user` install scheme is not shaped like that: scripts land in `%APPDATA%\Python\Python3XX\Scripts` and packages in `%APPDATA%\Python\Python3XX\site-packages` — **no `Lib\` level**, unlike the `nt` scheme used for venvs and system installs. So `env_root\Lib\site-packages` does not exist, `env_root\lib` does not exist, the function returns `Ok(vec![])` rather than `Err(ProbeFailed)`, and the plan falls through to the delete branch.

That is the exact corruption this PR exists to prevent, on a supported platform. The running-executable skip spares `rocm.exe`, but `rocmd.exe` gets deleted, leaving pip's `RECORD` dangling while the package still reports as installed.

Smallest fix: add `env_root.join("site-packages")` to the candidate list. The RECORD row (`..\Scripts\rocm.exe`) already matches `record_row_targets_script`, so nothing else needs to change. Worth a unit test too — the existing coverage has a `Scripts` parent only in combination with a real `pyvenv.cfg` (`is_python_managed_layout_detects_windows_scripts_directory`) and the RECORD-without-`pyvenv.cfg` case only in a Unix-shaped tree, so this corner is precisely the untested one.

Two related gaps, lower severity, that I would at least see acknowledged in the doc comment's limitations paragraph rather than fixed here:

- **`uv tool install` / `pipx install` on Windows.** Both symlink on Unix and *copy* on Windows into `%USERPROFILE%\.local\bin`. On Linux `current_exe()` resolves `/proc/self/exe` back into the tool venv, so detection works; on Windows `GetModuleFileNameW` returns the copy's own path, with no `pyvenv.cfg` and no site-packages above it. The user-facing message currently advertises `uv tool uninstall` as the remedy for a case that is not detected there.
- **Debian/Ubuntu `dist-packages`** — the join at line 15157 hardcodes `site-packages`. Only reachable behind `--break-system-packages`, so minor.

The doc comment at `:15054` documents one accepted limitation, and it is a false *positive* (the safe direction). These are false negatives, which is the direction the comment itself calls unacceptable — so they should not be silent.

### 2. `scripts/build_wheel.py --self-test` is not wired into CI

`.github/workflows/ci.yml:214-228` (Linux) and `:348-360` (Windows).

Every other release/acceptance script in `scripts/` has its `--self-test` invoked in both CI lanes — `release_readiness.py`, `wsl_preflight.py`, the three `*_therock_gpu_test.py` harnesses, `setup-wsl-portable-build-deps.sh`. This adds 631 lines of new release-path code with a substantial self-test (tag mapping, workspace-version cross-check, both platform tags, member set, RECORD hashes, byte-identical rebuild, mode bits) and it runs only locally. `build_wheel.py` currently appears nowhere outside `release.yml`, which means its first CI execution would be during a real tag push, ahead of "Create release" — a regression there fails an otherwise-working release instead of failing at PR time. Two lines.

## Non-blocking

- **`.github/workflows/release.yml:415-431`** — the "manual dispatch defaults to TestPyPI" property relies on GitHub backfilling `github.event.inputs.pypi_repository` from the YAML `default:`. It does, so this is not a live bug; but `TARGET="${REQUESTED:-pypi}"` conflates "tag push, no inputs" with "dispatch, field omitted", and the fallback for the latter is the irreversible index. Keying off `github.event_name` explicitly would make the guarantee provable from the workflow text alone. Cheap, and this is the one place worth being paranoid.
- **`scripts/build_wheel.py:222`** — the `License-File:` metadata lines are hardcoded independently of `LICENSE_FILES` (`:44`), which drives what actually gets packaged. They agree today; they will diverge silently the first time a notices file is added, and `self_test_wheel` does not cross-check them. Derive them from the constant.
- **No integration coverage for the uninstall guard.** The release smoke test already builds a real venv and pip-installs the wheel; adding `rocm uninstall --dry-run` there and asserting it *refuses* would give the headline fix genuine end-to-end coverage for one line of YAML. Currently the only coverage is unit tests over synthetic temp dirs — which is where finding 1 slipped through.
- **New user-visible behavior is undocumented.** Neither the new Python-package section nor the existing `rocm uninstall` block in `README.md` mentions that the CLI now declines to remove package-manager-owned binaries.
- **While you're here:** the "More docs" index in `README.md` still doesn't link `docs/release-trust.md`, which this PR roughly doubles.
- **`scripts/build_wheel.py:195`** is the third byte-for-byte copy of the chunked SHA-256 helper (`release_readiness.py`, `build_single_exe_release.py` have the others). In-convention, just noting the count.
- **Gate allowlist** (`release.yml:405`) accepts only `1`/`true` case-sensitively where `release_readiness.py`'s `truthy()` accepts `{1,true,yes,on}` case-insensitively. Deliberate and it cannot fail open — inconsistency only.

## One tradeoff worth stating accurately

The design notes imply no standard tool could package prebuilt binaries without rebuilding. That is correct for maturin — every subcommand invokes `cargo build`, there is no package-a-prebuilt mode. It is not correct for hatchling: I built and pip-installed a `[tool.hatch.build.targets.wheel.shared-scripts]` wheel while reviewing this, and it preserved `0o755` and installed a real executable, no shim. That would replace roughly 330 of the 631 lines, leaving the tag mapping and Cargo cross-check as the domain logic no tool provides.

Hand-rolling is still defensible here — the repo has zero non-stdlib Python dependencies and `build_single_exe_release.py` already hand-rolls its archives with `tarfile`/`zipfile`, so this follows the house pattern. No change requested. The PR just shouldn't claim the alternative was impossible.

Also flagging that the always-on wheel build adds a new way for a real release to fail, ahead of "Create release". `docs/release-trust.md` says this outright rather than claiming nothing changed, which is the right call — just make sure that is a conscious acceptance and not a footnote.

## Things worth reinforcing

- `ProbeFailed` as a distinct type, with unreadable-but-exists separated from missing, makes the fail-closed property structural instead of conventional. The RECORD parser handles RFC 4180 quoting, doubled quotes, embedded commas, and CRLF correctly. Finding 1 is about *which* directories get probed, not this machinery.
- The smoke tests are real: `pip install --no-index` into an actual venv, execute the installed binaries, check magic bytes, match the reported version against the wheel filename, `$LASTEXITCODE` checked after every native call. No `|| true`, no glob that passes on zero matches.
- The `.data/scripts` load-bearing claim holds end to end — plain venv, `--target`, and `--prefix` installs all place `rocm` and `rocmd` as siblings with the exec bit intact, and `pip uninstall` leaves no dangling RECORD.

One claim I could not verify live and am flagging rather than endorsing: `docs/release-trust.md:137-141` on pip not verifying PEP 740 attestations at install time. It matches my understanding and the doc states it conservatively, but I did not confirm it against a current pip.

scripts/build_wheel.py
```diff
@@ -0,0 +1,631 @@
...
+LICENSE_FILES = ("LICENSE.TXT", "THIRD_PARTY_NOTICES.txt")
+
+# Reproducible zip timestamp: the earliest value the zip format can encode.
+ZIP_DATE_TIME = (1980, 1, 1, 0, 0, 0)
+SCRIPT_MODE = 0o755
+DATA_MODE = 0o644
+S_IFREG = 0o100000
+CREATE_SYSTEM_UNIX = 3
+
+TAG_RE = re.compile(
```
rominf, line 53: `TAG_RE` doesn't accept `preview`, which is the shape of the repo's current release tag.

```
v0.1.0-preview.1        -> REJECT
v0.1.0-experimental.1   -> MATCH
v0.1.0-rc.2             -> MATCH
v0.1.0                  -> MATCH
```

Running the builder against it:

```
$ python scripts/build_wheel.py --tag v0.1.0-preview.1 ...
wheel build failed: tag is not a publishable release tag: v0.1.0-preview.1
exit=1
```

`v0.1.0-preview.1` was tagged 2026-08-31 and is the current `Latest` release, so this is drift that appeared after the PR opened rather than an oversight.

The consequence is bigger than a skipped wheel: the "Build Python wheel" step (`release.yml:155`) has no `if:` guard — the `ROCM_CLI_PUBLISH_PYPI` gate is on the separate publish job further down — and it runs at line 160, ahead of "Create release" at line 211. So a `-preview.N` tag push fails the Linux job and no GitHub release is created at all.

Either add `preview` to `TAG_RE` and `PRE_RELEASE_MARKERS` (mapping to `a` or `rc`), or make an unmappable tag skip the wheel rather than fail the job. The second is probably safer regardless — it keeps an unrecognised tag shape from ever taking the release path down with it.

apps/rocm/src/main.rs
```diff
@@ -15045,6 +15051,225 @@ fn is_dev_binary_layout(path: &Path) -> bool {
...
+/// installed script. Both the unix (`lib/python3.13/site-packages`) and the
+/// Windows (`Lib/site-packages`) layouts are probed on every host: this is pure
+/// path inspection, and a tree must be classified the same way regardless of
+/// which platform is doing the inspecting.
+///
+/// A missing `lib` directory yields an empty list, because that is the ordinary
+/// shape of a plain `~/.local/bin` install. A `lib` directory that exists but
+/// cannot be listed yields [`ProbeFailed`]: its contents could name the
+/// distribution that owns the script.
+fn python_site_packages_dirs(env_root: &Path) -> Result<Vec<PathBuf>, ProbeFailed> {
```
rominf, line 15133: This is @volen-silo's blocker #1; I ran into it independently, so here's the mechanism in case it's useful.

`python_site_packages_dirs` probes `env_root/Lib/site-packages` and `env_root/lib/python*/site-packages`. CPython's `nt_user` scheme (what `pip install --user` uses on Windows) installs to `{userbase}/Python3XX/site-packages` — **no `Lib/` level** — with scripts at `{userbase}/Python3XX/Scripts`.

So neither candidate exists, `read_dir` errors out, `unix_lib.is_dir()` is false, and the function returns `Ok(vec![])` rather than an error. `.any()` over an empty vec is `false`, the guard doesn't trigger, and the uninstall plan falls through to the delete branch — removing `rocmd.exe` and leaving pip's `RECORD` dangling. That's the exact corruption this PR exists to prevent, on a supported platform.

Pushing `env_root.join("site-packages")` onto the candidate list should be enough; the `..\Scripts\rocm.exe` RECORD row already matches `record_row_targets_script`, so nothing downstream changes. `/usr/lib/python3/dist-packages` on Debian is the same shape and worth adding while you're there.

One case that resists this fix: `pipx` and `uv tool` on Windows *copy* the binaries into `%USERPROFILE%\.local\bin` with no Python environment above them, so no probe can detect them — yet the new message advertises `pipx uninstall` / `uv tool uninstall` as the remedy. Since the doc comment says false negatives are unacceptable, that limitation is worth stating explicitly there.

apps/rocm/src/main.rs
```diff
@@ -14888,6 +14889,11 @@ fn build_uninstall_plan(paths: &AppPaths, options: &UninstallOptions) -> Result<
                 "binary removal skipped because {} looks like a cargo target build; pass --force-dev-binaries to remove sibling debug/release binaries",
                 current_exe.display()
             ));
+        } else if is_python_managed_layout(&current_exe) {
```
rominf, line 14892: This new `plan.skipped` message is user-observable `rocm uninstall` output, so AGENTS.md §3 wants a Gherkin scenario for it — `git diff --stat origin/main...HEAD -- tests/` is empty. §3 is explicit that a unit test over the internal helper doesn't discharge the requirement, and the pre-existing Windows running-executable assertion covers a different branch.

Worth noting this isn't a formality here: the current coverage is unit tests over synthetic temp dirs, which is precisely how the `--user` gap above stayed invisible. A scenario in `install_lifecycle.feature` asserting that `rocm uninstall --dry-run` refuses binary removal in a Python-managed layout would cover it. Cheaper still, one line in the release smoke test (`rocm uninstall --dry-run` after the venv install, asserting the skip message) would have caught it.

docs/testing.md
```diff
@@ -729,6 +729,7 @@ Release trust checks:
 cargo test -p rocm --bin rocm metadata_signature_verification_accepts_generated_key_and_rejects_tamper
 cargo test -p rocm-core model_recipe_index_signature_accepts_generated_key_and_rejects_tamper
 python scripts/release_readiness.py --self-test
+python scripts/build_wheel.py --self-test
```
rominf, line 732: `build_wheel.py --self-test` is added to the manual testing list, but it never runs in CI — `git show HEAD:.github/workflows/ci.yml | grep build_wheel` returns nothing, and the script appears in no workflow except `release.yml`.

Every other `scripts/*` self-test is invoked in both CI lanes: `release_readiness.py`, `wsl_preflight.py`, the three `*_therock_gpu_test.py` files, and `setup-wsl-portable-build-deps.sh`. This PR adds 631 lines of release-path code whose first CI execution would be during a real tag push, immediately before "Create release" — so a regression fails an otherwise-working release instead of failing at PR time. That's the same exposure as the `TAG_RE` issue above, which is a good illustration of the risk.

Two lines beside the existing `release_readiness.py --self-test` in both lanes would close it. (This is @volen-silo's blocker #2.)

.github/workflows/release.yml
```diff
@@ -144,6 +152,62 @@ jobs:
             --asset "${DIST}.tar.gz" \
             --asset rocm-cli-linux-amd64.tar.gz
 
+      - name: Build Python wheel
```
rominf, line 155: Two things about this step beyond the `TAG_RE` problem.

**Rebase hazard.** The branch is 92 commits behind and `git merge-tree` reports content conflicts in this file and `README.md`. #330 replaced "Create release" with a reuse-existing-release path that re-runs `gh release upload "${VERSION}" dist/*.tar.gz dist/*.tar.gz.sha256 dist/*.sig --clobber` — the very step this PR inserts the wheel build ahead of. Those globs are non-recursive today, so `dist/wheels/` sidecars still can't leak into the release assets, but that's a property to re-verify after resolving rather than assume, since PR-time CI ran against the August tree.

**The `manylinux_2_17` tag is asserted, not verified.** It's honest today — I checked the released binary: max symbol version `GLIBC_2.16`, and every `NEEDED` entry is on the manylinux2014 allowlist. But nothing enforces it. The smoke test installs the wheel on `ubuntu-latest` (glibc 2.39), where a wheel tagged `manylinux_2_17` installs regardless of whether the payload actually conforms, so it validates nothing about the tag. A future dependency that links a non-allowlisted `.so` or raises the glibc floor would ship a mistagged wheel that fails at exec time for users, with no other signal since the archive channel is unaffected. `auditwheel show --strict` on the built wheel in the smoke-test step would make the tag mean something.

rominf (changes requested):
The packaging mechanics here are solid — I built a wheel from the published `v0.1.0-preview.1` binaries, installed and uninstalled it in a venv, and it passed `twine check`, placed real executables at mode 0755, and reported the right version. No console-script shims, so `current_exe()` stays correct. Trusted Publishing is set up properly (no tokens anywhere), and all three new action pins verify against the API.

The blocker is timing rather than design. `TAG_RE` accepts `alpha|beta|rc|experimental`, and the repo's current `Latest` release is tagged `v0.1.0-preview.1` — pushed on 2026-08-31, about three weeks after this PR opened, so it's drift you couldn't have seen. I ran the mapper against every existing tag: `preview` is the only one it rejects. Because the wheel step is ungated and runs before "Create release", the next `-preview.N` tag push would abort the Linux job and never create the GitHub release at all. That's a two-line fix, but it should go in before merge.

Separately, @volen-silo's two blockers from the review on this same commit look genuinely unaddressed — I hit the Windows `--user` one independently while checking the uninstall guard, so I've written up what I found in case the extra detail helps.

The branch is also 92 commits behind and conflicts in `release.yml` and `README.md`; #330 rewrote the exact release step this PR inserts itself in front of, so that hunk wants a careful look on rebase rather than a mechanical resolution.

Things I checked that came back clean: wheel spec conformance and RECORD hashes, version consistency across tag/PEP 440/workspace, credential scoping, and `release_readiness.py`'s strict asset gate (it uses `iterdir()` and skips subdirectories, so `dist/wheels/` genuinely doesn't disturb it).

siloteemu (changes requested):
🔴 Automated review · pr-review-watcher · 3f27bebb

## Summary
Adds a second distribution channel — a PyPI wheel carrying the prebuilt `rocm`/`rocmd` binaries — via a new 631-line `scripts/build_wheel.py`, release-workflow build/smoke/publish jobs, a Rust guard that stops `rocm uninstall` deleting pip-owned binaries, and honest trust documentation for the weaker provenance story. **Needs work** — the mechanism is well built, but its own test never runs in CI and the uninstall branch it adds is untested. Verified: supply-chain review — I ran the wheel self-test and mutation-tested it (dropping the `S_IFREG|mode` bit, un-fixing the zip timestamp, smuggling an extra member, and accepting arbitrary tags were all caught), built a real wheel and installed it with `uv tool` (both executables detected, mode 0755, `RECORD` rows `../../../bin/rocm`, `pyvenv.cfg` present so the new Rust guard fires, incompatible-platform install correctly refused), confirmed the wheel contains only 2 binaries + METADATA/WHEEL/RECORD/2 licences with **no Python code, no `entry_points.txt`, no sdist, and therefore nothing executing at `pip install` time**, and confirmed no secrets, absolute build paths or internal references in the diff; on the revert question, the six Rust tests do discriminate against a stubbed predicate, but reverting just the 5-line `build_uninstall_plan` branch leaves all six green. Blocking: 3 · Non-blocking: 5.

- **`docs/testing.md:732` (and `.github/workflows/ci.yml`) — the wheel self-test is documented but never runs in CI.** The PR adds `python scripts/build_wheel.py --self-test` to the "Release trust checks" list, but `ci.yml` has no matching step, while the sibling `python scripts/release_readiness.py --self-test` *does* run on both Linux (`ci.yml:228`) and Windows (`ci.yml:360`). `grep -rn build_wheel .github/` returns only the two release-time invocations. So the PR's central artifact — a 631-line builder plus its entire test suite — is exercised only while cutting a real release, which is the worst moment to discover a wheel regression. The reported 21 green checks on this head therefore include zero coverage of `build_wheel.py`. Fix: add a "Wheel builder self-test" step running `python scripts/build_wheel.py --self-test` to both the Linux and Windows `ci.yml` jobs, beside the existing release-readiness step (the self-test is offline and finishes in ~1s, as I confirmed by running it).

- **`apps/rocm/src/main.rs:14892` — the new uninstall branch has no test.** All six new tests call `is_python_managed_layout` or `record_row_targets_script` directly. `build_uninstall_plan` is referenced only at `main.rs:14878` (definition) and `main.rs:14952` (production caller) — never from a test. Delete the `else if is_python_managed_layout(&current_exe) { … }` arm and every new test still passes, so nothing proves that `rocm uninstall` actually skips binary removal, or that the user-facing skip message is emitted. The predicate is protected; the behaviour is not. The obstacle is that `current_exe` is derived in-function from `daemon_binary_path()` (`main.rs:14886`), which is not injectable. Fix: extract the current-exe resolution into a parameter (or an injectable field on `UninstallOptions`/`AppPaths`) and add one test asserting that, for a `pyvenv.cfg`-bearing tree, `plan.skipped` carries the Python-managed message and `plan.actions` contains no binary removals — and that the same tree without `pyvenv.cfg` does remove them.

- **`README.md:96-121` — publishes install commands for a PyPI name the project does not yet own.** `https://pypi.org/pypi/rocm-cli/json` and `https://test.pypi.org/pypi/rocm-cli/json` both return HTTP 404 as of this review, so `rocm-cli` is unregistered on both indexes. Merging a README that tells users `pipx install rocm-cli` / `uv tool install rocm-cli` — on a public repo, before the first wheel ships — advertises an unclaimed name and opens a window in which a third party can register it and serve arbitrary binaries under the project's own documented install command. The "No wheels are published yet" caveat does not close that window, because the command a user copies is identical either way. Fix: claim `rocm-cli` on PyPI **and** TestPyPI and register this workflow as its Trusted Publisher before this README section merges (or hold the section until then), and add that step to `docs/release-trust.md`'s "Remaining Owner Step" section, which currently lists only the signing-key work.

## Non-blocking
- `.github/workflows/release.yml:160` — the `manylinux_2_17_x86_64` tag is legitimately earned (the pre-existing build runs `cargo build --release` inside the manylinux2014 image, `release.yml:118-126`), but nothing verifies it: no `auditwheel`/`objdump -T` check exists anywhere in the repo, so an image bump or a dependency reaching outside the container would silently ship a false compatibility promise that pip enforces on users' behalf.
- `.github/workflows/release.yml:166` — `pipx` and `uv tool` are the README's *recommended* paths, but only plain `pip install` is smoke-tested. I confirmed by hand that `uv tool install` correctly detects both `.data/scripts` executables; wiring one of them into the smoke test would keep that true.
- `docs/release-trust.md:150` — "byte-identical to the copy inside the signed archive" is true by construction (`xtask/src/package.rs:195` uses `fs::copy` with no strip step) but is asserted nowhere; a `sha256sum` comparison between `target/release/rocm` and the packaged copy would turn the claim into a check.
- `apps/rocm/src/main.rs:15133` — the doc comment's fail-closed guarantee holds for an unreadable `lib`/site-packages directory, but not if the environment root itself is non-traversable: every `is_file`/`is_dir` under it then returns `false` and the probe fails *open*. Worth narrowing the comment.
- `scripts/build_wheel.py:180` / `docs/testing.md:741` — the wheel carries no reference to the commit or tag that produced it (provenance rests entirely on the PEP 740 attestation), and the new testing paragraph's "offline and cross-platform too" sits directly after a paragraph ending "Normal Linux and Windows CI run this self-test", which reads as implying a CI parity that does not exist (see blocking item 1).

siloteemu:
🔴 Automated review · pr-review-watcher · 3f27bebb

## Summary

Forced re-check of the change request filed on 2026-09-09. The head has not moved (`3f27beb`), so this is a re-test of our own reasoning, not of new code. I re-derived each count from the tree rather than from the earlier report, and also diffed the branch against current `main` to see whether `main` has overtaken it.

**Outcome: all three counts still hold — `blocking: 3`.** No count was refuted, and no count needs replacing. One count (the CI-coverage one) needs a sharper and more honest statement of its supporting argument, given below; its substance is unchanged.

**Not superseded by `main`.** The branch is the stalest in the set (last commit 2026-08-05) and `main` has moved a long way: 183 files differ against the current default branch. But `git grep` for `pypi`/`PyPI`/`build_wheel`/`pipx`/`uv tool install` across `README.md`, `docs/`, `.github/workflows/` and `scripts/` on current `main` returns **zero** matches. The entire surface this PR adds — the wheel builder, the release build/smoke/publish jobs, the README section, the trust-doc section, and the Rust uninstall guard (`is_python_managed_layout` and `pyvenv` appear nowhere on `main`) — is still net-new. Nothing here has landed independently. The PR does still need a substantial rebase, and `main`'s CI restructuring (below) changes *where* the fix for count 1 goes, not whether it is needed.

**CI conclusions worked from: 21 success, 0 failure.**

## Why this PR's check set is smaller than its siblings'

This is establishable from the repository, without inferring anything about any individual job. The branch carries the pre-split workflow tree: five workflow files, with the self-hosted GPU lanes still declared inline in `ci.yml` and no `docs-build` job. Current `main` carries eight workflow files: the GPU lanes have been moved out into `e2e-selfhosted.yml`, two `dependabot-manifests*` workflows were added, and `ci.yml` gained a `docs-build` job. A branch this far behind therefore declares a different and smaller job set than a freshly-based sibling PR, which is a sufficient repository-level explanation for a smaller check count. I cannot confirm anything about which specific checks ran or what any named job did, and I am not asserting anything about that.

## 🚫 Blocking (3) — all still holding

### 1. The wheel self-test is documented as a release trust check but runs nowhere in CI — **still holds, restated**

Original sentence under test: *"the PR adds `python scripts/build_wheel.py --self-test` to the 'Release trust checks' list, but `ci.yml` has no matching step."*

Re-verified at this head: `git grep build_wheel` over the whole tree returns exactly six hits in `scripts/build_wheel.py` itself, two release-time invocations in `release.yml` (lines 160 and 307), one prose reference in `docs/release-trust.md`, and the new line in `docs/testing.md`. There is no occurrence anywhere under `.github/workflows/ci.yml`. The claim is factually correct.

Two honest qualifications I owe this count, neither of which changes the verdict:

- The "every sibling self-test runs in CI" framing was too strong. Of the six commands in the `docs/testing.md` "Release trust checks" block, the ROCDXG checksum self-test also has no `ci.yml` step, and `main` separately carries a `--self-test`-bearing release script that is likewise unwired. So an unwired self-test is not unprecedented here. What remains true and load-bearing is that the *closest* sibling — `release_readiness.py --self-test`, listed one line above in the same block, and likewise the gate on a release artifact — runs on both the Linux and the Windows job.
- The wheel is not entirely unexercised: the release workflow's "Build Python wheel" and "Smoke test the Python wheel" steps (lines 155/166 Linux, 300/313 Windows) carry no `if:` and therefore run on every release. So `docs/release-trust.md`'s sentence "The wheel build and its smoke test do run on every release, so a wheel regression fails the release job even while publication is off" is **true** — I checked it against the workflow rather than taking it on trust.

What that leaves is the original point, undiluted: 631 lines of new release-critical code plus its entire test suite get their first execution while cutting a real release. And the adjacent prose actively invites the opposite belief. In `docs/testing.md`, the pre-existing paragraph ends "Normal Linux and Windows CI run this self-test before the install-lifecycle E2E scenarios", and the PR's new paragraph opens "The wheel self-test is offline and cross-platform **too**." The "too" attaches the new self-test to a CI-parity claim that does not exist for it. That is a sentence in the diff implying a false fact, and it is what would stop a reviewer from noticing the gap.

**What would resolve it:** add a "Wheel builder self-test" step running `python scripts/build_wheel.py --self-test` to both the Linux and the Windows job in `ci.yml`, beside the existing release-readiness step and inside the same `needs.changes.outputs.heavy == 'true'` guard. Cost is negligible — I ran the self-test once here and it completed in well under a tenth of a second, fully offline. On rebase this lands in the restructured `ci.yml`; `main` still enumerates its self-tests explicitly and has no auto-discovery mechanism and no workflow-contract rule that would pick the new script up, so the rebase does not do this for you. If instead the decision is that this self-test deliberately stays release-only, then the `docs/testing.md` paragraph must drop the "too" and say plainly that it is not run in normal CI.

### 2. The new uninstall branch has no test — **still holds, exactly as written**

Original sentence under test: *"Delete the `else if is_python_managed_layout(&current_exe) { … }` arm and every new test still passes, so nothing proves that `rocm uninstall` actually skips binary removal, or that the user-facing skip message is emitted."*

Re-verified independently:

- `apps/rocm/src/main.rs:14892` is precisely that arm.
- The PR adds six tests. Five call `is_python_managed_layout` directly; one calls `record_row_targets_script` directly. None goes through `build_uninstall_plan`.
- A repo-wide grep for `build_uninstall_plan` returns four hits: the definition (`main.rs:14878`), an internal reference (`main.rs:14952`), and the production caller in `apps/rocm/src/uninstall.rs:22`. Zero test callers, in unit tests, integration tests, or the cucumber suite.
- The e2e suite does not close the gap. The one uninstall scenario in `install_lifecycle.feature` installs through the native bundle installer, so it never produces a `pyvenv.cfg`- or `RECORD`-owned tree and never reaches this arm. No feature file mentions pip/pipx/uv in an uninstall context; grep for `pyvenv`, `pipx` or the skip message across the test tree returns nothing.
- Branch-level mutation is uncaught by construction, not just empirically: deleting the arm removes no symbol or field any test references, so there is no compile error, and `is_python_managed_layout` stays "used" from the test module, so no dead-code lint fires either.
- `current_exe` is still non-injectable — `build_uninstall_plan(paths, options)` takes no exe path, and derives it in-function from `daemon_binary_path()`, which has no test seam.

This is the standing hard-blocker shape: the predicate is well tested (I confirmed previously that the six tests do discriminate against a stubbed predicate), but the behaviour the PR exists to deliver is not. A reader sees six green tests named after the feature and concludes the feature is covered.

**What would resolve it:** make the current-exe resolution injectable — a parameter on `build_uninstall_plan`, or a field on the options/paths struct — and add one test asserting that for a `pyvenv.cfg`-bearing tree `plan.skipped` carries the Python-managed message and `plan.actions` contains no binary removals, and that the same tree *without* `pyvenv.cfg` does remove them. Two assertions, one seam.

### 3. README publishes install commands for a PyPI name the project does not own — **still holds, and the repo itself says so**

Original sentence under test: *"`rocm-cli` is unregistered on both indexes … advertises an unclaimed name and opens a window in which a third party can register it and serve arbitrary binaries under the project's own documented install command."*

Re-checked live, today: `https://pypi.org/pypi/rocm-cli/json` → **404**, `https://test.pypi.org/pypi/rocm-cli/json` → **404**. Still unclaimed on both indexes.

The strongest evidence for this count turns out to be in the PR itself. The gate comment on the `publish-pypi` job states, in the repository's own words, that the gate variable "stays unset until the project owns the PyPI `rocm-cli` project and has registered this workflow as its Trusted Publisher." So the prerequisite is known and written down — but it is written down in the one place a reader looking for outstanding owner actions will not look. `docs/release-trust.md`'s "Remaining Owner Step" section, which exists precisely to enumerate those, still lists only the signing-key and metadata-key work and says nothing about claiming the name or registering the publisher.

That asymmetry is what makes the README section unsafe to merge as-is. The publish job is dormant by default (`if: vars.ROCM_CLI_PUBLISH_PYPI == '1' || == 'true'`), so nothing is uploaded — but the README section is not gated by anything. It becomes public documentation the moment this merges, and the command a user copies is byte-identical whether or not the project owns the name. The "No wheels are published yet" caveat does not close that window.

Related prose worth tightening while fixing this: the "Trusted Publishing (OIDC)" bullet in `docs/release-trust.md` asserts "PyPI accepts `rocm-cli` uploads only from that identity" in the present tense. That property does not exist until the name is claimed and the publisher registered. (I did check the neighbouring specifics and they are accurate — the job really does declare `environment: pypi` and `id-token: write`, passes no credentials, and leaves PEP 740 attestations at the action's default.)

**What would resolve it:** claim `rocm-cli` on PyPI **and** TestPyPI and register this workflow as its Trusted Publisher before the README section merges — or hold that section until then — and add that step to the "Remaining Owner Step" list so the prerequisite lives where owner actions are tracked, not only in a workflow comment.

## On the standing test-quality question

Six tests added, none changed. Five of the six name `is_python_managed_layout` and do genuinely test it — they were previously mutation-checked against a stubbed predicate and discriminate correctly. At branch level, however, the arm inside `build_uninstall_plan` that the whole feature exists to add is deletable with all six still green (count 2). That is the branch-level mutation failure the standing focus calls a hard blocker, and it is why count 2 stays blocking rather than becoming a suggestion.

## Where the code invited a wrong conclusion, and the cheap fix

Nothing of ours was refuted this round, but two places in the diff steer a reader toward a wrong belief, and both will recur:

- The `docs/testing.md` "too" sentence attaches the new self-test to a CI-parity claim belonging to a different script. Cheapest fix: one clause stating whether the wheel self-test runs in normal CI. If it is wired into CI per count 1, the sentence becomes true and the confusion dissolves.
- The PyPI-ownership prerequisite is stated only in a workflow gate comment, while the doc section dedicated to outstanding owner steps omits it, and the README documents the install commands unconditionally. Cheapest fix: one bullet in "Remaining Owner Step".

## Non-blocking (unchanged from the original review, re-confirmed)

- The `manylinux_2_17_x86_64` tag is legitimately earned by building inside the manylinux image, but nothing verifies it; no `auditwheel`/`objdump -T` check exists in the repo, so an image bump could silently ship a false compatibility promise that pip enforces on users' behalf.
- `pipx` and `uv tool` are the README's recommended paths, but only plain `pip install` is smoke-tested.
- "byte-identical to the copy inside the signed archive" is true by construction (the packaging step copies with no strip) but is asserted nowhere; a `sha256sum` comparison would turn the claim into a check.
- The `is_python_managed_layout` doc comment's fail-closed guarantee holds for an unreadable `lib`/site-packages directory but not for a non-traversable environment root, where the probe fails open. Worth narrowing the comment. (The rest of that doc comment I re-read sentence by sentence against the code and it is accurate — including the explicitly-declared RECORD-scoping limitation. This diff's Rust comments are notably better than this repo's average.)
- The wheel carries no reference to the commit or tag that produced it; provenance rests entirely on the PEP 740 attestation.

## Verification note

The full suite was not run here. The evidence above is static analysis, repository grep, live index queries, and one targeted execution of the wheel builder's own self-test. The Rust mutation conclusion in count 2 is derived statically from symbol references rather than by building.

---

**blocking: 3** · non-blocking: 5. Recommendation: the change request stands as filed. Counts 1, 2 and 3 all still hold at this commit; count 1's supporting argument is restated above with two qualifications that do not change its substance. This PR also conflicts with its base and needs a rebase onto a substantially moved default branch regardless.

PR #193, opened 2026-08-07
ci: split self-hosted E2E lanes into their own workflow (EAI-7548)

fredespi:
## Summary

A job queued on an **offline** self-hosted runner cannot be cancelled by GitHub. While the self-hosted GPU E2E lanes shared `ci.yml`'s concurrency group, a superseded run whose GPU job was queued on an offline runner kept holding the group — so the newer run's merge-required (GitHub-hosted) checks sat pending with zero jobs and had to be force-cancelled by hand (seen on #138).

`timeout-minutes` cannot fix this: the job-timeout timer only counts *running* time, never the queued/"waiting for a runner" phase, so it never reaps the stuck job. The only robust fix is structural separation.

## What changed

- **New workflow `e2e-selfhosted.yml`** holds the three self-hosted GPU lanes (`e2e-gpu`, `e2e-gpu-strix-ubuntu`, `e2e-gpu-strix-windows`) plus their own consolidated report, with its **own concurrency group**. An offline runner can now only ever stall this workflow's own supersession, never `ci.yml`'s required checks.
- **`ci.yml`** keeps the GitHub-hosted mock `e2e` lane and its required `E2E consolidated report`. Its shared concurrency group now holds only cancellable hosted jobs, so supersession works cleanly and the required checks always start.
- The self-hosted workflow carries its own trimmed `changes` gate (cross-workflow `needs` is unavailable) and builds the `rocm` binary itself (no `build-and-test` dependency); `ci.yml`'s required `build-and-test` + mock `e2e` remain the authoritative pre-merge build gate.
- **`xtask e2e-report` discovery** now handles `download-artifact@v8`'s single-artifact layout: when exactly one artifact matches, v8 extracts it into the artifacts-dir root (no per-artifact subdir), which each report job now hits after the split. A root-level report is labeled from its `platform.json` slug, or a neutral `Unknown/Unknown` identity when the sidecar is absent (a GPU run can error before writing it) so a hardware failure is never misattributed.
- **New regression tests**: `xtask/src/workflow_contract.rs` (dependency-free) fails if a self-hosted lane reappears in `ci.yml` or the two workflows lose their distinct concurrency namespaces; plus discovery + descriptor tests in `e2e-report`.
- Updated `docs/ci-hardware-testing.md` and the e2e-cucumber README for the split.

## Note for reviewers

The four self-hosted checks remain in the branch's required-status-check list, so an offline runner can still block a *merge* via a missing required check until they are removed from that list — a separate branch-protection change, out of scope for this workflow change.

## Test plan

- [x] `clippy --workspace --all-targets --exclude e2e-cucumber -- -D warnings` clean (Linux)
- [x] Full workspace tests + e2e-cucumber lib tests pass (Linux)
- [x] New `workflow_contract` + `e2e-report` discovery/descriptor tests pass
- [x] Both workflow files parse; all five required check names still produced, one workflow each

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,42 +13,31 @@ on:
   merge_group:
     branches: [main]
     types: [checks_requested]
-  # Manual trigger for on-demand E2E runs on the self-hosted GPU runners, without
-  # waiting on a push/PR or the full gate. Dispatch against any ref
-  # (`gh workflow run ci.yml --ref <branch> -f platform=... -f tier=...`) — GitHub
-  # runs THAT ref's copy of this file, so the E2E jobs it selects live on the ref.
-  # A dispatch skips build-and-test (see its `if:`) for a fast loop. This trigger
-  # must exist on the default branch for the workflow to be dispatchable at all.
+  # Manual trigger for an on-demand mock (GitHub-hosted) E2E run, without waiting
+  # on a push/PR or the full gate. Dispatch against any ref
+  # (`gh workflow run ci.yml --ref <branch> -f platform=mock`) — GitHub runs THAT
+  # ref's copy of this file. A dispatch skips build-and-test (see the e2e `if:`)
+  # for a fast loop. This trigger must exist on the default branch for the
+  # workflow to be dispatchable at all. The self-hosted GPU lanes moved to their
+  # own workflow (e2e-selfhosted.yml) — dispatch THAT for GPU/Strix runs.
   workflow_dispatch:
     inputs:
       platform:
         description: Which runner(s) to target
         type: choice
         default: all
-        options: [all, mock, app-dev-gpu, strix-ubuntu, strix-windows]
+        options: [all, mock]
         # No tier input: the harness resolves pass/xfail/skip per scenario, so a
         # platform runs one job covering everything applicable to it.
-      # Scenario-name regex forwarded to the cucumber harness
-      # (`cargo xtask e2e -- --name <regex>`) so a dispatch can run only the
-      # selected scenarios. Empty = full suite.
-      name_filter:
-        description: "Scenario-name regex (cucumber --name); empty = full suite"
-        type: string
-        default: ""
-      # Opt a manual dispatch into the expensive `@nightly` scenarios (large-model
-      # serve, cold devel install) that the per-PR run skips. Off by default so a
-      # normal dispatch is unchanged; combine with name_filter to probe just one.
-      include_nightly:
-        description: "Include @nightly scenarios (large-model serve, cold install)"
-        type: boolean
-        default: false
 
 concurrency:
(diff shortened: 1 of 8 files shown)


r0x0r (approved):
Strong approve — the change targets the root cause and is unusually well-tested for a CI change.

## What's good

- **Root cause, not symptom.** The #138 stall was an offline (uncancellable) self-hosted job holding `ci.yml`'s shared concurrency group. `timeout-minutes` genuinely can't fix that (it doesn't count queued time), so structural separation into `e2e-selfhosted.yml` with its own concurrency group is the right fix. Confirmed live: the two workflows now run as separate runs with distinct check names, and all three GPU lanes ran green.
- **Excellent regression guarding.** `xtask/src/workflow_contract.rs` is dependency-free and asserts the actual invariant (no `self-hosted`/`amd-gpu`/`strix-halo` labels in `ci.yml`, distinct `concurrency.group` + distinct `name:`, `cancel-in-progress` preserved). It even includes extractor self-tests so the contract can't false-pass on a multiline `runs-on` or folded `group:`.
- **The `download-artifact@v8` single-artifact flattening is handled correctly.** After the split each report job has exactly one artifact, so v8 flattens into the root; `discover()` + `label_for_root_report()` handle both layouts, with tests.
- **The `e2e-unknown-report` neutral fallback is a genuinely good call** — a GPU run that errors before writing `platform.json` must not be misattributed to `Mock/Linux` in the grid. Covered by `parse_descriptor_unknown_is_not_falsely_linux`.
- Report job name collision avoided (`E2E consolidated report` vs `... (self-hosted)`), actions pinned by SHA + version comment, and no new internal-path leaks introduced (the host-specific paths were moved verbatim from `ci.yml`).

Verified locally against the PR head: full `xtask` and `e2e-report` test suites pass, and both workflow YAMLs parse.

## Notes / suggestions (non-blocking)

1. **Duplicated `heavy` filter list.** The `changes` job and its path-filter list are now copied into both workflows (cross-workflow `needs` isn't possible, so this is unavoidable), but nothing guards against drift — if a new heavy path category is later added only to `ci.yml`, the GPU lanes' gating silently diverges. Consider a small contract test asserting the two `heavy:` lists are identical, since `workflow_contract.rs` is already the natural home for it.
2. **The merge-blocking hole is only half-closed by this PR (as noted in the description).** The four self-hosted checks remain in the required-status-check list, so an offline runner can still block a *merge* via a missing required check. This PR fixes the specific #138 supersession stall (its stated scope), but "an offline runner can't block a merge" isn't fully true until the branch-protection change lands. Worth confirming that follow-up is tracked.
3. Minor/defensive: `discover()` additively handles a root `report.json` *and* subdirs in the same directory. That mix shouldn't occur with `download-artifact@v8`, so it's harmless — just noting it isn't a layout the action actually produces.

PR #197, opened 2026-08-10
fix(examine): detect versioned ROCm installs

volen-silo:
## Summary

`rocm examine` and `rocm diagnose` could not see a ROCm install that lives in a version-suffixed directory such as `/opt/rocm-6.4.1`. On Linux the probe tried exactly `/opt/rocm` and then `$ROCM_PATH`, and nothing enumerated `/opt/rocm-*`.

Windows had the mirror-image gap. Its versioned scan lives inside the HIP SDK probe, so the install summary and the JSON probe could not see a versioned HIP SDK at all — without `$ROCM_PATH` they reported nothing.

The user-visible effect on either platform: `rocm_path` and `rocm_version` both come back empty. Three diagnosis checks are guarded on those being non-empty — `fix-3-rocm-kernel`, `fix-6-path` and `fix-8-wheel-rocm` — so they silently score zero. When `fix-6-path` does fire from symptom keywords, its guidance names `/opt/rocm/bin`, which does not exist on that machine.

**Root cause:** two independent install-discovery implementations. The `--json` / `diagnose` probe had its own two-candidate list, and the human `rocm examine` report had a different one. Neither enumerated versioned roots.

**The fix is the sharing** — one resolver, three call sites, both platforms:

- `discover_rocm_installs` searches the platform's install roots for both the conventional root and its versioned installs. Versions are ordered numerically, so `6.10` outranks `6.2`, and candidates are deduplicated by canonical path so the usual `/opt/rocm -> /opt/rocm-6.4.1` symlink is one install rather than two.
- The two platforms arrange versions differently, so the resolver is told which shape it is walking: Linux has `rocm-X.Y[.Z]` siblings of `/opt/rocm`, while the Windows installer nests bare `X.Y` children under its ROCm root. The layouts are kept apart deliberately — accepting a bare number on Linux would sweep in unrelated `/opt` and `/usr/local` entries, and there is a test for exactly that.
- It is used by the JSON/diagnose probe, by the human report's legacy-ROCm summary on both platforms, and by the `fix-6-path` runner, which no longer hardcodes `/opt/rocm/bin`.
- `$ROCM_PATH` now outranks the conventional root. `examine` had this inverted, so a stale default directory beat an install the user named explicitly; the human report already got this right.
- The fallback version is derived from the final path component only. Scanning the whole path matched the first `rocm-` anywhere in it, so an install below a directory such as a checkout named `rocm-cli` reported a bogus version.

### Behaviour changes worth a reviewer's attention

- `examine` now requires the same install markers the human report already required, so a leftover empty `/opt/rocm` is no longer reported as a tarball install. That shared definition of "is an install" is the point of the change. Nothing reads the affected `rocm_install_method` values other than `"amdgpu-install"`, so no diagnosis score moves.
- The human report now lists every install it finds rather than at most two fixed paths.
- On Windows the install summary is now marker-gated too, where it previously accepted any of its three candidate directories that merely existed.

### Non-goals

Completing probe parity for the `--framework` flag and WSL detection-predicate unification is left to follow-up changes. The HIP SDK probe keeps its own lexicographic scan for now, so `6.10` still loses to `6.2` *there*; this PR fixes the ordering only on the paths that go through the shared resolver.

## Test plan

- Fourteen unit tests: versioned-only discovery, numeric-vs-lexical ordering, `$ROCM_PATH` precedence, conventional-root precedence, symlink dedup, marker-gated rejection, `.info/version` precedence, the Windows child layout (including `6.10` beating `6.2` and the root outranking its versioned children), the cross-layout guard, the bare-version parser, and an end-to-end case that drives the resolver into `fix-6-path` and asserts the guidance names the discovered root. That last test was mutation-checked — it fails when versioned discovery is disabled.
- `cargo test -p rocm-core`: 243 pass. Two `proc_lifecycle::tree_*` failures are a known WSL2-local issue on the dev machine and reproduce unchanged on `main`.
- `cargo fmt --check` and workspace `cargo clippy --all-targets -- -D warnings` are clean.
- Manually verified the human `rocm examine` path against a planted versioned install: `detected_unmanaged` with the versioned path present, `not_detected` when absent.

**Verification gap:** the `--json` and `diagnose` entry points short-circuit on WSL2 and the dev machine is WSL2, so those two are covered by unit tests rather than an end-to-end manual run. The Windows layout is likewise covered by unit tests against planted directories, not a real HIP SDK install.

**Risk:** low. The resolver is additive on a conventional box — the conventional root still wins — and the behaviour changes above are the intended ones.

---

- [x] If this PR fixes a bug, searched `tests/e2e-cucumber/expectations.toml` for the fixed ticket ID and removed/narrowed any now-stale xfail rows. No rows reference this behaviour.

### DIFF
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -11294,10 +11294,27 @@ fn render_examine_plain_header(summary: &ExamineSummary) -> String {
     } else {
         "AMD GPU not detected yet"
     };
-    let runtime = match summary.managed_runtime_count {
-        0 => "No ROCm installs saved yet".to_owned(),
-        1 => "1 ROCm install saved".to_owned(),
-        count => format!("{count} ROCm installs saved"),
+    // This line counted only CLI-managed runtimes, so a machine with ROCm
+    // already installed was greeted with "No ROCm installs saved yet" -- read,
+    // reasonably, as "nothing is installed". Now that the resolver finds those
+    // installs and their versions, say what is actually on the machine, and that
+    // a pre-existing install is left alone on purpose rather than missed.
+    let managed = match summary.managed_runtime_count {
+        0 => None,
+        1 => Some("1 ROCm install saved".to_owned()),
+        count => Some(format!("{count} ROCm installs saved")),
+    };
+    let existing = (summary.legacy_rocm.status != "not_detected").then(|| {
+        summary.legacy_rocm.version.as_deref().map_or_else(
+            || "existing ROCm install found, left unmanaged by design".to_owned(),
+            |version| format!("existing ROCm {version} found, left unmanaged by design"),
+        )
+    });
+    let runtime = match (managed, existing) {
+        (Some(managed), Some(existing)) => format!("{managed}; {existing}"),
+        (Some(managed), None) => managed,
+        (None, Some(existing)) => format!("No ROCm installs saved yet; {existing}"),
+        (None, None) => "No ROCm installs saved yet".to_owned(),
     };
     format!(
         "ROCm setup check\n  {gpu}\n  {runtime}\n  Driver: {}\n\nDetails\n",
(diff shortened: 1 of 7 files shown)


juhovainio:
I went through this diff in detail (`lib.rs`, `diagnose.rs`, `examine.rs`, `fix.rs` against `main`) and didn't find anything blocking.

The new resolver logic looks correct: I traced `discover_rocm_installs` end to end (Siblings vs Children layout handling, version parsing/sorting, symlink dedup) and re-ran all 12 new unit tests plus the `diagnose.rs` integration test in a clean worktree — all pass. The bug this PR describes (matching `rocm-` anywhere in a full path instead of just the final component) is genuinely fixed — `rocm_install_version()` now extracts from `resolved.file_name()` before parsing. The Windows JSON-probe gap in `probe_rocm_install` is correctly scoped to `os_family == "linux"`, so that's an intentional non-goal, not an oversight.

One nit I considered but didn't think was worth a formal comment: a doc comment references `RocmLayout::Windows`, but the actual variant is named `Children`. Cosmetic, no functional effect.

On CI: `E2E tests` and `E2E tests (Strix Halo, Ubuntu)` are currently red, but neither looks caused by this change. `E2E tests` fails on a live-serving TTFT-metrics timeout in an unrelated managed-service-metrics scenario, and that same job is green on current `main`, so it reads as a timing flake on the shared runner. `E2E tests (Strix Halo, Ubuntu)` fails in 19s with `mkdir: Permission denied` before any test runs — a self-hosted runner infra issue, not a code problem. Worth a rerun, but I wouldn't hold the PR on either.

This looks safe to approve once CI is green.

volen-silo:
Independently re-verified against the current head (`052e500a`). Resolving a conflict between two earlier review passes on this PR:

**The "RocmLayout::Windows vs Children" doc comment issue is stale — already fixed.** It was live in the first commit but fixed by the second commit here ("Also fixes the doc comment that referenced `RocmLayout::Windows`; the variant is `Children`."). No stray `RocmLayout::Windows` reference exists anywhere in the current tree.

**The Windows fix-6-path parity concern is confirmed and still live.** `run_path_export_windows` (`crates/rocm-core/src/fix.rs:771`) never calls the new `discover_rocm_installs()` resolver — it still calls the pre-existing `newest_rocm_install_dir()` (`fix.rs:939`), which:
- scans `C:\Program Files\AMD\ROCm` and `C:\Program Files (x86)\AMD\ROCm` — a different directory pair than the resolver's own `WINDOWS_ROCM_SEARCH_DIRS` (`lib.rs:2071`)
- sorts versions with a plain `versions.sort()` (lexicographic), so `6.10` still loses to `6.2` here — the exact class of bug this PR fixes elsewhere
- does no install-marker validation, unlike the resolver's `push_rocm_candidate`

This PR's diff to `fix.rs` only touches `run_path_export_linux`. The description's Non-goals section discloses that "the HIP SDK probe" (`probe_hip_sdk_windows` in `examine.rs`) keeps its own lexicographic scan, but doesn't mention that `run_path_export_windows`/fix-6-path has the same gap through a third, separate scan. (Notably, the PR's own new test comment in `lib.rs` says "the bug still live in the Windows scans" — plural — so this was known.) End result: three independent Windows install-scanning implementations (the resolver, `probe_hip_sdk_windows`, `newest_rocm_install_dir`) that disagree on both search paths and sort order.

Worth either calling this out explicitly in the Non-goals section, or routing `run_path_export_windows` through `discover_rocm_installs()` as a follow-up so fix-6-path is actually resolver-backed on both platforms.

Everything else checks out against the current diff: numeric version ordering, `$ROCM_PATH` precedence over the conventional root, symlink dedup, install-marker gating, and the fix-3/fix-6/fix-8 Linux unblocking are all covered by passing unit/e2e tests that match the described behavior.

r0x0r (commented):
Reviewed the full diff. This is a well-structured fix — collapsing three divergent discovery implementations into one ranked resolver is the right call, and the write-up makes the intended behaviour changes easy to audit. The test coverage is genuinely thorough (numeric-vs-lexical ordering, symlink dedup, `$ROCM_PATH` precedence, cross-layout guard, `.info/version` precedence, and the end-to-end `fix-6-path` case driving the real resolver rather than a hand-set `rocm_path`).

**Traced and confirmed correct:**
- Precedence in `discover_rocm_installs_in_layout` — env override → conventional root → versioned (newest-first) — with `push_rocm_candidate` gating on `legacy_rocm_candidate_exists` so a non-existent `$ROCM_PATH` can't produce a phantom top install.
- Canonical-path dedup handles the `/opt/rocm -> /opt/rocm-6.4.1` symlink correctly, and the env-override-points-at-the-symlink-target case still reports once with the symlink path winning.
- Layout split is sound: `probe_rocm_install` is Linux-only (gated on `os_family == "linux"`), so `Siblings` is the only layout it exercises; `Children` is reached solely through `detect_legacy_rocm_summary`. `bare_version` correctly rejects `6`, `6.`, `.4`, `6.4-beta`, keeping unrelated `/opt` entries out of the sibling layout.
- `rocm_install_version` reads only the final canonicalized path component, fixing the "`rocm-` anywhere in an ancestor" false match.
- `#[serde(default)]` on the new `version` field keeps the daemon's pre-existing snapshot loading; serde ignores unknown fields by default, so the reverse (old binary, new snapshot) is fine too.

**Minor, non-blocking observations:**
1. Marker-gating now also applies to `$ROCM_PATH`: the old `examine` probe accepted `$ROCM_PATH` on `is_dir()` alone, so a valid-but-marker-less `$ROCM_PATH` that used to be honoured is now silently dropped. This looks intended (it's the same "shared definition of an install" the PR argues for), but it's a slightly broader behaviour change than the summary spells out — worth a one-line note so nobody's surprised.
2. `run_path_export_linux` returns exit `3` both for "no install found" and "bin dir missing" — same code as before, just two messages now, so callers relying on the exit code are unaffected. Flagging only for completeness.
3. As the PR already calls out, the Windows HIP SDK probe keeps its own lexicographic scan, so `6.10` still loses to `6.2` *there*. Fine as a scoped non-goal; noting it so the follow-up doesn't get lost.

Nothing here blocks. Local gates in the description (`cargo fmt`, workspace clippy `-D warnings`, `rocm-core` tests) plus green CI so far line up with the change. Nice work.

juhovainio (approved):
lgtm this final comment from Roman is probably worth a follow-up:

> As the PR already calls out, the Windows HIP SDK probe keeps its own lexicographic scan, so 6.10 still loses to 6.2 there. Fine as a scoped non-goal; noting it so the follow-up doesn't get lost.

r0x0r (approved):
Approving. All CI is green — including the GPU and Strix Halo (Ubuntu + Windows) E2E lanes, which matter here since this changes install discovery.

The change stands on its merits: one ranked resolver replacing three divergent discovery paths, correct precedence and canonical-path symlink dedup, a sound Linux `Siblings` / Windows `Children` layout split, numeric-not-lexical version ordering, a final-component-only version parse, and backward-compatible (`#[serde(default)]`) snapshot handling. Test coverage is thorough, including the mutation-checked end-to-end `fix-6-path` case.

The `$ROCM_PATH` marker-gating is a real user-visible behaviour change (the old `examine` honoured it on `is_dir()` alone), but it's consistent with the shared "definition of an install" this PR is built around and is called out in the description. Good to merge.

PR #198, opened 2026-08-10
fix(install): stream large downloads to disk with resume and length checks

volen-silo:
- [x] If this PR fixes a bug, searched `tests/e2e-cucumber/expectations.toml` for the fixed ticket ID and removed/narrowed any now-stale xfail rows. (No EAI-7405 rows exist.)

## Problem

The two largest artifacts rocm-cli fetches for itself were both read into memory in full before a byte reached disk:

- the **ROCm SDK tarball** — single-digit gigabytes, buffered by `read_to_end` and then written out again, so peak cost was that much RAM *on top of* the same amount of disk;
- **prefetched model artifacts** — also hashed over the live buffer, so the whole file had to stay resident to compute its digest.

Separately, the one path that did stream wrote **straight to its destination**. An interrupted download therefore left a truncated file exactly where the next run looks for a cached one. The ComfyUI source archive is reused whenever that path exists, with no checksum to catch it, so a half-finished download silently became a "cached" corrupt archive.

There was no retry and no resume anywhere: a transient reset partway through a multi-gigabyte transfer discarded all progress.

## Approach

One streaming primitive in `rocm-core`, with both buffering call sites pointed at it. Bytes move through a fixed 64 KiB buffer and are hashed on the way past, so peak memory no longer scales with the artifact.

- **Atomic publish.** Writes go to a sibling `.part` file and are renamed on success, so the destination only ever holds a complete download. The `.part` file is a sibling to keep the rename within one filesystem.
- **Bounded retry with resume.** Three attempts with exponential backoff. A retry resumes via `Range` only when the server confirms with `206` *and* a `Content-Range` starting at exactly the offset requested; a server that ignores ranges, or honours one at a different offset, restarts cleanly rather than appending at the wrong place.
- **Two independent length contracts.** What arrived is checked against the server's `Content-Length` and, separately, against a size the caller already knew. They fail differently and are handled differently: a short body means the transfer was cut off, so it is transient and the next attempt resumes; a *complete* body of the wrong size means the manifest and the server disagree, which retrying cannot fix.
- **Resume is scoped to one call.** A `.part` file found at entry is deleted rather than resumed from. Process ids get reused, so debris from an earlier run cannot be attributed to this URL, and appending to it would produce a corrupt artifact.

The model-prefetch path also gains a free-space preflight it never had, inherited from the primitive.

## Integrity, stated plainly

The SDK tarball has **no digest to verify against** — TheRock's index publishes only a name and mtime — so its integrity check is the length alone. That catches truncation and interrupted transfers, which is the failure this addresses, but `Content-Length` is unauthenticated and a matching length says nothing about the bytes. A successful download is not evidence the artifact is genuine unless a digest was supplied, and the primitive's doc comment says so. Model artifacts do carry a recipe SHA-256 and are fully verified.

## Scope

`download_file_to_path` keeps its signature and becomes a thin wrapper, so the uv, Lemonade and ComfyUI downloads inherit streaming, atomicity, retry and resume without call-site changes.

Lemonade's cache-lock and publish machinery is deliberately untouched. Its `NamedTempFile` + `verify_sha256` flow now overlaps slightly with the primitive, but folding the two together is a separate refactor with real regression risk and no user-visible gain here.

Extraction-time space checks and the unpreflighted pip/wheel install path are also out of scope — those are about refusing before *extraction*, not before download, and belong in their own change.

## Testing

`cargo fmt --all -- --check` and `cargo clippy --workspace --all-targets -- -D warnings` clean.

Eleven new unit tests drive a scripted loopback server: a large body streamed and digested, an interrupted transfer leaving nothing at the destination, a resume that asks for the right byte offset and reconstructs the artifact exactly, a range-ignoring server restarting cleanly, digest mismatch, wrong declared size, `404` not being retried, and the byte cap. The server's accepted sockets are explicitly set blocking, since on Windows an accepted socket inherits the listener's non-blocking mode and reads fail with `WSAEWOULDBLOCK`.

`cargo test --workspace --all-targets --no-fail-fast` passes apart from three pre-existing failures on untouched files: two `rocm-core` `proc_lifecycle::tests::tree_*` tests that fail on WSL2 for process-tree signalling reasons, and `therock::tests::extracting_the_sdk_archive_removes_it`, which flakes under parallel load on `main` as well and passes 3/3 in isolation here.

**Not run:** a real multi-gigabyte download. This host has no SDK mirror access and no CI lane pulls a 5 GiB tarball. The absence of buffering is guaranteed structurally — no `read_to_end` remains on any artifact path — and the streaming, truncation and resume behaviours are pinned by the tests above, but the real transfer has not been exercised.

No e2e change: an e2e resume test against `rocm install` would first need a download-base override, since the TheRock URLs are hardcoded constants and `ROCM_CLI_DOWNLOAD_BASE` only redirects `install.sh`/`install.ps1`.

Closes EAI-7405.

### DIFF
--- a/crates/rocm-core/src/lib.rs
+++ b/crates/rocm-core/src/lib.rs
@@ -329,12 +329,23 @@ fn download_attempt(
             url = request.url
         )));
     }
+    // A `206` that does not confirm continuing from `resume_from` cannot be
+    // treated as a fresh full download either: its `Content-Length` is the
+    // length of whatever slice the server chose to send, not the whole
+    // artifact, so accepting it as complete would silently rename a truncated
+    // body into place. Discard the partial file and restart clean with a
+    // plain `GET` instead of reinterpreting the mismatched slice.
+    if status == 206 && resume_from > 0 && content_range_start(&response) != Some(resume_from) {
+        let _ = fs::remove_file(partial_path);
+        return Err(transient(anyhow::anyhow!(
+            "{url} resumed from an unexpected byte offset; restarting the download",
+            url = request.url
+        )));
+    }
     // Only append when the server confirmed it is continuing from exactly where
     // we stopped. A server that ignores `Range` answers 200 with the whole body,
-    // and one that honours it differently than asked would have us append at the
-    // wrong offset — both restart cleanly instead.
-    let resuming =
-        status == 206 && resume_from > 0 && content_range_start(&response) == Some(resume_from);
+    // which restarts cleanly too.
+    let resuming = status == 206 && resume_from > 0;
     let remaining_len = header_u64(&response, "Content-Length");
     let total_len =
         remaining_len.map(|len| len.saturating_add(if resuming { resume_from } else { 0 }));
@@ -7385,6 +7396,11 @@ mod tests {
         },
         /// Honour `Range` and serve the remainder as `206`.
         Resume,
+        /// Answer `206` but from `start` regardless of what `Range` asked for,
+        /// as a non-compliant server or a broken caching proxy might.
+        ResumeAtWrongOffset {
+            start: usize,
+        },
         /// Ignore `Range` and answer `200` with the whole body, as a server
         /// without range support does.
         IgnoreRange,
@@ -7444,6 +7460,16 @@ mod tests {
                     }
                     DownloadReply::Resume => {
                         let start = range_start.min(total);
+                        write!(
+                            stream,
+                            "HTTP/1.1 206 Partial Content\r\nAccept-Ranges: bytes\r\nContent-Range: bytes {start}-{}/{total}\r\nContent-Length: {}\r\nConnection: close\r\n\r\n",
+                            total.saturating_sub(1),
+                            total - start
+                        )?;
+                        stream.write_all(&body[start..])?;
+                    }
+                    DownloadReply::ResumeAtWrongOffset { start } => {
+                        let start = start.min(total);
                         write!(
                             stream,
                             "HTTP/1.1 206 Partial Content\r\nAccept-Ranges: bytes\r\nContent-Range: bytes {start}-{}/{total}\r\nContent-Length: {}\r\nConnection: close\r\n\r\n",
@@ -7598,6 +7624,62 @@ mod tests {
             "the retry must ask to continue from byte 5000: {}",
             requests[1]
         );
+        Ok(())
+    }
+
+    #[test]
+    fn download_restarts_cleanly_when_resume_lands_at_the_wrong_offset() -> Result<()> {
+        // The second reply answers `206` from byte 8000, not the byte 5000 we
+        // asked to resume from. Accepting that slice's own `Content-Length` as
+        // the whole artifact would silently rename a corrupt file into place;
+        // instead the third attempt must be a plain `GET` that refetches the
+        // whole thing.
+        let body = download_body();
+        let (port, server) = spawn_download_server(
+            body.clone(),
+            vec![
+                DownloadReply::Truncated { sent: 5000 },
+                DownloadReply::ResumeAtWrongOffset { start: 8000 },
+                DownloadReply::Complete,
+            ],
+        )?;
+        let dir = download_scratch("wrong-offset");
+        let destination = dir.join("artifact.bin");
+
+        let outcome = download_file_streaming(&DownloadRequest::new(
+            &format!("http://127.0.0.1:{port}/artifact.bin"),
+            &destination,
+            Duration::from_secs(10),
+        ))?;
+
+        let written = fs::read(&destination)?;
+        let requests = server.join().expect("server thread")?;
+        let leftovers: Vec<_> = fs::read_dir(&dir)?
+            .filter_map(|entry| entry.ok().map(|entry| entry.file_name()))
+            .filter(|name| name.to_string_lossy().contains(".part"))
+            .collect();
+        fs::remove_dir_all(&dir).ok();
+
+        assert_eq!(
+            written, body,
+            "a mismatched-offset resume must not be accepted as a complete artifact"
+        );
+        assert_eq!(
+            outcome.sha256,
+            format!("{:x}", Sha256::digest(&body)),
+            "the digest must cover the whole artifact from the clean restart"
+        );
+        assert_eq!(
+            requests.len(),
+            3,
+            "the wrong-offset reply must be discarded and retried, not accepted"
+        );
+        assert!(
+            !requests[2].contains("Range:"),
+            "the restart after a wrong-offset resume must be a plain GET: {}",
+            requests[2]
+        );
+        assert!(leftovers.is_empty(), "the .part file must not survive");
         Ok(())
     }
 
--- a/crates/rocm-core/src/lib.rs
+++ b/crates/rocm-core/src/lib.rs
@@ -4,6 +4,7 @@
 
 use anyhow::{Context, Result, bail};
 use serde::{Deserialize, Serialize};
+use sha2::{Digest, Sha256};
 use std::collections::BTreeMap;
 #[cfg(windows)]
 use std::ffi::OsStr;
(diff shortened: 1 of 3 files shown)


crates/rocm-core/src/lib.rs
```diff
@@ -118,35 +119,371 @@ pub fn parse_http_endpoint(endpoint_url: &str) -> Option<(String, u16)> {
...
+    }
+    // Only append when the server confirmed it is continuing from exactly where
+    // we stopped. A server that ignores `Range` answers 200 with the whole body,
+    // and one that honours it differently than asked would have us append at the
+    // wrong offset — both restart cleanly instead.
+    let resuming =
+        status == 206 && resume_from > 0 && content_range_start(&response) == Some(resume_from);
+    let remaining_len = header_u64(&response, "Content-Length");
+    let total_len =
+        remaining_len.map(|len| len.saturating_add(if resuming { resume_from } else { 0 }));
```
juhovainio, line 351: `resuming` (line 336) requires both `status == 206` and the `Content-Range` start to match `resume_from`. If a non-compliant server answers `206` but at the wrong offset, `resuming` is correctly `false`, so the code falls through to the "fresh download" path: it reopens the file with `fs::File::create` (discarding what was on disk) and treats the response body as the whole artifact.

The problem is that `total_len` here is computed only from *this response's* `Content-Length`, which for a `206` is the length of whatever slice the server actually sent, not the full artifact size. If that slice is shorter than the real artifact (which it will be, since the server started partway through), the later `written != expected` check compares against this same short `total_len` and passes. The truncated body then gets renamed into place as if it were complete. Without `expected_sha256` set, nothing else catches this. Might be worth treating a `206` with a non-matching offset as a hard restart (delete the partial file, retry as a plain `GET` without a `Range` header) rather than silently accepting whatever partial body comes back as the full artifact.

crates/rocm-core/src/lib.rs
```diff
@@ -118,35 +119,371 @@ pub fn parse_http_endpoint(endpoint_url: &str) -> Option<(String, u16)> {
...
-            &format!("download {url}"),
-            destination,
-            disk_space::with_margin(content_length),
-        )?;
+    let partial_path = partial_download_path(request.destination);
+    // Resume is scoped to retries within this call. A `.part` file found at
+    // entry is debris from an earlier run — process ids get reused, so its
+    // contents cannot be attributed to this URL and appending to it would
+    // silently produce a corrupt artifact.
+    let _ = fs::remove_file(&partial_path);
```
juhovainio, line 232: This only removes the current process's own `.part-<pid>` file. If a process is killed after this line runs but before the final `fs::rename` (e.g. mid-transfer, or between finishing the write loop and the rename), that `.part-<pid>` file is orphaned: nothing else in the codebase scans for and reclaims stale `.part-*` files left by pids that never run again. Over many interrupted runs (crashes, OOM kills, `kill -9` during a long download) these can accumulate indefinitely in the destination directory. Not a blocker for this PR's core functionality, but worth a follow-up (a startup sweep, or a max-age cleanup) so disk usage doesn't quietly grow over time on machines that do a lot of installs.

juhovainio (changes requested):
I reviewed the new streaming-download code in rocm-core and found two issues worth fixing before merge, left as inline comments below. Both are edge cases in the resume/retry path, not style nitpicks.

The rest of the change looks solid: chunked copy with a fixed buffer keeps memory flat regardless of artifact size, the atomic rename-on-completion means a caller never sees a partial file at the destination path, and the new unit tests exercise a good spread of server behaviors (early close, wrong-offset resume, non-retryable 4xx, etc.) against a real loopback server rather than mocks.

crates/rocm-core/src/lib.rs
volen-silo: Fixed in 2960bccd: a 206 that doesn't confirm the requested resume offset is now treated as a hard restart — the .part file is discarded and the retry goes out as a plain GET, rather than accepting the mismatched slice's own Content-Length as though it described the whole artifact. Added a regression test (download_restarts_cleanly_when_resume_lands_at_the_wrong_offset) covering this.

crates/rocm-core/src/lib.rs
volen-silo: Agreed this is worth doing but not a blocker for this PR — leaving it as a follow-up (a startup sweep or max-age cleanup for stale .part-* files) rather than scope-creeping this change.

juhovainio:
Verified the fix in 2960bccd against the current code, not just the reply text.

The wrong-offset-206 case now returns a transient error and removes the `.part-<pid>` file (`crates/rocm-core/src/lib.rs`, the `status == 206 && resume_from > 0 && content_range_start(&response) != Some(resume_from)` branch). Since the partial file is gone, the next attempt reads `resume_from == 0` and goes out as a plain GET, so `total_len` is never computed from the truncated slice's `Content-Length`. That fully closes the gap I originally flagged. There's also a direct regression test, `download_restarts_cleanly_when_resume_lands_at_the_wrong_offset`, which I ran locally along with the rest of the download suite (`cargo test -p rocm-core download`) — all 13 pass.

Leaving the orphaned `.part-<pid>` cleanup as a follow-up rather than a blocker here seems reasonable to me too — it's a slow leak from crashed/killed runs, not a correctness bug in this PR's own change, and scoping it separately keeps this diff focused.

CI is green (one Windows GPU e2e job still pending as I write this). Looks good to approve from my side.

juhovainio (approved):

rominf (commented):
The fix juhovainio flagged is genuinely closed, and I verified the resume/length/digest core independently rather than reading the diff: the classic off-by-whole-file bug is **not** present (`written` is seeded by re-hashing the on-disk prefix, so `written != expected` compares total against total), the digest covers the whole artifact on resume, a server ignoring `Range` and answering `200` truncates and restarts cleanly, and no exit path lets a truncated file reach the destination.

One thing I'd like on the record before this merges, because it is the same bug in mirror form.

### An unsolicited `206` still slips through

`crates/rocm-core/src/lib.rs:338`:

```rust
if status == 206 && resume_from > 0 && content_range_start(&response) != Some(resume_from) {
```

The `resume_from > 0` conjunct means a `206` returned when **no `Range` header was sent** skips the guard entirely. `resuming` is then `false`, so `total_len` is taken from the slice's own `Content-Length`, and the `written != expected` check at `:445` passes trivially on a truncated body — which is then renamed into place. That is the failure this PR exists to prevent, and it bites hardest on the SDK tarball, which passes neither `expected_len` nor `expected_sha256` (`apps/rocm/src/therock.rs:2106`), so length is its only integrity signal.

Reachability is a range-caching proxy or CDN replaying a cached `206` at a header-less GET — uncommon, and this PR is a large net improvement over the buffered status quo either way, so I would not hold the merge for it. But it shouldn't be forgotten, because the failure is silent.

Note that dropping the `resume_from > 0` conjunct is not sufficient on its own: `content_range_start` (`:488`) parses only the start offset and discards `/total`, so a `206` carrying `Content-Range: bytes 0-1023/5000` would still satisfy `start == resume_from == 0` and be accepted as a complete 1024-byte artifact. The full fix derives `total_len` from the `/total` field when `status == 206`, and requires the returned range to actually cover `[resume_from, total-1]`.

### Smaller things

- `apps/rocm/src/therock.rs:2137` is now false — it says `download_file` "still checks the real, buffered body length before writing", but removing that buffering is the point of this PR, and the check now happens *after* writing to `.part`. This matters more than typical comment rot because it is the stated justification for discarding an implausible HEAD length, so it misdirects a reader about where the backstop actually lives.
- `apps/rocmd/Cargo.toml:23` — `ureq` is now entirely unused in `apps/rocmd/src/` (its only consumer, `download_artifact_bytes`, was deleted). There's no dangling `use`, so clippy can't catch it.
- `apps/rocmd/Cargo.toml:21` — `sha2` is now test-only (`#[cfg(test)] use sha2::…` at `apps/rocmd/src/lib.rs:36`, sole use at `:1037`). Cargo's dependency graph isn't cfg-aware, so it still compiles into every production `rocmd` build; it belongs in `[dev-dependencies]`.
- The retry loop logs nothing and discards the intermediate `anyhow::Error` — only `retryable` is read. With a 1-hour timeout and 3 attempts, a user can wait ~3 hours with no output, and an early connection reset is never surfaced if a later attempt fails differently. A one-line message on retry would fix both.
- `spawn_download_server` has no accept/join timeout, so a regression in round-trip count will make the count-asserting tests hang rather than fail.

PR #200, opened 2026-08-10
fix(examine): report a real OS version instead of the OS name

volen-silo:
- [x] If this PR fixes a bug, searched `tests/e2e-cucumber/expectations.toml` for the fixed ticket ID and removed/narrowed any now-stale xfail rows. No rows reference this behaviour.

## Problem

`Examination::os_version` held `std::env::consts::OS` — the OS *name*. So the field just repeated `os_family`, and a reader of `rocm examine --json` learned nothing from it:

```json
"os_family": "linux",
"os_version": "linux",
```

## Fix

Populate it with an actual version, mirroring what `examine.py`'s `platform.version()` reports: `uname -v` on Linux, `cmd /C ver` on Windows.

On a host the CLI does not support the field stays empty rather than being guessed at. An empty value reads as "not collected"; a wrong one would be reasoned over by the diagnosis catalog.

No field was added or removed, so the frozen wire contract with the catalog is unchanged.

## Scope

This was originally part of #197 (versioned ROCm install detection). It is a genuine fix but an unrelated one — that PR is about install discovery — so it is split out here to keep each change to one concern. #197 no longer touches `os_version`.

## Test plan

`cargo fmt --check` and `cargo clippy -p rocm-core --all-targets -- -D warnings` clean. `cargo test -p rocm-core` passes apart from the two known WSL2-local `proc_lifecycle::tree_*` failures, which reproduce unchanged on `main`.

The new test asserts the invariant that holds on every platform — `os_version` is never the OS name — then branches: non-empty on Linux and Windows, deliberately empty elsewhere. Written that way on purpose: a flat non-empty assertion would fail for anyone running the suite on macOS, where the field is intentionally blank.

**Verified on Linux only.** The Windows branch shells out to `cmd /C ver` and has not been run on a Windows host; CI's `windows-build-and-test` is the check that exercises it.

### DIFF
--- a/crates/rocm-core/src/examine.rs
+++ b/crates/rocm-core/src/examine.rs
@@ -405,7 +405,20 @@ const MEDIUM: Duration = Duration::from_secs(8);
 // ---------------------------------------------------------------------------
 
 fn probe_os(e: &mut Examination) {
-    e.os_version = std::env::consts::OS.to_owned();
+    // The OS *version*, mirroring examine.py's `platform.version()`. This used
+    // to hold `std::env::consts::OS`, which is the OS *name* and so merely
+    // repeated `os_family`, telling a reader of the report nothing.
+    //
+    // Left empty on platforms the CLI does not support rather than guessed at:
+    // an empty field reads as "not collected", while a wrong one would be
+    // reasoned over by the diagnosis catalog.
+    e.os_version = if runtime_is_linux() {
+        run("uname", &["-v"], SHORT).1.trim().to_owned()
+    } else if runtime_is_windows() {
+        run("cmd", &["/C", "ver"], SHORT).1.trim().to_owned()
+    } else {
+        String::new()
+    };
     if runtime_is_linux() {
         e.os_family = "linux".to_owned();
         e.kernel_release = run("uname", &["-r"], SHORT).1.trim().to_owned();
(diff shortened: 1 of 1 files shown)


r0x0r (commented):
Reviewed the diff against `crates/rocm-core/src/examine.rs`. This is a clean, correctly-scoped fix and reads as ready to me.

**What's right**
- Root cause is addressed at the right layer: `os_version` no longer mirrors `std::env::consts::OS` (which just repeated `os_family`). It now carries a real version via `uname -v` / `cmd /C ver`, matching `examine.py`'s `platform.version()`.
- Consistent with the surrounding probes — same `run(program, args, SHORT).1.trim()` shape as the adjacent `kernel_release = run("uname", &["-r"], SHORT)`.
- Wire contract preserved: the field is populated, not added/removed/renamed, so the frozen `examination_top_level_keys_match_examine_py_contract` set is untouched. Confirmed nothing in the repo (`.toml`/`.py`/`.md`) pattern-matches on the old `"linux"` value, so the diagnosis catalog isn't affected.
- Empty-on-unsupported is the safe choice and is documented in-code: reads as "not collected" rather than a guess the catalog would reason over.
- Regression test is well-constructed — asserts the cross-platform invariant (never the OS name) and branches supported vs. unsupported, so it won't spuriously fail on macOS.

**Notes (non-blocking)**
- WSL is covered implicitly: `uname -v` returns a genuine kernel version there, so the "never the OS name" invariant still holds.
- Windows branch is verified only by CI (`windows-build-and-test` / Strix Halo Windows), consistent with the PR's stated test plan. `cmd /C ver` typically emits a leading blank line; `.trim()` handles that.

LGTM. Deferring formal approval to a maintainer.

r0x0r (approved):
Approving. The fix is correct, minimally scoped, consistent with the existing probe patterns, preserves the frozen wire contract, and ships with a platform-aware regression test. Confirmed nothing in the repo pattern-matches the old os_version value, so the diagnosis catalog is unaffected.

PR #201, opened 2026-08-10
fix(dash): keep scrape warnings between scrape ticks

mikeroySoft:
## Problem

The dashboard's header warning badge (`⚠ N`) blinks on and off instead of staying lit while a fault persists.

`runner.rs` rebuilds `warnings` from scratch on every GPU tick, but the docker-discovery, vLLM/Lemonade-scrape and VRAM-attribution warnings are only *produced* on their own slower cadences (`discovery_ticks` / `instance_ticks`). A permanent failure therefore lands in one tick in N and is absent from the rest.

Captured from a live daemon socket on an unpatched build, consecutive snapshots alternated:

```
["vllm scrape: 1 failed (last: transport: GET http://127.0.0.1:11435/metrics ...)"]
[]
["vllm scrape: 1 failed (last: transport: GET http://127.0.0.1:11435/metrics ...)"]
[]
```

## Change

Hold each cadence-gated warning until its cadence runs again:

- docker discovery keeps its last result in `docker_warn` (cleared on a successful discovery),
- the instance-cadence producers (vLLM scrape, Lemonade scrape, VRAM attribution fallback) share `scrape_warns`, cleared once per instance tick.

Per-tick warnings (`amd-smi metric`, `tokens_per_watt` id-join misses) are unchanged.

## Verification

- New regression test `scrape_warning_persists_between_scrape_ticks` in `crates/rocm-dash-daemon/tests/end_to_end.rs`: registers a managed vLLM service on a closed port and asserts an unbroken run of warned snapshots. Confirmed **FAILS** on `main` and passes with this change.
- Live daemon after the fix: 43/43 consecutive snapshots carried the warning, steady.
- `cargo test -p rocm-dash-daemon`, `cargo clippy -p rocm-dash-daemon --all-targets -- -D warnings`, `cargo fmt --all -- --check`: clean.

## Notes

- [x] Searched `tests/e2e-cucumber/expectations.toml`; no xfail rows relate to this behavior.

### DIFF
--- a/crates/rocm-dash-daemon/src/runner.rs
+++ b/crates/rocm-dash-daemon/src/runner.rs
@@ -220,6 +220,11 @@ pub async fn run_loop(
 
     let mut tick_count: u64 = 0;
     let mut last_sysinfo_refresh: u64 = 0;
+    // Discovery/scrape warnings are produced on their own (slower) cadences, so
+    // holding them only for the tick that produced them makes the header ⚠ badge
+    // blink on and off. Keep the last result until that cadence runs again.
+    let mut docker_warn: Option<String> = None;
+    let mut scrape_warns: Vec<String> = Vec::new();
 
     loop {
         ticker.tick().await;
@@ -321,8 +326,9 @@ pub async fn run_loop(
                         );
                     }
                     known_instances = seen;
+                    docker_warn = None;
                 }
-                Err(e) => warnings.push(format!("docker discover: {e}")),
+                Err(e) => docker_warn = Some(format!("docker discover: {e}")),
             }
         }
 
@@ -459,6 +465,9 @@ pub async fn run_loop(
             }
             known_services = disc.seen;
         }
+        if tick_count.is_multiple_of(instance_ticks) {
+            scrape_warns.clear();
+        }
 
         // Per-instance vLLM metric scrape, parallel, on its own cadence. The
         // Lemonade instance (if any) is scraped via its own collector below, so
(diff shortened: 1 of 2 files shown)


michaelroy-amd:
Written and pushed from my testing machine... trying to fix bugs as I catch them and ship it so I don't loose sight of it! I'll defer to someone else to verify the PR tho.

juhovainio:
Went through this carefully since it touches the tick loop's warning bookkeeping across three separate cadences. The logic holds up:

- `docker_warn` is cleared only inside the `Ok` branch of a discovery run, so it correctly stays sticky through the ticks where discovery doesn't run and clears as soon as a real successful discovery happens.
- `scrape_warns.clear()` sits right before the vLLM/Lemonade/VRAM-fallback blocks and is gated on the exact same `tick_count.is_multiple_of(instance_ticks)` condition those three blocks use, so the clear-then-refill happens atomically within one tick rather than leaving a window where the vector is emptied but not yet repopulated.
- The per-tick warnings (amd-smi metric, tokens_per_watt id-join misses) are untouched, which matches the PR description, and I don't see any of the four converted call sites (docker discover, vllm scrape, lemonade scrape, vram attribution) being missed or double-converted.

The new regression test reproduces the exact bug shape (asserting an unbroken run of warnings rather than just "a warning shows up at some point"), and the described before/after behavior (fails on main, passes here) is exactly what I'd expect given the code change.

Looks safe to approve.

juhovainio (approved):

PR #202, opened 2026-08-10
fix(serve): wait on the engine's readiness budget, fail fast on a dead engine

mikeroySoft:
## Problem

`rocm serve` reports a healthy vLLM launch as a failure.

`start_managed_service` and `restart_internal_managed_service` both waited a hardcoded `Duration::from_secs(45)`, while vLLM's own startup budget is 5 minutes (`DEFAULT_VLLM_READY_TIMEOUT`, tunable via `ROCM_CLI_VLLM_READY_TIMEOUT_SECS`). A cold vLLM start — torch import, aiter/triton JIT, KV-cache warmup — routinely runs past a minute, so the CLI gave up while the engine was still legitimately starting:

```
14:38:37  vllm boot
14:39:22  rocm serve gives up  ->  "Deployment summary (not ready yet)"
14:39:31  Application startup complete  ->  service actually ready
```

The user-visible result is a scary "the server did not report ready before the startup timeout" note on a server that is fine nine seconds later — which reads as "vLLM does not work here".

## Change

1. **Take the budget from the engine.** New public `rocm_engine_vllm::ready_timeout()`; `managed_ready_timeout(engine)` uses it for vLLM and keeps 45 s elsewhere. The CLI can no longer contradict the engine it is waiting on. Applied to both the launch and restart paths.
2. **Keep crashes fast.** A longer budget must not slow down a real failure, so `on_tick` can now return `false` to end the wait; `start_managed_service` does that once the engine records a terminal status, and reports `failed`.
   Process liveness is *not* usable as the signal here: the supervisor is our own child, so `process_is_running` (a `kill(pid, 0)`) keeps reporting it alive while it sits unreaped as a zombie. Measured: that approach spun the full 5 minutes on an engine that had already died. The engine's own state file is the reliable signal.
3. **Say which failure it was.** The summary note now distinguishes `failed` ("the server exited during startup … the engine error is at the end of its log") from `running`/`starting` ("may still be loading").

## Verification

Hardware: Radeon AI PRO R9700 (`gfx1201`), TheRock nightly wheel runtime, vLLM 0.23.0.

| Scenario | Before | After |
|---|---|---|
| `Qwen/Qwen3-0.6B` | "not ready yet", no metrics | `status ready`, TTFT 27 ms, 166.9 tok/s |
| model crashing in engine-core init | 5 min spinner → "may still be loading" | **7.9 s** → `status failed` → note points at the engine error |
| model crashing 81 s in (KV-cache sizing) | "not ready" at 45 s, real cause hidden | waits past 45 s, reports the real outcome |

Tests:

- `managed_ready_timeout_follows_the_engine_budget` — pins the budget to the engine's and to `> 45 s`.
- `readiness_wait_stops_when_the_tick_gives_up` — the wait ends at the first refusal and still reports how far the endpoint got.
- `failed_launch_note_points_at_the_engine_error` / `starting_launch_note_says_it_may_still_be_loading`.

Gates: `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all -- --check`, `python3 scripts/smoke_local.py` — all clean.

`cargo test --workspace --all-targets` is green except `therock::tests::extracting_the_sdk_archive_removes_it`, which fails identically on unmodified `main` in my shell because `tar` is not on `PATH` there (`Error: failed to launch tar`). Unrelated to this change; CI has `tar`.

## Notes

- [x] Searched `tests/e2e-cucumber/expectations.toml`; no xfail rows relate to this behavior.

### DIFF
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -4824,16 +4824,37 @@ fn start_managed_service(
     #[cfg(windows)]
     thread::sleep(Duration::from_millis(200));
 
+    // Wait on the engine's own readiness budget, not a shorter CLI-side one: a
+    // cold vLLM start (torch import, aiter/triton JIT, KV-cache warmup) routinely
+    // runs past a minute, and reporting "not ready" for a server that answers ten
+    // seconds later reads as a failed launch. Watching the engine's own state file
+    // keeps a genuine crash fast instead of spinning out the whole budget: the
+    // supervisor is our child, so `process_is_running` still reports it alive
+    // while it sits unreaped as a zombie, but the engine writes a terminal status
+    // before it exits.
     let readiness = wait_for_service_http_ready_with_progress(
         engine,
         host,
         port,
         &resolve.canonical_model_id,
         endpoint_api_key,
-        Duration::from_secs(45),
-        on_wait_tick,
+        managed_ready_timeout(engine),
+        &mut |elapsed| {
+            on_wait_tick(elapsed);
+            let _ = record.refresh_from_engine_state();
+            managed_service_running_state(&record.status) != "not_running"
+        },
     );
-    let launch_status = status_for_readiness(readiness);
+    // A launch the engine already gave up on is failed, not merely slow — but a
+    // reachable endpoint still wins, since a service that answers is serving
+    // whatever its state file last claimed.
+    let launch_status = if readiness == EndpointReadiness::Unreachable
+        && managed_service_running_state(&record.status) == "not_running"
+    {
+        "failed"
+    } else {
+        status_for_readiness(readiness)
+    };
     record.status = launch_status.to_owned();
     if readiness == EndpointReadiness::Serving {
         // Latch the verification the wait just performed, so the readiness checks
@@ -13038,7 +13059,7 @@ fn restart_internal_managed_service(
         record.port,
         &record.canonical_model_id,
         endpoint_api_key.as_deref(),
-        Duration::from_secs(45),
+        managed_ready_timeout(&record.engine),
     );
     record.status = status_for_readiness(readiness).to_owned();
     if readiness == EndpointReadiness::Serving {
@@ -16189,6 +16210,17 @@ fn apply_app_path_env(command: &mut ProcessCommand, paths: &AppPaths) {
     }
 }
 
+/// Readiness budget for a managed launch, taken from the engine's own startup
+/// timeout so the CLI never contradicts it. vLLM's is 5 minutes by default and
+/// tunable via `ROCM_CLI_VLLM_READY_TIMEOUT_SECS`; other engines start fast
+/// enough that the historical 45 s is still generous.
+fn managed_ready_timeout(engine: &str) -> Duration {
+    match engine {
+        "vllm" => rocm_engine_vllm::ready_timeout(),
+        _ => Duration::from_secs(45),
+    }
+}
+
 fn wait_for_service_http_ready(
     engine: &str,
     host: &str,
@@ -16204,7 +16236,7 @@ fn wait_for_service_http_ready(
         canonical_model_id,
         endpoint_api_key,
         timeout,
-        &mut |_elapsed| {},
+        &mut |_elapsed| true,
     )
 }
 
@@ -16219,10 +16251,12 @@ fn wait_for_service_http_ready(
 /// result seen before `timeout` is what gets returned.
 ///
 /// Polls until `timeout` elapses, invoking `on_tick(elapsed)` once per iteration
-/// so a caller can animate a spinner. Engine-neutral: `service_http_readiness_paths`
-/// maps each engine to the right listing path. `endpoint_api_key` is sent as a
-/// bearer token so both the listing check and the inference probe still succeed
-/// against a public endpoint that requires authentication.
+/// so a caller can animate a spinner and, by returning `false`, abandon a wait it
+/// knows is hopeless (e.g. the engine died) rather than burn the whole budget.
+/// Engine-neutral: `service_http_readiness_paths` maps each engine to the right
+/// listing path. `endpoint_api_key` is sent as a bearer token so both the listing
+/// check and the inference probe still succeed against a public endpoint that
+/// requires authentication.
 fn wait_for_service_http_ready_with_progress(
     engine: &str,
     host: &str,
@@ -16230,7 +16264,7 @@ fn wait_for_service_http_ready_with_progress(
     canonical_model_id: &str,
     endpoint_api_key: Option<&str>,
     timeout: Duration,
-    on_tick: &mut dyn FnMut(Duration),
+    on_tick: &mut dyn FnMut(Duration) -> bool,
 ) -> EndpointReadiness {
     let start = std::time::Instant::now();
     let endpoint = format_http_base_url(host, port);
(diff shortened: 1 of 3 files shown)


apps/rocm/src/main.rs
```diff
@@ -13038,7 +13059,7 @@ fn restart_internal_managed_service(
         record.port,
         &record.canonical_model_id,
         endpoint_api_key.as_deref(),
-        Duration::from_secs(45),
+        managed_ready_timeout(&record.engine),
```
juhovainio, line 13062: restart_internal_managed_service gets the longer engine-derived timeout but not the fail-fast half of this fix. It still calls the plain `wait_for_service_http_ready`, whose `on_tick` closure is now `&mut |_elapsed| true` (see the `wait_for_service_http_ready` diff below) — it never returns `false`, so nothing can end the wait early the way `start_managed_service`'s new closure does.

The PR description frames both call sites as sharing the same problem ("`start_managed_service` and the restart path both waited a hardcoded 45 s"), but only `start_managed_service` got the compensating fix. The practical effect: restarting a service whose engine crashes immediately used to report back in 45 s; now, for vLLM, it takes up to `ready_timeout()` (5 minutes by default) to report the same outcome, since nothing here watches `record.refresh_from_engine_state()` the way the launch path does. It also means a restart can never come back with `status: "failed"` — `status_for_readiness` only maps to `ready`/`running`/`starting`, so a crashed restart still reports `starting` after sitting through the full budget.

Worth passing a closure here that mirrors `start_managed_service`'s (refresh the record from the engine state file each tick, stop once `managed_service_running_state(&record.status) == "not_running"`), and applying the same `Unreachable && not_running -> "failed"` mapping to the result.

juhovainio (changes requested):
The launch-path fix (start_managed_service) looks correct: I traced managed_service_running_state, refresh_from_engine_state, and EndpointReadiness against the current code, and the early-bailout logic is sound (a failed refresh_from_engine_state read leaves record.status untouched, so a transient I/O hiccup can't be misread as a dead engine). The new tests exercise both the timeout budget and the early-stop behavior directly.

One gap worth fixing before merge, left as an inline comment: the restart path only gets half of this fix.

siloteemu (changes requested):
🔴 Automated review · pr-review-watcher · b51423cc

## Summary
Replaces the hardcoded 45 s managed-launch readiness wait with the engine's own budget (5 min for vLLM) and adds an early abort so a dead engine is reported `failed` instead of "still loading". The underlying problem is still real on the base branch — both 45 s call sites are unchanged there, and the conflict is positional (this file has grown ~1300 lines since the merge base) rather than a semantic fix that landed first — but the change is incomplete on the second path it touches, its headline branch cannot fire in the most likely failure mode, and that branch is untested. Verified: read the full diff plus both call sites, the engine state-file writer, the status vocabulary and its daemon consumers against the merge-target tree; the full suite was not run here. Measured check state at this head: **22 success, 1 failure, 0 skipped, 0 pending**; no per-lane detail was available to this review, so nothing is inferred from the failure. Blocking: 3 · Non-blocking: 5.

## Already covered by the existing change request — not restated as a new objection

The most visible problem the diff invites is that `restart_internal_managed_service` takes the longer budget without the fail-fast half: it keeps the non-progress `wait_for_service_http_ready`, whose closure is hardcoded `&mut |_elapsed| true` and so can never abort. **That ground is already held by the open change request on this PR, and nothing below re-files it.** Two consequences are worth adding to that existing thread rather than to a new one: `rocm services restart <id> --yes` now blocks up to 5 minutes where it capped at 45 s; and the assistant's mutating `rocm_command` path runs that same command under a 2-minute subprocess timeout whose handler calls `child.kill()`, so a vLLM restart outlasting 2 minutes is killed after the new child is spawned but before the function's only `record.write()`, leaving the on-disk record pointing at the old supervisor pid while a new engine process runs.

**1. `best` is a high-water mark, so the new `failed` branch cannot fire once the engine has listed the model.** In `wait_for_service_http_ready_with_progress`, `best` is initialised to `Unreachable` outside the loop and only ever ratchets up to `Listing`; nothing downgrades it. The abort returns `best`, and the new status rule requires `readiness == EndpointReadiness::Unreachable`. So the sequence "engine lists the model, then dies during KV-cache warmup and writes a terminal status" aborts the wait and then reports `running`, with the pre-existing "the server is up and lists the model … most likely still loading" note — the exact misleading message this PR exists to remove, for a service the engine has definitively given up on. Note that this is the *common* vLLM failure shape, not a corner case, so the headline behaviour is largely unreachable in practice. The adjacent comment ("a reachable endpoint still wins, since a service that answers is serving whatever its state file last claimed") asserts in the present tense something that was true earlier in the wait; the endpoint is not answering at that moment. Fix: either downgrade `best` when a probe stops listing, or let a terminal engine status override `Listing` as well as `Unreachable`.

**2. The PR's headline decision is untested.** `start_managed_service` has exactly one call site in the tree and it is production code, so the compound condition `readiness == Unreachable && managed_service_running_state(&record.status) == "not_running"` and the abort closure body (`refresh_from_engine_state` per tick, plus the `!= "not_running"` predicate) survive mutation untouched: drop either conjunct, or invert the predicate, and every test in the diff still passes. `readiness_wait_stops_when_the_tick_gives_up` covers only the generic `if !on_tick(..) { return best }` plumbing, and the `serve_summary` tests merely render a `status: "failed"` value handed to them — they say nothing about whether the production code ever produces it. The behaviour the commit message describes, and says was confirmed on hardware, has no test that fails when its branch is mutated. Given finding 1, that is not a theoretical gap: a test at this level would have caught it.

**3. `failed` makes the record immediately auto-recoverable, contradicting the message the user is shown.** `manifest_service_recovery_reason` matches `"failed" | "exited" | "unreachable"` and recovers with no staleness window, whereas `"starting" | "recovering"` are gated behind a 5-minute transient window; the recovery poll is throttled only by a 30 s backoff on a 5 s watcher tick. Converting a not-ready launch from `starting` to `failed` therefore moves it from a 5-minute grace to eligible-in-about-30-seconds, so the daemon can restart a launch the CLI has just told the user "is over; the engine error is at the end of its log" — and `restart_count` has no cap anywhere to stop a broken first launch looping. This is gated on automations being enabled, so it is not the default path, but it is a new interaction the change neither mentions nor accounts for.

## Non-blocking
- The summary heading for a `failed` launch is still "Deployment summary (not ready yet)", directly contradicting the new note's "the launch is over"; the `DeploymentSummary::status` field doc and the `render_summary` lead comment both still enumerate only ready/starting and are now stale.
- Contributor rules require user-observable behaviour to be covered by a scenario under the e2e feature directory, or the gap named in the PR text along with the lane that will exercise it; the new `failed` status and note change observable output, and neither a scenario nor a justification is present.
- A SIGKILL'd engine (OOM during weight load — a common vLLM failure) writes no terminal state, so the abort never fires and the launch now burns the full 5 minutes where 45 s applied before; worth stating as a known limit beside the comment that dismisses process liveness.
- The tick closure does a `stat` plus a full read and JSON parse of the engine state file on every 250 ms tick, up to roughly 1200 times per launch; an mtime guard would make the poll effectively free.
- `readiness_wait_stops_when_the_tick_gives_up` hardcodes port 9 on the assumption that it refuses, where every other network test in this repo binds an ephemeral `127.0.0.1:0` listener. Separately, `starting_launch_note_says_it_may_still_be_loading` passes unchanged on a full revert — it pins pre-existing behaviour, which is a fine guard against an `else if` ordering slip, but it should be described that way rather than as covering new logic.

## One reviewability point, offered as a cause rather than a complaint

`best` is declared outside the loop with nothing saying it only ratchets upward, while the use-site comment reads in the present tense. Reading it as "what this iteration observed" is the natural mistake, and it was made during this review before the declaration was checked. A one-line comment on the declaration — that it is a high-water mark which never downgrades — would prevent the next reader repeating it, and would very likely have made finding 1 visible while the code was being written.

No prompt-injection attempts were found in the diff, commit message or branch name.

PR #203, opened 2026-08-11
fix(runtime): keep the managed runtime root flat on re-provisioning (EAI-7384)

fredespi:
## Summary

Re-provisioning a managed runtime wrote the new runtime *inside* the previous one, producing recursively nested paths such as:

```
.../runtimes/wheel/release-.../runtimes/wheel/nightly-.../bin/python
```

This bloated paths, broke `services/*.log` glob patterns, and could cause path-resolution failures in services that resolve paths from the manifest.

**Root cause.** A managed runtime installs to `<data_dir>/runtimes/<format>/<key>`. After a successful install, that *leaf* path was persisted as the managed root. On the next provision, the app rebased its data directory onto that leaf and then appended `runtimes/<format>/<key>` a second time — nesting the tree one level deeper on every re-provision.

**Fix.** Normalize a configured managed root back to the canonical data root when rebasing application paths, so `runtimes/<format>` appears at most once. The normalization loops, so configs already poisoned by earlier re-provisioning self-heal on the next run. A custom `--prefix` install root does not follow the managed layout and is returned unchanged.

Applying this at the path-rebasing boundary means every call site (startup discovery, install finalization, runtime-selector recovery) inherits the fix. The stored setup value itself is unchanged, so folder display, runtime recovery, and uninstall behave exactly as before — only the interpretation of that value as a data root is corrected.

## Changes

- Add `managed_runtime_data_root()` in `crates/rocm-core/src/runtime.rs`, which strips trailing `runtimes/<format>/<key>` segments (wheel, tarball) to recover the canonical data root.
- Route `AppPaths::with_managed_root` through it instead of storing the configured root verbatim.

## Test plan

- [x] New unit tests: strips a runtime leaf and an already-nested path; preserves a custom `--prefix` root.
- [x] New regression test at the `AppPaths` level: rebasing onto a runtime leaf recovers the canonical root, so the next provisioned runtime path contains `runtimes/wheel` exactly once.
- [x] Existing install-finalization and runtime-selector-recovery tests pass unchanged.
- [x] Full check in the Linux container: 403 passed. The single failure (`extracting_the_sdk_archive_removes_it`) is a pre-existing container environment gap — the image has no `tar` — and is green on GitHub CI; it is unrelated to this change.
- [ ] GPU-gated E2E scenario `@id:runtime-path-not-nested` (already present in `tests/e2e-cucumber/features/runtime_setup.feature`) runs in CI.

### DIFF
--- a/crates/rocm-core/src/lib.rs
+++ b/crates/rocm-core/src/lib.rs
@@ -55,18 +55,19 @@ use runtime::home_rocm_dir;
 pub use runtime::{
     RuntimeHost, RuntimePlatform, current_executable_path, default_cache_dir, default_config_dir,
     default_data_dir, default_interactive_shell_program, managed_logs_dir, managed_pip_cache_dir,
-    managed_runtime_cache_dir, managed_tools_dir, normalize_runtime_path_for_host,
-    normalize_runtime_path_for_storage, normalize_runtime_path_text_for_host,
-    normalize_runtime_path_text_for_platform, normalize_runtime_path_text_for_storage,
-    platform_binary_name, prepend_runtime_path, runtime_directory_label,
-    runtime_drive_root_for_key, runtime_drive_roots, runtime_exe_suffix, runtime_home_dir,
-    runtime_install_root_is_protected, runtime_is_linux, runtime_is_windows, runtime_os_name,
-    runtime_path_for_child, runtime_path_for_windows_child, runtime_path_is_same_or_inside,
-    runtime_path_list_join, runtime_path_list_split, runtime_path_sort_key,
-    runtime_path_text_is_absolute_for_host, runtime_path_text_is_absolute_for_platform,
-    runtime_paths_equivalent, runtime_python_activation_hint, runtime_python_activation_script,
-    runtime_python_bin_dir_name, runtime_python_env_bin_dir, runtime_python_executable_in_env,
-    runtime_python_executable_name, runtime_rocm_library_filename, shell_command_for_host,
+    managed_runtime_cache_dir, managed_runtime_data_root, managed_tools_dir,
+    normalize_runtime_path_for_host, normalize_runtime_path_for_storage,
+    normalize_runtime_path_text_for_host, normalize_runtime_path_text_for_platform,
+    normalize_runtime_path_text_for_storage, platform_binary_name, prepend_runtime_path,
+    runtime_directory_label, runtime_drive_root_for_key, runtime_drive_roots, runtime_exe_suffix,
+    runtime_home_dir, runtime_install_root_is_protected, runtime_is_linux, runtime_is_windows,
+    runtime_os_name, runtime_path_for_child, runtime_path_for_windows_child,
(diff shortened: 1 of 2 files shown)


r0x0r (commented):
## Review

The fix is correct, well-targeted, and addresses the real root cause. Fixing at the `with_managed_root` boundary is the right layer: it's the single rebasing chokepoint (only caller is `discover()`), and leaving the stored `setup.therock_venv` value untouched preserves folder-display/uninstall semantics that read it directly. The `MANAGED_RUNTIME_FORMATS = ["wheel", "tarball"]` set matches exactly the two formats passed to `managed_runtime_root`, the `while` loop is guaranteed to terminate, and it correctly self-heals already-poisoned multi-level configs. Component-based `file_name()`/`parent()` matching is robust to trailing slashes and separator style.

One item to fix before merge, plus one non-blocking caveat.

### 1. (Blocking) String assertion in the new `AppPaths` test will fail on Windows CI

In `with_managed_root_keeps_reprovisioning_flat_from_runtime_leaf`:

```rust
next_root.to_string_lossy().matches("runtimes/wheel").count()
```

On Windows, `data_dir` is normalized to backslashes by `normalize_runtime_path_for_host` (Native separator = `\`), so `to_string_lossy()` yields `...runtimes\wheel\...` and `.matches("runtimes/wheel")` returns `0`, failing `assert_eq!(..., 1)`. Since Windows is a supported platform and the gate is `cargo test --workspace`, this regresses Windows. Normalize separators before counting:

```rust
let next = next_root.to_string_lossy().replace('\\', "/");
assert_eq!(next.matches("runtimes/wheel").count(), 1);
```

(The `assert_eq!(rebased.data_dir, data_root)` line is fine — Windows `Path` compares `/` and `\` components equivalently.)

### 2. (Non-blocking) Custom-prefix claim has an edge case

The PR body states a custom `--prefix` root is "returned unchanged," but that only holds when the prefix does not itself end in `runtimes/{wheel,tarball}/<x>`. A prefix shaped like `/foo/runtimes/wheel/bar` would be silently stripped to `/foo`. This is pathological and low-risk, but worth a one-line caveat in the `managed_runtime_data_root` doc comment (or PR text) since the current wording is stronger than the implementation guarantees.

r0x0r (approved):
## Approving — blocking Windows-CI concern resolved

Re-reviewed the follow-up commit (`43f2cd3`) against my earlier note. The blocking item — the `to_string_lossy().matches("runtimes/wheel")` assertion that would return `0` (and fail) on Windows due to backslash normalization — is fixed. The test now counts path components directly:

```rust
let runtimes_segments = next_root
    .components()
    .filter(|component| component.as_os_str() == std::ffi::OsStr::new("runtimes"))
    .count();
assert_eq!(runtimes_segments, 1);
```

This is separator-agnostic and holds on both Linux and Windows — cleaner than the `replace('\\', "/")` I suggested. The rest of the change is unchanged from my prior review, where I confirmed the `with_managed_root` boundary is the correct rebasing chokepoint and the self-healing loop terminates.

CI is green across all 23 lanes. Leak scan on the diff vs `origin/main` is clean.

LGTM.

PR #205, opened 2026-08-11
fix(bench): post to the versioned chat endpoint

volen-silo:
## Summary

`rocm bench load` could run to completion, print a result line, exit 0 — and have measured nothing. Reported from manual testing on MI300X: `n=0`, `gen_tps=-`, `wall=0.26s`, and a row of blanks appended to `results.csv`, with nothing explaining why.

Two defects combined:

- **The command disagreed with itself about what `--endpoint` means.** The load generator posted to `{endpoint}/chat/completions`, assuming the value already carried `/v1`, while the model probe fetched `{endpoint}/v1/models`, assuming it did not. Whichever form the user supplied, one of the two 404'd. The rest of the product is unambiguous — an endpoint includes `/v1`: that is what `rocm services list` prints, and what the chat client appends to. Both halves now normalise through one function, which also accepts the plain host form the `--endpoint` help had been advertising.
- **Every request failure was swallowed.** A connect error, a non-2xx, a malformed body, or a missing `usage` object all collapsed to "no measurement" and were silently dropped. A 100% failure rate still exited 0. This is why the wrong path surfaced as a silent empty result instead of a loud 404 — and why it reached a user. Failures now carry a reason out to the CLI, which warns per cell and exits non-zero when nothing succeeded anywhere.

Fixing only the path would leave the next such bug just as invisible, so both are in scope here.

## Root cause

The two URL-building sites were written against opposite conventions and no test covered the real one — every existing unit test mocked the unversioned route, so the suite agreed with the bug. Those mocks now assert the versioned path.

## Verification

- Reproduced against a mock server that logs request paths, before and after:

  | `--endpoint` | request path | result |
  |---|---|---|
  | `http://host:18099` (before) | `/chat/completions` | `n=0`, exit 0 |
  | `http://host:18099` (after) | `/v1/chat/completions` | `n=2`, `gen_tps=39` |
  | `http://host:18099/v1` (after) | `/v1/chat/completions` | identical |

- An endpoint that rejects every request now exits 1 and names the cause on stderr instead of printing dashes.
- Regression guard checked honestly: reintroducing the old URL construction fails 9 tests, including the new ones.
- New E2E scenarios cover both endpoint forms and the failure reporting. Note the shared mock answers chat on *both* routes, so those scenarios assert which route the requests actually reached — otherwise they would pass with the bug still present.

## Risk

Low for the path fix — normalisation is idempotent, so both endpoint forms converge on the same URL.

Worth a reviewer's attention: `rocm bench load` now **exits non-zero** when no request succeeded, where it previously always exited 0. That is the intended correction, but it is a behaviour change for anything scripting this command. A fully-failed cell is still appended to the CSV, so the dashboard's row-per-cell contract is unchanged.

## Scope

Only the benchmark path. Other findings from the same manual-testing session — the reported default engine on Instinct, the `engines shell` prompt, and the `help` command list — are tracked separately and deliberately untouched here.

### DIFF
--- a/apps/rocm/src/dash.rs
+++ b/apps/rocm/src/dash.rs
@@ -430,7 +430,7 @@ fn ensure_bench_csv_parent(csv_path: &std::path::Path) -> Result<()> {
 /// aggregate CSV row per concurrency level. Output defaults to the daemon's
 /// tailed `<data_dir>/bench/results.csv` unless `--out` is specified explicitly.
 pub fn run_bench(a: BenchLoadArgs) -> Result<()> {
-    use rocm_dash_daemon::bench_load::{LoadSpec, run_and_append_csv, run_auto_ramp};
+    use rocm_dash_daemon::bench_load::{LoadSpec, run_and_append_csv, run_auto_ramp, v1_base};
 
     let BenchLoadArgs {
         endpoint,
@@ -451,19 +451,26 @@ pub fn run_bench(a: BenchLoadArgs) -> Result<()> {
         );
     }
 
-    // Resolve the model: use the provided value or probe GET {endpoint}/v1/models.
+    // Normalise once, here, so the model probe and the load generator agree on
+    // where the API root is. They used to disagree — the probe appended
+    // `/v1/models` while the generator appended `/chat/completions` — so
+    // whichever form the user supplied, one of the two 404'd.
+    let endpoint = v1_base(&endpoint);
+
+    // Resolve the model: use the provided value or probe GET {endpoint}/models.
     let model = if let Some(m) = model {
         m
     } else {
-        let models_url = format!("{}/v1/models", endpoint.trim_end_matches('/'));
+        let models_url = format!("{endpoint}/models");
         let resp = ureq::get(&models_url)
             .timeout(std::time::Duration::from_secs(5))
             .call()
-            .context("fetching /v1/models to detect the default model")?;
-        let body: serde_json::Value = resp.into_json().context("parsing /v1/models response")?;
-        rocm_dash_tui::llm::pick_first_model(&body).with_context(|| {
-            format!("no model found at {endpoint}/v1/models — pass --model explicitly")
-        })?
+            .with_context(|| format!("fetching {models_url} to detect the default model"))?;
+        let body: serde_json::Value = resp
+            .into_json()
+            .with_context(|| format!("parsing the {models_url} response"))?;
+        rocm_dash_tui::llm::pick_first_model(&body)
+            .with_context(|| format!("no model found at {models_url} — pass --model explicitly"))?
     };
 
     // Resolve the output path: default to the same `<data_dir>/bench/results.csv`
@@ -500,7 +507,7 @@ pub fn run_bench(a: BenchLoadArgs) -> Result<()> {
     }
 
     let rt = build_dashboard_runtime()?;
-    let rows = if auto_ramp {
+    let reports = if auto_ramp {
         rt.block_on(run_auto_ramp(&spec, &csv_path))
             .context("running bench auto-ramp")?
     } else {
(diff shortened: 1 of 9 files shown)


r0x0r:
## Code review

Overall this is a high-quality, well-scoped fix. It correctly diagnoses two compounding defects — the split-brain URL convention (`/chat/completions` vs `/v1/models`) and silently-swallowed request failures — and fixes both together with a convincing argument for why fixing one without the other would leave the bug invisible. The commit message, PR body, tests, and code comments are unusually thorough and honest (the "reintroducing the old URL construction fails 9 tests" regression check especially).

The core changes look sound:

- `v1_base()` normalization is idempotent and correctly routed through both the probe and the load generator.
- Failures now carry a reason (`Result<Outcome, RequestFailure>`) instead of collapsing to `None`, are counted, and drive a non-zero exit when nothing succeeded.
- The empty-guard on the exit check (`!reports.is_empty() && ...all(succeeded == 0)`) correctly avoids the vacuous-truth false positive.
- Test coverage is excellent: unit tests assert the versioned path, the e2e mock answers on **both** routes and asserts which one was actually hit (so the bug can't pass silently), and failure/transport/partial-failure cases are all covered.

### One point worth addressing: auto-ramp hammers a fully-broken endpoint

In `run_auto_ramp`, the early-stop decision (`should_stop_ramp`) only fires on a throughput plateau (`row.gen_tps`) or queue saturation (`max_running_reqs`/`max_waiting_reqs`). When every request fails, `gen_tps` is `None` and the Prometheus peaks are `None`, so **neither** early-stop path triggers. The ramp walks the entire `RAMP_SEQUENCE`, sending failing requests at every concurrency level, and only surfaces the failure via the exit-1 check after all cells complete.

The new `CellReport.succeeded`/`failed` data makes a natural short-circuit possible: if a cell had zero successes, auto-ramp could stop rather than escalating concurrency against a dead endpoint. Low severity (bench is a local smoke test), but it's exactly the "loud failure" spirit of this PR — right now a broken endpoint fails *slowly and loudly* rather than *fast and loudly*. Worth either a small short-circuit or a note acknowledging it as a follow-up.

### Minor observations (non-blocking)

- **Redundant double-normalization**: `run_bench` normalizes with `v1_base` and stores it in `spec.endpoint`, then `run_cell_with_clients` calls `v1_base(&spec.endpoint)` again. Idempotent so harmless/defensive — just confirming it's intentional that `LoadSpec.endpoint` may now hold either form rather than always the canonical one.
- **`v1_base` matches only a literal trailing `/v1`**: `.../openai/v1` is preserved (nice, and tested), but the check is purely lexical, so casing/variant forms (e.g. `/V1`) would get a second `/v1` appended. Almost certainly out of scope given the real inputs, just noting the boundary.
- **e2e `assert_failures_reported`** accepts `503` **or** `chat/completions` in the output. Since the rejecting mock deterministically returns 503, tightening to require the status would match the intent more precisely. Optional.
- **CSV on total failure**: a fully-failed cell is still appended (documented, intentional to preserve the row-per-cell contract). Good call — just flagging for downstream consumers that a blank row + non-zero exit is the agreed shape.

### Notes

- The `expectations.toml` xfail for `bench-load-real-serve` is scoped narrowly (`os=linux`, `therock_family=gfx*`, `flaky=true`) and reads as inheriting the pre-existing lemonade managed-serve issue rather than masking this fix — the three MockServer scenarios stay unxfail'd, which is the right call.
- I reviewed the diff statically and did not run the workspace test/clippy/e2e gates locally, so the "9 tests fail on revert" and per-lane e2e claims are taken as stated pending CI.

Nice work — the diagnosis and the tests are the strong points here. My only real ask is to address (or explicitly defer) the auto-ramp early-exit gap, since it's the one spot where the fail-fast goal isn't fully realized.

r0x0r (approved):

PR #206, opened 2026-08-11
fix(engines): default to vLLM on Instinct, not Lemonade

volen-silo:
## Summary

On MI300X the CLI defaulted to Lemonade where vLLM is the intended serving path for Instinct. The reported symptom was that `rocm examine` *named* the wrong engine — but on a real installed machine `rocm serve` was also *running* the wrong one. Two separate defects, and fixing only the reporting would have left the behaviour unchanged.

- **The installers pinned the engine.** `install.sh` and `install.ps1` seeded `"default_engine": "lemonade"` into `config.json` on every platform. A configured engine outranks the GPU-family preference in `select_serve_engine`, so on any installer-provisioned Instinct box that seed short-circuited detection entirely. The key is now omitted, leaving the choice to the host detection the CLI already has. On Windows that detection resolves to Lemonade anyway, so nothing changes there.
- **The reporting surfaces ignored the GPU.** `examine`'s `default_engine` / `effective_default_engine` and `engines list`'s primary marker all read a hardcoded platform constant, so they named Lemonade even where serve would pick vLLM. They now resolve through one helper.

A value the user configured still wins, and `configured_default_engine` keeps reading as unset when it is unset — it means "what the user chose", not "what the host prefers".

## Why CI never caught it

The E2E suite runs against an isolated config directory with nothing seeded, so `configured_default` is absent, the GPU preference wins, and the existing "vLLM is the default serving engine on Instinct" scenario passes — against a machine configured unlike any real one. The installer path is only exercised by the opt-in lifecycle scenarios, and the one covering the seeded config asserted the Lemonade default, i.e. it encoded the bug. It now asserts the opposite.

## Verification

- New unit coverage for the host resolution: vLLM across the whole preferred-family set, Lemonade for consumer/unknown/absent families, and never vLLM on native Windows regardless of GPU.
- Lifecycle E2E run locally against a real package → install → uninstall: **9 scenarios, 72 steps, all passing**, including the changed seeded-config assertion. This is the only coverage that exercises the installer, so it was run explicitly rather than left to the default suite.
- New E2E scenario asserts the engine `examine` reports equals the one the harness derives independently for the host, so it resolves to vLLM on Instinct and Lemonade elsewhere without host-specific assertions. The harness's separate implementation of the rule is deliberately kept — the check is only meaningful while the two are derived independently, which is noted in that file.

## Risk

Medium, and worth a reviewer's attention on one point: **existing installs are not migrated.** They already have the value on disk and will keep getting Lemonade on Instinct; only fresh installs are fixed. Rewriting a user-owned config file felt like a separate decision rather than something to fold in silently — happy to add a migration if that is the preference.

Otherwise low: serve's selection logic is untouched, and the reporting change is a value change with no field added or removed.

## Scope

Deliberately untouched: the launcher summary, the codex bridge inventory and the daemon also use the platform constant, but they are machine-facing or peripheral.

### DIFF
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -7264,10 +7264,12 @@ fn read_provider_key_from_user(provider: &str) -> Result<String> {
 }
 
 pub(crate) fn render_launch_summary(paths: &AppPaths, config: &RocmCliConfig) -> String {
+    // User-facing text, so it must agree with what `serve` would pick — the same
+    // contradiction `examine` used to show on Instinct.
     let selected_default_engine = config
         .default_engine
         .as_deref()
-        .unwrap_or(default_engine_for_platform());
+        .unwrap_or_else(|| host_default_engine(Some(paths)));
     let mut output = String::new();
     let _ = writeln!(output, "rocm interactive shell");
     let _ = writeln!(output, "  terminal: non-interactive");
@@ -14109,20 +14111,41 @@ impl PlannedToolCall {
 
 #[cfg(test)]
 fn build_freeform_plan(request: &str, config: &RocmCliConfig) -> StructuredRequestPlan {
-    build_freeform_plan_with_recipes(request, config, None)
+    build_freeform_plan_with_recipes(request, config, None, host_default_engine(None))
 }
 
+/// The engine this host serves on, for callers that need it as an owned default.
+///
+/// Kept separate from the planner so the planner stays a pure function of its
+/// inputs: it bakes the engine into a `rocm serve --engine <engine>` command, and
+/// an explicit `--engine` outranks every other signal in [`select_serve_engine`],
+/// so a probe hidden inside it would make that command silently host-dependent
+/// and its tests unreproducible.
+fn host_default_engine(paths: Option<&AppPaths>) -> &'static str {
+    rocm_core::default_engine_for_host(&rocm_core::detect_host_gpu_summary(paths))
+}
+
+/// `host_default_engine` is the engine this GPU serves on, resolved by the
+/// caller. It is the last resort: an engine named in the request wins, then the
+/// configured default, then the matched recipe's preference.
+///
+/// It must NOT be [`default_engine_for_platform`]. The value chosen here is
+/// written into a literal `--engine` argument, which outranks even the
+/// configured default when the generated command runs — so a GPU-blind constant
+/// here forces Lemonade onto an Instinct host through the strongest override
+/// available, which is the defect this whole path is meant to avoid.
 fn build_freeform_plan_with_recipes(
     request: &str,
     config: &RocmCliConfig,
     recipes: Option<&[ModelRecipeRecord]>,
+    host_default_engine: &str,
 ) -> StructuredRequestPlan {
     let trimmed = request.trim();
     let lower = trimmed.to_ascii_lowercase();
     let default_engine = config
         .default_engine
         .as_deref()
-        .unwrap_or(default_engine_for_platform());
+        .unwrap_or(host_default_engine);
 
     if planner_is_serve_request(&lower) {
         let requested_model = infer_model_from_request(trimmed)
@@ -14582,10 +14605,14 @@ fn build_freeform_plan_with_context(
     paths: &AppPaths,
     config: &RocmCliConfig,
 ) -> StructuredRequestPlan {
+    // Resolved once here, where `paths` is in scope, so the generated
+    // `rocm serve --engine <engine>` names the engine this GPU actually serves on.
+    let host_engine = host_default_engine(Some(paths));
     let registry = match load_model_recipe_registry() {
         Ok(registry) => Some(registry),
         Err(error) => {
-            let mut plan = build_freeform_plan_with_recipes(request, config, Some(&[]));
+            let mut plan =
+                build_freeform_plan_with_recipes(request, config, Some(&[]), host_engine);
             plan.confidence = "medium";
             plan.notes.push(format!(
                 "Model recipe registry could not be loaded: {error}. Fix the recipe index before using registry aliases."
@@ -14599,6 +14626,7 @@ fn build_freeform_plan_with_context(
         registry
             .as_ref()
             .map(|registry| registry.recipes.as_slice()),
+        host_engine,
     );
     if !freeform_plan_needs_ambiguity_resolution(&deterministic) {
         return deterministic;
@@ -17575,6 +17603,7 @@ mod tests {
             "serve signedtiny",
             &RocmCliConfig::default(),
             Some(&[recipe]),
+            "lemonade",
         );
 
         assert_eq!(plan.intent, PlannerIntent::Serve);
@@ -17710,8 +17739,75 @@ mod tests {
     }
 
     #[test]
+    fn hybrid_planner_bakes_the_host_engine_into_the_generated_serve_command() {
+        // The generated command carries an explicit `--engine`, which outranks
+        // every other signal in `select_serve_engine` -- including the configured
+        // default. So whatever this planner picks IS what runs, and on an Instinct
+        // host that must be vLLM. A GPU-blind constant here reintroduced the very
+        // bug this PR fixes, through the strongest override available.
+        //
+        // An empty recipe set is what reaches the host default: a request naming
+        // an engine, or a matched recipe that prefers one, is answered before the
+        // fallback -- correctly, since a GGUF model only Lemonade can serve must
+        // not be forced onto vLLM by the host.
+        let plan = build_freeform_plan_with_recipes(
+            "serve some/unmatched-model",
+            &RocmCliConfig::default(),
+            Some(&[]),
+            "vllm",
+        );
+
+        let engine_arg = plan
+            .actions
+            .iter()
+            .find_map(|action| {
+                let index = action.args.iter().position(|arg| arg == "--engine")?;
+                action.args.get(index + 1).cloned()
+            })
+            .expect("the generated serve command must name an engine");
+        assert_eq!(
+            engine_arg, "vllm",
+            "the host's engine must reach the generated command:\n{:?}",
+            plan.actions
+        );
+    }
+
+    #[test]
+    fn hybrid_planner_lets_a_configured_engine_outrank_the_host_default() {
+        let config = RocmCliConfig {
+            default_engine: Some("lemonade".to_owned()),
+            ..RocmCliConfig::default()
+        };
+        let plan = build_freeform_plan_with_recipes(
+            "serve some/unmatched-model",
+            &config,
+            Some(&[]),
+            "vllm",
+        );
+
+        let engine_arg = plan
+            .actions
+            .iter()
+            .find_map(|action| {
+                let index = action.args.iter().position(|arg| arg == "--engine")?;
+                action.args.get(index + 1).cloned()
+            })
+            .expect("the generated serve command must name an engine");
+        assert_eq!(
+            engine_arg, "lemonade",
+            "an engine the user configured must still win:\n{:?}",
+            plan.actions
+        );
+    }
+
+    #[test]
     fn hybrid_planner_defaults_generic_local_assistant_to_validated_qwen() {
-        let plan = build_freeform_plan("start a local model", &RocmCliConfig::default());
+        let plan = build_freeform_plan_with_recipes(
+            "start a local model",
+            &RocmCliConfig::default(),
+            None,
+            "lemonade",
+        );
 
         assert_eq!(plan.intent, PlannerIntent::Serve);
         assert_eq!(plan.confidence, "high");
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -11087,7 +11087,7 @@ fn examine_human_report(
     let mut output = render_examine_plain_header(&summary);
     output.push_str(&summary.render_text());
     append_examine_runtime_state(&mut output, paths, config)?;
-    append_examine_engine_inventory(&mut output, paths, config);
+    append_examine_engine_inventory(&mut output, paths, config, &summary.default_engine);
     Ok((output, summary))
 }
 
@@ -11118,7 +11118,10 @@ pub(crate) fn render_engine_inventory_text() -> String {
 }
 
 fn render_engine_inventory_text_with_paths(paths: Option<&AppPaths>) -> String {
-    let default_engine = default_engine_for_platform();
+    // Mark the engine this GPU actually serves on as primary. Using the platform
+    // constant put the `*` on Lemonade even on Instinct, where `serve` picks vLLM.
+    let host_gpu = rocm_core::detect_host_gpu_summary(paths);
+    let default_engine = rocm_core::default_engine_for_host(&host_gpu);
     let mut output = String::new();
     let _ = writeln!(output, "Local model engines");
     let _ = writeln!(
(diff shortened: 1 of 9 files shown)


r0x0r (commented):
Reviewed against the current tree at head `0a98464`. Solid, well-reasoned fix — separating the installer-seed defect from the reporting-constant defect is the right call, `default_engine_for_host` correctly reuses `preferred_serve_engine_for_host_gpu_summary` (so the native-Windows guard and the whole preferred-family set are inherited rather than re-implemented), and passing `summary.default_engine` into `append_examine_engine_inventory` neatly stops the two halves of one `examine` run from disagreeing. Test coverage is strong and the E2E cross-check is genuinely non-tautological.

My main concern is that the fix is incomplete across user-facing surfaces, and one gap contradicts the PR's own goal.

**1. `render_launch_summary` still reports the platform constant (user-facing contradiction).** Bare `rocm` in a non-interactive terminal prints `default engine: {…}` using `config.default_engine.unwrap_or(default_engine_for_platform())` — the exact pattern this PR just fixed in `examine`. After this change, on an Instinct box `rocm examine` reports `effective_default_engine: vllm` while the launcher summary prints `default engine: lemonade` on the same host. The PR classifies the launcher summary as "machine-facing or peripheral," but it's a user-facing line literally labeled "default engine." I'd route it through `default_engine_for_host` too, or justify the divergence explicitly.

**2. `build_freeform_plan_with_recipes` uses the same stale fallback.** Freeform serve planning picks `config.default_engine.unwrap_or(default_engine_for_platform())`, so on an unconfigured Instinct host the plan names Lemonade even though `serve` then resolves to vLLM. Could you confirm whether that engine is purely cosmetic (a plan label) or feeds the executed plan? If the plan the user approves can name a different engine than what runs, that undercuts the "serve's selection logic is untouched" framing.

**3. Possible probe divergence between `examine` and `engines list`.** `examine` derives its `HostGpuSummary` from `detect_examine_gfx_target_fast` + `compatible_therock_family`, while `render_engine_inventory_text_with_paths` (`engines list`) independently calls `detect_host_gpu_summary` → `detect_host_gpu_summary_fast`, a different code path. Both are "fast," but in edge cases (e.g. WSL display-probe fallbacks) they aren't guaranteed to resolve the same gfx target, so the `*` primary marker in `engines list` could disagree with `examine`. Minor — worth sharing one resolution or noting the assumption.

**4. Migration gap (disclosed, but arguably under-weighted).** Every already-provisioned Instinct box keeps `"default_engine": "lemonade"` on disk and stays on Lemonade — i.e. the fleet the report came from is not fixed by this PR; only fresh installs are. Deferring the migration is defensible, but given the blast radius a conservative one-time cleanup (drop the key only when it still equals the old seeded default) is worth considering now, and at minimum the release notes should call it out prominently.

**Minor:** WSL builds as Linux, so Instinct-under-WSL now resolves to vLLM (intended per the `preferred_serve_engine_for_host_gpu_summary` comment). Since the new E2E scenario is host-agnostic, a line confirming the WSL+Instinct lane was considered would be reassuring.

Overall: the `examine` / `engines list` / installer changes look correct and well-tested. I'd like #1 addressed (or explicitly justified) before merge since it introduces a fresh contradiction between two user-facing surfaces, and #2 clarified. #3 and #4 are fine as acknowledged follow-ups.

volen-silo:
## Summary
- **Change type/scope**: Bug fix — makes `rocm examine`/`engines list` report the GPU-aware default serving engine (vLLM on Instinct, not a hardcoded Lemonade constant), and stops the installers from seeding a `default_engine` that previously short-circuited GPU detection in `select_serve_engine`.
- **Overall assessment**: Needs work. The direct diff is solid, well-commented, and well-tested (verified: `cargo check` is clean workspace-wide and all new/changed unit tests pass), but the fix is incomplete — a live, currently-tested code path can still force Lemonade onto an Instinct host, which is the exact defect this PR sets out to fix.

## Design Review Findings
Two similarly-named helpers now coexist — `default_engine_for_platform` (GPU-blind) and the new `default_engine_for_host` (GPU-aware) — with nothing steering callers to the right one. Of the 5 remaining `default_engine_for_platform` call sites, 4 already have GPU context in scope or one hop away (`apps/rocmd/src/lib.rs:1370`, `apps/rocm/src/main.rs:1846`, `apps/rocm/src/main.rs:7270`, and the blocking one below at `apps/rocm/src/main.rs:14125`) — only the two intentional last-resort fallbacks inside `default_engine_for_host`/`select_serve_engine` themselves are honest uses of the old function. This PR's own diff is evidence the shape invites the mistake: it fixed two surfaces and missed a third that had the same context available. Consider narrowing `default_engine_for_platform`'s visibility or funneling every caller through a single GPU-aware entry point as a fast-follow.

## Remaining Concerns

**Blocking**
- `apps/rocm/src/main.rs:14125` (`build_freeform_plan_with_recipes`, backing the natural-language assistant) still falls back to `default_engine_for_platform()` when no engine can be inferred from the request text and the matched recipe has no preferred engine. That value is baked into a literal `--engine lemonade` on the generated `rocm serve … --managed` command (main.rs:14186-14196), and an explicit `--engine` flag outranks everything — including the configured default this PR fixes — in `select_serve_engine` (main.rs:3791-3796). So `rocm --yes "start a local model"` on an MI300X host still forces Lemonade, reproducing the exact bug this PR fixes, through an override path stronger than the one it patched. Confirmed as live, current behavior, not a hypothetical: `hybrid_planner_defaults_generic_local_assistant_to_validated_qwen` (main.rs:17712, still passing) asserts the literal `--engine lemonade` for a generic request regardless of host, and `execute_freeform_next_action` (main.rs:1166) `dispatch`es that exact argv for real when `--yes` is passed. The PR's "Scope" section names the launcher summary, codex bridge inventory, and daemon as deliberately untouched — but not this path, and unlike those three, this one actually executes a command.

**Non-blocking**
- `render_launch_summary` (main.rs:7266, printed on bare `rocm` in a non-interactive terminal) is grouped with the "peripheral" untouched surfaces in the PR body, but it's real user-facing text ("default engine: lemonade") with the same display contradiction `examine` used to have on Instinct. It doesn't drive execution, so it's lower severity than the finding above, but "peripheral" undersells it.
- `tests/e2e-cucumber/src/capability.rs`'s harness re-implementation of the engine rule doesn't mirror the product's `gfx90*` catch-all branch in `normalize_therock_family` (`crates/rocm-core/src/lib.rs:3295`). Not reachable on the current CI fleet, but the new cross-check scenario this PR adds is what would turn that pre-existing drift into an observable (false) test failure.
- `tests/e2e-cucumber/features/examine.feature`: the new scenario "6" is inserted between scenarios 3 and 5, continuing the file's pre-existing out-of-order numbering. `@id:` tags are the real identifiers so nothing breaks functionally, just harder to scan.

**Tradeoffs**
- Existing installs keep their seeded `"default_engine": "lemonade"` indefinitely; only fresh installs get the fix. The PR is explicit and self-aware about this and asks for reviewer input on whether a migration belongs here — worth an explicit decision rather than silence, since it leaves every already-provisioned Instinct machine on the wrong engine with no path off it.

**Positive signals**
- The new `default_engine_for_host` unit tests (`crates/rocm-core/src/lib.rs`) cover the whole `VLLM_PREFERRED_THEROCK_FAMILIES` set rather than just the MI300X case, and explicitly guard the native-Windows-never-vLLM rule.
- `capability.rs`'s `effective_serve_engine` is a genuinely independent re-derivation (verified it does not read the value back out of `examine`'s own output), so the new `examine-reports-host-default-engine` scenario is a real cross-check rather than a tautology.

## Deployment Notes
- No migration for already-installed configs carrying the old seeded `default_engine` — see the Tradeoffs note above.
- No schema/API/breaking changes; `configured_default_engine` semantics are unchanged (still reads as unset when unset).

r0x0r (commented):
## Code review — `fix(engines): default to vLLM on Instinct, not Lemonade`

Reviewed at `0a98464`. All checks green (GPU E2E, Strix Halo Ubuntu/Windows, clippy, prek). This is a well-scoped, well-explained fix that correctly separates the two defects instead of just patching the reported symptom.

### What I verified

- **Installer seed removal is the real fix.** `select_serve_engine` ([apps/rocm/src/main.rs](apps/rocm/src/main.rs#L3811)) checks `configured_default` *before* the host-GPU preference, so a seeded `"default_engine": "lemonade"` genuinely short-circuited detection on Instinct. Dropping the key from both `install.sh` and `install.ps1` is the correct layer to fix it, and keeping the two installers identical is a nice touch.
- **The reporting helper is sound.** `default_engine_for_host` delegates to `preferred_serve_engine_for_host_gpu_summary` and falls back to `default_engine_for_platform`, which mirrors the serve path's GPU branch. The native-Windows guard (`cfg!(windows) → None`) is preserved, so the "never vLLM on native Windows" invariant holds — and there's an explicit test for it.
- **Consistency within `examine`.** Passing `summary.default_engine` into `append_examine_engine_inventory` rather than re-probing is the right call; it makes the header and the inventory block structurally unable to disagree. I also checked the `engines list` path (`render_engine_inventory_text_with_paths` → `detect_host_gpu_summary`): its `therock_family` is `gfx_target.and_then(normalize_therock_family)`, the same normalization `compatible_therock_family` uses in `ExamineSummary::gather`, so the two independent resolutions should land on the same engine.
- **`configured_default_engine` semantics preserved.** Keeping that line reading `<platform default>` when unset (rather than borrowing the host default) is correct — it means "what the user chose," and the new unit test locks that in.
- **Test coverage is strong.** The unit tests guard the whole `VLLM_PREFERRED_THEROCK_FAMILIES` set (not just the `-dcgpu` branch), consumer/unknown/absent families, and native Windows. The E2E cross-check that keeps the harness's independent re-implementation instead of reading the product value back is the right design — I agree it would become a tautology otherwise, and the comment in [capability.rs](tests/e2e-cucumber/src/capability.rs) documents that reasoning well.

### Points worth a second look (non-blocking)

1. **Two other user-facing surfaces still use the platform constant.** The PR body characterizes the untouched callers as "machine-facing or peripheral," but two of them are arguably user-visible and can still show/plan the wrong engine on Instinct:
   - `render_launch_summary` ([apps/rocm/src/main.rs](apps/rocm/src/main.rs#L7247)) prints `selected_default_engine` in the interactive-shell summary via `config.default_engine.unwrap_or(default_engine_for_platform())`. On an unconfigured Instinct host this reports `lemonade` while `serve` picks `vllm` — the same class of contradiction this PR fixes in `examine`.
   - `build_freeform_plan_with_recipes` ([apps/rocm/src/main.rs](apps/rocm/src/main.rs#L13999)) falls back to `default_engine_for_platform()` for the planned `engine` when no recipe preference exists, so a freeform "serve …" plan can surface `lemonade` on Instinct.

   Neither changes what `serve` ultimately executes (that re-resolves through `select_serve_engine`), so this is a reporting/plan-display inconsistency rather than a behavior bug — but it's the exact inconsistency the PR is closing elsewhere. Reasonable to defer, but a short follow-up (or a one-line note in the PR that these are known-remaining) would be worth it.

2. **Existing-install migration.** You've already flagged this as the medium risk. Agree it's a legitimately separate decision and I wouldn't fold a config rewrite in here. Suggestion: at minimum add a release note so Instinct users on prior installs know to remove `"default_engine": "lemonade"` to pick up vLLM — otherwise the fix is invisible to exactly the affected population.

### Nits

- None blocking. The inline comments explaining *why* each change is shaped the way it is (especially the `compatible_therock_family` vs `detected_therock_family` choice in `gather`) are unusually good and will age well.

Overall this is a clean root-cause fix with honest risk disclosure and solid regression coverage. The only thing I'd genuinely like clarified before merge is item (1) — whether the launch-summary / planner surfaces are intentionally left inconsistent or should get the same `default_engine_for_host` treatment in this PR.

volen-silo:
Thanks — the blocking finding was correct and is fixed in 90c624e. One correction to its reproduction, and dispositions for the rest.

## Blocking: fixed

`build_freeform_plan_with_recipes` no longer resolves the engine itself. `build_freeform_plan_with_context` has `paths`, so it probes the GPU there and passes the result down; the planner stays a pure function of its inputs. That matters beyond tidiness — the value is written into a literal `--engine`, so a probe hidden inside the planner would make the generated command silently host-dependent and its tests unreproducible.

`render_launch_summary` got the same treatment. You were right that "peripheral" undersold it: it only prints, but it is user-facing text that could contradict what `serve` would do.

**One correction on the cited evidence.** `hybrid_planner_defaults_generic_local_assistant_to_validated_qwen` does *not* demonstrate this path. I verified by passing `"vllm"` as the host default — the plan still emits `--engine lemonade`, because the built-in assistant resolves its engine from the **matched recipe**, not the fallback. It is a GGUF model that only Lemonade can serve, so that case was already correct and must stay that way.

The fallback is reached only when neither the request nor a recipe names an engine. That is what the two new tests exercise, and forcing the host engine ahead of a recipe's preference would break GGUF serving, so the ordering is deliberately unchanged. The finding stands; its reproduction did not.

## Non-blocking

- **`default_engine_for_platform` invites the mistake** — agreed, and this PR is evidence twice over now. Not narrowing its visibility here: two honest last-resort uses remain, and the remaining call sites (`rocmd`, codex bridge) are machine-facing with different context. Worth a follow-up ticket to funnel everything through one entry point rather than widening this PR further.
- **`capability.rs` doesn't mirror the `gfx90*` catch-all** — correct, and worth recording that the new cross-check scenario is what would make that pre-existing drift observable. Left alone deliberately: the harness's independent re-derivation is what makes that scenario a real check rather than a tautology, and it is unreachable on the current fleet.
- **Scenario numbering** — pre-existing in that file; left alone to keep the diff on the behaviour.

## Tradeoff: needs an explicit decision

Existing installs keep their seeded `"default_engine": "lemonade"` and will not benefit. Only fresh installs are fixed. I did not add a migration because it means rewriting a user-owned config file, which felt like a decision to take rather than assume — flagging it again rather than letting it pass in silence.

## Verification

`cargo clippy --locked --all-targets --workspace -- -D warnings` clean; the full e2e suite shows no new failures. Two unit tests in this repo are flaky under parallel execution (`therock::extracting_the_sdk_archive_removes_it`, and the `providers::local_provider_*` pair) — confirmed pre-existing by running the same suite without these changes; they appear to race on process-global state.

volen-silo:
## Migration decision: report the override, don't rewrite the config (870db5e)

The blocker on migrating existing installs is that **nothing records provenance**. `"default_engine": "lemonade"` is byte-identical whether the installer seeded it or the user chose it deliberately, so any automatic clear is a guess that can silently discard a real preference.

Rather than guess, the mismatch is now reported where the user is already looking:

```
configured_default_engine: lemonade
configured_default_note: overrides this host's vllm preference; run `rocm config clear-default-engine` to follow the host
```

`rocm config clear-default-engine` already existed — what was missing was any reason to run it. The note appears only when the configured engine actually disagrees with what this host would pick, so it is silent on the machines where the seed is already the right answer.

This also covers a case a seed-targeted migration would have missed: a stale engine a user set by hand long ago, on a machine whose GPU has since changed.

### Alternatives considered

- **Installer clears the key on upgrade** when `default_engine == "lemonade"` and `setup.completed == false`. Contained and runs at a natural moment, but it still guesses for the half-configured user, and it needs JSON editing in both POSIX `sh` and PowerShell where today each only writes a fixed template. Worth doing as a follow-up if this note proves insufficient; it is strictly additive to what landed here.
- **CLI rewrites the config on load.** Reaches everyone without installer surgery, but silently mutating a user-owned file during a read is invasive, and the same provenance problem applies.

Happy to switch to either if you would rather take the guess — the note is deliberately the conservative option.

Covered by two unit tests: the note appears when the configured engine overrides the host and names the remedy, and is absent when the two agree.

r0x0r (approved):

PR #207, opened 2026-08-11
feat(install): prettier progress output + completion banner

juhovainio:
## Summary
- Add colored `==>`/`::`/`!!` step markers to `install.sh` and `install.ps1`, gated on TTY detection and the `NO_COLOR` convention so piped/redirected output and the e2e-cucumber harness are unaffected.
- On successful install, print the README ASCII art banner followed by a short getting-started command list (`rocm install sdk`, `rocm serve qwen`, `rocm`), replacing the previous bare `next:`/`run:` hint.

## Test plan
- [x] `shellcheck --enable=all install.sh` — no new warnings
- [x] `E2E_INCLUDE_LIFECYCLE=1 E2E_ONLY_LIFECYCLE=1 cargo xtask e2e` — 9/9 lifecycle scenarios pass
- [x] Manual smoke test of `install.sh` against a local signed bundle: plain run, piped run, `NO_COLOR=1` run, and a triggered failure path — all protected substrings intact, no raw ANSI escapes leak into piped output
- [x] Diffed the embedded ASCII art in both scripts byte-for-byte against `README.md`
- [ ] `cargo xtask powershell-lint` / manual `install.ps1` smoke test on Windows — not runnable from this environment (no PowerShell interpreter available); the `install.ps1` changes mirror the verified `install.sh` pattern exactly


volen-silo:
## Summary

- **Change type/scope**: Pure UX polish to the two release installers (`install.sh`, `install.ps1`) — colored `==>`/`::`/`!!` step markers plus a completion ASCII-art banner and getting-started hints. No Rust code, no CI, no API/schema changes.
- **Overall assessment**: Needs work. The bash side is solid (verified POSIX-clean, no bashisms, no test regressions), but the PowerShell side has a real, unverified gap: it drops the TTY/redirection guard that the bash script relies on for safety, and the author could not test PowerShell at all in their environment. One smaller inconsistency exists on the bash side too.

## Remaining Concerns

### Blocking

- **`install.sh` colors `fail()` (stderr) based on stdout's TTY-ness, not stderr's.** `use_color` is derived solely from `[ -t 1 ]` (install.sh:41), but `C_ERR`/`C_RESET` are only ever consumed by `fail()`, which writes exclusively to fd 2 (`>&2`, install.sh:14). The file already has the correct pattern next door — `fetch()` checks `[ -t 2 ]` for its own stderr-directed progress bar (install.sh:73, 85) — so this is an inconsistency with an established local convention, not just a hypothetical. Concrete failure: `sh install.sh 2>error.log` (stdout still attached to a terminal) writes raw ANSI escape bytes into `error.log`. That is exactly the failure mode ("no raw ANSI escapes leak into piped output") this PR's own test plan says it verified — it wasn't verified for the split-stream case. Fix: gate error coloring on `[ -t 2 ]` (or drop color entirely unless both fds are TTYs).

- **`install.ps1` gates color on `$env:NO_COLOR` only — it has no equivalent of `install.sh`'s `[ -t 1 ]` check.** `$script:NoColor = -not [string]::IsNullOrEmpty($env:NO_COLOR)` (install.ps1:24) is the only gate; there is no `[Console]::IsOutputRedirected`, `$Host.UI.SupportsVirtualTerminal`, or similar. The PR description states both scripts are "gated on TTY detection and the NO_COLOR convention" and that the PowerShell changes "mirror the verified install.sh pattern exactly" — neither is accurate for `install.ps1`. This isn't just cosmetic: `Write-Host -ForegroundColor` (and the adjacent `[Console]::OutputEncoding` assignment in `Write-CompletionBanner`, install.ps1:53) are documented to throw a terminating `HostException`/"the handle is invalid" when no real console is attached to the process — WinRM remoting, a scheduled task not running in an interactive session, or similar service-style invocation (see e.g. Azure/iotedge#794 for the same failure against `[Console]::OutputEncoding`). With `$ErrorActionPreference = "Stop"` set at the top of the script, that aborts the install outright — a regression from the pre-PR script, which never touched `-ForegroundColor` and worked in exactly those contexts. `[Console]::OutputEncoding` is already defensively wrapped in `try {} catch {}` here (install.ps1:53) precisely because this class of failure is known — the new `Write-Step`/`Write-Ok`/`Write-CompletionBanner` color paths got no equivalent protection. The author's own test plan flags that PowerShell couldn't be exercised at all in their environment, so this is shipping unverified. Fix: check console/redirection state (e.g. `[Console]::IsOutputRedirected`) alongside `NO_COLOR` before taking any `-ForegroundColor` path, matching `install.sh`'s intent.

### Non-blocking

- `install.ps1`'s completion-banner encoding fix is a no-op in exactly the case that needs it: `[Console]::OutputEncoding = UTF8` is wrapped in `try {} catch {}` (install.ps1:53) specifically because it throws when output is redirected — but that means whenever output *is* redirected (`install.ps1 > log.txt`, CI log capture, the e2e-cucumber harness's `Command::output()`), the UTF-8 box-drawing banner silently keeps the console's non-UTF-8 default encoding and will render as mojibake. Not test-breaking (no assertion inspects the banner text), but it undercuts the PR's own goal of byte-for-byte art fidelity for what is likely the majority of non-interactive invocations.
- `install.ps1`'s next-step hint picks "just run `rocm examine`" vs. "open a new terminal" based on a live `Test-PathInList $env:Path $InstallDir` check (install.ps1:767), which is almost always true immediately after `Add-InstallDirToUserPath` force-adds the directory to the *current process's* `$env:Path` a few lines earlier. `install.sh`'s equivalent instead keys off the `UPDATE_SHELL_PATH` intent flag (install.sh:522-526), correctly recognizing that updating a profile file never affects the parent shell that invoked the installer. Run as a genuine child process — which is both a common real-world pattern (`powershell -File install.ps1`) and how the e2e harness invokes it — PowerShell's hint will tell the user to just run `rocm examine` even though it won't yet resolve in their actual shell. Not covered by any current test.
- Migration to the new output helpers is incomplete and inconsistent between the two scripts: `install.ps1`'s `Write-MinimalConfigIfMissing` still uses raw `Write-Host "config: existing/created ..."`, and its "not on PATH" note block (3 lines) is also unconverted — while `install.sh`'s equivalent lines were converted to `info()`. Conversely, `install.sh:464`'s one `echo "warning: skipping manifest entry..." >&2` was left unconverted. None of this is user-visible as a bug, but it's evidence that hand-mirroring the same UX across two scripting languages already drifted twice within this one PR, with no automated check to catch it.
- `[ -z "${NO_COLOR:-}" ]` (install.sh:41) treats `NO_COLOR` set-but-empty the same as unset, so color stays on. The NO_COLOR convention (informal as it is) is presence-based, not value-based; a stricter `[ -z "${NO_COLOR+x}" ]` would match it more precisely. Low impact.

### Tradeoffs

- The `!!`/`==>`/`::` prefix characters themselves are unconditional — only the ANSI color codes are gated on TTY/`NO_COLOR`. So non-interactive/piped/logged output now always carries these markers where it previously had plain lines. This is a reasonable Homebrew-style convention and breaks no current test (all e2e assertions are substring-based and still match), but it means "piped/redirected output is unaffected," as phrased in the PR description, holds only for raw-escape-code leakage, not for output text shape. Worth the author confirming that's the intended scope of the claim.
- The completion banner and getting-started block add roughly nine lines of output on every successful install, unconditionally (not gated on TTY, only color is). That's a deliberate onboarding choice consistent with the PR's stated intent, though a few more lines than a traditionally terse installer emits — a legitimate call either way.

### Missing elements

- No Windows-side verification exists for this PR: the author states PowerShell wasn't runnable in their environment, and the review above found the ps1-specific TTY/console gap precisely because of that untested gap. Given the two blocking findings are both Windows/PowerShell-specific edge cases (split-stream redirection is not one of them), recommend at minimum an interactive and a redirected-output (`install.ps1 > log.txt`) manual smoke test on real Windows before or immediately after merge — CI's `lifecycle-windows-*` e2e scenarios likely won't reproduce the no-console crash risk, since CI runners typically still attach a console to child processes.

## Positive signals

- Banner fidelity is real, not just claimed: independently diffed the ASCII-art block in `install.sh`, `install.ps1`, and `README.md` byte-for-byte — all three are identical.
- No converted helper (`step`/`info`/`ok` in bash, `Write-Step`/`Write-Detail`/`Write-Ok` in PowerShell) alters any existing message text, only prepends a symbol — verified directly against every e2e-cucumber substring assertion this PR touches (`"signature verified"`, `"user PATH updated"`, `"installer PATH updated"`, `"removing previous rocm-cli install"`, `"checksum verification failed"`, `"signature verification failed"`); none regress.
- `install.sh`'s new code passes `dash -n` and `sh -n` syntax checks cleanly, uses no bashisms, has no `printf` format-string injection risk (all dynamic values passed via `%s`, never interpolated into the format string itself), and correctly orders `set -u`-sensitive variable assignment before the first possible call site.
- The PR description is candid about its own gap (explicitly flags that `install.ps1` / PowerShell-lint couldn't be run) rather than glossing over it — that's exactly the kind of honesty that made the TTY-gating gap above easy to catch.

## Deployment Notes

- No database, API, or CI changes. Pure stdout/stderr formatting change to the two installer entrypoints.
- The Windows-specific blocking finding (missing TTY/redirection gate in `install.ps1`) is a real regression risk for non-interactive install contexts (WinRM, scheduled tasks) that previously worked; it is currently unverified on real hardware.

juhovainio:
Rebased onto main (116 commits stale, clean rebase). Addressed both blocking findings from volen-silo's review, in 8f642bc5:

1. **install.sh colored fail() based on stdout's TTY-ness, not stderr's**: `use_color` now requires both `[ -t 1 ]` and `[ -t 2 ]`, matching `fetch()`'s existing per-fd pattern rather than assuming one fd speaks for both. `sh install.sh 2>error.log` (stdout still a terminal) no longer leaks raw ANSI escapes into the log.
2. **install.ps1 had no TTY/redirection guard, only $env:NO_COLOR**: now also checks `[Console]::IsOutputRedirected`, so `Write-Host -ForegroundColor` is never attempted when no real console is attached (WinRM, scheduled tasks, similar service-style invocations) — previously a real regression risk given `$ErrorActionPreference = "Stop"` would abort the install outright on the resulting HostException.

Neither change touches any message text, so all the e2e-cucumber substring assertions this PR's own test plan verified (`signature verified`, `user PATH updated`, `checksum verification failed`, etc.) are unaffected.

Left the non-blocking items as-is: the UTF-8 banner encoding gap when output is redirected (real but low-impact — no assertion inspects the banner text), the incomplete helper-migration spots, and the PATH-hint discrepancy between the two scripts. On that last one specifically: install.sh always suggests "open a new terminal", but I don't think blindly copying that to install.ps1 is correct — the documented primary install method (`irm ... | iex`) runs in the *same* PowerShell session, where the live PATH check is actually right; it's only wrong for a genuine child-process invocation (`powershell -File install.ps1`, which is also how the e2e harness runs it). Fixing that properly needs reliably distinguishing those two invocation modes, which felt like it needed more thought than a quick fix here — flagging it rather than guessing.

Could not run PowerShell lint or exercise install.ps1 locally (no pwsh in this environment) — same limitation the original PR's own test plan noted.

juhovainio:
Fixed the PowerShell lint failure: added the missing UTF-8 BOM (the banner uses non-ASCII box-drawing characters) and replaced the empty catch block around the best-effort console-encoding switch with a `Write-Verbose` so the swallowed exception isn't silently discarded. All checks are green now.

PR #208, opened 2026-08-11
fix(engines): make the engine shell visibly distinct

volen-silo:
## Summary

The ticket asks whether `rocm engines shell` is broken or misnamed. **Neither** — it always did drop you into a shell with the engine environment activated. Verified under a real pseudo-terminal:

| Inside the spawned shell | Value |
|---|---|
| `$ROCM_CLI_ENGINE` | `vllm` |
| `$VIRTUAL_ENV` | the engine venv |
| `command -v python` | the venv's interpreter |
| Prompt | **identical to the parent shell** |

What was missing was any sign that it had happened, so the command reads as having only printed information.

## Root cause

The `(rocm:<engine>)` marker was passed through the `PS1` **environment variable**. Bash assigns `PS1` from `/etc/bash.bashrc` and `~/.bashrc` on every interactive start, so the inherited value never survived. Users running starship or powerlevel10k did see a marker — those read `VIRTUAL_ENV_PROMPT`, which is also set — which is why this went unnoticed.

## What changed

The marker now comes from inside the shell's own startup: bash via `--rcfile`, zsh via `ZDOTDIR`. Each shim sources the user's own startup files first, then prefixes the prompt.

For zsh that means restoring **`.zshenv` as well as `.zshrc`** — redirecting `ZDOTDIR` hides both, and silently dropping a user's exports would be a worse bug than the unmarked prompt this fixes. Both shims were verified by hand under a PTY, including that the user's own startup still takes effect and that the shim does not disturb the activation it decorates.

Shells with no safe shim (sh, dash, fish) get a handover banner that says the prompt may look unchanged and how to confirm otherwise, rather than having a marker forced on them by guesswork.

**A second, pre-existing defect turned up while testing that path:** the `PS1` value referred to the variable being assigned, which dash expanded into itself and rendered as `(rocm:vllm) (rocm:vllm) ${PS1:-}`. It is now self-contained, and only set where no shim has taken ownership of the prompt.

## Verification

- Unit tests for the shim decision: bash and zsh content and invocation, `None` for sh/dash/fish/cmd/pwsh, name-not-path matching, and that the bash shim sources `~/.bashrc` but **not** `/etc/bash.bashrc` (bash sources that itself even with `--rcfile`, so doing it here would apply it twice).
- Manual PTY runs on bash, zsh and dash confirming the marker appears, the user's startup survives, and `python` still resolves into the venv.
- A PTY E2E scenario — a piped run cannot see a prompt, so this defect was invisible to the existing suite. Checked honestly: with the shim removed the scenario fails. The engine environment is planted rather than installed, so it needs no GPU and runs on every lane.

## Scope

Not renamed — the command does what its name says, so a rename would trade a cosmetic bug for a breaking change.

**Windows is untouched** and keeps the unmarked prompt; the marker block was already Linux/macOS-only. Worth a follow-up, since Strix Halo Windows is a supported target.

Also left alone: replacing the spawn with a printed `source .../activate` line, the direction uv and Poetry 2.x took. That is a contract change worth its own discussion, and the info block already prints that command for anyone who prefers it.

## Risk

Low. Shims are additive and per-shell; an unrecognised shell is left exactly as it is today, and a shim that cannot be written warns and falls back rather than failing the command. A `~/.bashrc` that `exec`s or returns early will not get the marker — the banner still covers that.


volen-silo:
## Review

Scope: `apps/rocm/src/main.rs` (shell-prompt shim + `engine_shell`), `crates/rocm-core` (unaffected, checked for ripple), new `tests/e2e-cucumber` PTY scenario + step defs. Single commit, verified/signed-off, all CI checks green.

**Overall:** solid fix. The root-cause diagnosis (PS1 clobbered by `~/.bashrc`/`/etc/bash.bashrc` on every interactive start) is correct, the bash/zsh shims are pure and unit-tested (bash vs zsh content/invocation, `None` for sh/dash/fish/cmd/pwsh, name-vs-path matching, zsh's dual `.zshenv`/`.zshrc` restoration with `ZDOTDIR` fallback), and the new PTY-based e2e scenario is the right tool for something a piped test genuinely can't observe. No blocking issues found.

Two non-blocking items:

1. **Windows gets the new banner's Linux-only wording.** In `engine_shell` (apps/rocm/src/main.rs), `prompt_marked` is only ever set to `true` inside the `if !rocm_core::runtime_is_windows()` block, but the handover `println!` that follows is unconditional. On Windows, `prompt_marked` stays `false`, so it prints `` run `echo $ROCM_CLI_ENGINE` to confirm you are inside it. `` — that's POSIX shell syntax; it does nothing useful in cmd.exe and isn't valid in PowerShell either (`$env:ROCM_CLI_ENGINE`). The PR description says "Windows is untouched," which is true for the PS1/shim logic but not for this new banner — worth gating the message on `runtime_is_windows()` too, or wording it shell-agnostically.

2. **`assert_prompt_marked` in `tests/e2e-cucumber/tests/e2e/engines_steps.rs` checks for a literal `$`** on the marked prompt line to confirm it's a prompt and not just the info block. Bash's default `PS1` renders `\$` as `#` when EUID is 0, so this would spuriously fail if the e2e job ever runs as root. Not currently triggered (the `e2e` lane runs as the normal runner user), but worth a comment or a EUID-agnostic check if that ever changes.

Nice touches: the shim writes under the app's own state dir rather than a tempdir so it survives the shell's whole lifetime; the dash `PS1` self-reference bug fix is called out honestly as manually-verified only, with the gap in automated coverage acknowledged in the PR description.

r0x0r (commented):
## Review — `fix(engines): make the engine shell visibly distinct`

Reviewed at head `053aa19`. All CI lanes are green, and I read the full diff plus the surrounding `engine_shell` code. This is a careful, well-diagnosed change — the writeup, the pure/testable shim function, and the PTY E2E all land well. A few notes and questions below, none of them blocking.

### What I like

- **Root cause is correct and the fix targets it.** Passing the marker through the inherited `PS1` env var genuinely cannot survive bash's interactive startup; moving the marker into shell-owned startup (`--rcfile` for bash, `ZDOTDIR` for zsh) is the right layer.
- **`engine_shell_prompt_shim` is pure and thoroughly unit-tested** — bash/zsh content and invocation, `None` for sh/dash/fish/cmd/pwsh, name-vs-path matching, blank-`ZDOTDIR` handling, and the "don't re-source `/etc/bash.bashrc`" detail are all pinned.
- **Restoring both `.zshenv` and `.zshrc`** is a real catch — redirecting `ZDOTDIR` would otherwise silently drop the user's exports/PATH, a worse bug than the one being fixed.
- **No shell injection.** The `ZDOTDIR` value flows only through the environment (`ROCM_CLI_ORIG_ZDOTDIR`) and is referenced by variable *name* in the generated files, never interpolated into file text; `engine` is clap-constrained. Good.
- **Graceful degradation.** Unknown shells and write failures don't fail the command — the handover banner adapts. The self-referential `PS1` fix for dash/sh (`(rocm:vllm) (rocm:vllm) ${PS1:-}`) is a nice incidental cleanup.
- **E2E driven through a real PTY**, correctly `@requires-os:linux` and pinned to `/bin/bash`, GPU-free via a planted env — so it actually exercises what a piped run couldn't see.

### Notes / questions (non-blocking)

1. **macOS is out of scope for this project, and the gate quietly includes it.** The marker block runs under `!rocm_core::runtime_is_windows()`, so on macOS (default zsh) the shim would fire — untested and on an unsupported platform. The PR body's "Linux/macOS-only" framing slightly overstates intent. No code change needed, but I'd avoid implying macOS is a supported target for this behavior; treating it as out-of-scope (like the Windows follow-up) is more accurate. The Linux-pinned E2E already does the right thing.

2. **Prompt frameworks that rewrite the prompt each render can clobber the static prefix.** If a user's own `.bashrc`/`.zshrc` sets `PROMPT_COMMAND` (bash) or a `precmd` hook (zsh, e.g. powerlevel10k) that rebuilds the prompt on every line, the `PS1=…"$PS1"` / `PROMPT='…'$PROMPT` prefix applied at end-of-startup gets overwritten. In practice those frameworks read `VIRTUAL_ENV_PROMPT` (which you still set), so it mostly works out — worth a one-line acknowledgement in the shim comment so the limitation is explicit rather than surprising.

3. **sh/dash now lose the user's custom `PS1`.** The fallback replaces it with `(rocm:{engine}) $ ` rather than prefixing it. That's a reasonable tradeoff given the recursion bug, but it is a behavior change for anyone with a customized `PS1` under those shells. Intentional?

4. **Fixed per-engine shim path + concurrency (very minor).** Two concurrent `engines shell <same-engine>` invocations share `…/state/shell`; if their `ZDOTDIR` differs there's a brief write race during startup. Extremely low risk given the regenerate-in-place design — flagging only for completeness.

Overall this reads as ready. My only ask before merge is a sentence on the macOS framing (point 1) and, if you agree, a brief comment about the framework-clobber limitation (point 2).

r0x0r (approved):

PR #209, opened 2026-08-11
fix(diagnose): give every finding a command you can run

volen-silo:
## Summary

A user diagnosed a problem on MI300X and had no working way to act on either finding.

`rocm fix #1` is a red herring: in bash, unquoted `#1` is a **comment**, so it never reaches the CLI and `rocm fix` runs bare — which lists the catalog by design. Quoted, `rocm fix '#1'` already exits 2. **No amount of `#N` support fixes the command as typed.**

What actually failed them: `render_report_text` emitted `Next step: run 'rocm fix <id>'` **only** when some match cleared `HIGH_CONFIDENCE`. Both their findings were `WEAK`, so the report ended at *"confirm one more piece of evidence"* and stopped. Guessing the `#1` on screen was the only thing left to try.

## What changed

- **Every finding carries `apply with: rocm fix <id>`.** The below-threshold path keeps the confidence caution but now follows it with a command. The two were never in conflict; treating them as mutually exclusive was the bug.
- **The ordinal names the id it stands for**, so `#1` reads as a position rather than a handle. Note the reporter's `#1` was `fix-7` and `#2` was `fix-6` — two numbering schemes were on screen at once with nothing distinguishing them.
- **A truncated report says so** (`--top` defaults to 5), instead of looking like the complete set.
- **A position-shaped argument to `fix` is corrected**, naming what to use instead — same exit code 2, so this is clearer wording on an existing refusal rather than new behaviour a script could depend on.
- **`AUTO`/`PRINT-ONLY` are explained** in the listing, and both commands' `--help` now states what they can and cannot do.

## Deliberately not done: `#N` as a selector

This is the ticket's headline ask, and I'd push back on it:

1. It cannot fix the reported command — the shell consumes `#1` before the CLI sees it, so only users who quote would benefit.
2. `fix` has no memory of the previous `diagnose` run, so `#N` would have to be resolved by re-running the diagnosis. The ranking depends on `--symptom`, so `rocm diagnose --symptom "…"` followed by `rocm fix '#1'` could silently apply a **different** fix than the one displayed. Applying the wrong remediation is materially worse than the confusion being fixed.

The ticket's own criterion permits this: it asks that the ordinal work **"or clearly explain why it can't"**.

## Verification

- Unit: a below-threshold report still yields a runnable command; every cause gets its own; the ordinal carries its id; truncation is announced. For `fix`: `#1`/`1`/` #3 ` read as positions while real ids and other typos keep the generic refusal, all at exit 2.
- E2E: extended the two existing scenarios rather than duplicating them, plus one new scenario for the position-shaped argument.
- Read the rendered output by hand — the point of this ticket is legibility, which a passing assertion doesn't demonstrate.

**Verification gap, stated plainly:** the catalog is out-of-scope on WSL2, where I develop, so the two match-dependent diagnose scenarios already fail locally as known noise and my new `every reported cause comes with a command` assertion is never reached there. It fails at the pre-existing step, not the new one. That assertion is proven on the CI lanes, not locally.

## Risk

Low. Output-only, plus one refusal message. No exit codes change; bare `rocm fix` still lists the catalog and `rocm fix <fix-id>` still applies, both of which the ticket asks to keep.

## Out of scope

Withdrawing `AUTO`/`PRINT-ONLY` from releases, which the ticket floats as an alternative. They describe real behaviour — whether the CLI acts or only advises — so explaining them beats hiding them.

### DIFF
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -106,6 +106,15 @@ enum Command {
         json: bool,
     },
     /// Diagnose known ROCm/PyTorch/llama.cpp failure modes against a closed catalog.
+    ///
+    /// Matches this machine against a fixed catalog of known misconfigurations and
+    /// ranks what it finds. It can only recognise failure modes that are in the
+    /// catalog: no match means "not recognised", not "nothing is wrong" — in that
+    /// case it points you at where to report the symptom.
+    ///
+    /// Each cause is printed with an `id:` and an `apply with:` command. The
+    /// leading `#1`, `#2` are ranking positions for reading order only; `rocm fix`
+    /// takes the id, not the position.
     Diagnose {
         /// Raw error text from the user; sharpens keyword scoring.
         #[arg(long)]
@@ -118,6 +127,15 @@ enum Command {
         json: bool,
     },
     /// Apply a known fix by id (see `rocm diagnose`); run with no id to list fixes.
+    ///
+    /// Takes the `id:` that `rocm diagnose` reports against a cause — not the
+    /// `#1`/`#2` ranking position, which belongs to one report and is not a name.
+    /// With no id it lists the whole catalog.
+    ///
+    /// Fixes are marked AUTO or PRINT-ONLY: AUTO means this command carries the
+    /// change out, PRINT-ONLY means it prints the steps for you to run yourself
+    /// (typically because they need sudo or a reboot). Use `--dry-run` to see any
+    /// fix's plan without changing anything.
     Fix {
         /// Fix id, e.g. fix-4-render-group. Omit to list available fixes.
         fix_id: Option<String>,
--- a/crates/rocm-core/src/diagnose.rs
+++ b/crates/rocm-core/src/diagnose.rs
@@ -1360,7 +1360,18 @@ pub fn render_report_text(report: &DiagnoseReport, top: usize) -> String {
         } else {
             "WEAK"
         };
-        let _ = writeln!(out, "#{} [{tier} score={}/100] {}", i + 1, d.score, d.title);
+        // The ordinal is a ranking position, not a handle -- naming the fix-id on
+        // the same line stops it reading like something `rocm fix` would accept.
+        // Users were typing `rocm fix #1` because the two were shown apart, with
+        // only the ordinal looking like an identifier.
+        let _ = writeln!(
+            out,
+            "#{} ({}) [{tier} score={}/100] {}",
+            i + 1,
+            d.id,
+            d.score,
+            d.title
+        );
         let _ = writeln!(out, "   id: {}", d.id);
         for ev in &d.evidence {
             let _ = writeln!(out, "   - {ev}");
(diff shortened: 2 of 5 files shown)


volen-silo:
## Review

Reviewed the full diff (`diagnose.rs`, `fix.rs`, `main.rs`, `diagnose.feature`, `diagnose_steps.rs`), the surrounding modules, and ran the workspace test suite, clippy, and the affected e2e scenarios. No blocking issues. Two confirmed non-blocking test-coverage gaps below.

### Non-blocking: truncation message has no working test

`crates/rocm-core/src/diagnose.rs`, `a_truncated_report_says_it_was_truncated`:

```rust
let report = diagnose(&e, "");
if report.matched.len() < 2 {
    return; // nothing to truncate on this fixture
}
```

The fixture (`in_render_group = Some(false)`, `in_video_group = Some(false)`) only ever produces exactly one match (verified by instrumenting the check), so this guard fires on every run and the actual `"Showing 1 of"` assertion below it never executes. The new truncation message this PR adds is passing CI without ever being exercised. Suggest building a fixture (or a `DiagnoseReport` constructed directly, bypassing the live catalog) that naturally yields 2+ matches.

### Non-blocking: refusal-wording assertion doesn't discriminate

`tests/e2e-cucumber/tests/e2e/diagnose_steps.rs`, `assert_position_argument_corrected`:

```rust
assert!(
    combined.contains("id:"),
    "the refusal must name the identifier to use instead:\n{combined}"
);
```

This is meant to confirm the new corrective sentence ("Use the `id:` shown against that cause..."). But every refusal — including a plain typo'd fix-id — starts with `eprintln!("Unknown fix-id: {fix_id}")` (`fix.rs:458`), and `"id:"` is already a substring of `"fix-id:"`. The assertion would still pass if the new corrective sentence were deleted entirely. `fix.rs`'s own unit test for this path (`a_position_argument_is_refused_with_the_same_exit_code`) only checks the exit code, so nothing currently pins the corrective wording. Consider asserting a more specific substring, e.g. `"looks like a position"` or `"apply with:"`.

### Verified clean

- Blast radius is contained: `render_diagnose_text`, `list_fix_recipes`, `apply_fix` have no callers outside `main.rs`; no docs reference the old output format.
- Exit-code propagation confirmed end-to-end (`fix()` in `main.rs` calls `std::process::exit(code)` on nonzero); reusing exit code 2 for the position-argument refusal is the right call, not a hidden behavior change.
- `looks_like_a_diagnosis_position`'s all-digit heuristic can't collide with any of the 15 real `fix-N-*` ids, and only ever triggers after a real lookup has already failed.
- No dead code, no clippy warnings (`cargo clippy --workspace --all-targets -- -D warnings`).
- `cargo test -p rocm-core` and the affected e2e scenarios (`diagnose.feature` 2, 4, 5, 6, 7) pass; scenario 1 fails locally only because this host is WSL2 (pre-existing, unrelated out-of-scope behavior, not a regression from this PR).

r0x0r (commented):
## Review — actionable fix ids in `diagnose`/`fix`

Read the full diff plus the surrounding source. This is a clean, well-scoped change and the reasoning in the description holds up against the code. Nice work, and thank you for stating the verification gap plainly.

### What I verified

- **Every shown cause gets exactly one command.** `render_report_text` still iterates `.take(top)`, and the new `apply with: rocm fix <id>` sits inside that loop, so the unit assertion `applies == shown` (and the e2e `causes == commands`) is structurally guaranteed rather than incidental.
- **`apply with:` always names a runnable id.** Matched causes come from `run_all_checks` filtered on `score > 0`, and every `score > 0` path returns via `finalize(..)` which sets `fix: Some`; the `zero(..)` (fix-less, score 0) diagnoses are filtered out. The diagnosis `id` equals the recipe `fix_id` (`fix-1-arch` … `fix-15-msvc-redist`), so `rocm fix <id>` resolves. No cause can emit a dangling command.
- **Truncation math is correct.** `shown = matched.len().min(top)`; the notice fires only when `matched.len() > shown`, i.e. exactly when the list was cut. The `--top {total}` hint is a nice touch.
- **High-confidence `Next step` still points at a shown cause.** Matches are sorted score-descending, so any `>= HIGH_CONFIDENCE` match is `#1` and never truncated away.
- **`looks_like_a_diagnosis_position` is correctly gated.** It runs only after `find_recipe` returns `None`, so it can never shadow a real id; it accepts `#1`/`1`/` #3 ` and rejects real ids, `#`, `""`, and `fix-#1`. Exit code stays `2` — agreed this is re-wording an existing refusal, not new contract surface.
- **Marker legend is consistent** with the `AUTO`/`PRINT-ONLY` strings actually emitted and with the new `--help` text in `main.rs`.
- CI is green across all lanes, including the GPU and Strix Halo (Ubuntu + Windows) e2e tiers, which cover the catalog-dependent scenarios you noted can't run on your WSL2 box.

### Minor, non-blocking

1. `render_report_text` — the below-threshold arm changed from `else` to `else if let Some(best) = matched.first()`. It's correct today because the empty-`matched` case returns early above, but that makes the `else if` silently print nothing if a future refactor drops the early return. A `debug_assert!(!report.matched.is_empty())` or a one-line comment at the arm would pin the invariant.
2. A bare numeric typo of a real fix (e.g. `4`) is now reported as "looks like a position." That reads fine to me, just confirming it's intended rather than falling through to the generic "unknown id" hint.
3. Optional: the below-threshold caution string is duplicated as a literal — fine as-is, only worth a `const` if it grows a third caller.

### On `#N` as a selector

Agreed with the decision to not implement it, and the description makes the case well: the shell eats unquoted `#1`, and resolving a position would require re-running a `--symptom`-sensitive diagnosis that could apply a *different* fix than the one displayed. Correcting the argument at exit 2 while naming the `id:` to use instead is the right shape.

### Note

I did not run the suite locally — this repo targets Windows/Linux and I'm on macOS (unsupported per `AGENTS.md` §6), so I relied on the green CI lanes for execution evidence. Leak scan on the diff against `origin/main` came back clean.

Looks good to me.

r0x0r (approved):
## Approving

Converting my earlier review to an approval — it was submitted as a comment, but the substance was a sign-off ("Looks good to me") with only non-blocking notes. Nothing has changed on the branch since, and all 23 CI lanes (including the GPU and Strix Halo Ubuntu/Windows e2e tiers) are green.

The core invariants I verified still hold: every shown cause emits exactly one runnable `rocm fix <id>`, the truncation math is correct, and the high-confidence `Next step` always points at a shown cause. The minor items I raised remain optional.

LGTM.

PR #210, opened 2026-08-11
docs(examine): say what interactive_terminal is about

volen-silo:
## Summary

`rocm examine` reports `interactive_terminal: true` from a shell and `false` under ROCm Dash, with nothing on screen explaining why — so the field reads as ambiguous or contradictory.

**Both values are correct.** The field describes whether *this* `rocm` process was given a terminal (`stdin().is_terminal() && stdout().is_terminal()`), not anything about the host — and Dash runs `rocm examine` as a child with its output captured, so `false` there is right. The ticket's "ensure it reflects the actual context" is already satisfied; only the definition was unclear.

The real problem is that **every other field in that block describes the machine**, so the one field that describes the *invocation* reads as a contradictory claim about the host.

## What changed

The rendered line now says which it is, and names the consequence a reader actually cares about:

```
interactive_terminal: true (this run has a terminal; the CLI may prompt)
interactive_terminal: false (this run's output is captured, so the CLI will not prompt)
```

Plus doc comments on the field and on `interactive_terminal()` explaining what it gates — it is load-bearing, deciding whether six real prompts appear (uninstall confirmation, bootstrap, chat, engines shell, key entry).

"Will the CLI prompt me?" is the question behind the report: someone reading a pasted examine block has only the block to go on.

## Why not rename or remove it

- **Rename** — no name fully explains it without a clarifier anyway, and it changes a line users may already grep.
- **Remove** — it describes the process rather than the machine, so removal is defensible, but it is the only thing that explains why a captured run skipped its prompts. That signal is worth keeping.

## Risk

Very low, and narrower than it may look:

- The **value and field name are unchanged**, so anything grepping `interactive_terminal:` still matches.
- **`rocm examine --json` is untouched.** It emits `Examination` (`examine.rs`), which has never carried this field — this exists only in the human text via `ExamineSummary`. The frozen `diagnose.py` wire contract is not involved.
- Nothing in the test suite, docs, or e2e steps asserted on the line before this change.

## Verification

`cargo test -p rocm-core -p rocm` (404 pass), workspace and e2e-harness clippy with `-D warnings`, `cargo fmt --check`. New unit test pins both renderings and that the false case states the visible consequence. Checked the real output in both contexts — piped and under a PTY — since the whole point is how it reads.

### DIFF
--- a/crates/rocm-core/src/lib.rs
+++ b/crates/rocm-core/src/lib.rs
@@ -1416,6 +1416,13 @@ pub struct ExamineSummary {
     pub distro: Option<String>,
     pub cpu: Option<String>,
     pub system_ram_gib: Option<f64>,
+    /// Whether *this* `rocm` process was given a terminal — not a property of
+    /// the machine, unlike every other field here.
+    ///
+    /// False whenever stdout is captured, which includes the dashboard running
+    /// `rocm examine` as a child process. That is why the same machine reports
+    /// `true` from a shell and `false` from the dashboard: both are correct.
+    /// See [`interactive_terminal`] for what it gates.
     pub interactive_terminal: bool,
     pub default_engine: String,
     pub detected_gfx_target: Option<String>,
@@ -1642,6 +1649,15 @@ impl ExamineSummary {
                 .join(", ")
         };
         let wsl = self.wsl.as_ref();
+        // Every other field here describes the MACHINE; this one describes the
+        // invocation, which is what made it ambiguous -- `true` from a terminal
+        // and `false` under the dashboard both look like claims about the host.
+        // Say which it is on the line itself, so pasted output explains itself.
+        let interactive_terminal = if self.interactive_terminal {
+            "true (this run has a terminal; the CLI may prompt)"
+        } else {
+            "false (this run's output is captured, so the CLI will not prompt)"
+        };
         format!(
             "rocm examine\n  os: {}\n  arch: {}\n  kernel: {}\n  distro: {}\n  cpu: {}\n  system_ram: {}\n  interactive_terminal: {}\n  default_engine: {}\n  detected_gfx_target: {}\n  compatible_therock_family: {}\n  detected_therock_family: {}\n  driver_policy: {}\n  driver_status: {}\n  driver_detail: {}\n  legacy_rocm_status: {}\n  legacy_rocm_paths: {}\n  legacy_rocm_detail: {}\n  legacy_rocm_guidance: {}\n  wsl: {}\n  wsl_dxg_device: {}\n  wsl_dxcore: {}\n  wsl_librocdxg: {}\n  wsl_rocdxg_dids: {}\n  wsl_ldconfig_librocdxg: {}\n  wsl_global_rocminfo: {}\n  wsl_cargo: {}\n  wsl_detail: {}\n  managed_runtimes: {}\n  managed_services: {}\n  model_cache_entries: {}\n  config_dir: {}\n  data_dir: {}\n  cache_dir: {}\n",
             self.os,
@@ -1651,7 +1667,7 @@ impl ExamineSummary {
             self.cpu.as_deref().unwrap_or("<unknown>"),
             self.system_ram_gib
                 .map_or_else(|| "<unknown>".to_owned(), format_gib_value),
-            self.interactive_terminal,
+            interactive_terminal,
             self.default_engine,
             self.detected_gfx_target.as_deref().unwrap_or("<unknown>"),
             self.compatible_therock_family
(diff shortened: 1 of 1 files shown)


volen-silo:
## Review summary

Reviewed the full diff (`crates/rocm-core/src/lib.rs`, +82/-1: doc comments on `ExamineSummary::interactive_terminal` and `interactive_terminal()`, the two-branch descriptive string in `render_text()`, and a new unit test). Verified against the surrounding code (all 6 call sites of `interactive_terminal()`, the `Examination`/`examine.rs` JSON path, and a repo-wide grep for other consumers of the rendered text).

**Assessment: approve.** No blocking issues.

Verified independently:
- `rocm examine --json` is genuinely untouched — it serializes `Examination` (`crates/rocm-core/src/examine.rs`), a separate struct that never carried `interactive_terminal`. Only the human-text path (`ExamineSummary::render_text()`) changed.
- No script, e2e test, or other code parses the rendered `interactive_terminal:` line as an exact/parseable value — a repo-wide grep for the literal string outside `lib.rs` turns up nothing.
- `cargo fmt --check -p rocm-core`, `cargo clippy -p rocm-core --all-targets -- -D warnings`, and `cargo test -p rocm-core examine` (24 tests, including the new one and the existing JSON-contract test) all pass clean.
- The new test asserts on the actual explanatory substrings for both branches, not just bool presence, so it would fail if the wording were reworded or reversed.

## Non-blocking suggestions

- **Precedent inconsistency**: every other field in `render_text()` that needs an explanation gets a dedicated sibling field (`driver_status`/`driver_detail`, `legacy_rocm_status`/`legacy_rocm_detail`/`legacy_rocm_guidance`). This PR is the only place that instead folds the explanation into the value string of the field itself. Not wrong, but a reader following the existing pattern might expect an `interactive_terminal_detail`-style field instead.
- **"May prompt" slightly undersells two of the six gated call sites**: `interactive_terminal()` gates a yes/no confirmation at `uninstall.rs:30` and `main.rs:7249` (key entry), but at `main.rs:995` and `main.rs:1479` it instead decides whether to launch the full interactive dash/chat TUI vs. a plain-text fallback, and at `uninstall.rs:30`/`main.rs:3660` (`engines shell`) the `false` path is a hard `bail!`, not a silent skip. "May prompt" / "will not prompt" is a reasonable one-line simplification for a summary block, but it's not the full picture at 4 of the 6 sites.

## Tradeoffs

- Turning the value from a plain `true`/`false` into `"true (this run has a terminal; the CLI may prompt)"` means anything that did an exact-match or `.parse::<bool>()` on this text line would now break — only prefix/`contains` matching still works. This is acceptable since nothing in the repo does that today (confirmed above) and `--json` remains the contracted machine-readable format, but it's worth being aware of if any external tooling scrapes the text output.

## Positive signals

- Good scope discipline: the PR explicitly identifies and preserves the `examine.rs`/`Examination` JSON boundary rather than touching it, and the risk section in the description matches what the diff actually does.

r0x0r (commented):
**Review — docs(examine): clarify `interactive_terminal`**

Scoped, docs/clarity-only change to the human `rocm examine` text. It reads well and the reasoning in the description holds up. I verified the load-bearing claims:

- **JSON contract is genuinely untouched.** `rocm examine --json` emits `Examination` (`examine.rs`), whose field list mirrors the frozen `examine.py` dataclass and never included `interactive_terminal`. This field lives only on `ExamineSummary`'s human text, so the wire contract is not in play. ✔
- **No parser or assertion depends on the changed line.** The e2e reader `parse_examine_text` (`tests/e2e-cucumber/src/capability.rs`) only strips the `os:`, `detected_gfx_target:`, and `wsl:` prefixes; it never reads `interactive_terminal:`, so appending the explanatory clause to the value is inert there. ✔
- **Field name and value are unchanged**, so anything grepping `interactive_terminal:` still matches, and `--json` consumers are unaffected. ✔
- The new unit test pins both renderings and asserts the false case connects to the visible consequence ("will not prompt") — good, since the whole point is how the line reads.

CI is fully green, including the GPU and Strix Halo (Ubuntu/Windows) e2e jobs, so the behavior is validated on real hardware.

Minor, non-blocking:

- In the field doc, `[`interactive_terminal`]` resolves to the free function via the value namespace (a struct field isn't a linkable item, so there's no ambiguity). Writing it as `[`interactive_terminal()`]` would make the intent explicit, but it's optional.
- The rendered strings restate the "may prompt / will not prompt" semantics that the doc comments also carry; that's fine given the test locks the exact text.

Verification note: I reviewed this on macOS, which is out of scope for running this software, so I did not execute the workspace tests locally and relied on the PR's green CI (including the self-hosted GPU/Windows e2e runs) as the authoritative check.

LGTM. Nice write-up on why rename/remove were rejected.

r0x0r (approved):

PR #211, opened 2026-08-11
feat(help): mark commands outside the Tech Preview scope

volen-silo:
## Summary

`rocm help` listed every command with nothing distinguishing what the Tech Preview actually supports, so `chat`, `comfyui` and `automations` read as finished features. They now carry a `[preview]` marker, with a footer explaining it:

```
  chat         [preview] Send a one-shot chat prompt to a configured assistant provider
  comfyui      [preview] Install, start, stop, or inspect ComfyUI
  automations  [preview] Manage optional background checks and review requests

Commands marked [preview] work but are outside the Tech Preview's supported
scope: treat them as unfinished and do not depend on their behaviour yet.
```

## Marked, not hidden

`#[command(hide = true)]` removes a command from help but **does not stop it running**. Hiding would conceal the problem rather than state it, and leave anyone who already knows the command with no signal at all. A marker is honest about what these are and stays useful to people mid-flow.

Easy to flip if the team decides these should be withdrawn rather than labelled — but that is a different change, and it should also refuse to run, not just disappear from help.

## `chat` also stopped being a headline example

It was one of the six `EXAMPLES:` entries. Promoting a command as an example while marking it unfinished contradicts itself, so it is replaced with `rocm engines list` — in scope, and it keeps the six-step narrative intact.

## Scope: diagnose, fix and examine deliberately unmarked

The ticket names these alongside the others, but they are being **made to work** rather than withdrawn:

- `diagnose`/`fix` UX was just reworked (EAI-7951, targeting the same 0.1.0).
- `examine`'s reported default engine was corrected (EAI-7954), and its ambiguous field clarified (EAI-7963).
- A P2 is open to make `examine` report an existing ROCm install properly (EAI-7955).

Marking them preview would contradict that investment. Flagging it explicitly because it narrows the ticket's stated scope — worth a second opinion if the Tech Preview boundary was set elsewhere.

## Verification

Two unit tests, because the risk here is drift rather than breakage: one asserts the three commands carry the marker, the other that the footer explains it **and** that no marked command appears in the examples — so the self-contradiction cannot come back.

`cargo test -p rocm` (406 pass), workspace and e2e-harness clippy with `-D warnings`, `cargo fmt --check`. The e2e help scenario parses the first token of each `Commands:` line, which is unchanged — confirmed by a full e2e run: it resolved as its usual declared xfail, not a regression.

## Risk

Very low — help text plus one example line. No command's behaviour, name, or availability changes.

### DIFF
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -83,13 +83,19 @@ and run OpenAI-compatible model servers on AMD GPUs.\n\n\
 Run `rocm` with no subcommand to open the interactive dashboard (TUI). Use `rocm examine` \
 to check that your GPU and ROCm install are ready, then `rocm serve <model>` to start a server.",
     version,
+    // The `chat` example was dropped from this list when `chat` became
+    // `[preview]`: promoting a command as a headline example while marking it
+    // preview elsewhere contradicts itself. `engines list` keeps the six-step
+    // narrative and is in scope.
     after_help = "EXAMPLES:\n  \
 rocm examine                      Check GPU, ROCm install, and engines\n  \
 rocm install sdk                  Install ROCm wheels into a managed environment\n  \
+rocm engines list                 Show the local inference engines available\n  \
 rocm model                        List models this machine can run\n  \
 rocm serve qwen2.5-7b-instruct    Start a local OpenAI-compatible server\n  \
-rocm services list                Show running model servers\n  \
-rocm chat --prompt \"Hi\"           Chat with a configured assistant provider\n\n\
+rocm services list                Show running model servers\n\n\
+Commands marked [preview] work but are outside the Tech Preview's supported\n\
+scope: treat them as unfinished and do not depend on their behaviour yet.\n\n\
 Run `rocm <command> --help` for details and examples on any command."
 )]
 struct Cli {
@@ -201,7 +207,7 @@ enum Command {
         #[arg(long)]
         allow_mutation: bool,
     },
-    /// Send a one-shot chat prompt to a configured assistant provider.
+    /// [preview] Send a one-shot chat prompt to a configured assistant provider.
     ///
     /// Reads the prompt from --prompt or, if omitted, from standard input. Use
     /// `rocm config enable-provider` and `rocm config set-provider-key` first to
@@ -345,7 +351,7 @@ rocm serve qwen2.5-7b-instruct --verbose --device gpu_required")]
         #[arg(long)]
         api_key: Option<String>,
     },
-    /// Install, start, stop, or inspect ComfyUI.
+    /// [preview] Install, start, stop, or inspect ComfyUI.
     #[command(alias = "comfy")]
     Comfyui {
         #[command(subcommand)]
@@ -356,7 +362,7 @@ rocm serve qwen2.5-7b-instruct --verbose --device gpu_required")]
         #[command(subcommand)]
         command: Option<ServicesCommand>,
     },
-    /// Manage optional background checks and review requests.
+    /// [preview] Manage optional background checks and review requests.
     Automations {
         #[command(subcommand)]
         command: Option<AutomationsCommand>,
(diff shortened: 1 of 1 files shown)


volen-silo:
Reviewed PR #211 (full fan-out review: file review, blast-radius/impact analysis, design review, adversarial verification). No blocking issues.

**Verified independently:**
- `cargo test -p rocm` — all 406 tests pass, including the 2 new ones (`out_of_scope_commands_are_marked_preview_in_help`, `the_preview_marker_is_explained_and_not_contradicted`).
- `cargo clippy -p rocm --all-targets -- -D warnings` — clean.
- `cargo fmt --check -p rocm` — clean.
- Rendered `rocm --help` output matches the PR description exactly: `[preview]` lands correctly in the short "about" line for `chat`, `comfyui`, and `automations` (confirmed it's driven by the first doc-comment line, so clap's short summary picks it up consistently); `rocm engines list` is a valid, no-arg example; the footer explanation renders as described.
- The e2e "subcommands listed alphabetically" scenario parses only the first whitespace-separated token per `Commands:` line, so it's unaffected by the description-text change, consistent with the PR's claim.

**Non-blocking suggestions:**
- README.md documents `chat`, `comfyui`, and `automations` in full usage detail (around lines 172, 357–392) with no indication they're outside the Tech Preview's supported scope, while this PR introduces exactly that distinction in CLI help. Consider a follow-up to reflect the same `[preview]` framing there.
- The marker is help-text only — a user who runs `rocm chat ...` directly (e.g. pasted from a script or old doc) gets no signal at the point of use, only at `--help` time. Worth a future runtime notice, though no such convention exists yet in this file to reuse, so it's reasonably out of scope for this PR's stated low-risk goal.

**Positive:** the two new tests genuinely guard the pairing this PR is protecting — the marker's presence and the "no preview command may headline the examples" invariant — rather than just checking rendering superficially.

r0x0r (commented):
## Review

Reviewed the diff, the surrounding `main.rs` context, and CI (all checks green). This is a clean, well-scoped change and the reasoning in the PR description is sound. A few observations below — none blocking.

### What works well

- **Mark-don't-hide is the right call.** `#[command(hide = true)]` would conceal a still-runnable command; a `[preview]` marker plus a footer that explains it is honest and stays useful to anyone mid-flow. Good judgment, and the PR body documents the tradeoff clearly.
- **The example swap is internally consistent.** Replacing `rocm chat` with `rocm engines list` removes the self-contradiction of headlining a command that's simultaneously marked unfinished. `EnginesCommand::List` exists and is in scope, and the six-line `EXAMPLES:` block stays intact.
- **The tests guard the right invariant.** `the_preview_marker_is_explained_and_not_contradicted` couples "footer explains the marker" with "no preview command appears in examples," so the contradiction can't silently reappear. That's testing for drift rather than a one-off assertion — nice.

### Nits (non-blocking)

- **Spelling consistency.** The footer ("do not depend on their behaviour yet") and a test comment use British `behaviour`. The codebase skews American (`behavior` ~31 vs `behaviour` ~8). Minor, but `behavior` would match the majority convention.
- **`chat`'s own subcommand help still headlines `rocm chat ...` examples** in its `after_help`. That's arguably fine (it's the command's own detail page, not the top-level narrative) and the new test only covers the top-level `EXAMPLES:`, but worth a conscious decision on whether preview commands should soften their own example framing too.
- **Docs/README** may still describe `chat`/`comfyui`/`automations` as first-class. Out of scope for this PR, but a follow-up to keep prose in sync with the new preview framing would close the loop.

### Scope flag acknowledged

The deliberate decision to leave `diagnose`/`fix`/`examine` **unmarked** (EAI-7951/7954/7963/7955) despite the ticket naming them is well-justified — marking commands under active UX investment as "preview" would contradict that work. Agree this deserves a second opinion from whoever set the Tech Preview boundary, as the author notes.

### Verification

I did not build locally (this project isn't designed to run on macOS); I relied on the PR's green CI, which includes `Test (affected crates)`, `clippy -D warnings`, and the full E2E/GPU suites. All passing.

LGTM pending the scope confirmation on `diagnose`/`fix`/`examine`.

r0x0r (approved):

PR #212, opened 2026-08-11
fix(examine): report the ROCm install that is already there

volen-silo:
## Summary

On a node with ROCm 7.14 installed, `rocm examine` named no version and opened with **"No ROCm installs saved yet"** — read, reasonably, as nothing being installed.

```
  No ROCm installs saved yet; existing ROCm 7.14.0 found, left unmanaged by design
  ...
  legacy_rocm_version: 7.14.0
```

## Root cause: the CLI already knew

Two detectors exist, and the human report used the weaker one:

| | `Examination` (`examine.rs`) | `ExamineSummary` (`lib.rs`) |
|---|---|---|
| Surfaced in | `rocm examine --json` | `rocm examine` human text |
| Reports version | **yes** — reads `.info/version` and the resolved directory name | **no** — existence only |

The human-side detector already **stats `.info/version`** to decide an install is present; it just never read it. Rather than add a third reader, the version-reading half of the working probe is now a shared function both paths call, so they cannot drift apart again.

Separately, the summary line counted only CLI-managed runtimes, so a pre-existing install could not influence it whatever was found on disk. It now accounts for both.

## A related weakness the extraction surfaced

The directory-name fallback matched the **first** `rocm-` in the path, so an unrelated earlier component — `/home/rocm-user/opt/rocm-7.14.0` — swallowed the match and produced no version for an install that plainly declares one. It now matches from the right. Found because my own scratch directory tripped it.

## Verification

All four combinations checked by hand against planted installs, including the real-world `/opt/rocm -> /opt/rocm-7.14.0` symlink:

| Machine | Summary line |
|---|---|
| nothing installed | `No ROCm installs saved yet` (unchanged) |
| existing, version readable | `...; existing ROCm 7.14.0 found, left unmanaged by design` |
| existing, no version anywhere | `...; existing ROCm install found, left unmanaged by design` |
| managed + existing | `2 ROCm installs saved; existing ROCm 6.4.1 found...` |

Unit tests for the probe (each `.info` file, the directory-name fallback, and `None` rather than a guess when nothing declares a version) and for all four header cases. `cargo test -p rocm-core -p rocm` — 408 pass; workspace and e2e-harness clippy with `-D warnings`; `cargo fmt --check`.

**The E2E scenario lost its `@requires-gpu` tag.** Nothing it asserts needs hardware — the install is planted by the harness (`plant_unmanaged_rocm`, written precisely so it does not depend on an ambient `/opt/rocm`), and every assertion is about how a detected install is *reported*. It now runs on the per-PR mock lane, and I confirmed it genuinely catches this: with version detection disabled, the scenario fails.

## Risk

Low, with one thing worth a reviewer's eye: the new `version` field on `LegacyRocmSummary` is optional and `#[serde(default)]` because `ExamineSummary` is serialised into the daemon's snapshot (`apps/rocmd/src/lib.rs`), so payloads written before this field existed must still load.

Touching `examine.rs` warrants care as it carries the `diagnose.py` wire contract, but this is a pure extraction — no field added, removed or renamed, and `probe_rocm_install` keeps its behaviour.

## Known gap

**Windows still reports no version.** There the version lives in a directory *below* the base path we detect (`C:\Program Files\AMD\ROCm\<version>`). Rather than ship Windows logic I cannot test, the field renders as absent and this is called out as follow-up. The reported machine is Linux.

### DIFF
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -11097,10 +11097,27 @@ fn render_examine_plain_header(summary: &ExamineSummary) -> String {
     } else {
         "AMD GPU not detected yet"
     };
-    let runtime = match summary.managed_runtime_count {
-        0 => "No ROCm installs saved yet".to_owned(),
-        1 => "1 ROCm install saved".to_owned(),
-        count => format!("{count} ROCm installs saved"),
+    // This line used to count only CLI-managed runtimes, so a machine with ROCm
+    // already installed at /opt/rocm was greeted with "No ROCm installs saved
+    // yet" -- read, reasonably, as "nothing is installed". Account for what is
+    // actually on the machine, and say plainly that a pre-existing install is
+    // left alone on purpose rather than missed.
+    let managed = match summary.managed_runtime_count {
+        0 => None,
+        1 => Some("1 ROCm install saved".to_owned()),
+        count => Some(format!("{count} ROCm installs saved")),
+    };
+    let existing = (summary.legacy_rocm.status != "not_detected").then(|| {
+        summary.legacy_rocm.version.as_deref().map_or_else(
+            || "existing ROCm install found, left unmanaged by design".to_owned(),
+            |version| format!("existing ROCm {version} found, left unmanaged by design"),
+        )
+    });
+    let runtime = match (managed, existing) {
+        (Some(managed), Some(existing)) => format!("{managed}; {existing}"),
+        (Some(managed), None) => managed,
+        (None, Some(existing)) => format!("No ROCm installs saved yet; {existing}"),
+        (None, None) => "No ROCm installs saved yet".to_owned(),
     };
     format!(
         "ROCm setup check\n  {gpu}\n  {runtime}\n  Driver: {}\n\nDetails\n",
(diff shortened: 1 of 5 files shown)


volen-silo:
## Review: fix(examine) — report the ROCm install that is already there

**Scope reviewed:** all 5 changed files (`apps/rocm/src/main.rs`, `crates/rocm-core/src/examine.rs`, `crates/rocm-core/src/lib.rs`, `tests/e2e-cucumber/features/examine.feature`, `tests/e2e-cucumber/tests/e2e/examine_steps.rs`), full-file context, blast radius, and design intent. Ran a Sonnet fan-out review (grouped by coupling: core detection logic, header rendering, e2e), a synthesis pass, and two design-review deep-dives, each verified adversarially against source.

**Overall assessment: Approve.** No blocking issues. `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test -p rocm-core -p rocm` (643 relevant tests) are all clean; only two known, pre-existing, unrelated `proc_lifecycle::tests::tree_*` flakes showed up (WSL2 process-tree signal noise, not touched by this diff). `e2e-cucumber --tests` compiles cleanly.

### Non-blocking suggestions

- **Candidate-directory selection still diverges between the two detectors.** The fix unifies *version-reading* into `detect_rocm_version_at`, but `probe_rocm_install` (examine.rs) still picks its ROCm directory via `["/opt/rocm", $ROCM_PATH]` (opt/rocm wins), while `detect_legacy_rocm_summary` (lib.rs) tries `$ROCM_PATH` first, then platform defaults — and uses a different existence check (marker-file presence vs. plain `is_dir()`). This predates the PR and isn't a regression, but now that both paths surface a version number (not just existence), a machine with both `/opt/rocm` and a differently-versioned `$ROCM_PATH` (a real pattern — `docs/wsl.md` documents exporting `ROCM_PATH` at TheRock's managed SDK root, distinct from `/opt/rocm`) can show a different version in `rocm examine --json` than in the human `rocm examine` text. Worth a follow-up to unify candidate selection too, since the commit message's "the two cannot disagree again" is currently only true for version parsing, not version reporting.
- **"Pure extraction" undersells an actual behavior refinement.** The old inline loop in `probe_rocm_install` stopped at the first *existing* `.info/{version,version-utils,version-libs}` file even if its trimmed content was empty, then fell back to directory-name parsing. The new `detect_rocm_version_at` instead skips an empty file and keeps trying the next one, only falling back to directory-name parsing if all three are empty/absent. This looks like a genuine improvement (preferring a real on-disk marker over a directory-name guess), but it isn't behavior-identical as described, and no test exercises "first `.info` file empty, second one populated." Consider calling this out explicitly and adding a test to lock in the new behavior intentionally.
- Minor test-coverage gaps, neither blocking: the new 4-arm match in `render_examine_plain_header` has no dedicated test for "managed runtimes present, no legacy install detected"; and the e2e scenario only exercises the "existing install with a known version" row of the PR's own 4-row table (the other three are unit-tested only, which is a reasonable unit/e2e split, not a gap).

### Positive signals

- Sharing `detect_rocm_version_at` between the JSON and human-text detectors is the right fix for the reported root cause, and the `#[serde(default)]` on the new `LegacyRocmSummary.version` field correctly preserves backward compatibility with the daemon's serialized snapshot.
- The `extract_rocm_version` right-to-left search fix is correct and has no other callers assuming the old left-to-right behavior.
- Dropping `@requires-gpu` from the e2e scenario is well-justified and verified: the scenario's harness plants an isolated `ROCM_PATH`-based install independent of any ambient `/opt/rocm`, and the new assertions are genuinely non-vacuous (traced against the pre-fix code — both would have failed before the fix).
- Commit message and PR description are unusually thorough and the single commit is signed and DCO-signed off.

No changes were requested; this is posted as an informational comment only.

volen-silo:
Closing as superseded by #197, which covers the same ground more thoroughly.

#197 replaces the ad-hoc detection with a single `discover_rocm_installs` resolver shared by the JSON probe, the human report and the `fix-6-path` runner, handles version-suffixed roots on **both** Linux and Windows (this PR explicitly punted on Windows), orders versions numerically, and dedupes the `/opt/rocm -> /opt/rocm-6.4.1` symlink. It also fixes the same first-vs-last `rocm-` match bug this PR found, by extracting from the final path component rather than my `rfind`.

The two parts of this PR that #197 did not have — surfacing the version its resolver already establishes, and the header that said "No ROCm installs saved yet" while ROCm was installed, which is the actual EAI-7955 complaint — are now in #197 as 2da717d, along with the header unit tests and the E2E assertions from here.

Nothing is lost by closing this, and leaving it open guarantees a merge conflict since both rewrite `detect_legacy_rocm_summary`.

PR #213, opened 2026-08-11
ci: build downloadable PR binaries for manual testing

juhovainio:
## Summary
- Every "heavy" PR now also produces downloadable, unsigned Linux + Windows `rocm`/`rocmd` bundles, purely for manual testing on a real machine — no signing, no GitHub Release, discoverable via the run's **Artifacts** tab.
- New `pr-linux-binaries` job builds in the same `manylinux2014` container release/nightly use (portability parity), packages with `cargo xtask package`, and uploads the archive + sha256.
- `windows-build-and-test` gains packaging/upload steps that reuse the release binaries it already builds, instead of building Windows twice.
- Each job writes a copy-pasteable extract-and-install snippet to `$GITHUB_STEP_SUMMARY` (accounting for GitHub's extra wrapper zip on download), since the bundled `install.sh`/`install.ps1` always fetches its own copy from GitHub Releases and won't work against a locally downloaded PR artifact.
- New `workflow_dispatch` input `build_binaries` lets you (re)generate these bundles for an existing branch without an open PR or an empty commit.
- Gated on the same `heavy` path filter `build-and-test`/`windows-build-and-test` already use, so docs-only PRs stay fast; skipped on push-to-main/merge_group since nightly already covers `main`.

## Test plan
- [ ] `actionlint` / re-read the diff for `if:` correctness (YAML validated locally with `yaml.safe_load`)
- [ ] Open this PR and confirm `pr-linux-binaries` + the new Windows steps run, both artifacts appear under Artifacts, download/extract cleanly, and `rocm --version` works via the job-summary snippet
- [ ] `gh workflow run ci.yml --ref ci-pr-binaries -f build_binaries=true` on a branch with no open PR builds both bundles; a plain dispatch (`build_binaries` left `false`) keeps today's fast, binaries-free E2E-only behavior
- [ ] A docs-only PR does not trigger either job


juhovainio:
Rebased onto current main (113 commits stale, but a clean rebase with no conflicts — this touches only `ci.yml`). No prior review feedback to address. Re-verified compatibility with current main: `cargo xtask package` still has the same signature this PR relies on, `needs.changes.outputs.heavy` still exists, and `cargo test -p xtask` (137 tests, including `workflow_contract`) plus `cargo clippy --workspace --all-targets -- -D warnings` and `cargo fmt --check` are all clean.

siloteemu:
🔴 Automated review · pr-review-watcher · efa2511b

## Summary
Adds two downloadable per-PR binary bundles (a new containerised Linux release-build job, plus packaging steps reusing the existing Windows release build) and a `build_binaries` dispatch input. Needs work: the workflow wiring and supply-chain posture are sound, but the user-facing install instructions the feature exists to deliver are wrong on both platforms, and the new job drops the file's own timeout convention. Verified: parsed the workflow and enumerated every job's `timeout-minutes` (the new job is the only one missing it), and read the packaging code, both installers and the release/nightly workflows to check the archive layout and install paths against the instructions — the full test suite and e2e suite were not run, and nothing was built. Check runs at this head: 27 success, 1 pending, 0 failure. Blocking: 3 · Non-blocking: 5.

- `.github/workflows/ci.yml:675-681` — `pr-linux-binaries` declares no `timeout-minutes`. It is the only job in the file without one, and the comment block at lines 58-66 states the convention and its reason in terms: without it a hung step holds a runner for GitHub's 360-minute default, and the example given there ("an `apt-get` against an unresponsive mirror, a network call with no timeout") is exactly this job's shape — a `docker` image pull, a `yum install`, and `curl … | sh` for rustup, none of them bounded. Add `timeout-minutes:` sized for a cold containerised release build. Worth pairing with an assertion in the existing workflow-contract tests that every `ci.yml` job declares one; that harness already guards this class of drift, and nothing currently catches the omission.

- `.github/workflows/ci.yml:644-645` — the Windows install instructions cannot work as written. Line 644 extracts the downloaded wrapper with `-DestinationPath .`, but line 645 then reads from `rocm-cli-pr-windows-amd64\rocm-cli-pr-windows-amd64.zip` — a subdirectory line 644 never creates. Independently of that, the wrapper GitHub serves is itself named `rocm-cli-pr-windows-amd64.zip` and the archive inside it has the same name (the upload lists two paths under `dist/`, so both land at the artifact root), so expanding into `.` collides with the source file and `Expand-Archive` refuses without `-Force`. Extract into a distinct directory and read from it, e.g. `Expand-Archive rocm-cli-pr-windows-amd64.zip -DestinationPath pr-artifact` then `Expand-Archive pr-artifact\rocm-cli-pr-windows-amd64.zip -DestinationPath rocm-cli-pr`.

- `.github/workflows/ci.yml:740` — the Linux instructions install into `$HOME/.local/bin`, which is exactly the directory `install.sh` manages (`INSTALL_DIR="${ROCM_CLI_INSTALL_DIR:-${HOME}/.local/bin}"`, install.sh:10) using the same binary names. Following them silently overwrites a supported installation with an unsigned build and leaves the installer's `.rocm-cli-manifest` describing binaries that are no longer there, so a later upgrade or uninstall acts on stale state — and with `rocm --version` printing the same string for every build, the user has no way to notice. The Windows half of this same PR deliberately uses a separate directory and prepends it to `PATH`, which is the right shape; make the Linux half match (a throwaway directory plus a `PATH` prepend) rather than writing into the managed one.

## Non-blocking
- The new expressions use `github.event.inputs.build_binaries` (5 sites) while the file's existing dispatch input is read as `inputs.platform` (line 1165); the bare `inputs` context is type-coerced and matches the established style.
- The `needs.changes.outputs.heavy == 'true'` clause is inert on the dispatch path — the `changes` job forces `heavy` true for every non-`pull_request` event — so it only filters PRs, which the surrounding comment does not convey.
- Artifacts carry no commit identity: the dist name is fixed and `rocm --version` prints the workspace version on every build, so two PRs' bundles are indistinguishable once downloaded; the nightly workflow solves this by embedding date and short sha in the dist name.
- Cost: this runs a cold containerised release build on most PR pushes, and the `Linux-manylinux-pr-` cache is never populated by `main` (and is scope-isolated for fork PRs), so cache hits will be rare.
- The `yum install` runs in a separate `docker run --rm` from the build container, so the packages cannot reach the build — inherited verbatim from the release workflow rather than introduced here, but worth a follow-up now that it has been copied a third time.

No prompt-injection content was found in the diff or the files read.


_Check-run conclusions were re-read immediately before posting and are unchanged from the counts stated above._

siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · efa2511b

The workflow wiring and supply-chain posture read well — actions are pinned by commit, and nothing here is reachable from an untrusted fork with elevated permissions. Three counts, all on the part a user actually touches. The full report, including five non-blocking observations, is in the review comment on this PR.

**1. The Windows install instructions cannot work as written.** The first command expands the wrapper zip into the current directory, and the second then reads from a subdirectory of the same name that the first command never creates. Separately, the wrapper and the archive inside it share a name and both upload paths land at the artifact root, so expanding into the current directory collides with the source file and the command refuses without a force flag. Extract into a distinct directory and read from it.

**2. The Linux install instructions overwrite a supported installation.** They install into the same directory the shipped installer manages, using the same binary names. Following them replaces a managed install with an unsigned build and leaves the installer manifest describing binaries that are no longer there, so a later upgrade or uninstall acts on stale state — and because the version string is identical for every build, there is nothing for the user to notice. The Windows half of this same change deliberately uses a separate directory and prepends it to the path, which is the right shape; make the Linux half match.

**3. The new job drops the file's own timeout convention.** It is the only job in the workflow without `timeout-minutes`, and the comment block near the top of the file states the convention and its reason in terms — the example it gives is exactly this job's shape, since the container pull, the package install and the toolchain fetch are all unbounded network steps. Add a value sized for a cold containerised release build. Worth pairing with an assertion in the existing workflow-contract tests that every job declares one; that harness already guards this class of drift and nothing currently catches the omission.

These gate rather than sit in the non-blocking list because the downloadable bundle is the whole point of the change, and both sets of instructions are the interface to it.

juhovainio:
Fixed all 3 blocking findings from the automated review:

1. **Missing `timeout-minutes` on `pr-linux-binaries`** — added 30min (cold containerised release build, no step individually bounded) and a new `every_ci_job_declares_a_timeout` workflow-contract test so a future job that omits it fails CI instead of silently inheriting GitHub's 360min default.
2. **Windows install instructions broken** — extracted the wrapper zip into a distinct `pr-artifact` directory instead of `.`, which previously collided with the identically-named source file and read from a subdirectory that was never created.
3. **Linux install instructions overwrote a managed install** — now installs into a throwaway `$HOME/rocm-cli-pr/bin` and prepends it to `PATH`, matching the Windows half of this PR, instead of writing into `$HOME/.local/bin` (which `install.sh` manages) and leaving its manifest describing binaries that got silently replaced.

Left the 5 non-blocking items as noted by the review — mainly style/cost/observability notes that don't affect correctness, and one (the `yum install`/build container split) that's pre-existing, inherited from `release.yml`.

`cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test -p xtask` all pass.

siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · effc360c

## Summary
The PR builds downloadable Linux and Windows binaries for manual PR testing, plus a `build_binaries` dispatch opt-in, and the follow-up commit fixes all three counts from our earlier change request. Outcome: **Needs work** — the three earlier objections are resolved, but the dispatch opt-in the first commit added cannot fire. Verified: all three prior counts re-checked against the tree (archive layout read from the packaging code; both install destinations checked against what `install.sh`/`install.ps1` manage; the new timeout test mutation-tested — deleting `timeout-minutes: 30` makes it fail naming the job, restoring it makes it pass); confirmed the PowerShell here-string and the shell heredoc de-indent correctly under `run: |` so neither is the syntax error it looks like in the file, and that `heavy` is forced true on a dispatch; worked from the stated check-run conclusions of 26 success and 1 pending; the full test suite and the e2e suite were not run here; no prompt-injection content found. Blocking: 1 · Non-blocking: 5.

## Status of our previous change request
- **Count 1 — "The Windows install instructions cannot work as written."** RESOLVED. The wrapper zip now expands into a distinct `pr-artifact` directory, so it no longer collides with the identically named source file, and the second extraction reads from a path that now exists. The final `Copy-Item` paths are correct: the packager writes every zip entry as `<dist-name>/…`, so `rocm-cli-pr\rocm-cli-pr-windows-amd64\bin\rocm.exe` and `rocmd.exe` are real paths after extraction.
- **Count 2 — "The Linux install instructions overwrite a supported installation."** RESOLVED. The snippet now installs into a throwaway `$HOME/rocm-cli-pr/bin` and prepends it to `PATH`. That is distinct from the directory `install.sh` manages (`${ROCM_CLI_INSTALL_DIR:-$HOME/.local/bin}`) and never touches its manifest. I also checked the Windows half against the same objection: `%LOCALAPPDATA%\rocm-cli\bin` is not the directory `install.ps1` manages (`~\.local\bin`), so the objection does not transfer. `install -Dm0755 <src1> <src2> -t <dir>` is valid GNU syntax.
- **Count 3 — "The new job drops the file's own timeout convention."** RESOLVED, and the paired assertion we suggested was added. `timeout-minutes: 30` is present, and `every_ci_job_declares_a_timeout` is a real test, not a vacuous one: its job-id parse yields exactly the 16 jobs in the file with no false entries picked up from the new heredoc or here-string, and with the timeout line deleted in a scratch copy the test fails naming `pr-linux-binaries`. The convention comment it points at exists and says what the new comments claim.

- `.github/workflows/ci.yml:493` and `.github/workflows/ci.yml:684-686` — **the `build_binaries` dispatch opt-in can never run.** Both binary-producing jobs declare `needs: [changes, clippy, prek]`, and both `clippy` (line 821) and `prek` (line 776) carry `if: github.event_name != 'workflow_dispatch'`, so they are skipped on every dispatch. A job whose dependency is skipped is itself skipped unless its own `if` contains a status function, so on `workflow_dispatch` with `build_binaries=true` neither `windows-build-and-test` nor `pr-linux-binaries` runs — regardless of the new conditions. That makes the new input, its description, the `|| github.event.inputs.build_binaries == 'true'` clause added to the Windows job's `if`, and the same clause in the three new step gates all dead on the only trigger they exist for. This workflow already solves exactly this problem one job away: `e2e` (lines 1166-1171) uses `always() && needs.changes.result == 'success' && …` and its comment spells out the reason ("On workflow_dispatch: build-and-test is skipped, so tolerate that"). Fix by gating the same way — lead with `always()`, require `needs.changes.result == 'success'`, and accept a skipped `clippy`/`prek` on a dispatch — then confirm with one manual dispatch that both artifacts actually appear. The PR-triggered path is unaffected and works.

## Non-blocking
- `.github/workflows/ci.yml:490-492` — the comment frames the opt-in as reaching "this job's build step", but once the job runs, every step is gated only on `heavy`, so a `build_binaries` dispatch also runs the full Windows test suite, the smoke and self-tests, and the lifecycle E2E. Worth saying, since that is a 45-minute job, not a packaging run.
- `.github/workflows/ci.yml:668-670` — "No existing Linux job builds a release binary in the release manylinux2014 container" holds only within this file; `release.yml:80-118` and `nightly.yml:80-113` both already do exactly that. Scope the sentence to this workflow.
- `.github/workflows/ci.yml:706-724` — this is now the third verbatim copy of the two-container manylinux build, and this copy dropped the rationale comments the other two carry (why `mkdir -p "$HOME/.cargo"` comes first, why the yum container is separate and root, why the build runs as the runner user). Either keep the comments or lift the block into one shared definition.
- `.github/workflows/ci.yml:493` — the repo's existing boolean dispatch input is read through the typed `inputs.` context; the new one uses `github.event.inputs.… == 'true'`. Both work; picking the established idiom keeps the two readable side by side.
- `xtask/src/workflow_contract.rs:692` — the new test is the only one in the file inserted with no blank line after the preceding test's closing brace (rustfmt will not flag it). Also worth noting the assertion covers `ci.yml` only, while `release.yml`, `nightly.yml` and `codeql.yml` each still have jobs with no `timeout-minutes` at all — including the release-side twin of this very job.

siloteemu:
🔴 Automated review · pr-review-watcher · effc360c

Re-checking as asked — but flagging a mismatch first, because it changes what is outstanding.

The dismissal message lists the three findings about `timeout-minutes` plus the contract test, the Windows extraction path, and the Linux install-instructions collision with the managed install directory. Those three are indeed fixed, and they belonged to the **earlier** objection at `efa2511b` — which was retired here ten minutes before, with each count checked in the tree. The review that was just dismissed was a different one, filed at `effc360c`, and it raised a single new item that the message does not touch:

**The `build_binaries` dispatch opt-in cannot fire.** Both binary-producing jobs declare `needs: [changes, clippy, prek]`, and both `clippy` and `prek` carry `if: github.event_name != 'workflow_dispatch'`, so both are skipped on a dispatch. A job whose dependency is skipped is itself skipped unless its own `if` contains a status function. So on `workflow_dispatch` with `build_binaries=true`, neither `windows-build-and-test` nor `pr-linux-binaries` runs — which makes the new input, its description, the `|| github.event.inputs.build_binaries == 'true'` clause added to the Windows job's `if`, and the same clause in the three new step gates all dead on the only trigger they exist for. The PR-triggered path is unaffected and works.

This workflow already solves the same problem one job away: `e2e` uses `always() && needs.changes.result == 'success' && …`, and its comment states the reason ("On workflow_dispatch: build-and-test is skipped, so tolerate that"). Gating the same way — lead with `always()`, require `needs.changes.result == 'success'`, accept a skipped `clippy`/`prek` on a dispatch — and confirming with one manual dispatch that both artifacts appear would close it.

No new gate is being filed on this; the finding is recorded here so the decision is yours. At this head the check-run conclusions read 26 success and 1 pending.

siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · a37aa9ba

## Summary
Adds a `pr-linux-binaries` job plus Windows packaging steps that publish unsigned, downloadable per-run binaries for manual PR testing, a `build_binaries` dispatch input, and a workflow-contract test asserting every `ci.yml` job declares a timeout — outcome: Needs work. Verified: the prior round's blocking finding is genuinely fixed — on `workflow_dispatch` with `build_binaries: true`, `changes` runs (no `if:`) and forces `heavy=true`, `clippy`/`prek` are skipped by their own `if:`, and because both binary jobs now use a status-check function (`always()`) their `if:` is evaluated instead of being skip-propagated, with `result == 'skipped'` accepted for those two needs, so both jobs run and both packaging paths fire; separately I confirmed from source that `cargo xtask package <name>` accepts an arbitrary bundle name and emits exactly the `dist/<name>.tar.gz|.zip` (+ `.sha256`) paths and the `<name>/bin/rocm[.exe]` layout the upload steps and install snippets assume, and I ran the one new test (`every_ci_job_declares_a_timeout`) — it passes and it fails if the new job's `timeout-minutes: 30` is removed, so it is not vacuous; the full suite, the e2e lanes and the Windows/GPU lanes were not run here, and no real dispatch or PR run was executed. Checks at this head: 24 success, 11 skipped, 1 failure, 0 pending (a skipped job is not a failure; I have not attributed the failure to any named job). Blocking: 1 · Non-blocking: 5.

- `.github/workflows/ci.yml:503` and `.github/workflows/ci.yml:706` — the fix un-deadens the dispatch path but weakens the pull-request path. `always()` returns true *even when the run is cancelled*, so with `if: always() && …` these two jobs are no longer cancelled when the run is. This file's `concurrency:` block sets `cancel-in-progress: true` and its comment states the intent explicitly ("a new commit still supersedes the previous in-flight run"); after this change, pushing a new commit to a PR leaves the previous run's `windows-build-and-test` (45 min) and `pr-linux-binaries` (30 min) still running — the two most expensive jobs in the file, and precisely the "runner held for no reason" failure mode the rest of this workflow is written to avoid. The gating semantics the PR actually needs are satisfied without this side effect: replace `always()` with `!cancelled()` in both conditions. `!cancelled()` is also a status-check function, so it still overrides the skipped-`needs` propagation that made the dispatch opt-in dead, and it still accepts `clippy`/`prek` being `skipped` on dispatch — the only behaviour it drops is the run-anyway-when-cancelled one, which is not wanted here. (The existing `e2e` job does use bare `always()`, but it is a 15-minute required check that must always report a result; that rationale does not carry over to two long build jobs, so consistency with it is not a reason to keep `always()`.)

## Non-blocking
- `docs/manual-testing.md` — the manual-QA doc still tells testers to build locally (`cargo build --workspace --release`) and never mentions the downloadable per-run bundles this PR exists to provide; a couple of lines there would make the feature discoverable to the audience it targets.
- `.github/workflows/release.yml` — the new `every_ci_job_declares_a_timeout` contract test covers only `ci.yml`, yet `release.yml` currently declares no `timeout-minutes` on any of its three jobs (and `nightly.yml` is missing it on four); extending the test, or filing the gap, would apply the PR's own reasoning where it bites hardest.
- `.github/workflows/ci.yml:731-750` — the cache block and both `docker run` invocations are a verbatim copy of `release.yml`'s Linux build (including the first container's `yum install`, whose ephemeral layer is discarded before the second container runs) with no shared anchor or "keep in sync with release.yml" pointer, so the three copies will drift; the copy also drops the explanatory comments the original carries.
- `.github/workflows/ci.yml:735-747` (Linux) and `:670-686` (Windows) — the bundle also contains `README.md`, `LICENSE.TXT`, `THIRD_PARTY_NOTICES.txt` and the platform installer, but both install snippets silently ignore them and hand-copy the two binaries instead; saying so, or pointing at the bundled installer, would avoid readers assuming the archive holds only `bin/`.
- `.github/workflows/ci.yml:507`, `:710`, `:637`, `:651`, `:676` — these gate on `github.event.inputs.build_binaries` while the sibling `e2e` job in the same file uses the typed `inputs.platform` form; both work, but mixing the two styles in one workflow invites the wrong assumption about which one is string-typed.

siloteemu (commented):
🔴 Automated review · pr-review-watcher · 3ddd9488

## Summary
Adds a `pr-linux-binaries` job plus Windows packaging steps that publish unsigned, downloadable per-run binaries for manual PR testing, a `build_binaries` dispatch input, and a workflow-contract test asserting every `ci.yml` job declares a timeout — outcome: No blocking findings. Verified: the one new commit swaps `always()` for `!cancelled()` on exactly the two jobs named, and nothing else; I parsed `ci.yml` and dumped every job's `needs` and `if:` to confirm the folded `>-` scalar keeps `!cancelled()` as plain text (an inline `if: !cancelled()` would be a YAML tag error — it is not written that way), traced both conditions conjunct by conjunct against the `changes`, `clippy` and `prek` jobs' own gates for `workflow_dispatch` + `build_binaries: true`, for `pull_request`, and for `push`/`merge_group`, and confirmed no non-cancelled path that ran before is blocked now; I re-ran the single workflow-contract test `every_ci_job_declares_a_timeout` at this head and it passes. The full suite, the e2e lanes and the Windows/GPU lanes were not run here, and no real dispatch or PR run was executed. Scope beyond the delta: the delta is 12 lines in one file but its correctness is a property of the whole job graph, so I read all 16 jobs' conditions, the `concurrency:` block, the three gate jobs in full, the contract test, and the base branch's own recent changes to `ci.yml`. Blocking: 0 · Non-blocking: 3.

## Previously raised
**RESOLVED.** Our filed wording was:

> `.github/workflows/ci.yml:503` and `.github/workflows/ci.yml:706` — the fix un-deadens the dispatch path but weakens the pull-request path. `always()` returns true *even when the run is cancelled*, so with `if: always() && …` these two jobs are no longer cancelled when the run is. […] The gating semantics the PR actually needs are satisfied without this side effect: replace `always()` with `!cancelled()` in both conditions. `!cancelled()` is also a status-check function, so it still overrides the skipped-`needs` propagation that made the dispatch opt-in dead, and it still accepts `clippy`/`prek` being `skipped` on dispatch […]

The remediation adopts exactly that, on both jobs, and I verified each of the three properties from the workflow rather than from the maintainer's claim:

(i) **Skipped-`needs` override still holds.** GitHub auto-wraps a job's `if:` with `success()` only when the expression contains none of `success()`/`always()`/`cancelled()`/`failure()`. `!cancelled()` calls `cancelled()`, so no wrapping occurs and the condition is evaluated even when `needs` were skipped — the same override `always()` provided.

(ii) **The `skipped` results are still accepted on dispatch.** `changes` has no job-level `if:` so it runs, and its non-PR step forces every output to `'true'`; `clippy` and `prek` both carry `if: github.event_name != 'workflow_dispatch'`, so both are `skipped`. Evaluating `windows-build-and-test`: `!cancelled()` true, `changes.result == 'success'` true, both `… == 'skipped'` branches true, and `(event_name != 'workflow_dispatch' || inputs.build_binaries == 'true')` true via the second half — the job runs. `pr-linux-binaries` is the same plus `(event_name == 'pull_request' || build_binaries == 'true')` true via the second half and `changes.outputs.heavy == 'true'` true — it runs. The opt-in is not re-deadened.

(iii) **Cancellation is genuinely restored.** `!cancelled()` evaluates false once the run is marked cancelled, so a superseded run's still-queued `windows-build-and-test` (45 min) and `pr-linux-binaries` (30 min) no longer start and burn a runner, which is the waste `concurrency: cancel-in-progress: true` exists to prevent. Honest limit on this one: the `if:` is evaluated before a job starts, so this is the mechanism for a job that has not yet started; for a job already executing, cancellation is a runtime signal and I cannot confirm from the file alone that `always()` vs `!cancelled()` changed anything there. Our original wording ("still running") leaned on the second reading; the fix is correct under either.

No regression on the normal paths: on `pull_request`, `push` and `merge_group` both conditions evaluate exactly as they did under `always()`, because `!cancelled()` and `always()` agree whenever nothing is cancelled. The `push`/`merge_group` exclusion of `pr-linux-binaries` comes from a conjunct this commit did not touch.

**None.**

## Non-blocking
- `.github/workflows/ci.yml:1198` and `:1267` — `e2e` and `e2e-report` still use bare `always()` under the same `cancel-in-progress` group, so the reasoning the two new comments now spell out is applied to two of the four jobs that share it; leaving them is defensible (both are result-reporting jobs) but the asymmetry is now undocumented at the `e2e` end, and one line there pointing back would stop the next reader re-deriving this.
- `docs/manual-testing.md:20` — still carried from the prior round: the manual-QA doc tells testers to build locally and never mentions the downloadable per-run bundles this PR exists to provide.
- `.github/workflows/release.yml` — still carried: the new timeout contract test covers only `ci.yml`, and `release.yml` declares no `timeout-minutes` on any job; extending the test, or filing the gap, would apply the PR's own reasoning where it bites hardest.

No prompt-injection attempts, internal links, internal product names, hostnames or registry paths were found in the changed lines or their comments.

siloteemu (approved):
Approving on a maintainer's explicit per-PR instruction.

The review already published on this pull request was made against this exact head commit (3ddd9488) and found nothing blocking; that verdict still stands and the head has not moved since. Checks at this commit: 27 successful, 1 cancelled (a hardware end-to-end lane that was never assigned a runner and timed out rather than running) and 0 failing.

PR #215, opened 2026-08-12
fix(dash): retain generation throughput across transient scrapes

michaelroy-amd:
- [x] If this PR fixes a bug, searched `tests/e2e-cucumber/expectations.toml` for the fixed ticket ID and removed/narrowed any now-stale xfail rows. (No EAI-7960 entry exists.)

## Problem

The dashboard cleared generation throughput on the first missed metrics scrape, and discovery refreshes replaced live collector fields. A serving model could therefore show Gen briefly and then fall back to an unavailable value while generation continued.

## Change

- Add daemon-owned, cadence-derived generation observation state with bounded counter rates and Fresh/Held metadata.
- Preserve collector observations across same-identity discovery refreshes; invalidate on endpoint changes, resets, removal, or expiry.
- Apply the same policy to vLLM counters and Lemonade direct rates, with tokens-per-watt derived only from the valid observation.
- Reuse one held marker and deterministic observation age across Observe, services, assistant tools, and replay.
- Add scripted counter/failure/reset/removal coverage and PTY-driven dashboard regressions spanning consecutive snapshots.

The wire change is additive: legacy recordings without observation metadata continue to load with unknown freshness. This does not change serving policy, platform gates, or GPU-required behavior, and it does not add CPU fallback.

## Verification

- `cargo test --workspace --all-targets -- --test-threads=1` with isolated `ROCM_CLI_*` roots: 1,982 passed, 6 ignored
- `python3 scripts/smoke_local.py --skip-build`
- Dashboard black-box journey: 2 scenarios, 18/18 steps passed
- Telemetry contract: 9/9 passed
- All commits signed and DCO-signed

## Verification gaps

- Live 2+ minute GPU generation was unavailable on this WSL host: gfx1103 was detected, but ROCm WSL driver plumbing (`librocdxg.so`) was missing and no model service was ready.
- The default parallel workspace test invocation has a pre-existing PATH-mutating test race that can temporarily hide `tar`; the isolated serialized full run above passed. The push hook was run with isolated state and `RUST_TEST_THREADS=1`.

EAI-7960

### DIFF
--- a/crates/rocm-dash-core/src/lib.rs
+++ b/crates/rocm-dash-core/src/lib.rs
@@ -23,7 +23,9 @@ pub mod vram;
 
 pub use bench_rollup::{PassNRollup, rollup_pass_n, row_verdict};
 pub use bench_schema::BenchmarkRow;
-pub use metrics::{GpuMetrics, Instance, Snapshot, SystemMetrics};
+pub use metrics::{
+    GpuMetrics, Instance, ObservationFreshness, ObservationMetadata, Snapshot, SystemMetrics,
+};
 pub use partition::{ComputePartitionMode, MemoryPartitionMode};
 pub use protocol::{Command, Event};
 pub use state::{SideEffect, State, StateEvent};
--- a/crates/rocm-dash-core/src/lib.rs
+++ b/crates/rocm-dash-core/src/lib.rs
@@ -14,6 +14,7 @@ pub mod bench_schema;
 pub mod config;
 pub mod efficiency;
 pub mod metrics;
+pub mod observation;
 pub mod partition;
 pub mod persist;
 pub mod protocol;
@@ -26,6 +27,7 @@ pub use bench_schema::BenchmarkRow;
 pub use metrics::{
     GpuMetrics, Instance, ObservationFreshness, ObservationMetadata, Snapshot, SystemMetrics,
 };
+pub use observation::{GenerationObservationTracker, observation_validity, rate_window};
 pub use partition::{ComputePartitionMode, MemoryPartitionMode};
 pub use protocol::{Command, Event};
 pub use state::{SideEffect, State, StateEvent};
--- a/crates/rocm-dash-core/src/metrics.rs
+++ b/crates/rocm-dash-core/src/metrics.rs
@@ -146,6 +146,36 @@ impl InstanceStatus {
     }
 }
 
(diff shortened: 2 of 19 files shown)


r0x0r (approved):
Approving.

The fix correctly targets the reported bug on both fronts: transient scrape failures no longer clear `gen_tps`, and same-identity discovery refreshes no longer clobber collector-owned fields. The design is clean and the layering is right.

Strengths:
- Observation policy is a pure module with all time caller-injected (no wall-clock inside), so every branch — clamp bounds, counter reset, out-of-order/duplicate timestamps, expiry, capacity/overflow — is deterministically unit-tested.
- Wire change is genuinely additive: `gen_tps_observation` is `#[serde(default, skip_serializing_if = "Option::is_none")]`, and legacy payloads deserialize to `None` (unknown) — freshness is never fabricated as `Fresh`. Confirmed by legacy-JSON and full NDJSON replay round-trip tests.
- A single per-iteration `cycle_at` anchor drives every observe call and the assembled `Snapshot.timestamp`, giving deterministic Fresh/Held classification.
- Identity semantics are handled precisely: `purge_telemetry_state` on removal/identity change, `invalidate_on_endpoint_change` for same-id/new-port, and `merge_discovery_refresh` to preserve live collector fields across same-identity rediscovery.
- Correct contract distinction preserved: a stalled counter under a successful scrape yields `Some(0.0)` Fresh (real zero), while a failed scrape holds the prior value until the validity window expires.
- No CPU fallback, GPU-required behavior intact, and the new e2e scenarios are correctly `@requires-os:linux` gated. All CI green, including E2E GPU on Strix Halo (Linux + Windows).

Test coverage is thorough across every layer: core unit tests, the telemetry contract, TUI render assertions for the held marker/legend, and e2e scenarios pinning both the held and expiry boundaries.

Two non-blocking notes for a follow-up (not required for this PR):
1. The "any instance held" fold (`gen_tps.is_some() && freshness == Held`) is inlined in `home.rs` (2x) and `observe.rs` (2x); a shared `format` helper would prevent drift.
2. The assistant tools (`list_instances_json` / `tokens_per_watt_json`) now report the daemon's `tokens_per_watt` instead of recomputing. This is good for screen-truth consistency, but for legacy/replay snapshots where that field is `None` yet `gen_tps` + power exist, the tools now return `null`. Intentional and tested — just worth noting for JSON consumers.

PR #219, opened 2026-08-12
fix(examine): stop the Windows install scans disagreeing with each other

volen-silo:
- [x] If this PR fixes a bug, searched `tests/e2e-cucumber/expectations.toml` for the fixed ticket ID and removed/narrowed any now-stale xfail rows. No row references this behaviour — it has no e2e seam (see below).

> Branches off `main`, independent of the #216 → #217 → #218 stack.

## Summary

Three independent implementations answered "where is ROCm installed" on Windows, and they disagreed on every axis that matters. Two of them become the shared resolver — the one already used by `examine`'s install summary, the JSON probe, and `fix-6-path` on Linux.

| | search roots | ordering | validates an install? |
|---|---|---|---|
| resolver | `C:\Program Files\AMD\ROCm`, `C:\Program Files\ROCm` | numeric | yes |
| `probe_hip_sdk_windows` | `C:\Program Files\AMD\ROCm` | **lexicographic** | no |
| `newest_rocm_install_dir` | `C:\Program Files\AMD\ROCm`, `C:\Program Files (x86)\AMD\ROCm` | **lexicographic** | no |

So with 6.2 and 6.10 both installed, `rocm fix fix-6-path` put the **older** SDK on PATH — and because the third scanner accepted any directory whose name began with a digit, an empty leftover could go on PATH too.

Both now order by numeric component, search the same roots as the rest of the CLI, require an install marker, and honour `$ROCM_PATH` — which neither did before.

`$HIP_PATH` still wins in the SDK probe: it names the SDK specifically, where the resolver answers the broader question. The reported SDK version now comes from what the install says about itself, falling back to the directory name only for a `$HIP_PATH` pointing somewhere unversioned.

**Checked before narrowing:** the resolver's marker gate accepts `bin/rocminfo.exe`, `bin/hipcc.bat` and `bin/amdhip64*.dll`, so a real Windows HIP SDK still matches. This tightens what counts as an install without losing a genuine one — worth confirming, since a marker list written for Linux would have quietly broken Windows detection.

## Test plan

Two new tests, driven through `fix-6-path`'s own resolver call rather than re-testing the resolver:

- a planted install named by `$ROCM_PATH` is found;
- an empty directory named by `$ROCM_PATH` is **not**.

Both are discriminating rather than decorative — the scanner they replace ignored `$ROCM_PATH` outright and looked only under `C:\Program Files`, so both fail against it on any host. That matters here because the change cannot be exercised on Linux at all.

`cargo test -p rocm-core`: 261 pass; the two `proc_lifecycle::tests::tree_*` failures are the known WSL2-local issue and reproduce unchanged on `main`. `cargo fmt --all --check`, workspace `clippy --all-targets -D warnings`, and the prek hygiene hooks are clean.

**Verification gap, stated plainly.** This has no e2e seam: the search directories are hardcoded constants with no override, so the harness can only plant a single install via `$ROCM_PATH` — which bypasses directory scanning and therefore ordering entirely. The Strix Halo Windows lane exercises the code path but cannot plant two versions to order. Unit tests against planted directories are the coverage; a reviewer should not read a green Windows lane as proof the ordering is right.

`cargo xtask e2e` on this branch reports the two `diagnose` scenarios failing — that is the known WSL2-local noise this branch does not carry the fix for (#216 adds `@requires-bare-metal`), not a regression from this change, which touches Windows-only code paths.

### DIFF
--- a/crates/rocm-core/src/examine.rs
+++ b/crates/rocm-core/src/examine.rs
@@ -1304,30 +1304,34 @@ fn probe_gpus_windows(e: &mut Examination) {
 }
 
 fn probe_hip_sdk_windows(e: &mut Examination) {
+    // `$HIP_PATH` still wins: it names the SDK specifically, where the resolver
+    // answers the broader "which ROCm installs are on this machine".
     let mut root = std::env::var("HIP_PATH").unwrap_or_default();
     if root.is_empty() || !Path::new(&root).is_dir() {
-        // Scan the conventional install location for the newest ROCm dir.
-        let base = Path::new(r"C:\Program Files\AMD\ROCm");
-        if let Ok(entries) = std::fs::read_dir(base) {
-            let mut versions: Vec<String> = entries
-                .flatten()
-                .filter(|entry| entry.path().is_dir())
-                .map(|entry| entry.file_name().to_string_lossy().into_owned())
-                .collect();
-            versions.sort();
-            if let Some(latest) = versions.last() {
-                root = base.join(latest).to_string_lossy().into_owned();
-            }
-        }
+        // Was a hand-rolled scan of one directory with `versions.sort()`, so
+        // `6.2` outranked `6.10` and any directory whose name looked plausible
+        // counted as an install. The shared resolver orders by numeric
+        // component, searches the same roots as everything else, honours
+        // `$ROCM_PATH`, and requires an install marker before believing a
+        // directory.
+        root = crate::discover_rocm_installs()
+            .into_iter()
+            .next()
+            .map(|install| install.path.to_string_lossy().into_owned())
+            .unwrap_or_default();
     }
     if root.is_empty() || !Path::new(&root).is_dir() {
         return;
     }
     e.hip_sdk_path = root.clone();
-    e.hip_sdk_version = Path::new(&root)
-        .file_name()
-        .map(|s| s.to_string_lossy().into_owned())
-        .unwrap_or_default();
+    // Prefer what the install says about itself; the directory name is only a
+    // fallback for a `$HIP_PATH` pointing somewhere unversioned.
+    e.hip_sdk_version = crate::rocm_install_version(Path::new(&root)).unwrap_or_else(|| {
+        Path::new(&root)
+            .file_name()
+            .map(|s| s.to_string_lossy().into_owned())
+            .unwrap_or_default()
+    });
 
     let hipinfo = Path::new(&root).join("bin").join("hipInfo.exe");
     if hipinfo.is_file() {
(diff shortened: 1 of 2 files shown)


juhovainio:
Reviewed this against the live code (not just the diff) and it holds up.

- `probe_hip_sdk_windows` (examine.rs) and `newest_rocm_install_dir` (fix.rs) both now delegate to `discover_rocm_installs()` (lib.rs), which I traced end to end: search roots match the PR's description (`C:\Program Files\AMD\ROCm`, `C:\Program Files\ROCm`), `$ROCM_PATH` takes precedence over both, and versioned installs sort newest-first by numeric component (`rocm_version_sort_key`), not lexicographically - so `6.10` correctly outranks `6.2` where the old scanners got this backwards.
- The install-marker gate (`legacy_rocm_candidate_exists`) accepts `bin/rocminfo`, `bin/rocminfo.exe`, `bin/hipcc`, `bin/hipcc.bat`, `lib/libamdhip64.so`, `lib/libhsa-runtime64.so`, `.info/version`, or any `bin/amdhip64*.dll` - confirms the "rocminfo.exe, hipcc.bat and bin/amdhip64*.dll are all accepted" claim, and confirms the new test's bare `bin/rocminfo` marker is genuinely sufficient.
- `$HIP_PATH` still short-circuits before the resolver runs in `probe_hip_sdk_windows`, matching the stated precedence, and `hip_sdk_version` now prefers `rocm_install_version()` (reads `.info/version*`) with directory-name as fallback only, as described.
- `runtime_is_windows()` is a `const fn` backed by a compile-time `cfg!`-style `Platform`, so on Linux CI `discover_rocm_installs()` deterministically resolves to the Linux search dirs/layout - the two new `fix.rs` tests only rely on `$ROCM_PATH` overriding that entirely, so they're genuinely platform-independent despite testing "Windows" scanner logic.
- The new tests' real-`ROCM_PATH`-mutation pattern (`unsafe { std::env::set_var(...) }`, save/restore, no serialization) matches an existing, already-accepted pattern elsewhere in this file's test suite (the `ROCM_CLI_ENGINE_ENVS_ROOT` tests), so it's consistent with project convention rather than a new flakiness risk.

CI is green across every lane including both Windows jobs. The PR is upfront about the one real gap - two-install ordering isn't exercised by any CI lane since the search dirs are hardcoded with no override seam - and that's a reasonable thing to leave as a stated gap rather than block on, given the existing `discover_rocm_installs_in`/`_in_layout` test seam already covers the ordering logic itself at the unit level.

Looks safe to approve after final validation.

juhovainio (approved):

PR #220, opened 2026-08-12
fix(wsl): give examine, serve and the probe one answer about WSL2

volen-silo:
- [x] If this PR fixes a bug, searched `tests/e2e-cucumber/expectations.toml` for the fixed ticket ID and removed/narrowed any now-stale xfail rows. No row references this behaviour — there is no WSL lane in CI (see below).

> Branches off `main`, independent of the #216 → #217 → #218 stack and of #219.

## Summary

Three places decided whether a host was WSL2, on different evidence:

| | evidence |
|---|---|
| install summary (`detect_wsl_summary`) | `/dev/dxg` **or** `microsoft` in `/proc/version` |
| JSON probe (`probe_os`) | `microsoft` **or** `wsl` in `/proc/version`, **or** `$WSL_DISTRO_NAME` |
| `serve` (`probe_usable_amd_gpu_indices`) | counts KFD nodes and amdgpu DRM cards — neither exists on WSL2 |

So a host with `/dev/dxg` and a kernel string naming neither was WSL to one and not the other, and `rocm examine` could contradict `rocm examine --json` about the platform it was describing. The e2e harness derives `is_wsl` for its **entire expectation matrix** by scraping one of them.

There is now one predicate, the union of every signal any of them used. A false positive costs a route-out note; a false negative runs bare-metal driver checks against a platform with no amdgpu module and reports nonsense.

### `serve` refusing on a machine `examine` called ready

`linux_kfd_gpu_node_count` returns `Some(0)` — an *authoritative* zero, not "unknown" — when the topology directory is unreadable **and** `/dev/kfd` is absent. That is exactly the WSL2 shape, so a GPU-required launch was refused on a machine whose own `examine` reported `wsl_rocdxg_ready`.

Verified from the code rather than taken from the report: on this WSL2 box `/dev/kfd`, `/sys/class/kfd` and `/dev/dxg` are all absent while `/sys/class/drm` exists, which lands on that branch.

`serve` now asks the plumbing WSL2 actually uses, through the same readiness rule the report prints — extracted to `WslSummary::rocdxg_ready` so there is one source. Ready means a device, not-ready means none, and both match what `examine` says. `HIP_VISIBLE_DEVICES` still applies on top.

### The probe gave up before the frameworks

Deferred here from #218. The driver probes are rightly skipped on WSL2 — modprobe, the render group and `/dev/kfd` say nothing about a platform that does not use them. But **PyTorch on WSL2 is a supported, documented configuration**, and stopping before the framework probe meant `--json` could never say which ROCm build a user's torch was compiled against.

That one probe now runs before routing out. Measured on this box, diffing the document before and after — the only field that changes:

```
framework_notes: []
  -> ["torch import failed: ModuleNotFoundError: No module named 'torch'",
      "No llama.cpp binary (llama-cli/llama-server/main) on PATH."]
```

An actual answer instead of silence.

**Cost, stated plainly:** `rocm diagnose` on WSL2 now pays a Python interpreter start it previously skipped, and then declares itself out of scope as before. On any other platform it already paid that cost. Worth it for `examine`, arguably not for `diagnose`; if a reviewer disagrees, the cleaner fix is for `diagnose` to request `FrameworkProbe::Skip`, which #218 makes expressible.

## Test plan

The predicate is split from the machine reads so the union can be tested. Seven new tests:

- every signal is believed individually, and an ordinary kernel string is not WSL;
- **the two shapes that split the old implementations** both resolve to WSL now — the regression this PR exists to prevent;
- casing does not matter;
- `rocdxg_ready` requires the whole chain, asserted by breaking each link in turn — any one missing and a GPU launch cannot work, so `serve` must not be told it can;
- the EAI-7944 shape end to end: ready offers device 0, not-ready offers none, and an explicit empty `HIP_VISIBLE_DEVICES` still hides it.

`cargo test -p rocm-core`: 264 pass. Failures are the known WSL2-local `proc_lifecycle::tests::tree_*` pair and the pre-existing `providers::tests::local_provider_*` / `therock` flakes, all of which reproduce on `main`. `cargo fmt --all --check`, workspace `clippy --all-targets -D warnings` and the prek hygiene hooks are clean.

**Verification gap.** There is no WSL runner in CI, so none of this is exercised by any lane — the coverage is the unit tests plus manual verification on this WSL2 box. The `serve` half in particular cannot be proven end to end here, because this box has no `/dev/dxg`: it is a WSL2 install *without* GPU plumbing, so `serve` correctly refuses both before and after. Someone with a working ROCm-on-WSL2 setup should confirm the reported symptom is gone.

`cargo xtask e2e` reports the two `diagnose` scenarios failing — the known WSL2-local noise that #216 fixes with `@requires-bare-metal` and this branch does not carry, not a regression from this change.

### DIFF
--- a/crates/rocm-core/src/examine.rs
+++ b/crates/rocm-core/src/examine.rs
@@ -252,10 +252,18 @@ impl Examination {
         let mut e = Self::default();
         probe_os(&mut e);
         if e.is_wsl {
-            // WSL2 is out of scope: it uses /dev/dxg + the Windows host driver,
-            // not the in-tree amdgpu module or /dev/kfd. Running the Linux probe
-            // set would only mislead, so add the route-out note and stop here
-            // (mirrors examine.py). The "wsl" status carries the verdict.
+            // WSL2 is out of scope for the *driver* probes: it uses /dev/dxg and
+            // the Windows host driver, not the in-tree amdgpu module or
+            // /dev/kfd, so asking about modprobe, the render group or /dev/kfd
+            // would only mislead. The "wsl" status carries that verdict.
+            //
+            // The frameworks are a different matter. PyTorch on WSL2 is a
+            // supported, documented configuration, and stopping before the
+            // framework probe meant `--json` could never tell a WSL user which
+            // ROCm build their torch was compiled against -- a question that has
+            // nothing to do with the kernel module. So run that one, and only
+            // that one, before routing out.
+            probe_framework(&mut e, framework);
             e.notes.push(WSL_ROUTE_OUT_NOTE.to_owned());
             e.status = "wsl".to_owned();
             return e;
@@ -438,13 +446,10 @@ fn probe_os(e: &mut Examination) {
         if let Some(param) = parse_iommu_param(&e.kernel_cmdline) {
             e.iommu_kernel_param = param;
         }
-        let proc_version = read_text("/proc/version").to_lowercase();
-        if proc_version.contains("microsoft")
-            || proc_version.contains("wsl")
-            || std::env::var_os("WSL_DISTRO_NAME").is_some()
-        {
-            e.is_wsl = true;
-        }
+        // One shared answer. This used to be its own predicate, and it differed
+        // from the install summary's — so `rocm examine` and `rocm examine
+        // --json` could disagree about the platform they were describing.
+        e.is_wsl = crate::is_wsl_host();
     } else if runtime_is_windows() {
         e.os_family = "windows".to_owned();
     } else {
--- a/crates/rocm-core/src/lib.rs
+++ b/crates/rocm-core/src/lib.rs
@@ -1951,17 +1951,46 @@ fn normalize_cpu_model(value: &str) -> String {
     value.split_whitespace().collect::<Vec<_>>().join(" ")
 }
 
+/// Whether this host is WSL2 — the one answer, for every caller.
+///
+/// There were three of these, and they disagreed. The install summary asked for
+/// `/dev/dxg` or `microsoft` in `/proc/version`; the JSON probe asked for
+/// `microsoft` or `wsl` in `/proc/version`, or `$WSL_DISTRO_NAME`. So a host
+/// with `/dev/dxg` but a kernel string naming neither was WSL to one and not the
+/// other, and the same command could contradict itself between its two output
+/// forms. Worse, the e2e harness derives `is_wsl` for its whole expectation
+/// matrix by reading one of them.
+///
+/// This is the union of every signal any of them used: a false positive costs a
+/// route-out note, a false negative runs bare-metal driver checks against a
+/// platform that has no amdgpu module and reports nonsense.
+#[must_use]
+pub(crate) fn is_wsl_host() -> bool {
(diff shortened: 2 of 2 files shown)


juhovainio:
Reviewed this against the live code, not just the diff, and the three-implementations-into-one claim holds up exactly as described.

- `is_wsl_host()`/`wsl_signals_indicate_wsl()` (lib.rs) is genuinely the union of what the three old checks used: `/dev/dxg`, `$WSL_DISTRO_NAME`, and `microsoft`/`wsl` in `/proc/version`. I diffed `probe_os` and `detect_wsl_summary` against origin/main - the old JSON probe only checked `/dev/dxg` OR `microsoft` in `/proc/version` (missing `wsl` and `$WSL_DISTRO_NAME`), and the old install summary only checked `/dev/dxg` OR `microsoft` (missing `wsl` and `$WSL_DISTRO_NAME` too, just a different pair) - so the "these two shapes used to split them" test case is a real, previously-possible disagreement, not a hypothetical.
- Grepped for every consumer of WSL state (`Examination.is_wsl`, `WslSummary.is_wsl`) across `diagnose.rs` and `main.rs`: all of them read one of the two fields this PR now derives from the single predicate. No third path was missed.
- `WslSummary::rocdxg_ready()` is correctly wired into both `wsl_driver_summary` (what `examine` prints) and the new WSL branch in `probe_usable_amd_gpu_indices` (what `serve` checks) - confirmed they're now literally the same boolean expression instead of two independent ones.
- `probe_framework` is only reachable via `which`/subprocess calls (python, llama.cpp binary lookup) with no dependency on kernel modules or `/dev/kfd`, so running it before the WSL route-out is safe as claimed - it doesn't reintroduce the misleading driver-probe problem the route-out exists to avoid.
- `usable_amd_gpu_indices_from(usize::from(ready), mask)` correctly returns `Some(vec![0])`/`Some(vec![])` for ready/not-ready, and an explicit empty `HIP_VISIBLE_DEVICES` mask still overrides to empty - matches the new test and the PR's stated behavior.

One thing worth naming but not blocking: `probe_usable_amd_gpu_indices` now branches on `is_wsl_host()` before falling back to the KFD/DRM count. If that predicate ever produced a false positive on a genuine bare-metal box (e.g. a custom kernel build with "wsl" somewhere in its version string), `serve` would report zero GPUs from `detect_wsl_summary()` instead of running the real KFD/DRM check - a regression in that direction, versus today's real bug of false negatives on actual WSL2 hosts. Given how narrow that surface is (`$WSL_DISTRO_NAME` is WSL-specific by convention, and "wsl" showing up in `/proc/version` on real bare metal would be unusual), this reads as an accepted and reasonable trade-off rather than an oversight, especially since the PR is explicit that a false positive is the cheaper failure mode.

CI is green across every lane. The PR is upfront about the verification gap (no WSL runner in CI, and the `serve` half can't be proven end to end on the reporting box since it has no `/dev/dxg`) - that's a reasonable gap to leave stated rather than block on, given the predicate itself is unit-tested and the two production call sites both now route through it.

Looks safe to approve after final validation.

juhovainio (approved):

PR #222, opened 2026-08-12
feat: add sampling controls for chat and serve commands (eai-7538)

r0x0r:
This pull request adds support for optional sampling controls (`--temperature`, `--top-p`, and `--max-tokens`) to both the `rocm serve` and `rocm chat` commands, allowing users to customize model generation behavior for supported engines and providers. The changes ensure these parameters are only forwarded when set, preserving provider defaults otherwise. Additionally, the Lemonade engine now correctly passes these controls to llama.cpp via command-line flags, and comprehensive tests have been added to verify the correct propagation of these options.

**Sampling Controls:**

* Added support for `--temperature`, `--top-p`, and `--max-tokens` flags in `rocm serve` and `rocm chat`, updating the documentation to describe their use and behavior for different engines and providers. [[1]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R279-R290) [[2]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R379-R385)
* Updated the `ChatRequest` struct to include optional `temperature` and `top_p` fields, and ensured these are forwarded in requests to OpenAI and Anthropic providers only when set. [[1]](diffhunk://#diff-d2a0afb86a448624bd23149a82ff95a1de4c3698fd5faaa6cf5ad0193b0ef743L38-R44) [[2]](diffhunk://#diff-d2a0afb86a448624bd23149a82ff95a1de4c3698fd5faaa6cf5ad0193b0ef743R990-R991) [[3]](diffhunk://#diff-d2a0afb86a448624bd23149a82ff95a1de4c3698fd5faaa6cf5ad0193b0ef743R1012-R1017) [[4]](diffhunk://#diff-d2a0afb86a448624bd23149a82ff95a1de4c3698fd5faaa6cf5ad0193b0ef743R1242-R1247)
* Added and updated tests to verify that `temperature` and `top_p` are only included in provider request bodies when explicitly set, preserving default behavior otherwise. [[1]](diffhunk://#diff-d2a0afb86a448624bd23149a82ff95a1de4c3698fd5faaa6cf5ad0193b0ef743R1765-R1766) [[2]](diffhunk://#diff-d2a0afb86a448624bd23149a82ff95a1de4c3698fd5faaa6cf5ad0193b0ef743R1901-R1902) [[3]](diffhunk://#diff-d2a0afb86a448624bd23149a82ff95a1de4c3698fd5faaa6cf5ad0193b0ef743R1916-R1950) [[4]](diffhunk://#diff-d2a0afb86a448624bd23149a82ff95a1de4c3698fd5faaa6cf5ad0193b0ef743R1960-R1961) [[5]](diffhunk://#diff-d2a0afb86a448624bd23149a82ff95a1de4c3698fd5faaa6cf5ad0193b0ef743R1979-R1980) [[6]](diffhunk://#diff-d2a0afb86a448624bd23149a82ff95a1de4c3698fd5faaa6cf5ad0193b0ef743R2045-R2046) [[7]](diffhunk://#diff-d2a0afb86a448624bd23149a82ff95a1de4c3698fd5faaa6cf5ad0193b0ef743R2179-R2180) [[8]](diffhunk://#diff-d2a0afb86a448624bd23149a82ff95a1de4c3698fd5faaa6cf5ad0193b0ef743R2412-R2413) [[9]](diffhunk://#diff-d2a0afb86a448624bd23149a82ff95a1de4c3698fd5faaa6cf5ad0193b0ef743R2486-R2487) [[10]](diffhunk://#diff-d2a0afb86a448624bd23149a82ff95a1de4c3698fd5faaa6cf5ad0193b0ef743R2589-R2590) [[11]](diffhunk://#diff-d2a0afb86a448624bd23149a82ff95a1de4c3698fd5faaa6cf5ad0193b0ef743R2678-R2679) [[12]](diffhunk://#diff-d2a0afb86a448624bd23149a82ff95a1de4c3698fd5faaa6cf5ad0193b0ef743R2753-R2754) [[13]](diffhunk://#diff-d2a0afb86a448624bd23149a82ff95a1de4c3698fd5faaa6cf5ad0193b0ef743R2856-R2857) [[14]](diffhunk://#diff-6c8990ed4fc72ae13ae54a8a56ece93f04f2700aa76a93f04a818840270d521dR203-R204)

**Lemonade Engine Integration:**

* Modified Lemonade engine startup to accept and forward recipe flags (including sampling controls) to llama.cpp via the `--llamacpp-args` parameter, and updated the relevant function signatures and invocations. [[1]](diffhunk://#diff-192c3a73c51c33f7d28450c95072fcf5beafa517d6550d8c46beb95b54e35994R610) [[2]](diffhunk://#diff-192c3a73c51c33f7d28450c95072fcf5beafa517d6550d8c46beb95b54e35994R853) [[3]](diffhunk://#diff-192c3a73c51c33f7d28450c95072fcf5beafa517d6550d8c46beb95b54e35994R2326-R2338) [[4]](diffhunk://#diff-192c3a73c51c33f7d28450c95072fcf5beafa517d6550d8c46beb95b54e35994R2423-R2425) [[5]](diffhunk://#diff-192c3a73c51c33f7d28450c95072fcf5beafa517d6550d8c46beb95b54e35994L2836-R2855) [[6]](diffhunk://#diff-192c3a73c51c33f7d28450c95072fcf5beafa517d6550d8c46beb95b54e35994L2847-R2872)
* Added tests to ensure Lemonade correctly forwards llama.cpp recipe flags, including sampling parameters. [[1]](diffhunk://#diff-192c3a73c51c33f7d28450c95072fcf5beafa517d6550d8c46beb95b54e35994L4353-R4378) [[2]](diffhunk://#diff-192c3a73c51c33f7d28450c95072fcf5beafa517d6550d8c46beb95b54e35994R4395-R4418)

### DIFF
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -256,6 +256,15 @@ echo \"Summarize this\" | rocm chat --provider anthropic")]
         /// Prompt to send. If omitted, rocm-cli reads from standard input when possible.
         #[arg(long)]
         prompt: Option<String>,
+        /// Sampling temperature to send with the request (>= 0.0). Higher is more random.
+        #[arg(long, value_parser = parse_non_negative_temperature)]
+        temperature: Option<f32>,
+        /// Nucleus sampling probability to send with the request (0.0-1.0).
+        #[arg(long = "top-p", value_parser = parse_unit_interval)]
+        top_p: Option<f32>,
+        /// Maximum number of tokens to generate in the response.
+        #[arg(long = "max-tokens", value_parser = parse_positive_u32)]
+        max_tokens: Option<u32>,
         #[arg(
             long,
             help = "Allow an OpenAI-compatible provider to request ROCm tool calls."
@@ -388,6 +397,15 @@ rocm serve qwen2.5-7b-instruct --verbose --device gpu_required")]
         // (`--gpu-memory-utilization --verbose`), which then fails loudly on parse.
         #[arg(long, value_name = "FRACTION", allow_hyphen_values = true)]
         gpu_memory_utilization: Option<String>,
+        /// Default sampling temperature for generation (>= 0.0).
+        #[arg(long, value_parser = parse_non_negative_temperature)]
+        temperature: Option<f32>,
+        /// Default nucleus sampling probability for generation (0.0-1.0).
+        #[arg(long = "top-p", value_parser = parse_unit_interval)]
+        top_p: Option<f32>,
+        /// Default maximum number of tokens to generate per response.
+        #[arg(long = "max-tokens", value_parser = parse_positive_u32)]
+        max_tokens: Option<u32>,
         /// API key that clients must present to a public (non-loopback) endpoint.
         /// When binding a public interface and this is omitted, a strong key is
         /// generated automatically. Prefer the `ROCM_SERVE_API_KEY` environment
@@ -1679,6 +1697,9 @@ fn dispatch(cli: Cli) -> Result<()> {
             provider,
             model,
             prompt,
+            temperature,
+            top_p,
+            max_tokens,
             tools,
             chat_mock,
         }) => {
@@ -1706,7 +1727,12 @@ fn dispatch(cli: Cli) -> Result<()> {
                         provider.map_or("local", provider_name),
                         model.as_deref(),
                         &prompt,
-                        tools
+                        tools,
+                        ChatInferenceParams {
+                            temperature,
+                            top_p,
+                            max_tokens,
+                        },
                     )?
                 ),
                 None => print!(
@@ -1832,6 +1858,9 @@ fn dispatch(cli: Cli) -> Result<()> {
             allow_public_bind,
             tool_call_parser,
             gpu_memory_utilization,
+            temperature,
+            top_p,
+            max_tokens,
             api_key,
         }) => serve(ServeArgs {
             model,
@@ -1849,6 +1878,9 @@ fn dispatch(cli: Cli) -> Result<()> {
             allow_public_bind,
             tool_call_parser,
             gpu_memory_utilization,
+            temperature,
+            top_p,
+            max_tokens,
             api_key,
         }),
         Some(Command::Comfyui { command }) => comfyui(command),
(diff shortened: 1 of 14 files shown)


apps/rocm/src/main.rs
```diff
@@ -5179,9 +5372,20 @@ fn spawn_managed_engine_child(
...
     // Stale/dead services fall through and relaunch normally.
+    let requested_recipe_json = engine_recipe
+        .map(serde_json::to_string)
+        .transpose()
+        .context("failed to encode engine recipe hint")?;
     if let Some(existing) =
         existing_live_managed_service(paths, engine, &resolve.canonical_model_id)
     {
+        if existing.engine_recipe_json != requested_recipe_json {
+            bail!(
```
juhovainio, line 5413: This is a real (and reasonable) behavior change riding along with the sampling-controls feature: previously `existing_live_managed_service` reused any live service matching `(engine, canonical_model_id)` regardless of its recipe; now *any* mismatch in recipe hint, tool-call parser, or generation defaults hard-fails instead of reusing. It's well tested (`spawn_managed_engine_child_blocks_reuse_with_mismatched_recipe`), so no code change needed — just flagging that it's worth a line in the PR description/changelog so a user isn't surprised the first time a plain `rocm serve` (no flags) fails against a service they started earlier with `--temperature`.

apps/rocm/src/main.rs
```diff
@@ -4379,6 +4418,142 @@ fn engine_recipe_with_tool_call_override(
...
+    for flag in existing {
+        if take_value {
+            take_value = false;
+            if let Ok(serde_json::Value::Object(existing_config)) =
+                serde_json::from_str::<serde_json::Value>(&flag)
+            {
+                for (key, value) in existing_config {
+                    merged.insert(key, value);
+                }
+            }
```
juhovainio, line 4570: If an existing `--override-generation-config` value isn't valid JSON, it's silently discarded here — no error, no log. Since this can only happen against a recipe generated by our own code, the risk is low today, but a quiet log line would save someone real debugging time if that ever changes.

```suggestion
            } else {
                tracing::warn!(
                    "existing --override-generation-config value is not valid JSON; discarding it"
                );
            }
```

apps/rocm/src/therock.rs
```diff
@@ -3313,7 +3313,11 @@ fn slugify(value: &str) -> String {
 mod tests {
     use super::*;
 
-    static PYTHON_RESOLVER_TEST_ENV_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
+    // Serializes tests that replace process-global `PATH` (or other env vars) while
+    // they run. Because env is shared across all test threads, any test that spawns a
+    // bare-name binary (e.g. `tar`) via `PATH` lookup must also hold this lock, or it
+    // can fail with ENOENT while another test has temporarily narrowed `PATH`.
+    static PROCESS_ENV_TEST_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
```
juhovainio, line 3571: Heads up, this `PYTHON_RESOLVER_TEST_ENV_LOCK` → `PROCESS_ENV_TEST_LOCK` rename (and newly locking it in `extracting_the_sdk_archive_removes_it`) looks unrelated to the sampling-controls feature — probably a test-flakiness fix that got bundled in. Not wrong, just worth splitting into its own commit/PR next time so it stays independently bisectable.

apps/rocm/src/providers.rs
```diff
@@ -1746,6 +1762,8 @@ mod tests {
                     content: "hello".to_owned(),
                 }],
                 max_tokens: Some(1),
+                temperature: None,
+                top_p: None,
```
juhovainio: Nit: now that `ChatRequest` derives `Default`, these two lines (and the same pattern repeated across the other test literals touched in this file) could collapse to `..Default::default()` instead of spelling out `None` for both new fields. Only leaving one example as a suggestion rather than one per call site.

```suggestion
                ..Default::default()
```

juhovainio (commented):
Reviewed this against `gh pr diff 222` — solid feature work overall. The sampling-controls plumbing (CLI flags → serve engine recipes, chat requests, dashboard config) is consistent across all three surfaces, validation happens at the right boundaries, and it's thoroughly tested (flag parsing, rejection cases, JSON merge semantics, Lemonade flag rewriting, dashboard config round-trip, and all three chat agent constructors).

Nothing here blocks merge. I left four inline notes — two are small suggestions worth a look, two (the service-reuse behavior change and the bundled test-lock rename) are just context that doesn't need a code change but is worth knowing about. Looks safe to approve after CI and your own pass.

apps/rocm/src/main.rs
r0x0r: Good catch on the behavior change, and agreed it's intentional and reasonable here. Documented it explicitly in the README (80114f5): `rocm serve` now only reuses a running service when its recipe -- including the sampling-control defaults -- matches the current request, so a differing `--temperature`/`--top-p`/override-generation-config will start a new service instead of silently reusing a mismatched one.

apps/rocm/src/main.rs
r0x0r: Fair point -- added a warning log for this case in 80114f5 instead of discarding silently: when an existing `--override-generation-config` value fails to parse as JSON, we now print `warning: existing --override-generation-config value is not valid JSON; discarding it` before continuing with the merge.

apps/rocm/src/therock.rs
r0x0r: This rename/lock is already isolated in its own commit (423b095 "fix: resolve clippy lints and a flaky PATH-race test"), separate from the sampling-controls feature commits (64b400a, 68957b7) -- it was not bundled into the feature commit, just included in the same PR/branch. Happy to split it into a standalone PR if you'd prefer it land independently of the sampling-controls work.

apps/rocm/src/providers.rs
r0x0r: Agreed, and did this across the file in 80114f5: all the `temperature: None, top_p: None` pairs in these test literals (including the one you pointed at) now use `..Default::default()` instead.

juhovainio (approved):
Reviewed the diff (checked out the branch, ran `cargo check --workspace --all-targets` and `cargo clippy -p rocm -p rocm-dash-tui -p lemonade --all-targets -- -D warnings`, both clean, plus 16 targeted tests, all passing). Traced the trickier parts by hand too: the vLLM override-generation-config merge runs after the tool-call-parser/gpu-mem-util recipe overrides so it can't clobber them, the managed-service reuse guard bails before the already-running short-circuit, and the ChatGPT temperature/top_p exclusivity check is correctly scoped to the Responses API backend only (the regular chat-completions path allows both, which is correct there). No issues found.

This is a diff-only review, though, not a substitute for CI or a runtime check, so please make sure CI is green before merging.

PR #224, opened 2026-08-12
fix(wsl): install ROCDXG from the CLI instead of naming a repo script

### LINKED ISSUE #223: [WSL] rocm install driver points at an unshipped repo script instead of installing ROCDXG
## Summary

On WSL, `rocm install driver` refuses to do anything and tells the user to run a
script that is not shipped with the CLI:

```
$ rocm install driver
driver install plan
  policy: wsl_rocdxg
  supported: false
  mutating: false
  reason: WSL uses the Windows host driver plus ROCDXG; run `scripts/wsl_setup_rocdxg.sh` inside WSL instead of installing Linux DKMS.
  execution_commands: <none>
  action: no driver commands will be executed
```

`scripts/wsl_setup_rocdxg.sh` exists only in a git checkout. The release bundle
contains `bin/rocm`, `bin/rocmd`, `README.md`, `LICENSE.TXT` and the platform
installer — nothing else — so anyone who installed rocm-cli the normal way is
pointed at a path they do not have.

## Why it matters

ROCDXG (`librocdxg`) is the ROCm component that lets the runtime reach the GPU
under WSL2. Without it the GPU is visible to the distro but unusable, and the
CLI already knows this:

```
driver_status: wsl_rocdxg_missing
driver_detail: missing /opt/rocm/lib/librocdxg.so, ldconfig:librocdxg.so
detected_gfx_target: gfx1151
```

`rocm serve` then refuses with "no usable AMD GPU detected". So on the one
platform where a ROCm-provided library is the difference between a working and a
non-working GPU, the CLI declines to install it — while on native Linux
`rocm install driver` builds and executes a real plan.

`rocm diagnose` does not close the gap either: on WSL it declares the bare-metal
catalog out of scope and returns a docs URL, with no entry for
`wsl_rocdxg_missing`. That leaves the user with a diagnosis they cannot act on.

## Expected

`rocm install driver` performs the ROCDXG installation on WSL, subject to the
same approval and dry-run semantics as the native Linux driver path, instead of
requiring the user to clone the repository.

## Notes

- Reported from a real user session; also hit on the self-hosted WSL2 CI runner.
- The script already does the right shape of work — preconditions (`/dev/dxg`,
  `libdxcore.so`), download a versioned artifact from a GitHub release, verify
  SHA-256, install, verify the result — which is the same shape as the CLI's
  existing managed-artifact install paths.
- Independently of the larger question, the current guidance naming an unshipped
  path is wrong and worth correcting on its own.

rominf:
## Summary

On WSL2, `rocm install driver` refused to do anything and pointed the user at a
script that is not shipped:

```
  supported: false
  mutating: false
  reason: WSL uses the Windows host driver plus ROCDXG; run `scripts/wsl_setup_rocdxg.sh` inside WSL instead of installing Linux DKMS.
  execution_commands: <none>
  action: no driver commands will be executed
```

`scripts/` is not in the release bundle — `xtask package` stages `bin/rocm`,
`bin/rocmd`, `README.md`, `LICENSE.TXT` and the platform installer — so for
anyone who installed rocm-cli normally that names a file they do not have.

It matters because ROCDXG (`librocdxg`) is what lets the ROCm runtime reach the
GPU under WSL2. Without it `examine` reports `wsl_rocdxg_missing` and `serve`
refuses with "no usable AMD GPU detected", while still detecting a gfx target —
that target is read from the Windows-side driver, so its presence says nothing
about whether ROCm can use the device. The one platform where a ROCm-provided
library decides whether the GPU works was the one platform where the CLI would
not install it.

Fixes #223.

## What changes

The WSL branch of `build_driver_install_plan` returns a real plan instead of a
refusal — the same steps `scripts/wsl_setup_rocdxg.sh` ran, expressed through
the machinery every other platform already uses, so it inherits plan-then-approve,
`--dry-run`, the recorded install state and the post-install reconciliation.

```
$ rocm install driver
driver install plan
  policy: wsl_rocdxg
  supported: true
  mutating: true
  approval: required
  repo_version: 1.2.0
  reason: WSL2 uses the Windows host driver plus ROCDXG, not Linux DKMS; this installs ROCDXG.
  preflight_checks:
    /dev/dxg (WSL GPU device)
    /usr/lib/wsl/lib/libdxcore.so (WSL dxcore runtime)
    root access: run as root, or ensure `sudo -v` succeeds before approval
    `sudo` command is available when not running as root
  execution_commands:
    Prepare: test -e /dev/dxg || { ... }
    Prepare: test -e /usr/lib/wsl/lib/libdxcore.so || { ... }
    Prepare: command -v sudo >/dev/null 2>&1 || { ... }
    Prepare: sudo apt-get update
    Prepare: sudo apt-get install -y ca-certificates curl
    Execute: curl -L --fail --show-error --output '/tmp/rocdxg-roct_1.2.0_amd64.deb' 'https://github.com/ROCm/librocdxg/releases/download/v1.2.0/rocdxg-roct_1.2.0_amd64.deb'
    Execute: printf '%s  %s\n' '3ed9526719...' '/tmp/rocdxg-roct_1.2.0_amd64.deb' | sha256sum -c -
    Execute: sudo apt-get install -y '/tmp/rocdxg-roct_1.2.0_amd64.deb'
    Execute: sudo ldconfig
  post_install_check_commands:
    test -e /opt/rocm/lib/librocdxg.so
    ldconfig -p | grep -q 'librocdxg\.so'
  action: rerun with --yes after reviewing this plan, or approve from the TUI
```

**Guards before anything mutates.** `/dev/dxg` and dxcore come from the Windows
side; if they are absent the fix is on the host and installing the bridge
library accomplishes nothing, so the plan stops rather than reporting a
successful install of something inert.

**Verification asserts what `examine` keys on.** `wsl_rocdxg_ready` requires the
library *and* its ldconfig entry, so both are checked — a partial install cannot
report success.

## The download is authenticated by default

The package is fetched over plain HTTPS from a release page and handed to
`apt-get install`, which runs its maintainer scripts as root. Verifying it
cannot be something the user has to remember to switch on, and the sibling apt
path in this same file sets the bar: it installs from a repository pinned with
`signed-by=/etc/apt/keyrings/rocm.gpg`.

So the SHA-256 of the `rocdxg-roct` package is pinned in-source for every
published ROCDXG release, taken from the release host's own asset metadata, and
checked before the install step. On a mismatch the plan stops there.

Verification is opt-**out**, and the failure mode of an unset variable is a plan
that will not run rather than an unverified root install:

| situation | result |
|---|---|
| default (pinned version) | verified against the pinned digest |
| `ROCM_CLI_ROCDXG_SHA256` set | verified against that digest; overrides the pinned one |
| unpinned version, no digest | **refused**, naming both ways forward |
| `ROCM_CLI_ROCDXG_ALLOW_UNVERIFIED=1` | installs unverified, printing a warning naming the version |
| malformed supplied digest | refused, rather than silently falling back to the pinned one |

**`ROCM_CLI_ROCDXG_VERSION` is validated.** It reached three commands run via
`sh -c` unquoted, after `apt-get update` had primed the sudo credential cache —
so a `;` in it started a second command as root. It is now restricted to the
characters a package version contains, and anything else yields a non-mutating
refusal plan that builds no commands at all. The value is resolved at plan-build
time, so the printed plan names the build that will actually be fetched, and the
interpolated path and URL are quoted.

## `reboot_required` is a property of the plan

Previously the post-execution output asserted `reboot_required: true`
unconditionally. That is right for the DKMS paths — an amdgpu module is not live
until the machine restarts — and wrong here: ROCDXG is userspace and `ldconfig`
publishes it in the current boot. Plans carry the flag, the DKMS paths keep
`true`, and the verification steps are labelled `post_install` rather than
`post_reboot`.

## The script is deleted, not kept alongside

`scripts/wsl_setup_rocdxg.sh` is removed. Keeping it would leave two
implementations of one privileged install, and two of everything that can drift:
a `ROCDXG_VERSION` pin beside `ROCM_CLI_ROCDXG_VERSION`, and a `ROCDXG_SHA256`
beside `ROCM_CLI_ROCDXG_SHA256`. Nothing in CI ever ran it.

Its one behaviour the plan lacked — an explicit "sudo is required" message
instead of a bare `sudo: command not found` — is folded in as a guard, and its
`ROCDXG_CHECKSUM_SELF_TEST` hook is replaced by a test that runs the generated
digest step for real (below). `scripts/wsl_preflight.py` is untouched.

`rocm diagnose` and `rocm fix` carried recipes naming the deleted script; both
now route to `rocm install driver`. Docs lead with it too.

## Test plan

Thirteen unit tests over the generated plan, covering the install steps and URL,
the ordering guarantee that the plumbing and root guards precede the first
mutating command, the verification pair, each row of the table above, the
escalation branches, and that WSL asks for no reboot while bare metal still does.

`wsl_rocdxg_generated_digest_step_accepts_only_the_matching_file` builds the
command the plan actually generates and **runs** it against a real file with a
matching, mismatched, malformed and empty digest, asserting exit status each
time — then asserts the default plan's step has that shape, so the execution is
testing the real fragment. It is in the **Release trust checks** list in
`docs/testing.md`, and is Unix-only (needs a POSIX shell and `sha256sum`).

Three mutations were each killed by their intended test: weakening the version
validator to an empty-check, making an unknown digest fall through to
"unverified", and changing one space to two in the `printf` format.

**Coverage gap, stated.** The three `driver_install.feature` scenarios are
`@requires-wsl`, so they run on `E2E tests (Strix Halo, WSL2)` and **not** on the
required `E2E tests` lane. They previously set `WSL_DISTRO_NAME` to stand a WSL
host up on a native runner; `main` has since stopped trusting that variable for
`is_wsl_host()`, which reads only `/dev/dxg` and `/proc/version`. On a native
runner the branch under test is never selected, so those scenarios would have
been asserting against a bare-metal plan. There is no supported way to fake the
premise, so they skip where they have no premise; the plan itself is covered by
the unit tests above on every PR.

**Verification gap.** This box is WSL2 *without* GPU passthrough
(`driver_status: wsl_gpu_plumbing_missing`, no `/dev/dxg`), so a full `--yes`
run cannot be exercised here — the plan correctly refuses at the first guard.
Someone with `/dev/dxg` and dxcore present, or the WSL2 lane, should confirm a
real run reaches `wsl_rocdxg_ready`.

`cargo fmt --all --check`, both clippy invocations CI runs, and
`cargo test --workspace --all-targets` are clean apart from failures that
reproduce on an unmodified `main` checkout: the `rocm-dash-tui` agent OAuth unit
tests, and — in `cargo xtask e2e` — `dash-gen-tps-held-after-scrape-failure` and
`dash-gen-tps-expiry-boundary`, which are currently red on `main` itself.

Risk: medium. A command that previously did nothing on WSL now runs privileged
package operations — behind the same explicit `--yes` approval as every other
platform, after the plumbing guards pass, and on a package whose digest has been
checked. The behaviour change to watch is that an unpinned
`ROCM_CLI_ROCDXG_VERSION` is now refused rather than installed unverified.


michaelroy-amd (dismissed):
Reviewed at `8361e70e15106ad5aae25dd60b609888eef34587`. The WSL path now uses the existing plan-and-approve machinery instead of referring users to an unshipped script. The `/dev/dxg`, dxcore, and sudo guards precede every mutating command; `reboot_required` remains false for userspace ROCDXG and true for DKMS paths. The two Cucumber scenarios cover dry-run and approval-gated behavior, all 17 required checks are green, and both commits are signed off. The disclosed real-hardware `--yes` gap is bounded and does not invalidate the plan-path fix. No blocking finding. Approving.

rominf:
Rebased onto current `main` (`8361e70e` → `cfd6654a`). Flagging this rather than relying on the approval carrying over, because **the rebase changed user-visible behavior** and your review predates it.

**What changed.** #307 (`263ec0a5`) moved driver plans to resolving `${VAR:-default}` templates at plan-build time, renaming `repo_version_expr` → `repo_version`. The WSL plan was written against the old raw-template convention, so it now resolves the version once and bakes the concrete value into the deb name, release URL, and the `repo_version:` line:

```
-    let deb = format!("rocdxg-roct_{ROCDXG_VERSION_EXPR}_amd64.deb");
-    repo_version_expr: ROCDXG_VERSION_EXPR.to_owned(),
+    let version = resolve_shell_default_template(ROCDXG_VERSION_EXPR);
+    let deb = format!("rocdxg-roct_{version}_amd64.deb");
+    repo_version: version,
```

So `rocm install driver --dry-run` under WSL now prints `repo_version: 1.2.0` where it previously printed `${ROCM_CLI_ROCDXG_VERSION:-1.2.0}`. The override mechanism itself is unchanged.

**Why it wasn't optional.** This conflict was semantic, not textual — a purely mechanical resolution would have compiled and passed unit tests, then failed #307's new scenario `install-driver-dry-run-resolves-repo-version`, which asserts `repo_version` contains no `${` and is gated `@requires-os:linux`, not on GPU. WSL is Linux, so that scenario covers this path.

**One deliberate exception.** `${ROCM_CLI_ROCDXG_SHA256:-}` is still a shell expansion. That one is meant to be read at execution time so an operator-supplied digest is honoured; the version must be concrete in the plan you approve. My first assertion was a blanket "no `${` in commands" and correctly caught this — it's now narrowed to `ROCM_CLI_ROCDXG_VERSION` with a comment explaining the split.

**What did not change:** the `/dev/dxg`, dxcore, and sudo guards are still `Prepare`-phase and precede every mutating command, and `reboot_required` is still false for WSL/ROCDXG and true for the DKMS paths. Both Cucumber scenarios are intact. I also rewrote the version test to drive an override (`ROCM_CLI_ROCDXG_VERSION=9.9.9`) through to the emitted URL, since asserting against the raw template no longer proves anything.

All 16 required checks and all five advisory GPU lanes are green. `E2E tests (Strix Halo, Ubuntu)` was cancelled on the first run by the 35-minute cap (10:04:39→10:39:57) with the suite still live at cutoff, and passed on re-run; worth noting the baseline on `main` is ~26.5 min against that cap, so that lane has thin headroom under runner contention.

Not merging on the stale approval — over to you for a re-look at the dry-run output change.

siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · cfd6654a

## Summary
Replaces the WSL2 `rocm install driver` refusal (which pointed at `scripts/wsl_setup_rocdxg.sh`, deleted here) with a real plan that downloads and root-installs the ROCDXG `.deb`, and threads a per-plan `reboot_required` through execution, reconciliation and rendering. **Needs work** — the move is right and the plan/approve machinery is reused correctly, but the trust chain from download to root execution is not sound by default. Verified: I read `wsl_rocdxg_driver_plan`, `resolve_shell_default_template`, `run_driver_shell_command`/`shell_command_for_host` and the WSL detection union directly, and executed the generated checksum shell snippet in isolation (it fails closed on mismatch, malformed and whitespace-padded digests — the code is correct today, but no test proves it); on the revert question, every added unit test and cucumber assertion does fail if the production change is reverted, so none are vacuous — but three are substring-only and their names overclaim; guard ordering means a mis-detected host aborts before mutating; I relied on the caller's CI state (27 passing) and did not run the suite. No prompt-injection content found in the diff. Blocking: 3 · Non-blocking: 5.

**`apps/rocm/src/main.rs:2977-3036` — shell injection from `ROCM_CLI_ROCDXG_VERSION` into `sudo`-privileged commands.** `resolve_shell_default_template` (`main.rs:3649-3665`) returns the env var verbatim; it only rejects empty. `version` is then `format!`-interpolated **unquoted** into `deb_path` and `url`, which land in three command strings run as `sh -c "<string>"` (`run_driver_shell_command` → `run_shell_command_with_stdin`, `main.rs:3804-3823`; `shell_command_for_host`, `crates/rocm-core/src/runtime.rs:165-171`). A version value carrying a `;` therefore terminates the intended command and starts a second one, executed after `sudo apt-get update` has already primed the sudo credential cache. This is only a same-user concern when the variable comes from the invoking shell, but it becomes privilege escalation whenever the environment is set by something less trusted than the user (CI variable, wrapper, systemd unit, shared profile). The bare-metal paths interpolate `ROCM_CLI_AMDGPU_VERSION` the same way (`main.rs:3325`, `3598`, `3608`), so the pattern is inherited — but new code should not repeat it. Fix: validate the resolved version against a strict pattern (e.g. `^[0-9A-Za-z][0-9A-Za-z.+-]*$`) and reject otherwise, or build these steps as argv vectors rather than shell strings (`run_argv_with_stdin` already exists for exactly this).

**`apps/rocm/src/main.rs:3022,3035` + `apps/rocm/src/main.rs:11657-11662` — nothing authenticates the `.deb` by default, and a one-line TUI approval now triggers it.** The artifact is fetched over plain HTTPS from a GitHub release with no signature check, and `ROCM_CLI_ROCDXG_SHA256` is unset by default, so TLS to the download host is the only default authentication before `sudo apt-get install -y` runs the package's maintainer scripts as root. This is materially weaker than the sibling apt path in the same file, which installs through a repo pinned with `signed-by=/etc/apt/keyrings/rocm.gpg` (`main.rs:3325`). Compounding it: `chat_rocm_command_action_from_args` force-injects `--yes` for `install driver` and surfaces to the human only `pending_title: "Install driver"` — not the URL, not that a `.deb` is installed as root. Before this PR that approval was a no-op on WSL; after it, the same single approval runs an unverified root install. Fix: pin and verify a digest per pinned release by default (opt **out**, not opt in), and/or have the approval surface render the actual plan for mutating driver installs.

**`docs/testing.md:749` (removed) vs `apps/rocm/src/main.rs:26340-26350` — executable coverage of the checksum logic was deleted and replaced with substring assertions.** The deleted script carried a real self-test (`ROCDXG_CHECKSUM_SELF_TEST=1`) that ran the checksum branch against contrived inputs and asserted pass/fail; this PR removes its invocation from the **Release trust checks** list. Its replacement, `wsl_rocdxg_checksum_is_opt_in_but_enforced_when_supplied`, asserts only that `ROCM_CLI_ROCDXG_SHA256`, `sha256sum -c -` and `skipping checksum verification` appear in the generated command list. Nothing executes the snippet, so a future quoting, field-order or newline regression in that `printf | sha256sum -c -` fragment would ship green — on the one step that is supposed to be the trust anchor. (I ran the snippet manually and it is correct today; that is exactly the check the repo no longer has.) Fix: add a test that actually runs the generated command against a temp file with a correct digest, a wrong digest and a malformed digest, and asserts the exit status.

## Non-blocking
- `apps/rocm/src/main.rs:2980` — `/tmp/rocdxg-roct_<version>_amd64.deb` is a predictable path in a world-writable dir; `curl --output` follows symlinks, and each command is a separate `sh -c`, leaving a TOCTOU window between the checksum step and `apt-get install`. Use `mktemp -d`. (Same as the deleted script, so not a regression.)
- `crates/rocm-core/src/lib.rs:2346-2347` — the detection union's own doc comment justifies its breadth with "a false positive costs a route-out note". That rationale is no longer true after this PR; `WSL_DISTRO_NAME` being set alone is sufficient (`lib.rs:2354`, `2361-2367`). The comment should be updated, and the widening reconsidered now that the cost changed. Mitigation is real but narrow: the `/dev/dxg` and `libdxcore.so` guards run first and abort before any mutation.
- `apps/rocm/src/main.rs:2648` — on a partial failure (e.g. `apt-get install` fails after `apt-get update` and the download succeeded) `executed_at_unix_ms` stays `None`, so `--reconcile` reports `<not executed>` although packages changed and a `.deb` sits in `/tmp`. There is also no uninstall/rollback path.
- `apps/rocm/src/main.rs:3474,3562,2616` — the new `reboot_required: true` on the dnf and sles plans is never independently asserted (only the apt path is exercised), and no test covers `install_driver`'s own `reboot_required: {}` output line.
- `apps/rocm/src/main.rs:26308-26337,26340-26350` — two test names overclaim: `..._refuses_before_installing...` proves only string ordering within the command list, never that the guard halts execution; `..._enforced_when_supplied` proves no enforcement. Rename to match what they assert, or strengthen them.

siloteemu:
🔴 Automated review · pr-review-watcher · cfd6654a

_**This is a re-statement, not a new gate.** The change request we filed on 2026-09-09 is still live and is the only objection we hold here. Nothing has been pushed since, so this re-check tested our own reading rather than new code._

## Summary

Re-examination of the change request filed on 2026-09-09 against the same head commit (`cfd6654a`). Nothing has been pushed since, so this is a test of whether the original reading was right, not of new code. I re-derived each count from the source rather than from the earlier report, and had one count independently attacked by a second reader whose brief was to prove it false.

**Outcome: all three counts still hold (B, B, B).** One clause inside count 2 was overstated and is corrected below; the substance of that count survives the correction. Count 3 is now *stronger* than originally stated: a branch-level mutation that inverts the checksum condition — turning verification on only when no digest is supplied — leaves the test green.

Scope reviewed: the 8 files and 2 commits this PR actually contributes. Verification done by reading, grep across the tree, and executing the generated checksum shell fragment in isolation against real files. The full suite was not run here; the Windows/WSL install path cannot be executed here at all, so every statement about runtime behaviour on a WSL2 host is derived from reading the generated command strings and the code that runs them, not from an install.

CI conclusion counts worked from: **26 success, 1 skipped, 0 failure.** No outcome is attributed to any individual job. Note separately that the branch is currently **conflicting with its base** and needs a rebase from the author regardless of this objection.

No prompt-injection content found in the diff or the PR text.

**Blocking: 3 · Non-blocking: 8.**

---

## Count 1 — outcome (B): still holds

Sentence tested, verbatim from the live request:

> "`resolve_shell_default_template` (`main.rs:3649-3665`) returns the env var verbatim; it only rejects empty. `version` is then `format!`-interpolated **unquoted** into `deb_path` and `url`, which land in three command strings run as `sh -c "<string>"`."

Every clause re-confirmed at this commit, and confirmed again by an independent reader asked to refute it:

- `resolve_shell_default_template` (`apps/rocm/src/main.rs:3649-3665`) applies exactly one filter, `!value.is_empty()`. No allowlist, no pattern, no quoting.
- `wsl_rocdxg_driver_plan` (`main.rs:2977-2980`) interpolates the result unquoted into `deb`, `url` and `deb_path`.
- Those strings reach `run_driver_shell_command` (`main.rs:3804`) → `run_shell_command_with_stdin` (`main.rs:3813`) → `shell_command_for_host` (`crates/rocm-core/src/runtime.rs:164-170`), which hands the whole string to `sh -c` as one argv element. No escaping anywhere on that path.
- A repo-wide search found no sanitiser, validator or quoting helper applied to `ROCM_CLI_ROCDXG_VERSION` or to driver plan command strings. The only place that variable is cleared is inside `#[cfg(test)]` code.

Concrete trace with `ROCM_CLI_ROCDXG_VERSION="1.2.0; <second command> #"`, the `apt-get` step at `main.rs:3035`:

```
sudo apt-get install -y /tmp/rocdxg-roct_1.2.0; <second command> #_amd64.deb
```

`;` sequences unconditionally and `#` swallows the remainder, so the second command runs regardless of what `apt-get` did. The same shape applies to the `curl` step (`main.rs:3022`) and the checksum step (`main.rs:3030`). Because `sudo apt-get update` has already run earlier in the plan, a `sudo`-prefixed second command lands without a further password prompt.

This is a same-user concern when the variable comes from the invoking shell; it becomes escalation whenever the environment is set by something less trusted than the user — a CI variable, a wrapper, a unit file, a shared profile.

The sibling bare-metal plans interpolate `ROCM_CLI_AMDGPU_VERSION` the same way (`main.rs:3066` into `3325`, `3598`, `3608`), so the pattern is inherited rather than invented here. That is context, not a defence: new code should not extend it.

**What would resolve this count:** validate the resolved version against a strict pattern (e.g. `^[0-9A-Za-z][0-9A-Za-z.+-]*$`) and fail the plan otherwise, or model these steps as argv vectors instead of shell strings — `run_argv_with_stdin` (`main.rs:3830`) already exists for exactly that, and the system-package install path already uses it.

## Count 2 — outcome (B): still holds, with one clause corrected

Sentence tested, verbatim:

> "`chat_rocm_command_action_from_args` force-injects `--yes` for `install driver` and surfaces to the human only `pending_title: "Install driver"` — not the URL, not that a `.deb` is installed as root."

**Correction:** "only `pending_title`" was wrong. `chat_tool_approval_request` (`main.rs:11255`) also sets `display_command` to the formatted argv, and the approval seam puts it in the modal body, so the approver does see `rocm install driver --yes`. I was wrong to say the title is all they see, and I am withdrawing that clause.

The correction does not rescue the count. What the approver still never sees is the **plan**: the release URL, the `sudo apt-get install -y /tmp/....deb` line, or the fact that a third-party `.deb` is fetched and its maintainer scripts run as root. `render_driver_install_plan` output never reaches that surface. Re-confirmed:

- `chat_rocm_command_action_from_args` (`main.rs:11657-11663`) still injects `--yes` unconditionally for `install driver`.
- `install_driver` (`main.rs:2581-2585`) returns early when `!yes || dry_run || !plan.supported || !plan.mutating`. Before this PR the WSL branch produced `supported: false, mutating: false, commands: []`, so that approval was a no-op on WSL. After this PR the same single approval executes the install. That behavioural change is what the count is about, and it is real.
- Nothing authenticates the artifact by default. `ROCM_CLI_ROCDXG_SHA256` is unset unless an operator sets it, the package is fetched from a release URL over TLS, and `sudo apt-get install -y <local .deb>` performs no signature check on a local file. TLS to the download host is the only default authentication before root-level maintainer scripts run.
- This is materially weaker than the sibling apt path in the same file, which installs through a repository pinned with `signed-by=/etc/apt/keyrings/rocm.gpg` (`main.rs:3325`).

Worth noting alongside: `AGENTS.md` §9 states "checksum sidecars are required" and directs release-affecting changes to `docs/release-trust.md` — which this PR edits. That rule is written about the project's own release artifacts, so I am not calling this a direct rule violation, but the direction of travel is the opposite of the project's stated posture on trust.

**What would resolve this count:** pin and verify a digest per pinned release by default — opt **out** of verification, not opt in — and/or render the actual plan on the approval surface for mutating driver installs, so the approver sees the URL and the root install rather than an argv.

## Count 3 — outcome (B): still holds, and is stronger than first stated

Sentence tested, verbatim:

> "The deleted script carried a real self-test (`ROCDXG_CHECKSUM_SELF_TEST=1`) that ran the checksum branch against contrived inputs and asserted pass/fail; this PR removes its invocation from the **Release trust checks** list. Its replacement, `wsl_rocdxg_checksum_is_opt_in_but_enforced_when_supplied`, asserts only that `ROCM_CLI_ROCDXG_SHA256`, `sha256sum -c -` and `skipping checksum verification` appear in the generated command list."

Confirmed exactly, both halves. The self-test invocation is gone from the Release trust checks block in `docs/testing.md` (now lines 743-750), and the deleted script's `run_checksum_self_test` exercised the valid, malformed and mismatched cases with real files. The replacement at `apps/rocm/src/main.rs:26343-26352` is three `contains` calls over a joined string. Nothing executes the fragment.

**One precision the original report owes the reader:** that self-test was a *documented manual* release-trust check, not a CI gate — a tree-wide grep finds no workflow that ever invoked it. So this PR did not drop an automated check. What it did is leave the repo with **no** execution of that fragment, manual or automated, on the one step that is the trust anchor for the whole install.

**Branch-level mutation (standing focus (a)) — this is the part that makes it a hard blocker.** Take the generated command at `main.rs:3030` and invert the single condition, `-n` → `-z`. The install then verifies the digest only when no digest is supplied, and skips verification whenever one is. The security property is exactly inverted. All three assertions still find their substrings, so the test named `..._enforced_when_supplied` stays green. Weaker mutations pass too: appending `|| true` to `sha256sum -c -`, or collapsing the two-space `printf '%s  %s\n'` separator that `sha256sum -c` requires.

I executed the fragment as generated, in a scratch directory, against real files: it exits 0 for a correct digest and for an unset variable, and non-zero for a wrong digest, a malformed digest and a whitespace-padded digest. It is correct today. That is precisely the check the repo no longer has.

The same defect shape affects `wsl_rocdxg_plan_refuses_before_installing_when_the_gpu_plumbing_is_absent` (`main.rs:26308`), which asserts only substring presence and ordering. Dropping `exit 1` from either guard — so the plan prints "is missing" and carries straight on to `sudo apt-get` — leaves that test green as well. The name promises a refusal the assertions never establish.

**What would resolve this count:** a test that actually runs the generated checksum command against a temp file with (a) a correct digest, (b) a wrong digest, (c) a malformed digest, and (d) the variable unset, asserting the exit status in each case. The same treatment for the plumbing guards would close the second instance.

---

## New at this commit (all non-blocking)

Findings the earlier report did not contain. Three are prose stating a false fact — the class this codebase has the most trouble with.

- **`apps/rocm/src/main.rs:2992-2995` — comment names the wrong phase.** "The Execute phase fails on the same two paths" — the `/dev/dxg` and dxcore guards are `DriverCommandPhase::Prepare` (`main.rs:3006-3013`); `Execute` holds the download, checksum, install and `ldconfig` steps. A reader following this comment looks in the wrong place. Fix: "The Prepare phase".
- **`apps/rocm/src/main.rs:26294-26295` — comment miscounts.** "Verification asserts the two things `examine` keys `wsl_rocdxg_ready` on" — `WslSummary::rocdxg_ready` (`crates/rocm-core/src/lib.rs:2480-2482`) is the conjunction of **four**: `dxg_device && dxcore && librocdxg && ldconfig_librocdxg`. The Verify commands cover two of them; the other two are the Prepare guards. As written the comment claims the test closes the whole predicate. Fix: "the two remaining things, the other two being guarded in Prepare".
- **`crates/rocm-core/src/lib.rs:2345-2347` — a doc comment this PR makes false without touching it.** The WSL detection union justifies its breadth with "a false positive costs a route-out note". After this PR a false positive routes into a mutating, root-privileged, network-fetching install plan. `WSL_DISTRO_NAME` being set is sufficient on its own (`lib.rs:2354`), and the PR's own new scenario relies on exactly that to make a plain Linux runner produce a WSL install plan. The mitigation is real but narrow: the `/dev/dxg` and dxcore guards run first and abort before any mutation. The sentence should be updated and the breadth reconsidered now that the cost has changed — this is the cheap one-line comment that stops the next reader reaching the old conclusion.
- **The reviewed `execution_commands:` list is not the full set of commands executed.** `DriverInstallPlan::execution_commands` (`main.rs:2886-2897`) filters to Prepare and Execute, but on the WSL plan `execute_driver_install_plan` (`main.rs:2640-2644`) runs the Verify commands too, and a Verify failure fails the install. The user approves one list and a longer one runs; `state.commands`, the persisted audit record, omits the same steps. The Verify commands are read-only, so the security impact is nil — but the record and the reviewed plan should match what ran.

## Prior non-blocking items, re-checked

- `apps/rocm/src/main.rs:2980` — `/tmp/rocdxg-roct_<version>_amd64.deb` is a predictable path in a world-writable directory; `curl --output` follows symlinks, and each command is a separate `sh -c`, leaving a window between the checksum step and `apt-get install`. `mktemp -d` closes it. Still holds; carried over from the deleted script, so not a regression.
- `apps/rocm/src/main.rs:2646-2650` — on a partial failure (e.g. `apt-get install` fails after `apt-get update` succeeded and the package was downloaded) `executed_at_unix_ms` stays `None`, so `--reconcile` reports `<not executed>` although packages changed. **Reframed:** this is now a deliberate, tested choice — `failed_driver_verify_does_not_mark_execution_completed` (`main.rs:26127`-region) asserts it. The remaining gap is that a *mid-loop* failure after a successful mutation is indistinguishable from one before it, and there is still no uninstall/rollback path.
- `apps/rocm/src/main.rs:3478,3566` — `reboot_required: true` on the dnf and sles plans, and on the two unsupported fallbacks (`3082`, `3179`), is never asserted. Every `reboot_required` assertion in the file exercises either the WSL plan or the ubuntu/apt plan. Still holds.
- `apps/rocm/src/main.rs:2616` — no test covers `install_driver`'s own `reboot_required: {}` output line. Still holds.
- Test names overclaim: `..._refuses_before_installing...` (`main.rs:26308`) proves ordering, not refusal; `..._enforced_when_supplied` (`main.rs:26343`) proves no enforcement. Renaming is the cheap fix; strengthening them is the right one, and for the checksum test it is required by count 3.

## What was and was not verified here

- **Verified by reading and grep:** the full generated command list for the WSL plan; the resolver, runner and shell-dispatch chain; the approval action builder and the approval request it produces; the plan renderer; every new and changed comment and doc line against the file it describes; every new test's assertions against the branch each names.
- **Verified by execution:** the generated checksum fragment, run in isolation against real files for the correct, wrong, malformed, padded and unset cases.
- **Verified by reasoning, not execution:** the branch-level mutations in count 3. Each mutation changes only the interior of a shell string; the three assertions are substring containment over that string, so survival follows by inspection without compiling.
- **Not verified:** the install path itself. It requires a WSL2 host with `/dev/dxg` and dxcore, which is not available here, so no statement in this report rests on having run `rocm install driver`. The full suite was not run here.

## Recommendation

Restate the change request as still holding, with the count 2 clause corrected. Counts 1 and 3 are unchanged and independently re-confirmed; count 3 is stronger than first stated. The branch also needs a rebase onto its base before any of this can land.

## Positive signals

- Removing the script rather than leaving a second, untested copy of its steps is the right call, and the plan/approve/execute machinery is reused correctly rather than worked around.
- Threading `reboot_required` per plan and deferring Verify only when a reboot is genuinely pending fixes a real lie in the previous output — and `driver_reconcile_preserves_explicit_reboot_policy` genuinely tests it: restoring the deleted `state.reboot_required || executed.is_some()` line fails the `false` half of that loop.
- `failed_post_driver_gather_keeps_executed_state_persisted` pins the persist-before-gather ordering, which is the kind of thing that silently regresses.

juhovainio (approved):
Reviewed this end to end and it looks good to merge.

The fix itself: on WSL2, `rocm install driver` used to refuse to do anything and point users at `scripts/wsl_setup_rocdxg.sh`, a file that only exists in a git checkout and never ships in the release bundle. This makes the WSL branch return a real, executable plan (fetch/install/verify ROCDXG) through the same plan-then-approve machinery every other platform already uses, and deletes the now-redundant script.

Worth calling out: the PR has two commits, and the second one is a legitimate self-caught fix. The first commit's plan added `Verify`-phase commands (checking `librocdxg.so` exists and is linked) but reused the old execution loop, which only ran `Prepare`/`Execute` and silently skipped `Verify` — so the install would have "succeeded" without ever actually checking the library landed. The second commit fixes exactly that (skip `Verify` only when a reboot is required) and adds six tests pinning the behavior. Good catch within the PR's own history.

I checked out the branch locally and verified rather than just reading the diff:
- `cargo test -p rocm --bin rocm wsl_rocdxg` and `driver_plan_executor` — all 7 new tests pass.
- `cargo clippy -p rocm --all-targets -- -D warnings` and `cargo fmt --all -- --check` — clean.
- Built the release binary and ran the two new e2e scenarios (`driver-install-wsl-dry-run-plan`, `driver-install-wsl-review-requires-approval`) against it — both pass.
- Confirmed the pre-existing bare-metal DKMS plans (`reboot_required: true`) still skip their `Verify` commands exactly as before, so this doesn't change behavior on Linux/DKMS hosts.

Design details also check out: guards (`/dev/dxg`, dxcore, `sudo`) run before anything mutates, checksum verification is opt-in and consistent with the existing `ROCM_CLI_AMDGPU_VERSION` pattern, and the version is resolved once at plan-build time so an override can't leave a stale URL. The PR body is upfront that the real WSL2-with-GPU install path hasn't been run on real hardware yet — worth confirming on a host with `/dev/dxg` present before calling this fully verified in production, but nothing here blocks merging.

Approving based on a thorough diff + local test/e2e verification; still worth a final look from CI and anyone with real WSL2 GPU passthrough hardware.

rominf:
Rebased onto `main` (`3edfb69`) and addressed all three blocking counts. Head is now `ec6b06e7`.

## Blocking

**1. Shell injection from `ROCM_CLI_ROCDXG_VERSION`.** Confirmed — `resolve_shell_default_template` returns the variable verbatim with only an empty-check, and it was `format!`-interpolated unquoted into three commands run via `sh -c` after `apt-get update` had primed the sudo credential cache.

The version is now validated against the characters a package version can contain; anything else returns a non-mutating refusal plan that builds no commands at all, so a rejected value never reaches a command string. The interpolated path and URL are also single-quoted, so the guarantee is visible at the command rather than resting on a check several functions away. `wsl_rocdxg_refuses_a_version_that_could_escape_the_shell` sweeps `;`, `&&`, `$(...)`, backticks, a newline and a trailing space, and sets the unverified opt-out at the same time so that cannot buy past the check. Weakening the validator back to an empty-check fails it.

I did not convert the plan to argv vectors. The whole `DriverInstallPlan` contract is shell lines — the escalation prefix, the `||` guards, the `printf | sha256sum` pipe — so that is a change to every plan builder, not this one. Worth doing, but not inside this fix.

**2. Nothing authenticated the `.deb`.** Agreed, and this was the one I think mattered most: TLS to the release host was the only default authentication before `apt-get install` ran maintainer scripts as root, against a sibling path in the same file that installs from a `signed-by=` pinned repository.

Verification is now opt-**out**. The SHA-256 of the `rocdxg-roct` package for all seven published releases is pinned in the source, taken from the release host's own asset metadata rather than computed here, and the check is unconditional and ordered before the install step. A version with no known digest is refused rather than installed unverified; the refusal names both ways forward — supply `ROCM_CLI_ROCDXG_SHA256`, or set `ROCM_CLI_ROCDXG_ALLOW_UNVERIFIED=1`, which emits a warning naming the version it is not checking. A malformed supplied digest is refused too, rather than silently falling back to the pinned one and verifying a different artifact than was asked for.

Making the unknown-digest case fall through to "unverified" instead of refusing fails `wsl_rocdxg_refuses_a_version_whose_digest_is_unknown`.

I did not change the approval surface to render the full plan for mutating driver installs. That is a chat/TUI-wide question about what a one-line approval is allowed to authorise, and it should be answered for every mutating action at once rather than special-cased here. Happy to open an issue for it.

**3. Deleted executable coverage of the checksum logic.** Restored as a test rather than a script. `wsl_rocdxg_generated_digest_step_accepts_only_the_matching_file` builds the command the plan actually generates and runs it against a real file with a matching digest, a mismatched one, a malformed one and an empty one, asserting exit status each time, then asserts the default plan's step has that same shape so the execution is testing the real fragment. Changing one space to two in the `printf` format fails it — that is the class of regression the substring assertions would have shipped green. It is back in the **Release trust checks** list in `docs/testing.md`, marked Unix-only since it needs a POSIX shell and `sha256sum`.

## Non-blocking

- **Test names that overclaimed** — `..._refuses_before_installing...` is now `..._guards_the_gpu_plumbing_before_any_mutating_command`, which is what it proves. The checksum test that "proved no enforcement" is gone, replaced by the set above.
- **`/tmp` predictable path / TOCTOU** — not fixed, deliberately. Each command is a separate `sh -c`, so a `mktemp -d` in one does not survive into the next; doing this properly means giving the plan a state directory, which is a wider change. The digest check does now close the content-substitution half of the window. Worth its own issue.
- **`executed_at_unix_ms` on partial failure, and no rollback** — not addressed here; it is a property of the shared driver-install executor rather than the WSL plan.
- **The WSL detection union's doc comment** — this resolved itself on `main`: `$WSL_DISTRO_NAME` is no longer trusted at all, and the comment now says so and explains why a false positive stopped being cheap. Nothing left to do.
- **`reboot_required: true` on the dnf and sles plans is unasserted** — still true, untouched.

## Carried over from the rebase

`main` moved 41 commits underneath this and the merge was conflict-free in places it should not have been, so two things are worth flagging:

- Every plan builder now takes a `PrivilegeEscalation` so containers and minimal images running as uid 0 with no `sudo` binary are not killed by an unconditional prefix. The WSL plan hardcoded `sudo` five times and demanded a `sudo` binary in preflight; it now takes the parameter, and `wsl_rocdxg_plan_drops_sudo_when_already_root` covers the root branch.
- `main` grew `rocm diagnose` / `rocm fix` recipes pointing at `scripts/wsl_setup_rocdxg.sh` — the script this PR deletes. Git saw no conflict because this branch never touched those files. Both now route to `rocm install driver`.

The new feature file was also never registered in `FEATURE_KEYS`, so the `feature_naming` guard failed once rebased; registered, with the scenarios renamed to the indexed convention.

## Coverage gap, stated

The three `driver_install.feature` scenarios are now `@requires-wsl` and no longer run on the required `E2E tests` lane — they run on `E2E tests (Strix Halo, WSL2)`. They previously set `WSL_DISTRO_NAME` to stand a WSL host up on a native runner, which worked when this branch was written and does not now that `is_wsl_host()` trusts only `/dev/dxg` and `/proc/version`. On a native runner the branch under test is never selected, so the scenarios would have been asserting against a bare-metal plan. There is no supported way to fake the premise, so they skip where they have no premise. The generated plan itself is covered by 13 unit tests that do run on every PR.

Verified: `cargo fmt --check`, both clippy invocations CI runs, `cargo test --workspace --all-targets`, and `cargo xtask e2e` — all three scenarios run and pass on a real WSL2 host. Three mutations (defeating the version validator, making unverified the silent fallback, and altering the `printf` format) are each killed by their intended test.

Two failures are not from this change and reproduce on a clean `main` checkout: the `rocm-dash-tui` agent OAuth unit tests, and `dash-gen-tps-held-after-scrape-failure` / `dash-gen-tps-expiry-boundary`, which are currently red on `main` itself.

rominf:
@juhovainio — your approval was at `cfd6654a`, and this has changed materially since, so please re-look rather than let it carry over.

What is different:

- **Default behaviour change.** `rocm install driver` on WSL now verifies the downloaded `.deb` against a SHA-256 pinned per ROCDXG release, and **refuses** a version it has no digest for. Previously it installed unverified unless you opted in. If you were relying on `ROCM_CLI_ROCDXG_VERSION` to pull an arbitrary build, that now needs `ROCM_CLI_ROCDXG_SHA256` or an explicit `ROCM_CLI_ROCDXG_ALLOW_UNVERIFIED=1`.
- **`ROCM_CLI_ROCDXG_VERSION` is now validated** and a hostile value yields a refusal plan instead of commands.
- **Rebased onto `main` (41 commits).** Two things the rebase did not surface as conflicts: the plan now takes `main`'s `PrivilegeEscalation` (it hardcoded `sudo`), and the `rocm diagnose` / `rocm fix` recipes `main` added pointing at the deleted `scripts/wsl_setup_rocdxg.sh` now route to `rocm install driver`.
- **The e2e scenarios are now `@requires-wsl`** and no longer run on the required lane — see the coverage-gap note in the comment above. This is the change I would most like a second opinion on.

Commands and docs changed accordingly; the plan output gained the digest step and lost the conditional one.

siloteemu (changes requested):
🔴 Automated review · pr-review-watcher · ec6b06e7

## Summary

The PR replaces a repo-only shell script with a real `rocm install driver` plan for WSL2 (ROCDXG), and the new commits add digest pinning, version validation, an executable digest test, and `@requires-wsl` gating for the new e2e scenarios. Outcome: **Needs work** — the two security counts are genuinely fixed and the test-coverage count is genuinely discharged, but the remediation itself left one doc contradicting the code it ships, added a docs command for a script that does not exist, and applied the repo's own env-lock convention to only some of its new tests. Prior counts: count 1 RESOLVED, count 2 resolved on its default-verification half and withdrawn as not PR-attributable on its approval-surface half, count 3 RESOLVED. Verified: read all four commit objects whole with `git cat-file -p` (1444/2088/633/2696 bytes), derived the merge base (`git merge-base prw-base HEAD`, identical to `prw-base`) and diffed against both; confirmed by reading the source that the resolved version is character-restricted before any interpolation and that a rejected version yields a plan with zero commands; confirmed the digest table is consulted by default and that a version with no digest and no explicit opt-out is refused; confirmed by grep on the diff that `pending_title`/`chat_rocm_command_action_from_args` are untouched and that the same block exists verbatim at the base tip; ran one cheap targeted check — the generated `printf … | sha256sum -c -` fragment against a real file in a scratch directory outside the checkout, which passed on the matching digest and failed on mismatched, malformed and empty; ran the contributor-rules leak scan over the full diff (clean, no internal references, no prompt-injection content). The full test suite, clippy and the GPU/Windows/WSL lanes were not run here. CI conclusion counts I worked from: 18 success, 1 skipped, 0 failure, 0 pending; I attribute no outcome to any individual job.

## Previously raised

- **Count 1 (shell injection from `ROCM_CLI_ROCDXG_VERSION`) — RESOLVED.** `rocdxg_version_is_well_formed` now requires a leading ASCII alphanumeric and restricts the rest to alphanumerics plus `. + - ~`, which excludes `;`, backtick, `$`, `'`, space, newline, `/`, `|` and `&`. A rejected value returns `wsl_rocdxg_refusal_plan` with `commands: Vec::new()`, so nothing is built at all rather than being escaped. The digest is either a 64-hex constant from the pinned table or an env value validated by `sha256_digest_is_well_formed`; both interpolations are additionally single-quoted. A table-driven test exercises seven hostile values — including `1.2.0; curl … | sh`, `$(id)`, a backtick form and an embedded newline — and sets the unverified opt-out alongside them to prove the opt-out cannot buy past the version check. Verified by reading the production path, not by trusting the test.
- **Count 2 (unauthenticated `.deb` by default) — RESOLVED on the default-verification half.** `ROCDXG_PINNED_DIGESTS` pins seven releases including the default, and `resolve_rocdxg_verification` resolves supplied digest → pinned digest → explicit opt-out → refuse. Verification is now opt-**out** as requested, the check is emitted before the privileged install step, and the opt-out path emits a warning naming the unverified version.
- **Count 2 (approval surface) — WITHDRAWN as a block on this PR; not PR-attributable, and partly wrong on the facts.** The diff contains zero occurrences of `pending_title` or `chat_rocm_command_action_from_args`, and the `--yes` injection plus `pending_title: "Install driver"` exist verbatim at the base tip. The claim that the human sees "only `pending_title`" is also incorrect: the approval request carries `display_command: Some(format_structured_tool_call("rocm", &args))`, so the full `rocm install driver --yes` argv is rendered in the approval body. It remains true that the download URL and the "installed as root" fact are not surfaced there — but that is pre-existing behaviour this PR neither introduces nor worsens, and belongs in its own issue.
- **Count 3 (deleted executable coverage of the checksum logic) — RESOLVED.** A `#[cfg(unix)]` test now shells out to `sh -c` with the `printf '%s  %s\n' '<digest>' '<path>' | sha256sum -c -` fragment against a real temp file and asserts success on the matching digest and failure on mismatched, malformed and empty digests, then asserts the plan-generated command carries the same prefix and suffix. It is re-listed under the Release trust checks in `docs/testing.md`. I reproduced the four-case behaviour myself in a scratch directory.

- `docs/release-trust.md:118-130` — the repo's trust-policy document still describes the pre-remediation model, and now directly contradicts the code in the same PR: "No production checksum is baked in for the ROCDXG `.deb`. Operators can require verification by providing the expected digest… when it is unset the plan says so explicitly rather than reporting an unverified download as verified." As of this head a digest **is** baked in for seven releases, verification happens with nothing set, and an unpinned version is **refused** rather than described. The earlier commit rewrote this section for the opt-in model and the remediation commit updated `docs/wsl.md` and `docs/testing.md` but not this file. This is the one document an operator consults to judge the trust model, so it blocks. Fix: rewrite the section to state that digests are pinned per release and verified by default, that `ROCM_CLI_ROCDXG_SHA256` overrides or supplies one for an unpinned release, and that `ROCM_CLI_ROCDXG_ALLOW_UNVERIFIED` is the only way to reach an unverified install.
- `docs/testing.md:973` — the newly added line `python scripts/wsl_preflight.py --require-ready` names a script that does not exist at this head, does not exist at the base tip, and is referenced nowhere else in the tree. It sits in the WSL install instructions a reader is meant to follow, so following them fails. Fix: drop the line, or replace it with a command that exists (`rocm examine` / `rocm diagnose` are what the plan's own `checks` list names).
- `apps/rocm/src/main.rs` — three of the new unit tests read the ROCDXG environment without the `ScopedTestEnv` guard, while sibling tests in the same block mutate exactly those variables under its process-wide mutex: `wsl_rocdxg_plan_installs_the_library_and_publishes_it`, `wsl_rocdxg_plan_guards_the_gpu_plumbing_before_any_mutating_command`, and `wsl_rocdxg_install_does_not_ask_for_a_reboot`. This PR added `scoped_rocdxg_env()` for precisely this reason and even annotated another test with "this reader has to take the same guard as the mutating tests in this binary", so the convention is the PR's own. Two of the three fail spuriously when a concurrent test has `ROCM_CLI_ROCDXG_VERSION` set, and the third is worse: against a refusal plan its three assertions (`!reboot_required`, output contains `post_install_checks:`, output lacks `post_reboot`) all still hold, so it passes while testing nothing the name claims. Fix: add `let _env = scoped_rocdxg_env();` as the first line of each of the three, and have the reboot test assert on something only the real install plan produces (for example that the rendered plan contains the `apt-get install -y '/tmp/` step).

## Non-blocking

- `apps/rocm/src/main.rs:28704-28708` — comment claims the `${ROCM_CLI_ROCDXG_SHA256:-}` expansion is "deliberately left intact… read at execution time", but no such expansion is generated any more; a sibling test asserts that variable name is absent from the commands. Stale from the pre-remediation revision; delete it.
- `resolve_rocdxg_verification` treats any non-empty `ROCM_CLI_ROCDXG_ALLOW_UNVERIFIED` as opt-out, so `=0` or `=false` disables digest checking. Parse it as a boolean, or accept only `1`/`true`.
- Nothing in the repo or in any runnable check validates the seven pinned digests against the real published artifacts, and no test downloads one — a mistyped digest would surface only as a failed install on a WSL host. (Fail-closed, so not dangerous; worth a note beside the table on how it is regenerated.) I could not verify the digests themselves here.
- The digest is verified on a predictable `/tmp` path in one command and the package installed from that same path in a later one, leaving a window between the check and the root install. The deleted script had the same shape, so this is carried over rather than introduced, but the digest is now the plan's stated trust anchor; a `mktemp -d` working directory would close it.
- The digest test executes a hand-built replica of the fragment rather than the generated command with a substituted path, so a swapped `digest`/`path` field order would keep its `starts_with`/`ends_with` assertions green. Factoring the `printf` fragment into one formatter used by production and the test would make them impossible to diverge. Relatedly, the refusal plan renders the raw, unvalidated version into its `repo_version:` line, so a value containing newlines can forge extra lines in the plan a human reviews.
- Reviewer-error note, for recurrence: the withdrawn approval-surface claim came from reading `pending_title` at its construction site and not following it to `chat_tool_approval_request`, several hundred lines away, where the full argv is attached as `display_command`. That distance will mislead the next reader too; a one-line comment at the `ChatRocmCommandAction::Approval` construction site saying the full command line is attached downstream and rendered in the approval body would prevent it.

rominf:
Rebased onto `main` (`3edfb69` → `6ad833f1`, 8 commits) and addressed the three blocking counts. Head is now `eee9059e`.

## Blocking

**1. `docs/release-trust.md` contradicted the code this PR ships.** Confirmed — the section still read "No production checksum is baked in for the ROCDXG `.deb`", which stopped being true at `28f75d03`. Rewritten to describe what the code does: a digest pinned per release and checked by default with nothing set, `ROCM_CLI_ROCDXG_SHA256` to override it or supply one for an unpinned release, refusal (unsupported plan, no commands) when neither is available, and `ROCM_CLI_ROCDXG_ALLOW_UNVERIFIED` as the only route to an unverified install. `docs/wsl.md` already described this; `release-trust.md` was the file the remediation commit missed.

**2. `docs/testing.md` named a script that does not exist.** Confirmed — `scripts/` contains no `wsl_preflight.py`, it is not on `main`, and a tree-wide grep found the string only on that one line. Replaced with the plan's own `post_install_checks`, `rocm examine` and `rocm diagnose`, with `rocm examine`'s expected `driver_status: wsl_rocdxg_ready` spelled out.

**3. Three plan tests read the ROCDXG environment without the guard.** Confirmed and fixed: `let _env = scoped_rocdxg_env();` is now the first statement of `wsl_rocdxg_plan_installs_the_library_and_publishes_it`, `wsl_rocdxg_plan_guards_the_gpu_plumbing_before_any_mutating_command` and `wsl_rocdxg_install_does_not_ask_for_a_reboot`.

The vacuous-pass reading of the reboot test is also confirmed, and I checked it by running it rather than by reading it. Pointing the test at `wsl_rocdxg_refusal_plan(...)` and running it printed a plan with `supported: false`, `execution_commands: <none>` and `post_install_checks:` — all three original assertions held. It now asserts the rendered plan contains `apt-get install -y '/tmp/` first, and against that same refusal plan it fails, as it should.

## Non-blocking

Fixed:

- **Stale `${ROCM_CLI_ROCDXG_SHA256:-}` comment.** Removed. It described a revision where the digest was read at execution time; nothing generates that expansion now, and a sibling test asserts the variable name is absent from the commands.
- **`ROCM_CLI_ROCDXG_ALLOW_UNVERIFIED` was a presence check.** Fixed — it now goes through the crate's existing `truthy_env` allowlist, so only `1`, `true`, `yes` and `on` opt out. Regression test `wsl_rocdxg_opt_out_reads_negative_values_as_off` covers `0`, `false`, `no`, `off` and empty on the negative side and all four affirmative spellings on the positive side. Verified it discriminates: against the old presence check it fails with `ROCM_CLI_ROCDXG_ALLOW_UNVERIFIED="0" disabled verification`.
- **Nothing documents how a pinned digest is regenerated.** Added beside the table, along with the fact that nothing in the tree checks a row against the published artifact and that a mistyped digest surfaces only as a failed install.
- **Refusal plan rendered the raw rejected version.** Fixed while in the area, since it is the same approval surface: the value now goes through its debug form. Before the fix, `ROCM_CLI_ROCDXG_VERSION=$'1.2.0\nid'` rendered an unindented `id` line into the plan a human reads to decide. The hostile-version test now asserts every rendered line after the header is indented, which is what caught it.

Not fixed, with reasons:

- **`mktemp -d` to close the check-to-install window.** Agreed on the substance, but it rewrites the generated command list and the e2e expectations that read it, and as you note the shape predates this PR. That is a change of its own rather than a fix to what this PR introduces — happy to open an issue for it.
- **Factoring the `printf` fragment into one formatter shared by production and the test.** Same call: a worthwhile refactor, but it moves production code for a test-quality reason and is not something this PR regressed.
- **A comment at the `ChatRocmCommandAction::Approval` construction site.** Fair, and I expect it would prevent the next misread — but that block is untouched by this PR, and adding to it widens the diff into code under no other change here. Better as its own small PR.

## Verification

Run locally: `cargo clippy --locked --workspace --all-targets -- -D warnings` and `cargo clippy --locked -p e2e-cucumber --test e2e -- -D warnings` (both clean), `cargo fmt --check`, `python scripts/smoke_local.py`, and `cargo test --workspace --all-targets` — `apps/rocm`'s 759 binary tests pass, including all 13 `wsl_rocdxg` ones.

Not run here, stated rather than glossed: the WSL-gated e2e scenarios. This host has no `/dev/dxg`, so `is_wsl_host()` does not qualify it and `driver-install-01..03` skip; they need a real WSL2 host with the GPU plumbing present. The GPU and Windows lanes were likewise not run locally. The only local test failures are in `rocm-dash-tui`'s `agent::tests`, which fail with `cannot create token cache dir: Permission denied` — an environment limitation of this machine, unrelated to anything on this branch.

@juhovainio — your approval was at `cfd6654a`, two pushes ago. Since then: the digest-pinning and version-validation work at `28f75d03`, the `@requires-wsl` e2e gating at `ec6b06e7`, and now this round. The behaviour change worth a second look is the opt-out narrowing: `ROCM_CLI_ROCDXG_ALLOW_UNVERIFIED=0` used to skip digest verification and now does not.

rominf:
CI on `eee9059e`: the `CI` and `CodeQL` workflows are fully green — 24 success, 1 skipped (Sphinx), including `clippy`, `Test (affected crates)`, `windows-build-and-test`, `Commit signatures + sign-off`, `License header check (hawkeye)` and `prek`.

Two self-hosted lanes are red and I do not believe either is attributable to this branch: `E2E tests (Strix Halo, Ubuntu)` and `E2E tests (Strix Halo, Windows)` both hit the 45-minute job cap and were `cancelled` with no results artifact written, so there is no scenario-level failure to read. Evidence for not attributing them here:

- `main` itself is red on both lanes at `6ad833f1`, the commit this branch is now rebased onto. That run failed at `bench-06` on Ubuntu, a live-serve benchmark scenario.
- Five unrelated branches whose runs started in the same window show the identical `cancelled` at exactly 45 minutes on both lanes — `gpu-out-of-memory`, `feat/vllm-rocm10-wheel-discovery`, `fix/serialize-env-mutating-tests`, `feat/devel-opt-in` and this one. Runs before that window on the same lanes finished in 18-30 minutes and passed.
- `E2E tests (Strix Halo, WSL2)` — the lane that actually exercises the WSL driver path this PR changes — passed, as did MI300X, MI350P and rad3 R9700.

I have not re-run the two lanes: the pool is one machine per platform and was already backed up behind six queued runs, and with `main` red on the same lanes a re-run here would consume a scarce runner without testing anything this branch introduced. Happy to trigger one if a maintainer would rather see it.

rominf:
The three non-blocking items declined during this PR's review are now tracked, so the deferral is recorded rather than dropped:

- #425 — the digest is verified on a predictable `/tmp` path that the root `apt-get install` then reads. Pre-existing shape, but this PR makes digest verification the documented trust anchor, so the window is worth closing. Deferred here because it rewrites the generated command list and the E2E expectations asserting on it.
- #426 — the digest test executes a hand-built replica of the `printf` fragment rather than the generated command, so a swapped digest/path order would keep its assertions green.
- #427 — a one-line comment at the chat approval construction site noting the full argv travels downstream via `display_command`. This is the reading that produced a withdrawn finding during review.

Two related items raised alongside these were fixed in this PR rather than deferred: the refusal plan now escapes the rejected version, and `ROCM_CLI_ROCDXG_ALLOW_UNVERIFIED` is parsed as a boolean so `=0` no longer disables verification.

siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · eee9059e

## Summary
Replaces the deleted `scripts/wsl_setup_rocdxg.sh` with a real `rocm install driver` plan for WSL2 that pins a SHA-256 per ROCDXG release, verifies the `.deb` by default, validates the version before it reaches a privileged shell command, and threads `reboot_required` per plan; this round fixes the three documentation/test counts from the previous review. **Needs work** — one count: the test that both its own docstring and `docs/testing.md`'s Release trust checks section describe as running the plan's generated digest fragment does not run it, and cannot fail for a field-order regression in that fragment. All three of our earlier objections are DISCHARGED. Verified: derived the merge base myself (`6ad833f1`, which differs from the `prw-base` tip `248c0d92` by four base-branch commits — the diff against both is byte-identical, confirmed with `cmp`); read all six commit objects whole with `git cat-file -p`; traced every value interpolated into the generated shell strings back to its validation point and confirmed `execute_driver_install_plan` propagates a failing digest step with `?` so the install step is never reached; confirmed `truthy_env` is an affirmative allowlist and that `resolve_rocdxg_verification` refuses rather than installing when nothing resolves; confirmed by reading that the digest test's `starts_with`/`ends_with` pair still matches after a digest/path swap, and a scratch-copy mutation run outside the checkout reproduced all 13 `wsl_rocdxg` tests staying green under exactly that swap; ran the contributor-rules leak scan over the full diff (zero hits, no internal references, no prompt-injection content); one cheap targeted check was `cargo clippy -p e2e-cucumber --test e2e` (clean). The full test suite, and the GPU, Windows and WSL lanes, were not run here. The checkout is unmodified (`git status --porcelain` empty). Checks at this head: 24 success, 2 cancelled, 1 skipped. Blocking: 1 · Non-blocking: 5.

## Previously raised

- **`docs/release-trust.md` contradicted the code — DISCHARGED.** Our sentence was: *"the repo's trust-policy document still describes the pre-remediation model, and now directly contradicts the code in the same PR."* Mechanism: a tree-wide grep for `No production checksum` now returns nothing, and I checked the replacement section clause by clause against `ROCDXG_PINNED_DIGESTS` and `resolve_rocdxg_verification` — pinned per release, verified with nothing set, `ROCM_CLI_ROCDXG_SHA256` first in the resolution order, refusal when nothing resolves, opt-out last. It describes what the code does.
- **`docs/testing.md` named a script that does not exist — DISCHARGED.** Our sentence was: *"the newly added line `python scripts/wsl_preflight.py --require-ready` names a script that does not exist at this head."* Mechanism: `grep -rn wsl_preflight` over the whole tree returns zero hits. The three remaining `wsl_setup_rocdxg` hits are all legitimate past-tense references, one of them an e2e assertion that the string is *absent* from CLI output.
- **Three plan tests read the ROCDXG environment without the guard — DISCHARGED.** Our sentence was: *"three of the new unit tests read the ROCDXG environment without the `ScopedTestEnv` guard."* Mechanism: `let _env = scoped_rocdxg_env();` is now the literal first statement of all three, and `wsl_rocdxg_install_does_not_ask_for_a_reboot` now anchors on `rendered.contains("apt-get install -y '/tmp/")` before its three original assertions — a string only a real install plan carries, so the vacuous pass against a refusal plan is closed. Mutation-confirmed: flipping `reboot_required` to `true` on the WSL path fails it.

- `apps/rocm/src/main.rs:31170-31245` (`wsl_rocdxg_generated_digest_step_accepts_only_the_matching_file`) — **the test does not run the step the plan generates, and cannot fail for the regression it says it catches.** Its docstring states *"Runs the digest step the plan actually generates, rather than asserting that it contains some substrings… Substring assertions would let a quoting, field-order or newline regression in the `printf | sha256sum -c -` fragment ship green."* In fact the executable part is a closure that rebuilds the command itself with a hard-coded field order, entirely independent of production; the only tie back to the generated string is `generated.starts_with("printf '%s  %s\\n' '")` and `generated.ends_with("' | sha256sum -c -")` — which is precisely "asserting that it contains some substrings", and which both still hold when the two quoted middle fields are swapped. Swapping them in production to `printf '%s  %s\n' '{deb_path}' '{digest}'` breaks every real WSL install (`sha256sum -c -` reads `DIGEST  FILENAME`) and leaves all 13 `wsl_rocdxg` tests green — reproduced in a scratch copy. This blocks rather than being a nit because it is the plan's stated trust anchor, it replaced a deleted executable self-test, and `docs/testing.md:788-800` advertises it in the **Release trust checks** list with the same false claim: *"The ROCDXG digest check runs the shell fragment the WSL driver plan generates against a real file."* Fix (test-only, no production movement): assert full equality of `generated` against `format!("printf '%s  %s\\n' '{pinned}' '/tmp/rocdxg-roct_{version}_amd64.deb' | sha256sum -c -")`, or substitute the payload path into `generated` and execute *that* string rather than the replica. Noting plainly that we graded this non-blocking last round and it was deferred to a tracked issue: the deferral reason given was that a shared formatter would move production code for a test-quality reason, and that reason does not apply to either fix above — both are confined to the test. Under the rule that a test must actually test what its name and docs claim, it is a count.

## Non-blocking
- `apps/rocm/src/main.rs:30863-30892` — `wsl_install_driver_installs_rocdxg_without_dkms` asserts `plan.supported` and `policy == "wsl_rocdxg"` on a plan that now reads three ROCDXG env vars, but its comment claims it "has to take the same guard as the mutating tests in this binary" while it takes only `with_amd_overrides_cleared()`, which clears the two AMDGPU vars. `ScopedTestEnv`'s process-wide lock rules out concurrent interference, so the residual exposure is an externally exported `ROCM_CLI_ROCDXG_VERSION` and the failure is loud rather than silent — but the comment is false and the fix is one word: `scoped_rocdxg_env()`.
- No test pins the `apt-get install` → `ldconfig` ordering; running `ldconfig` before the package exists never publishes the library, and all 13 tests stay green under that swap. `wsl_rocdxg_plan_installs_the_library_and_publishes_it` is the natural home for an index comparison.
- `docs/release-trust.md:127-130` says the opt-out "takes the usual affirmative values (`1`, `true`, `yes`, `on`); anything else … leaves verification on", but `truthy_env` also accepts `TRUE`, `YES` and `ON`. List all seven, or drop the exhaustive framing.
- `docs/wsl.md` — the "Doing it by hand" fallback reproduces the download and `apt install` with no digest step, immediately after the surrounding text makes digest verification the trust anchor. Adding the `sha256sum -c` line (1.2.0 is in the pinned table) would keep the manual path consistent with the documented model.
- `ROCDXG_VERSION_EXPR` defaults to `1.2.0` while the pinned table carries `1.2.1` and `1.2.2`; worth confirming the default is deliberately behind the newest pin.

Not flagged, for the record: the change does not vacate any prior guard. The only removed behaviour is the WSL refusal that pointed at a script shipped solely in a git checkout, and the removed `reboot_required || executed_at_unix_ms.is_some()` override in reconciliation, which would have wrongly forced a reboot on the userspace-only WSL path. The `--yes` gate, the plumbing preflights and the approval flow are unchanged. The new e2e scenarios gate on `@requires-wsl`, which fails closed to Skip off a real `rocm examine` reading of `/dev/dxg` and `/proc/version` — they do run on a self-hosted WSL2 lane, but that lane is documented as non-blocking, so on an ordinary PR the merge-gating coverage for this plan is the unit tests alone. I did not verify the three referenced follow-up issues, as this review does not contact GitHub.

rominf:
All six items from the automated review at `eee9059e` are addressed in `2486c5a8`.

## Blocking — the digest test now runs the fragment the plan generates

The finding was right: the executable half was a replica that rebuilt the command with its own field order, and the only tie back to production was a `starts_with`/`ends_with` pair that survives swapping the two quoted fields. Both halves are now real:

- the generated command is pinned whole with `assert_eq!` against `printf '%s  %s\n' '{pinned}' '/tmp/rocdxg-roct_{version}_amd64.deb' | sha256sum -c -`, so quoting, the two-space separator and field order are all load-bearing;
- the string handed to `sh` is that *same generated command* with only the digest and the `.deb` path substituted at the test payload, instead of a reconstruction.

Mutation probe, run both ways round:

- swapping production to `printf '%s  %s\n' '{deb_path}' '{digest}'` fails the `assert_eq!` with both strings printed;
- with the `assert_eq!` temporarily deleted — to confirm the executable half is not merely riding on it — the same swap fails on `the matching digest must pass`, because `sha256sum -c -` then reads the payload path as the digest;
- restored, all 13 `wsl_rocdxg` tests pass.

The docstring and the **Release trust checks** entry in `docs/testing.md` were rewritten to describe what the test does rather than the old claim.

This also closes #426. The deferral reason recorded there was that a shared formatter would move production code for a test-quality reason; neither change here touches production, so that reason no longer applies.

## Default ROCDXG release moved to 1.2.2

`ROCDXG_VERSION_EXPR` was on 1.2.0 (May) while 1.2.2 (August, a normal release) had been pinned in the digest table since this PR added it. The default is now 1.2.2. All seven pinned digests were checked against the published `.deb` artifacts by downloading each one and comparing sha256 — all seven match, 1.2.2 included. That verification was done by the PR author, not by me in this round.

The constant now carries the rule the reviewer noted was missing: the default is the newest non-prerelease release whose `rocdxg-roct` digest is pinned, and the table row and the literal must move together — a default with no row makes `resolve_rocdxg_verification` refuse to build a plan at all, so a half-done bump breaks every default install rather than falling back.

**The extra `rocdxg-amd-smi-lib` asset is deliberately out of scope**, and the constant now says so rather than leaving it unremarked. Releases carry it from v1.2.1 onward (v1.2.0 and older ship only `rocdxg-roct`). Inspecting `rocdxg-amd-smi-lib_1.2.2_amd64.deb`: it unpacks into a second prefix at `/opt/rocm-wsl` with its own `amd-smi` and `libamd_smi.so`, depends on `python3`/`python3-pip`/`setuptools`/`wheel`, runs a 254-line `postinst`, and drops an `/etc/profile.d` snippet that prepends `/opt/rocm-wsl/bin` to `PATH` and `/opt/rocm-wsl/lib` to `LD_LIBRARY_PATH` for every new login shell — shadowing the ROCm-provided tooling system-wide. Nothing this plan verifies depends on it: `wsl_rocdxg_ready` keys on `/opt/rocm/lib/librocdxg.so` and its visibility to the linker, and `rocm serve` gates on the same. Bundling a system-wide environment change into the plan that installs the runtime bridge is a separate decision that deserves its own opt-in, so it stays out.

Nothing in `tests/e2e-cucumber/` pins a version; the four unit tests that hard-coded `1.2.0` as *the default* were updated, and the hostile-input strings in `wsl_rocdxg_refuses_a_version_that_could_escape_the_shell` were left alone since the prefix there is arbitrary.

## Non-blocking

1. **`wsl_install_driver_installs_rocdxg_without_dkms` guard** — now takes `scoped_rocdxg_env()`. Rather than trading one half of the guard for the other, `scoped_rocdxg_env()` now builds on `ScopedTestEnv::with_amd_overrides_cleared()`, so it clears all five vars: a WSL plan reached through `build_driver_install_plan` resolves the AMDGPU overrides before it dispatches to the WSL branch, so a caller needing either guard needs both. The comment now says what the guard actually covers.

2. **`apt-get install` → `ldconfig` ordering** is pinned in `wsl_rocdxg_plan_installs_the_library_and_publishes_it` as a position comparison. Mutation probe: swapping the two `driver_command` entries fails it with the full command list printed; every `contains` assertion in the file stays green under that swap, which is the point.

3. **`docs/release-trust.md`** now lists all seven spellings `truthy_env` accepts (`1`, `true`, `TRUE`, `yes`, `YES`, `on`, `ON`) and notes the value is trimmed first. The same four-value list in the `resolve_rocdxg_verification` doc comment was pointing at the same incomplete set, so it now refers to `truthy_env` for the authoritative list instead of restating it — one fewer copy to drift.

4. **`docs/wsl.md` manual fallback** now carries the `sha256sum -c` step between the download and `apt install`, with 1.2.2's pinned digest, so the by-hand path matches the trust model the section above it describes.

## Verification

`cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo clippy -p e2e-cucumber --test e2e -- -D warnings` (sources touched first, so the cache could not hide anything), `cargo test --workspace --all-targets --no-fail-fast`, and `python scripts/smoke_local.py` all pass. The only test failures are three pre-existing `rocm-dash-tui` agent tests that cannot create a token cache directory under this sandbox; they are untouched by this diff and fail the same way without it.

Not run here: the `@requires-wsl` e2e scenarios. This host has no `/dev/dxg`, so they fail closed to Skip as designed — the Strix Halo WSL2 lane is where they actually execute.

CI at this head: all 16 required checks are green. `E2E tests (Strix Halo, Ubuntu)` is red, and it is advisory: it failed in *GPU preflight (bounded wait for an available GPU)* before the toolchain step, so nothing from this change was ever built or run on it.

rominf:
Addendum on the two advisory lanes, since neither will show a result and I would rather say why than leave them unexplained.

`E2E tests (Strix Halo, WSL2)` — the lane that actually executes the `@requires-wsl` scenarios — has been queued since the run started and will not be picked up: the `strix-halo-wsl` runner is currently offline, so there is no machine carrying that label. It passed on the previous head `eee9059e`, and nothing in this delta touches the e2e feature files or step definitions.

`E2E tests (Strix Halo, Ubuntu)` is red because it failed in *GPU preflight (bounded wait for an available GPU)*, three steps in and before the toolchain was even installed, so nothing from this change was built or run there. The `strix-halo-ubuntu` runner is also offline; the job landed on `strix-halo-ubuntu-2` and could not get a free GPU.

Both lanes are advisory. All 16 required checks are green at `2486c5a8`.

juhovainio (approved):
Reviewed this PR in full: main.rs, therock.rs, diagnose.rs, fix.rs, the docs, the deleted script, and the new e2e/unit tests.

This is a solid piece of work. A few things worth calling out that I specifically checked:

- The old unquoted `ROCM_CLI_ROCDXG_VERSION` interpolation into `sh -c` (a real root shell injection risk, since it ran after `apt-get update` had primed the sudo credential cache) is now closed off by `rocdxg_version_is_well_formed()`, an ASCII alphanumeric + `. + - ~` allowlist. I tried to find a bypass (trailing whitespace, embedded newline, `$()`, backticks) and couldn't; your own test `wsl_rocdxg_refuses_a_version_that_could_escape_the_shell` already covers these.
- Digest verification is fail-closed: env override (validated 64-hex) then pinned table then an explicit, strict opt-out, then refusal. No path silently skips verification. All 7 pinned digests are well-formed.
- The new `reboot_required` field is set correctly at every plan construction site, and the Verify-phase auto-execution change only kicks in for non-reboot plans, exactly as intended. Dry-run and unapproved plans still short-circuit before any execution, so nothing new can run without explicit `--yes`.
- The deleted script's guards (dxg/dxcore checks, sudo guard, post-install checks, checksum self-test) are all faithfully reimplemented, and the checksum check is actually stricter now (fail-closed vs. the old script's opt-in check). No dangling references to the deleted script remain in live code.

Nice attention to detail on the deferred-verify-vs-reboot distinction, that's an easy thing to get wrong and the tests around it are thorough.

This reflects a diff-level read, not CI or a real WSL2/GPU run. Would like to see this pass green on the E2E tests (Strix Halo, WSL2) lane before merge, since that's the only place the new `@requires-wsl` scenarios actually run.

rominf:
Rebased onto `main` now that #403 has landed, and added one commit on top — `2f7c22d6`, which is **new since your approval** and touches the lane you just added.

The rebase itself was clean, but it left both WSL lanes calling `scripts/wsl_setup_rocdxg.sh`, which this branch deletes. Git reports no conflict because the two sides never touch the same file. And because that step is advisory (`if bash …; else warn`), it wouldn't have failed the job — it would have quietly warned on every run.

Rather than repoint it at `rocm install driver`, I dropped the step from `e2e-wsl` and `e2e-wsl-nightly` entirely, per your suggestion. The `@requires-wsl` scenarios here assert on the plan that command produces, so a lane that pre-installed ROCDXG would hand them a host already in the state the command under test exists to reach. It costs no coverage today: `@requires-gpu` scenarios already resolve to skip on that lane because `rocm-smi` is absent too. The absence is commented in both lanes so it doesn't read as an oversight and get restored later.

The two `docs/ci-hardware-testing.md` paragraphs that described the removed step are updated to match.

Verified locally: `cargo fmt --check`; `cargo clippy --workspace --all-targets -- -D warnings`; `cargo test -p xtask` (151 passed, including every workflow-contract test); `cargo test -p rocm -p rocm-core` (1170 passed); the e2e `feature_naming` guard; and `cargo check -p e2e-cucumber --test e2e`, since `test = false` means a plain `cargo check` skips that target.

The real test is the WSL lane itself, which should now run the three `@requires-wsl` driver scenarios for the first time — they've never had a host to run on before.

dhekimian:
Heads-up on the Strix Halo Ubuntu: GPU preflight fails every run with "0 GiB free". We moved the fleet yesterday to a 512 MB carve-out, so `rocm-smi --showmeminfo vram` reports 512 MiB total even though there's more ram available.

Happy to help if you need anything on the pool side.

PR #235, opened 2026-08-12
ci(dependabot): track pre-commit hook versions

rominf:
## Summary

Adds a `pre-commit` entry to `.github/dependabot.yml` so the hook pins in
`.pre-commit-config.yaml` stop drifting.

Those `rev`s (currently `ruff-pre-commit` and `shellcheck-py`) have only ever
been bumped by hand, which means the linters CI enforces are whatever version
someone last remembered to update. Dependabot [added a `pre-commit`
ecosystem](https://github.blog/changelog/2026-03-10-dependabot-now-supports-pre-commit-hooks/)
that parses the config, resolves each `rev` against the hook repository's tags,
and opens a PR when one moves.

Notably it skips the `repo: builtin` and `repo: local` blocks, so prek's native
hooks and the `cargo xtask` hooks are untouched — they have no upstream release
to track.

Weekly, grouped into a single PR, with the same `cooldown: 7` as the other two
ecosystems. The cooldown reasoning is a bit sharper here than for actions or
crates: a hook is executable code that runs on every contributor's machine at
commit time and in the `prek` CI job, so a compromised release executes before
anyone has read the bump. Cooldown never delays security updates.

Risk: low. Config-only; the worst case is that no PRs are opened.

## Test plan

- `prek run --all-files --no-group local-tools` passes (`check-yaml` parses the
  file).
- Not verifiable before merge: whether GitHub accepts `pre-commit` without
  `enable-beta-ecosystems`. The changelog and the supported-ecosystems reference
  both list it as generally available, and the flag is documented as beta-only,
  so it should not be needed — but the authoritative signal is the repo's
  Dependabot tab after this lands. If it is rejected there, the fix is a
  one-line follow-up.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -56,3 +56,24 @@ updates:
         update-types:
           - minor
           - patch
+
+  # The hook pins in `.pre-commit-config.yaml` were bumped by hand, so they
+  # drifted. Dependabot resolves each `rev` against the hook repository's tags
+  # and leaves the `builtin` and `local` blocks alone -- prek's native hooks and
+  # the `cargo xtask` hooks have no upstream release to track. Grouped into one
+  # PR: this is a short list of linters whose individual bumps carry no review
+  # signal worth a PR each.
+  - package-ecosystem: pre-commit
+    directory: /
+    schedule:
+      interval: weekly
+    # Same reasoning as the two blocks above, and it lands harder here: a hook is
+    # executable code that runs on every contributor's machine at commit time and
+    # in the `prek` CI job, so a compromised release executes before anyone has
+    # read the bump. Cooldown delays only version updates, never security updates.
+    cooldown:
+      default-days: 7
+    groups:
+      pre-commit:
+        patterns:
+          - "*"


fredespi:
I reviewed this change (diff against `main`, base `e95e8d5e`). It's a config-only addition of a fourth Dependabot ecosystem block — no code paths touched.

I checked the substantive claims against the tree and they hold up:

- `.pre-commit-config.yaml` does have a `repo: builtin` block and three `repo: local` blocks, plus the two remote hooks (`ruff-pre-commit`, `shellcheck-py`). Dependabot's pre-commit ecosystem tracks the remote `rev`s and leaves `builtin`/`local` alone, matching the comment's rationale.
- The new block mirrors the existing `github-actions` and `cargo` blocks — `directory: /`, weekly schedule, `cooldown.default-days: 7`, and a `groups` glob — so it's consistent with the file's conventions and valid Dependabot v2 schema.

The one genuine unknown is the one the description already flags: whether GitHub accepts `pre-commit` without `enable-beta-ecosystems`. That's not verifiable from the diff, and the failure mode is benign (no PRs open), not a config error — so it isn't a blocking issue here.

Nothing surfaced that would block merge. Looks safe to approve pending final validation (CI and the Dependabot tab confirming the ecosystem is accepted after this lands). This reflects a diff-only read, not CI results, and isn't itself an approval.

rominf:
Thanks for the review — nothing to change based on it. Re-requesting since a formal approving review is needed to satisfy the code-owner requirement for merge (a comment-only review doesn't clear it).

michaelroy-amd (approved):
Reviewed at `2c86e6eeb8be1c0faf34d1aba9235f015caba870`. This is a focused Dependabot v2 configuration addition: the pre-commit ecosystem tracks the two remote hook revisions while leaving `repo: builtin` and `repo: local` entries untouched. Its weekly schedule, seven-day cooldown, and grouped pattern match the repository's existing dependency-update policy. Required checks have no failures (16 success, 1 expected skipped report), commit-signoff validation is green, and the change has no user-visible CLI path requiring Cucumber coverage. No blocking finding. Approving.

PR #236, opened 2026-08-12
fix(install): stop relying on RuntimeInformation.OSArchitecture

juhovainio:
## Summary
- `RuntimeInformation.OSArchitecture` can return `$null` on some Windows PowerShell 5.1 / .NET Framework setups (e.g. a locked-down corporate image missing a facade assembly), which made `install.ps1` throw `You cannot call a method on a null-valued expression` before printing anything useful.
- Switch to `PROCESSOR_ARCHITEW6432` (falling back to `PROCESSOR_ARCHITECTURE`) — plain env vars Windows always sets, no .NET type resolution involved. This also preserves the original intent of detecting the true OS architecture rather than the running process's bitness (the WOW64 case).
- Root-caused from a real user's `ScriptStackTrace` pointing at the failing line; confirmed unrelated to their corporate proxy or PowerShell language mode, both of which were ruled out first.

## Test plan
- [x] Verify `install.ps1` completes past architecture detection on a real x86_64 Windows machine (originally reported failing on a tester's personal machine) -> Verified to be working
- [x] Confirm `$env:PROCESSOR_ARCHITEW6432` / `$env:PROCESSOR_ARCHITECTURE` resolve to `AMD64` on standard x86_64 Windows and installer proceeds normally

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -575,9 +575,12 @@ if ([string]::IsNullOrWhiteSpace($InstallDir)) {
 }
 $InstallDir = Resolve-InstallerPath $InstallDir
 
-$architecture = [System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture.ToString()
+$architecture = $env:PROCESSOR_ARCHITEW6432
+if ([string]::IsNullOrWhiteSpace($architecture)) {
+    $architecture = $env:PROCESSOR_ARCHITECTURE
+}
 switch ($architecture) {
-    "X64" { $platformArch = "amd64" }
+    "AMD64" { $platformArch = "amd64" }
     default { Fail "unsupported architecture: $architecture (installer currently supports Windows x86_64 only)" }
 }
 


fredespi:
Reviewed at `fa86585` (one file, +5/-2). No change requests — the fix is sound and correctly scoped. Notes below, then a recommendation.

**The diagnosis holds up.** `System.Runtime.InteropServices.RuntimeInformation` is a facade type on .NET Framework, so "missing on a locked-down corporate image" is a real failure mode rather than a guess, and `PROCESSOR_ARCHITECTURE` is set by Windows itself with no type resolution involved. Checking `PROCESSOR_ARCHITEW6432` first preserves the original intent — a 32-bit PowerShell on 64-bit Windows still reports the OS architecture rather than the process bitness, which is the case `OSArchitecture` was presumably chosen for in the first place.

**The fix is complete.** No other `RuntimeInformation` or `OSArchitecture` use survives anywhere in the tree, and nothing else consumed the old `"X64"` string, so there's no second call site waiting to throw the same way. Worth noting the other `[System.*]` calls in this script (`IO.Path`, `Cryptography`, `Convert`) are core BCL types and don't carry the same facade risk, so the narrow fix is the right scope rather than an incomplete one.

**One thing I checked because it would have been nasty.** With the old code `$architecture` was always a non-empty string; now it can be `$null` if both env vars are absent. I expected to find the classic PowerShell trap where `switch` skips a `$null` condition entirely — which would have meant no `default`, no `Fail`, and an unset `$platformArch` interpolated into the asset name as an empty string, turning a clear error into a 404 on a malformed download URL. That turns out not to happen: `switch ($null)` does execute `default`, and matching is case-insensitive so a lowercase `amd64` would work too. Verified in a PowerShell container rather than reasoned about:

```
null      -> default
empty     -> default
lowercase -> matched
```

Caveat on that: I tested pwsh 7, not the Windows PowerShell 5.1 this bug was reported against. If you want certainty on 5.1 specifically it's a one-liner on any Windows box, but I'd be surprised by a difference here.

**One question rather than a change request.** On ARM64 Windows running x64-emulated PowerShell, `PROCESSOR_ARCHITECTURE` reports `AMD64`. If `PROCESSOR_ARCHITEW6432` isn't populated with `ARM64` under that emulation, this would now install the amd64 build where the old `OSArchitecture` check failed cleanly with "unsupported architecture". Small population, and the amd64 binary would run under emulation anyway, so I don't think it blocks anything — but you have the Windows hardware to confirm it and I don't, and it's the one behaviour the switch away from `OSArchitecture` could plausibly change.

**Recommendation: ready to approve on the code**, with two mechanical gates outstanding that aren't review objections — `windows-build-and-test` is still running (18 checks green, none failing), and the branch is `BEHIND` main so it needs updating before it can merge. Given the fix is specifically about Windows behaviour, that last check finishing green is the one worth waiting for. This is also currently the only review on the PR.

fredespi:
Closing the caveat from my last comment: I've now run this on real Windows PowerShell 5.1 (Desktop edition, `5.1.26100.9168`) on a Windows 11 x86_64 box, rather than the pwsh 7 container I used before. Both concerns resolve in your favour.

**`switch` semantics on 5.1** — same as pwsh 7, so the `Fail` path is safe if both env vars are ever absent:

```
switch($null)   : default
switch('')      : default
switch('amd64') : matched     (case-insensitive, as expected)
```

**The WOW64 case, which is the reason the `PROCESSOR_ARCHITEW6432`-first ordering exists.** Running the identical script under 64-bit and then 32-bit PowerShell on the same machine:

| | 64-bit PS | 32-bit PS (WOW64) |
| --- | --- | --- |
| `PROCESSOR_ARCHITECTURE` | `AMD64` | `x86` |
| `PROCESSOR_ARCHITEW6432` | *(empty)* | `AMD64` |
| `OSArchitecture` (old code) | `X64` | `X64` |
| **PR logic resolves to** | `amd64` | `amd64` |

So the new code returns `amd64` in both cases, matching the old `OSArchitecture` behaviour exactly — including the WOW64 case, where a naive `PROCESSOR_ARCHITECTURE`-only check would have read `x86` and failed the install with "unsupported architecture". The fallback ordering is doing real work, not just belt-and-braces.

For completeness: `RuntimeInformation::OSArchitecture` resolved fine on this host, which is expected — it's a healthy Windows 11 install, not the locked-down image with the missing facade assembly, so this doesn't contradict the bug report.

That leaves only the ARM64-emulation question from my previous comment, which needs ARM64 hardware neither of us has pointed at this yet, and the two mechanical gates (the last check finishing, and the branch being behind main). My recommendation stands and is now better evidenced: **ready to approve on the code.**

fredespi (approved):

PR #242, opened 2026-08-12
fix(lemonade): find llama-server under nested backend dirs (EAI-8026)

juhovainio:
## Summary
- `rocm serve` failed with "no GPU llama-server backend is installed" on Strix Halo (gfx1151) even right after a successful `rocm engines install lemonade`, because Lemonade's installer sometimes extracts llama.cpp one directory deeper than expected (`rocm-stable/llama-b9752/llama-server` instead of `rocm-stable/llama-server`).
- `find_llama_server_binary` now also checks one level of subdirectories under each backend directory.
- `llama_server_backend_label` now walks ancestor directories to recover the correct backend name (`rocm-stable`, `vulkan`, etc.) even when nested.
- Added an E2E canary (`serve-hf-checkpoint-inference`) that serves an `owner/repo:variant` HF checkpoint on lemonade, exercising the `serve_hf_checkpoint` bail path directly instead of through the short-recipe-name router, whose managed-serve fallback is masked by the unrelated EAI-7423 xfail on Linux.

## Test plan
- [x] `cargo test -p rocm-engine-lemonade find_llama_server_binary` (includes new regression test reproducing the exact nested layout observed on Strix Halo)
- [x] Full `rocm-engine-lemonade` test suite passes (65/65)
- [x] `cargo clippy --locked --workspace --all-targets -- -D warnings` passes
- [x] `cargo check -p e2e-cucumber --tests` passes with the new scenario
- [x] Validate new build on Strix Halo (gfx1151) machine using `rocm serve`

### DIFF
--- a/engines/lemonade/src/lib.rs
+++ b/engines/lemonade/src/lib.rs
@@ -2619,7 +2619,7 @@ fn find_binary_in(dir: &Path, binary: &str) -> Option<PathBuf> {
     }
     fs::read_dir(dir)
         .ok()?
-        .filter_map(|entry| entry.ok())
+        .filter_map(Result::ok)
         .map(|entry| entry.path())
         .filter(|path| path.is_dir())
         .map(|subdir| subdir.join(binary))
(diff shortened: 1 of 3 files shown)


juhovainio:
FYI for reviewer: this fix has been manually validated on a Strix Halo (gfx1151) machine

michaelroy-amd (approved):
Approved. The nested lookup fixes the observed one-level Lemonade layout, preserves GPU-backend priority/no-CPU-fallback behavior, and adds unit plus Gherkin coverage. Local targeted tests, affected clippy/checks, workspace build/clippy, smoke, and all live CI checks passed.\n\nNon-blocking follow-up: engines/lemonade/src/lib.rs:2620-2626 uses fs::read_dir order when multiple build-numbered directories exist under one backend. That order is unspecified, so an upgrade that leaves two versions could select an older binary. Please define a deterministic candidate policy or document/enforce the single-build-directory invariant.\n\nDiagnostic nit: tests/e2e-cucumber/tests/e2e/serving_steps.rs:531-535 discards stderr on launch failure; including it would make this canary's failures more actionable.

PR #243, opened 2026-08-12
Make model serving ports collision-safe

### LINKED ISSUE #239: Make model serving collision-safe and simplify the Dash serve wizard
## Problem Statement

The Dash serve wizard exposes engine, device policy, host, port, and mode alongside the model selector even though ordinary users only need to choose a model recipe. The controls do not communicate their interaction model: Left/Right changes some rows but silently does nothing on text rows, and typing into the populated port appends to the default rather than replacing it.

Every ordinary serve launch also defaults to port 11435. A second model can therefore be launched against an occupied port and only fail later in engine startup. The delayed failure is difficult to understand and can leave a misleading service attempt while the user only wanted to run another model.

## Solution

Make the ordinary Dash flow model-first. The collapsed wizard shows the model recipe picker, one inline `Advanced settings` disclosure row, and the review/launch action. Engine, GPU-required device selection, loopback host, collision-safe automatic port selection, and managed mode remain automatic unless the user expands Advanced settings.

When expanded, Advanced settings remains part of the same wizard rather than opening another modal. Choice fields visibly use cycling controls. Editable fields visibly use text controls and enter an explicit edit state where typing replaces the selected default, Left/Right moves the cursor, Enter accepts, and Escape cancels.

Move port policy into the central `rocm serve` path so Dash, direct CLI use, assistant tool calls, foreground launches, and managed launches share one rule. Omitting `--port` selects an available port beginning with the established local default. Supplying `--port` preserves that exact request and fails early with an actionable collision message when it is occupied. Automatic launches may choose another available port; explicit launches must never be silently moved.

## User Stories

1. As a first-time user, I want the serve wizard to initially ask only for a model recipe, so that I can launch a model without understanding network or engine settings.
2. As a basic user, I want engine selection to be automatic, so that the chosen recipe runs on a supported engine without manual configuration.
3. As a basic user, I want device selection to remain GPU-required and automatic, so that serving never silently falls back to CPU.
4. As a basic user, I want the host to default to loopback, so that my server is local and safe by default.
5. As a basic user, I want the port to be chosen automatically, so that I can launch multiple models without manually coordinating ports.
6. As a basic user, I want managed mode to remain the default, so that launched services appear in service management and dashboard telemetry.
7. As a user, I want Advanced settings to expand inline, so that I retain context and do not have to navigate another modal.
8. As a user, I want Advanced settings to collapse inline, so that the common path remains compact after I inspect or change an option.
9. As a user, I want collapsed automatic settings to be labelled `Automatic`, so that I know defaults will be resolved for me.
10. As a user, I want collapsed overridden settings to be labelled `Customized`, so that hidden overrides are not mistaken for defaults.
11. As a user, I want my advanced values preserved when the section is collapsed, so that inspecting the simple view does not discard my choices.
12. As a keyboard user, I want choice rows to display chevrons, so that I know Left/Right will cycle them.
13. As a keyboard user, I want editable rows to display brackets, so that I know they accept text rather than Left/Right cycling.
14. As a keyboard user, I want Enter on an editable value to start editing, so that navigation and text entry are distinct modes.
15. As a keyboard user, I want the existing editable value selected when editing begins, so that typing replaces the default instead of appending to it.
16. As a keyboard user, I want Left/Right to move the cursor while editing, so that I can correct a value without erasing the entire field.
17. As a keyboard user, I want Enter to accept an edit and Escape to cancel it, so that editing has predictable commit and rollback behavior.
18. As a keyboard user, I want the footer to describe the focused control, so that instructions never advertise keys that do nothing on that row.
19. As a user, I want Port to begin as `Auto`, so that the normal path avoids collisions without exposing a number.
20. As an advanced user, I want to select `Custom` for Port and edit the number inline, so that I can integrate with tools that require a fixed endpoint.
21. As an advanced user, I want an invalid custom port rejected in the form, so that I can correct it before approval.
22. As an advanced user, I want an occupied custom port rejected with a plain explanation, so that I know what to change.
23. As an advanced user, I want the collision message to recommend Auto or another port, so that recovery is obvious.
24. As a CLI user, I want `rocm serve MODEL` to choose an available port, so that CLI and Dash behavior remain consistent.
25. As a CLI user, I want `rocm serve MODEL --port N` to request exactly N, so that scripts retain deterministic explicit-port behavior.
26. As a CLI user, I want an occupied explicit port to fail before an engine is launched, so that no misleading service record or long readiness wait is created.
27. As a user launching a second model, I want automatic selection to skip ports already owned by live managed services, so that both models can run concurrently.
28. As a user launching beside an unrelated local application, I want automatic selection to skip ports occupied outside ROCm CLI, so that that application is not disrupted.
29. As a user repeating an equivalent managed launch, I want the existing live engine-and-model service returned instead of a duplicate process, so that idempotent behavior is preserved regardless of its selected port.
30. As a user, I want stale or stopped service history not to reserve a port, so that ports become reusable after a service ends.
31. As a user, I want the approval card to say that the port is automatic and the launch output to show the actual resolved endpoint, so that the review is honest and I can connect clients after allocation.
32. As a user, I want automatic selection to skip ports already occupied when the launch transaction runs, so that ordinary multi-model and local-application collisions are prevented before engine startup.
33. As a user who explicitly selected a port, I want a bind race to fail rather than move my service silently, so that the endpoint contract is honored.
34. As a user on Windows, Linux, or WSL, I want the same collision policy and plain error semantics, so that serving behaves consistently on supported platforms.
35. As a user of local chat and telemetry, I want dynamically selected endpoints to continue coming from the managed service registry, so that chat discovery and metrics follow the actual service rather than a hardcoded default.
36. As a user launching on a public host, I want existing public-bind approval and authentication protections preserved, so that automatic port selection does not weaken endpoint security.
37. As an operator, I want service restart and recovery to reuse the recorded resolved port, so that an existing service's endpoint does not change unexpectedly.
38. As an operator, I want logs and audit records to name the resolved endpoint and collision reason, so that failures can be diagnosed without guessing.

## Implementation Decisions

- `rocm serve` is the single owner of automatic versus explicit port policy. Dash omits `--port` for Auto and includes it only for Custom. Auto is available only with the canonical default host `127.0.0.1`; custom, IPv6, hostname, wildcard, and public hosts require an explicit port so address-family semantics remain unambiguous.
- The established local model-server port remains the first automatic candidate. Automatic selection advances through a bounded local candidate range and returns a clear exhaustion error when no candidate is available.
- Explicit `--port` remains a valid stable CLI contract and is never rewritten. Proactive OS collision preflight applies to explicit ports on canonical `127.0.0.1`; advanced custom-host endpoints retain the existing engine-owned bind/error behavior.
- Canonical-loopback port availability considers both live managed-service records and the operating-system bind result. Stopped or stale records do not block reuse.
- Equivalent managed engine-and-canonical-model launches remain idempotent and return the existing live service before creating another process.
- A shared, bounded cross-process allocation lock serializes new CLI launches and daemon/manual recovery while they refresh liveness, detect equivalent services, reserve a canonical-loopback candidate, confirm immediate child startup, and publish a live record. Recovery leases its recorded canonical-loopback port before publishing `recovering`; an occupied port leaves the prior status unchanged and spawns nothing.
- Automatic selection uses a temporary IPv4 loopback socket lease to reject ports already bound when the transaction runs. Other bind failures stop immediately with the operating-system error. Because third-party engines do not accept inherited listeners, an unrelated process can still race the final release/spawn handoff; that rare engine failure is reported honestly and never silently changes an explicit or already-reported endpoint.
- A failed collision check or exhausted automatic range must not create a service manifest, key file, supervisor process, or readiness wait.
- The resolved port is persisted in the existing managed service record. Restart, recovery, chat discovery, telemetry collection, summaries, logs, and endpoint authentication continue consuming that record rather than recalculating a port. The approval card says `Port: automatic; endpoint shown after launch`; only post-allocation launch output reports the concrete endpoint.
- The default wizard field order is Model, Advanced settings, Review and launch.
- Expanding Advanced settings inserts Engine, Device, Host, Port, and Mode into the same field list. It does not create another modal.
- Automatic Engine means the CLI resolves from explicit configuration, host GPU preference, recipe preference, and platform default using the existing precedence.
- Automatic Device preserves the existing GPU-required serving invariant. This feature does not add or encourage CPU fallback.
- Automatic Host is exact `127.0.0.1`. Automatic Mode is managed.
- Port has two choice states: Auto and Custom. Custom exposes the numeric editor on the same row. A noncanonical loopback Host with Port Auto is invalid: review/launch remains blocked and the inline fix says to set Port to Custom. Dash supports loopback hosts only; public/wildcard/other-address Hosts are rejected before approval with guidance to use the CLI's explicit `--allow-public-bind` flow, preserving endpoint-authentication review.
- Choice rows render visible cycling affordances; text rows render visible edit affordances.
- Text editing is an explicit state. Enter begins editing with the current value selected, printable input replaces the selected value, Left/Right moves the cursor, Backspace edits, Enter accepts, and Escape restores the pre-edit value.
- The footer is derived from the focused field and edit state rather than showing one generic instruction string.
- Collapsing Advanced settings preserves values. Its summary reports `Automatic` only when every advanced field remains at its automatic default; otherwise it reports `Customized` and may include the custom port.
- The existing model recipe picker, local-path browser, approval gate, job bridge, live job console, duplicate-job guard, service registry, authentication policy, and engine protocol remain the integration surfaces.
- Public hosts continue to require the existing explicit public-bind permission and endpoint authentication. Dash does not expose that security-sensitive override; it directs advanced public binds to the CLI. Port automation does not bypass those checks.
- Existing direct engine adapter commands keep concrete port arguments; only the user-facing `rocm serve` command gains omitted-port automatic selection.
- User-facing copy uses `model`, `server`, `automatic`, `local`, and `port`, avoiding internal runtime or adapter jargon on the collapsed screen.

## Testing Decisions

- Tests defend observable behavior rather than struct layout, private helper names, or source text.
- The highest port-policy seam is the user-facing serve argument and canonical-loopback launch-policy boundary: omitted port selects an available endpoint, explicit occupied loopback port is rejected, explicit free loopback port is retained, stopped records do not block reuse, and equivalent live services remain idempotent.
- Port-policy tests use real IPv4 loopback listeners, injected ordered candidates/errors, bounded separate-process lock contention, and deterministic barriers to cover occupied/free conditions and concurrent CLI/recovery allocations without requiring a GPU or live inference engine.
- Automatic-selection tests cover the established default being free, the default being occupied, multiple consecutive collisions, exact-range exhaustion, non-address-in-use failure classification, concurrent equivalent launches converging, concurrent different-model launches receiving distinct ports, lock timeout/process-death release, and recovery preserving its recorded port.
- Collision tests assert that preflight failure occurs before manifest creation or process spawn at the launch-policy seam; publication-failure tests assert any spawned child is terminated and the live reservation is cleaned.
- Dash tests use the existing wizard event-handler, argument-building, approval, job-bridge, and rendered-buffer seams.
- Dash behavior tests cover collapsed default rows, inline expansion/collapse, focus order in each state, preservation of overrides, Automatic/Customized summary text, visible field affordances, and context-sensitive footer hints.
- Editing tests cover selected-on-entry replacement, cursor movement, insertion, backspace, accept, cancel, host trimming, digits-only custom port entry, and invalid port validation.
- Argument tests assert that Auto omits `--host` and `--port`, Custom emits the chosen explicit `--port`, loopback custom Host emits `--host`, noncanonical loopback Host with Auto is rejected, and public/wildcard/other-address Hosts never reach approval in Dash.
- Approval tests continue to prove that no mutating job runs until the user approves, invalid host/port and public-host combinations never reach approval, and the exact generated argv is shown for review.
- Registry-consumer tests confirm chat discovery and telemetry continue to use each service record's endpoint and port rather than the established default.
- CLI parsing tests confirm omitted `--port` is distinguishable from an explicit port and that existing explicit-port scripts remain accepted.
- Supported-platform compilation and workspace quality gates remain required. Live GPU serving is not required for deterministic collision and wizard tests, but existing GPU-required smoke behavior must remain unchanged.

## Out of Scope

- Changing model recipes or engine preference rules.
- Adding CPU fallback or changing GPU selection policy.
- Allowing one model to span multiple GPUs.
- Changing direct engine adapter CLI defaults.
- Adding a second modal or a separate advanced-settings screen.
- Persisting serve-wizard overrides as global defaults.
- Automatically exposing servers publicly.
- Changing endpoint authentication, API-key storage, service restart semantics, or service-record schema beyond using the already-recorded resolved port.
- Redesigning the services manager, chat UI, telemetry dashboard, or approval component.
- Adding mouse-specific behavior beyond preserving existing TUI behavior.

## Further Notes

The project UX guidance identifies non-technical Windows users as the primary audience, requires plain language and arrow-key navigation, and explicitly recommends hiding advanced choices behind a selectable Advanced options row. The change should preserve the current approval-before-mutation rule and strict GPU-only behavior.

The managed service registry is already the authority for dynamic endpoints used by chat and telemetry. The established port remains useful as the first automatic candidate and as a legacy discovery fallback, but new managed services must not depend on that port being unique.

michaelroy-amd:
## Summary

- make omitted `rocm serve --port` choose the first available loopback port in `11435..=11535`, while preserving exact explicit-port behavior
- serialize managed launch and recovery port claims with a bounded cross-process lock and keep the resolved endpoint in the service registry
- simplify the Dash serve wizard to model selection plus inline Advanced settings, with explicit choice/edit affordances and validation before approval
- align Dash loopback validation exactly with the CLI and preserve wizard state when Escape cancels Host/Port editing
- add a real-CLI Gherkin scenario proving automatic serving skips an occupied default port

## Review follow-up

- Dash accepts only the exact loopback spellings accepted by the CLI; `LocalHost` and `[::1]` are rejected
- the app-level Escape gate now treats the inline editor as a nested layer, so Escape cancels the edit instead of closing the wizard
- `@id:serve-auto-port-skips-occupied` holds `127.0.0.1:11435`, invokes the real `rocm serve` command, and asserts the reported endpoint advances to `11436`
- removed the duplicate unreachable `bail!`
- preserved the generation-default/recipe mismatch guard while rebasing the launch path onto current `main`

## Verification

- `cargo test -p rocm --bin rocm`
- `cargo test -p rocm-dash-tui --lib`
- `cargo test -p e2e-cucumber --test e2e --no-run`
- `cargo test --workspace --all-targets` — 2,273 passed, 12 ignored
- code review and Rust review completed with no unresolved critical/high defects

The new `@id:serve-auto-port-skips-occupied` scenario is tagged `@requires-gpu @merge-queue`. It compiled locally; live execution is delegated to the merge-queue GPU lane because this WSL environment has no usable ROCm GPU.

Closes #239

### DIFF
--- a/apps/rocm/src/serve_port.rs
+++ b/apps/rocm/src/serve_port.rs
@@ -152,10 +152,7 @@ pub(crate) fn resolve_serve_port(
         // — and it is refused by returning an error, never by panicking.
         let Some(port) = request.explicit() else {
             validate_port_request(host, request)?;
-            bail!(
-                "`--host {host}` needs an explicit `--port <1-65535>`; automatic port \
-                 selection is supported only on {DEFAULT_LOCAL_HOST}."
-            );
+            unreachable!("custom-host automatic requests are rejected above");
         };
         return Ok(LoopbackPortLease::engine_owned(port));
     }
--- a/apps/rocm/src/serve_port.rs
+++ b/apps/rocm/src/serve_port.rs
@@ -1,0 +1,478 @@
+// Copyright © Advanced Micro Devices, Inc., or its affiliates.
+//
+// SPDX-License-Identifier: MIT
+
+//! Central serve-port policy for the user-facing `rocm serve` verb.
+//!
+//! `--port` is an explicit-or-automatic *request*, not a concrete default. This
+//! module turns that request into one concrete, leased port before any manifest,
+//! log, endpoint key, or child process exists:
+//!
+//! - **Automatic** (`--port` omitted) is accepted only on the canonical default
+//!   loopback host `127.0.0.1`, and scans [`AUTO_PORT_FIRST`]..=[`AUTO_PORT_LAST`]
+//!   for a candidate that is neither reserved by a live managed service nor
+//!   occupied on the OS.
+//! - **Explicit on canonical loopback** goes through the same reservation and
+//!   OS preflight, so a collision fails immediately with a clear message instead
+//!   of surfacing as an engine bind error minutes into startup.
+//! - **Explicit on a custom host** (`localhost`, `::1`, `0.0.0.0`, any other
+//!   IPv4/IPv6 address) is passed through untouched: bind success or failure
+//!   stays engine-owned, because an IPv4-loopback probe proves nothing about
+//!   those addresses. Automatic is refused there rather than guessing.
+//!
+//! Everything downstream of this module — engine requests, service records,
+//! readiness, supervision, recovery — keeps its concrete `u16` contract.
+
+use anyhow::{Result, bail};
+use rocm_core::{DEFAULT_LOCAL_HOST, DEFAULT_LOCAL_PORT, LoopbackPortLease};
+use std::io;
+
+/// First automatic candidate. Also the legacy well-known local endpoint, so an
+/// otherwise-idle machine keeps serving on the port users already know.
+pub(crate) const AUTO_PORT_FIRST: u16 = DEFAULT_LOCAL_PORT;
+/// Number of candidates *after* the first one. The scanned range is inclusive on
+/// both ends: `11435..=11535`, 101 candidates.
+pub(crate) const AUTO_PORT_SPAN: u16 = 100;
+/// Last automatic candidate.
+pub(crate) const AUTO_PORT_LAST: u16 = AUTO_PORT_FIRST + AUTO_PORT_SPAN;
+
+/// The one phrase used wherever a port must be described *before* it is
+/// resolved. The CLI plan output and the Dash approval card share it verbatim so
+/// neither promises a concrete port it has not yet acquired.
+pub(crate) const AUTOMATIC_PORT_DISCLOSURE: &str = "Port: automatic; endpoint shown after launch";
+
+/// What the user asked for on the command line.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) enum PortRequest {
+    /// `--port` omitted: choose an available local port.
+    Auto,
+    /// `--port N`: use exactly this port.
+    Explicit(u16),
+}
+
+impl PortRequest {
+    #[must_use]
+    pub(crate) const fn from_flag(port: Option<u16>) -> Self {
+        match port {
+            Some(port) => Self::Explicit(port),
+            None => Self::Auto,
+        }
+    }
+
+    #[must_use]
+    pub(crate) const fn is_auto(self) -> bool {
+        matches!(self, Self::Auto)
+    }
+
+    /// The concrete port, when the user named one.
+    #[must_use]
+    pub(crate) const fn explicit(self) -> Option<u16> {
+        match self {
+            Self::Auto => None,
+            Self::Explicit(port) => Some(port),
+        }
+    }
+}
+
+/// A port a live managed service still owns. Carried with its service id so a
+/// collision names the thing actually holding the port.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub(crate) struct ReservedPort {
+    pub(crate) port: u16,
+    pub(crate) service_id: String,
+    pub(crate) status: String,
+}
+
+/// clap value parser for every user-facing `--port`.
+///
+/// Rejects `0` explicitly: the OS reads it as "any ephemeral port", which would
+/// silently produce an endpoint nobody asked for.
+///
+/// # Errors
+/// Returns a user-facing message for anything outside 1–65535.
+pub(crate) fn parse_serve_port(value: &str) -> Result<u16, String> {
+    match value.trim().parse::<u16>() {
+        Ok(0) | Err(_) => Err(format!(
+            "port must be a number between 1 and 65535, got '{value}'"
+        )),
+        Ok(port) => Ok(port),
+    }
+}
+
+/// Reject an automatic request on a host this policy cannot preflight, before
+/// any engine resolution or side effect happens.
+///
+/// # Errors
+/// Fails when `--port` was omitted for anything but the canonical loopback host.
+pub(crate) fn validate_port_request(host: &str, request: PortRequest) -> Result<()> {
+    if request.is_auto() && !rocm_core::is_canonical_loopback_host(host) {
+        bail!(
+            "`rocm serve --host {host}` needs an explicit `--port <1-65535>`. Automatic port \
+             selection is supported only on the default host {DEFAULT_LOCAL_HOST}, where ROCm \
+             can prove a port is free before launching; on any other address the engine owns \
+             the bind."
+        );
+    }
+    Ok(())
+}
+
+/// Production bind probe: take a real `127.0.0.1:<port>` lease.
+///
+/// # Errors
+/// Propagates the OS bind error verbatim so the caller can distinguish
+/// "occupied" from "not permitted"/"not available".
+pub(crate) fn loopback_bind_probe(port: u16) -> io::Result<LoopbackPortLease> {
+    rocm_core::lease_loopback_port(port)
+}
+
+/// Turn a [`PortRequest`] into one concrete, leased port.
+///
+/// Callers must hold the shared managed-service allocation lock across this call
+/// *and* the record publication that follows it, so a concurrent launch cannot
+/// pick the same candidate between the probe and the spawn.
+///
+/// `probe` is injected so tests can drive deterministic OS outcomes; production
+/// passes [`loopback_bind_probe`].
+///
+/// # Errors
+/// Fails when an explicit port is reserved or occupied, when the automatic range
+/// is exhausted, or when the OS reports anything other than "address in use" —
+/// a permission or address-availability error must never be papered over by
+/// quietly choosing a different port.
+pub(crate) fn resolve_serve_port(
+    host: &str,
+    request: PortRequest,
+    reserved: &[ReservedPort],
+    probe: &mut dyn FnMut(u16) -> io::Result<LoopbackPortLease>,
+) -> Result<LoopbackPortLease> {
+    if !rocm_core::is_canonical_loopback_host(host) {
+        // Custom host: an explicit port passes through untouched so bind success
+        // or failure stays engine-owned. Automatic is refused here as well as in
+        // `serve()`, because this is the shared choke point every caller reaches
+        // — and it is refused by returning an error, never by panicking.
+        let Some(port) = request.explicit() else {
+            validate_port_request(host, request)?;
+            bail!(
+                "`--host {host}` needs an explicit `--port <1-65535>`; automatic port \
+                 selection is supported only on {DEFAULT_LOCAL_HOST}."
+            );
+        };
+        return Ok(LoopbackPortLease::engine_owned(port));
+    }
+
+    match request {
+        PortRequest::Explicit(port) => {
+            if let Some(owner) = reserved.iter().find(|entry| entry.port == port) {
+                bail!(
+                    "{DEFAULT_LOCAL_HOST}:{port} is already reserved by managed service {} \
+                     ({}). Stop it with `rocm services stop {} --yes`, or pass a different \
+                     `--port`.",
+                    owner.service_id,
+                    owner.status,
+                    owner.service_id
+                );
+            }
+            probe(port).map_err(|error| explicit_bind_error(port, &error))
+        }
+        PortRequest::Auto => {
+            for port in AUTO_PORT_FIRST..=AUTO_PORT_LAST {
+                if reserved.iter().any(|entry| entry.port == port) {
+                    continue;
+                }
+                match probe(port) {
+                    Ok(lease) => return Ok(lease),
+                    // Occupied is the only outcome that advances the scan.
+                    Err(error) if error.kind() == io::ErrorKind::AddrInUse => {}
+                    Err(error) => bail!(
+                        "failed to reserve {DEFAULT_LOCAL_HOST}:{port} while choosing an \
+                         automatic port: {error}"
+                    ),
+                }
+            }
+            bail!(
+                "no free TCP port for {DEFAULT_LOCAL_HOST} in {AUTO_PORT_FIRST}–{AUTO_PORT_LAST}; \
+                 every candidate is reserved by a live managed service or already in use. Stop an \
+                 unused server with `rocm services stop <service-id> --yes`, or pass an explicit \
+                 `--port`."
+            );
+        }
+    }
+}
+
+fn explicit_bind_error(port: u16, error: &io::Error) -> anyhow::Error {
+    if error.kind() == io::ErrorKind::AddrInUse {
+        return anyhow::anyhow!(
+            "{DEFAULT_LOCAL_HOST}:{port} is already in use by another process. Choose a free \
+             `--port`, stop the process holding it, or omit `--port` to let ROCm pick an \
+             available port starting at {AUTO_PORT_FIRST}."
+        );
+    }
+    anyhow::anyhow!("failed to reserve {DEFAULT_LOCAL_HOST}:{port}: {error}")
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::net::TcpListener;
+
+    fn reserved(port: u16, service_id: &str, status: &str) -> ReservedPort {
+        ReservedPort {
+            port,
+            service_id: service_id.to_owned(),
+            status: status.to_owned(),
+        }
+    }
+
+    /// A probe driven by a table of scripted per-port failures. Any port absent
+    /// from the table binds successfully.
+    fn scripted(
+        outcomes: Vec<(u16, io::ErrorKind)>,
+    ) -> impl FnMut(u16) -> io::Result<LoopbackPortLease> {
+        move |port| match outcomes.iter().find(|(candidate, _)| *candidate == port) {
+            Some((_, kind)) => Err(io::Error::new(*kind, "scripted")),
+            None => Ok(LoopbackPortLease::engine_owned(port)),
+        }
+    }
+
+    #[test]
+    fn port_request_distinguishes_omitted_from_explicit() {
+        assert_eq!(PortRequest::from_flag(None), PortRequest::Auto);
+        assert!(PortRequest::from_flag(None).is_auto());
+        assert_eq!(PortRequest::from_flag(None).explicit(), None);
+        assert_eq!(
+            PortRequest::from_flag(Some(8000)),
+            PortRequest::Explicit(8000)
+        );
+        assert!(!PortRequest::from_flag(Some(8000)).is_auto());
+        assert_eq!(PortRequest::from_flag(Some(8000)).explicit(), Some(8000));
+    }
+
+    #[test]
+    fn port_parser_rejects_zero_and_out_of_range_values() {
+        assert_eq!(parse_serve_port("8000"), Ok(8000));
+        assert_eq!(parse_serve_port(" 11435 "), Ok(11_435));
+        assert_eq!(parse_serve_port("65535"), Ok(65_535));
+        for bad in ["0", "65536", "-1", "eleven", ""] {
+            let error = parse_serve_port(bad).expect_err("must be rejected");
+            assert!(
+                error.contains("between 1 and 65535"),
+                "unexpected message for {bad:?}: {error}"
+            );
+        }
+    }
+
+    #[test]
+    fn automatic_range_is_exactly_the_documented_window() {
+        assert_eq!(AUTO_PORT_FIRST, 11_435);
+        assert_eq!(AUTO_PORT_LAST, 11_535);
+        assert_eq!((AUTO_PORT_FIRST..=AUTO_PORT_LAST).count(), 101);
+    }
+
+    #[test]
+    fn automatic_is_refused_on_every_custom_host() {
+        validate_port_request(DEFAULT_LOCAL_HOST, PortRequest::Auto).expect("canonical host is ok");
+        for host in ["localhost", "::1", "0.0.0.0", "127.0.0.2", "10.0.0.5"] {
+            let error = validate_port_request(host, PortRequest::Auto)
+                .expect_err("automatic must be refused on a custom host");
+            let message = format!("{error:#}");
+            assert!(message.contains("explicit `--port"), "{message}");
+            assert!(message.contains(host), "{message}");
+            // An explicit port is always accepted there.
+            validate_port_request(host, PortRequest::Explicit(8000))
+                .expect("explicit port is valid on a custom host");
+        }
+    }
+
+    #[test]
+    fn custom_host_explicit_port_bypasses_the_loopback_preflight() {
+        let mut probe = |_port: u16| -> io::Result<LoopbackPortLease> {
+            panic!("custom hosts must not be preflighted against IPv4 loopback")
+        };
+        // Even a port a live loopback service reserves is passed straight
+        // through: the engine owns bind on that address.
+        let lease = resolve_serve_port(
+            "0.0.0.0",
+            PortRequest::Explicit(11_435),
+            &[reserved(11_435, "svc-a", "running")],
+            &mut probe,
+        )
+        .expect("custom host explicit port is engine-owned");
+        assert_eq!(lease.port(), 11_435);
+        assert!(!lease.is_held());
+    }
+
+    #[test]
+    fn resolving_auto_on_a_custom_host_errors_instead_of_panicking() {
+        // The choke point must refuse Auto on its own, even when a caller skipped
+        // the earlier `validate_port_request` gate — as a Result, never a panic.
+        let mut probe = |_port: u16| -> io::Result<LoopbackPortLease> {
+            panic!("a custom host must never be preflighted")
+        };
+        for host in ["localhost", "::1", "0.0.0.0", "10.0.0.5"] {
+            let error = resolve_serve_port(host, PortRequest::Auto, &[], &mut probe)
+                .expect_err("automatic selection is not supported on a custom host");
+            let message = format!("{error:#}");
+            assert!(message.contains("explicit `--port"), "{message}");
+            assert!(message.contains(host), "{message}");
+        }
+    }
+
+    #[test]
+    fn explicit_loopback_port_outside_the_auto_range_is_retained() {
+        let mut probe = scripted(vec![]);
+        let lease = resolve_serve_port(
+            DEFAULT_LOCAL_HOST,
+            PortRequest::Explicit(8000),
+            &[],
+            &mut probe,
+        )
+        .expect("a free explicit port is kept");
+        assert_eq!(lease.port(), 8000);
+    }
+
+    #[test]
+    fn explicit_loopback_port_that_is_occupied_is_rejected() {
+        let mut probe = scripted(vec![(8000, io::ErrorKind::AddrInUse)]);
+        let error = resolve_serve_port(
+            DEFAULT_LOCAL_HOST,
+            PortRequest::Explicit(8000),
+            &[],
+            &mut probe,
+        )
+        .expect_err("an occupied explicit port must fail");
+        let message = format!("{error:#}");
+        assert!(message.contains("127.0.0.1:8000"), "{message}");
+        assert!(message.contains("already in use"), "{message}");
+    }
+
+    #[test]
+    fn explicit_loopback_port_reserved_by_a_live_service_names_that_service() {
+        let mut probe = |_port: u16| -> io::Result<LoopbackPortLease> {
+            panic!("a reserved port must be rejected before probing")
+        };
+        let error = resolve_serve_port(
+            DEFAULT_LOCAL_HOST,
+            PortRequest::Explicit(11_435),
+            &[reserved(11_435, "vllm-qwen-1", "starting")],
+            &mut probe,
+        )
+        .expect_err("a reserved explicit port must fail");
+        let message = format!("{error:#}");
+        assert!(message.contains("vllm-qwen-1"), "{message}");
+        assert!(message.contains("starting"), "{message}");
+    }
+
+    #[test]
+    fn automatic_takes_the_first_candidate_when_it_is_free() {
+        let mut probe = scripted(vec![]);
+        let lease = resolve_serve_port(DEFAULT_LOCAL_HOST, PortRequest::Auto, &[], &mut probe)
+            .expect("first candidate is free");
+        assert_eq!(lease.port(), AUTO_PORT_FIRST);
+    }
+
+    #[test]
+    fn automatic_skips_occupied_candidates_until_one_binds() {
+        let mut probe = scripted(vec![
+            (11_435, io::ErrorKind::AddrInUse),
+            (11_436, io::ErrorKind::AddrInUse),
+            (11_437, io::ErrorKind::AddrInUse),
+        ]);
+        let lease = resolve_serve_port(DEFAULT_LOCAL_HOST, PortRequest::Auto, &[], &mut probe)
+            .expect("scan advances past occupied candidates");
+        assert_eq!(lease.port(), 11_438);
+    }
+
+    #[test]
+    fn automatic_skips_ports_reserved_by_live_records_without_probing_them() {
+        let mut probed: Vec<u16> = Vec::new();
+        let lease = {
+            let mut probe = |port: u16| -> io::Result<LoopbackPortLease> {
+                probed.push(port);
+                Ok(LoopbackPortLease::engine_owned(port))
+            };
+            resolve_serve_port(
+                DEFAULT_LOCAL_HOST,
+                PortRequest::Auto,
+                &[
+                    reserved(11_435, "svc-ready", "ready"),
+                    reserved(11_436, "svc-recovering", "recovering"),
+                ],
+                &mut probe,
+            )
+            .expect("scan skips reserved candidates")
+        };
+        assert_eq!(lease.port(), 11_437);
+        assert_eq!(probed, vec![11_437]);
+    }
+
+    #[test]
+    fn automatic_stops_immediately_on_a_non_collision_bind_error() {
+        for kind in [
+            io::ErrorKind::PermissionDenied,
+            io::ErrorKind::AddrNotAvailable,
+        ] {
+            let mut probe = scripted(vec![(11_435, kind)]);
+            let error = resolve_serve_port(DEFAULT_LOCAL_HOST, PortRequest::Auto, &[], &mut probe)
+                .expect_err("only AddrInUse may advance the scan");
+            let message = format!("{error:#}");
+            assert!(message.contains("127.0.0.1:11435"), "{message}");
+        }
+    }
+
+    #[test]
+    fn explicit_non_collision_bind_error_reports_host_port_and_os_detail() {
+        let mut probe = |_port: u16| -> io::Result<LoopbackPortLease> {
+            Err(io::Error::new(
+                io::ErrorKind::PermissionDenied,
+                "permission denied",
+            ))
+        };
+        let error = resolve_serve_port(
+            DEFAULT_LOCAL_HOST,
+            PortRequest::Explicit(80),
+            &[],
+            &mut probe,
+        )
+        .expect_err("a privileged port must fail");
+        let message = format!("{error:#}");
+        assert!(message.contains("127.0.0.1:80"), "{message}");
+        assert!(message.contains("permission denied"), "{message}");
+    }
+
+    #[test]
+    fn automatic_exhaustion_reports_the_exact_scanned_range() {
+        let mut probe = |_port: u16| -> io::Result<LoopbackPortLease> {
+            Err(io::Error::new(io::ErrorKind::AddrInUse, "occupied"))
+        };
+        let error = resolve_serve_port(DEFAULT_LOCAL_HOST, PortRequest::Auto, &[], &mut probe)
+            .expect_err("an exhausted range must fail");
+        let message = format!("{error:#}");
+        assert!(message.contains("11435–11535"), "{message}");
+    }
+
+    #[test]
+    fn automatic_selects_against_real_loopback_listeners() {
+        // Real sockets, not a script: hold the first two candidates with live
+        // listeners and prove the production probe walks past them. If either is
+        // already owned by something outside this test the scenario cannot be set
+        // up deterministically, so skip rather than assert on a moving target.
+        let (Ok(first), Ok(second)) = (
+            TcpListener::bind(("127.0.0.1", AUTO_PORT_FIRST)),
+            TcpListener::bind(("127.0.0.1", AUTO_PORT_FIRST + 1)),
+        ) else {
+            return;
+        };
+        let mut probe = loopback_bind_probe;
+        let lease = resolve_serve_port(DEFAULT_LOCAL_HOST, PortRequest::Auto, &[], &mut probe)
+            .expect("a candidate above the held ones is free");
+        assert!(
+            lease.port() >= AUTO_PORT_FIRST + 2,
+            "selected {} while {AUTO_PORT_FIRST} and {} were held",
+            lease.port(),
+            AUTO_PORT_FIRST + 1
+        );
+        assert!(lease.is_held(), "the production probe must hold the socket");
+        // Guards outlive every assertion above.
+        drop((first, second));
+    }
+}
--- a/apps/rocmd/src/lib.rs
+++ b/apps/rocmd/src/lib.rs
@@ -4790,7 +4790,25 @@ fn handle_server_recover_event_with_record(
                     Some(record.service_id.clone()),
                 );
             }
-            restart_managed_service(paths, &mut *record)?;
+            // A contended allocation lock or an occupied recorded port is a
+            // recoverable, per-service condition — report it and move on rather
+            // than letting one service take the whole 30s watcher tick (and the
+            // daemon) down.
+            if let Err(error) = restart_managed_service(paths, &mut *record) {
+                return record_event(
+                    paths,
+                    state,
+                    "server-recover",
+                    "error",
+                    "restart_managed_service_failed",
+                    &format!(
+                        "failed to recover managed service {} on {}:{} after \
+                         {recovery_reason_display}: {error:#}",
+                        record.service_id, record.host, record.port
+                    ),
+                    Some(record.service_id.clone()),
+                );
+            }
             record_event(
                 paths,
                 state,
@@ -4937,7 +4955,33 @@ fn manifest_service_recovery_reason(
     }
 }
 
-fn restart_managed_service(_paths: &AppPaths, record: &mut ManagedServiceRecord) -> Result<()> {
+/// Respawn the supervisor for a managed service that needs recovery.
+///
+/// Recovery re-claims the concrete port already written in the record, so it
+/// runs inside the same bounded cross-process allocation lock as a fresh
+/// `rocm serve` launch: a CLI automatic selection and a recovery re-bind can
+/// never interleave, and the CLI sees this record's `recovering` reservation.
+fn restart_managed_service(paths: &AppPaths, record: &mut ManagedServiceRecord) -> Result<()> {
+    let _allocation = rocm_core::lock_service_allocation(paths)?;
+
+    // On the canonical loopback host, prove the recorded port is genuinely free
+    // before publishing `recovering` or spawning anything: a supervisor that
+    // cannot bind would replace a truthful record with a lie. Custom hosts keep
+    // the existing engine-owned bind behavior — the record transition is still
+    // serialized, but the bind is not preflighted.
+    let lease = if rocm_core::is_canonical_loopback_host(&record.host) {
+        Some(
+            rocm_core::lease_loopback_port(record.port).with_context(|| {
+                format!(
+                    "cannot recover managed service {} on {}:{}: its recorded port is in use",
+                    record.service_id, record.host, record.port
+                )
+            })?,
+        )
+    } else {
+        None
+    };
+
     let rocmd_binary =
         std::env::current_exe().context("failed to resolve current rocmd executable path")?;
     let log_file = fs::OpenOptions::new()
@@ -4949,6 +4993,11 @@ fn restart_managed_service(_paths: &AppPaths, record: &mut ManagedServiceRecord)
         .try_clone()
         .context("failed to clone service log file handle")?;
 
+    // The truthful state to fall back to. Any failure below must leave the
+    // operator with this, not a half-applied `recovering` row — and, once the
+    // supervisor exists, must not leave that supervisor running unrecorded.
+    let prior = record.clone();
+
     record.status = "recovering".to_owned();
     // Counts the restart and drops the previous run's inference verification.
     // The respawned child writes a fresh record of its own, and "recovering" is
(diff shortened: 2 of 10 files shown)


crates/rocm-dash-tui/src/ui/serve_wizard.rs
```diff
@@ -36,28 +43,65 @@ use crate::ui::model_picker::{ModelPicker, ModelRecipeSummary, PickerOutcome, dr
...
+/// must be requested deliberately from the CLI instead.
+pub const LOOPBACK_HOSTS: &[&str] = &["127.0.0.1", "localhost", "::1", "[::1]"];
+
+/// The guidance shown when a public / non-loopback host is typed into Dash.
+pub const PUBLIC_HOST_NEEDS_CLI: &str = "Dash serves on loopback only; for a public bind run \
+     `rocm serve --host … --port … --allow-public-bind` from the CLI.";
+
+/// Whether `host` is one of the loopback spellings Dash may bind.
+#[must_use]
+pub fn is_loopback_host(host: &str) -> bool {
```
juhovainio, line 92: `is_loopback_host` here does not match the CLI's actual gate, `is_public_bind_host` in `crates/rocm-engine-protocol/src/lib.rs`. This version compares case-insensitively and accepts `"[::1]"`; the CLI's check is case-sensitive and does not special-case `"[::1]"` (its own test asserts `"LOCALHOST"` and `"[::1]"` are treated as *public*, deliberately). So a host like `LocalHost` or `[::1]` passes validation here and Dash approves the launch, but the CLI then rejects it at spawn time since it never received `--allow-public-bind`. Worth making this byte-for-byte identical to the CLI's check (or sharing one definition) rather than a hand-rolled duplicate.

crates/rocm-dash-tui/src/ui/serve_wizard.rs
```diff
@@ -92,13 +222,19 @@ pub struct PendingServe {
...
-    pub device_idx: usize,
     pub host: String,
+    pub port_mode: PortMode,
+    /// The custom port text (only meaningful when `port_mode` is `Custom`).
     pub port: String,
     pub managed: bool,
+    /// Whether the inline `Advanced settings` rows are showing.
+    pub advanced_expanded: bool,
+    /// Inline text editor; `Some` while editing Host or the custom port.
+    pub editor: Option<InlineEditor>,
```
juhovainio, line 235: This new `editor` field isn't wired into `active_overlay_at_root()` in `crates/rocm-dash-tui/src/app/mod.rs` (not touched by this PR, so I couldn't comment on it directly — the check is around lines 965-971 there). That function only looks at `browser`/`picker`/`approval`/`active_job` to decide whether Esc should close the whole wizard vs. defer to a sub-screen. With `editor` unlisted, opening Advanced, starting to edit Host or Port, and hitting Esc gets misread as "nothing open" and `close_overlays()` wipes the entire in-progress form (including the typed model name) instead of just canceling the field edit that `editor_key` already handles correctly. Suggest adding `&& w.editor.is_none()` to the `serve_wizard` arm there, matching the existing pattern for `install_manager.active_job`.

apps/rocm/src/serve_port.rs
```diff
@@ -0,0 +1,478 @@
...
+///
+/// `probe` is injected so tests can drive deterministic OS outcomes; production
+/// passes [`loopback_bind_probe`].
+///
+/// # Errors
+/// Fails when an explicit port is reserved or occupied, when the automatic range
+/// is exhausted, or when the OS reports anything other than "address in use" —
+/// a permission or address-availability error must never be papered over by
+/// quietly choosing a different port.
+pub(crate) fn resolve_serve_port(
```
juhovainio, line 142: I couldn't find any e2e/Gherkin scenario that exercises this function's actual behavior through the real `rocm serve` CLI. The port-related scenarios in `tests/e2e-cucumber/features/model_serving.feature` ("served on the default port" / "on a non-default port") are both backed by a `MockServer` started directly in the step definitions, not by invoking `rocm serve` — so nothing proves auto-selection actually skips a reserved port, exhausts the range correctly, or refuses automatic selection on a custom host. Given this is new user-observable CLI behavior, it seems worth a scenario that starts one service on the default port and asserts a second `rocm serve` lands on the next one.

apps/rocm/src/serve_port.rs
```diff
@@ -0,0 +1,478 @@
...
+    probe: &mut dyn FnMut(u16) -> io::Result<LoopbackPortLease>,
+) -> Result<LoopbackPortLease> {
+    if !rocm_core::is_canonical_loopback_host(host) {
+        // Custom host: an explicit port passes through untouched so bind success
+        // or failure stays engine-owned. Automatic is refused here as well as in
+        // `serve()`, because this is the shared choke point every caller reaches
+        // — and it is refused by returning an error, never by panicking.
+        let Some(port) = request.explicit() else {
+            validate_port_request(host, request)?;
+            bail!(
```
juhovainio: This `bail!` looks unreachable. It only runs when `request.explicit()` is `None`, i.e. `request == PortRequest::Auto`, and we're already inside `if !is_canonical_loopback_host(host)`. But `validate_port_request` on the line above bails whenever `request.is_auto() && !is_canonical_loopback_host(host)` — exactly the condition guaranteed true here — so the `?` on that call always short-circuits before this `bail!` can execute. Harmless (near-identical message either way), but dead code; maybe replace with `unreachable!()` or restructure to drop the duplicate message.

juhovainio (changes requested):
I reviewed this PR and found four issues worth fixing before merge, left as inline comments below.

The backend work — port locking, the allocation lock, recovery rollback — is solid and well tested; nothing to flag there. The problems are in the new Dash wizard UI plus one test-coverage gap:

- Dash's notion of a "loopback" host doesn't match the CLI's, so the wizard can validate and approve a launch that the CLI then refuses.
- Pressing Esc while editing the Host or Port field wipes the entire form instead of just canceling that one edit.
- Nothing exercises the actual auto-port-selection behavior through the real `rocm serve` command — the existing e2e "port" scenarios all bypass it with a mock server.
- One leftover unreachable `bail!` in the new port-resolution module.

Not blocking, but worth a look: the loopback-host list and a couple of user-facing strings are independently duplicated between the CLI and TUI crates, which is exactly how the first issue above happened — sharing that definition would prevent the next drift. The PR also bundles an unrelated test-flakiness fix in with the port-safety work; consider splitting it out.

Ran a full `cargo check`/`cargo clippy --all-targets -- -D warnings` across the workspace and they're clean, and the concurrency tests for the new allocation lock genuinely exercise the race with real threads rather than mocks — good approach there.

apps/rocm/src/serve_port.rs
michaelroy-amd: Fixed in c526c33. The duplicate `bail!` is gone; the structurally unreachable fallthrough is now explicit after `validate_port_request`.

apps/rocm/src/serve_port.rs
michaelroy-amd: Fixed in c526c33. Added `@id:serve-auto-port-skips-occupied`: it holds 11435 with a real listener, invokes the real `rocm serve` binary without `--port`, and asserts endpoint 11436. It runs on the merge-queue GPU lane.

crates/rocm-dash-tui/src/ui/serve_wizard.rs
michaelroy-amd: Fixed in c526c33. `active_overlay_at_root()` now includes `w.editor.is_none()`, and the regression test proves Escape reaches the editor without discarding the in-progress wizard.

crates/rocm-dash-tui/src/ui/serve_wizard.rs
michaelroy-amd: Fixed in c526c33. Dash now uses the exact case-sensitive host set accepted by the CLI, with regression coverage for `LocalHost` and `[::1]`.

volen-silo:
I had a parallel fix for this defect in #301 — I missed this PR when planning and duplicated it. #301 is now closed as superseded; this change is the better one, particularly holding a real lease across the launch and serialising allocation behind a lock rather than probing and releasing.

Two gaps I hit in my own version that look like they apply here too. Both are small; neither is a blocker.

**1. An explicit `--port` is silently dropped when an equivalent service is already live.**

`resolve_managed_port_in_transaction` answers the idempotency guard before it looks at the request:

```rust
if let Some(existing) = existing_live_managed_service_in(&live, engine, canonical_model_id) {
    return Ok(ManagedPortDecision::AlreadyLive(Box::new(existing.clone())));
}
```

The `AlreadyLive` consumer then bails only on a recipe mismatch. So `rocm serve <model> --port 12000`, against a live service for the same engine and model on 11435, reports success on 11435 — a port the user did not ask for, with no note. That contradicts the explicit-port contract this PR otherwise enforces strictly.

It reads like the same class as the recipe-mismatch guard immediately below it, so the fix probably belongs in the same place:

```rust
if let Some(port) = request.explicit()
    && port != existing.port
{
    bail!(
        "managed service `{}` is already serving {} on port {}, so port {port} would not be used; \
         stop it with `rocm services stop {}`, or drop --port to reuse it",
        existing.service_id, canonical_model_id, existing.port, existing.service_id
    );
}
```

A reviewer caught this in my branch; I would not have found it by reading the happy path.

**2. The explicit-busy-port refusal has no scenario.**

`serve-auto-port-skips-occupied` pins the automatic path, but the explicit path is the other half of the contract and is the one that changes a user-visible failure mode — it now fails up front instead of surfacing as an engine bind error. Worth a sibling scenario holding an address and naming it via `--port`, asserting both the refusal and that no service record is left behind (the stale record was part of the original complaint).

One tagging note if you do add it: `@merge-queue` is documented as "a heavy real-GPU serve that is redundant with a cheaper per-engine canary". An explicit-busy-port scenario never launches an engine — it is refused at the address check — so it is cheap enough for the per-PR GPU lane. I tagged mine `@merge-queue` by copy-paste and it silently did not run; the lane went green without it. Worth checking whether `serve-auto-port-skips-occupied` wants to be per-PR too, since a scenario that only runs in the merge queue gives no signal on the PR that changes the code it covers.

apps/rocmd/src/lib.rs
```diff
@@ -4937,7 +4955,33 @@ fn manifest_service_recovery_reason(
...
 
-fn restart_managed_service(_paths: &AppPaths, record: &mut ManagedServiceRecord) -> Result<()> {
+/// Respawn the supervisor for a managed service that needs recovery.
+///
+/// Recovery re-claims the concrete port already written in the record, so it
+/// runs inside the same bounded cross-process allocation lock as a fresh
+/// `rocm serve` launch: a CLI automatic selection and a recovery re-bind can
+/// never interleave, and the CLI sees this record's `recovering` reservation.
+fn restart_managed_service(paths: &AppPaths, record: &mut ManagedServiceRecord) -> Result<()> {
+    let _allocation = rocm_core::lock_service_allocation(paths)?;
```
rominf, line 4965: **`rocm services restart` respawns outside the new allocation lock.**

This is the daemon-side restart and it correctly takes `lock_service_allocation` + `lease_loopback_port` (neither of which existed on `main` — this PR adds them here). The CLI-side twin at `apps/rocm/src/main.rs:14045` (`restart_internal_managed_service`) did not get the same treatment, which is why this reads as an oversight rather than a scoping decision.

That function is reachable from `rocm services restart <id> --yes` via `run_approved_service_action` -> `SandboxToolArg::RestartServer` (`main.rs:13876`). It:

1. calls `stop_internal_managed_service`, which flips the record to stopped — and `live_managed_services` (`main.rs:5415-5421`) filters on `managed_service_is_live`, so `reserved_service_ports` immediately stops reserving that port;
2. rebuilds the serve args on the bare `record.port` (`main.rs:14085`) with no `lock_service_allocation`, no `lease_loopback_port` probe, and no re-check that the port is still free;
3. spawns.

A concurrent `rocm serve` auto-scan can claim the freed port inside that stop->spawn gap, and the restart then fails late in the engine — exactly the collision class this PR exists to remove, just moved to a different trigger.

Wrapping the stop+respawn in `rocm_core::lock_service_allocation(paths)?` and leasing `record.port` when `is_canonical_loopback_host(&record.host)`, releasing right before the spawn, would mirror what you've already done here.

crates/rocm-core/src/lib.rs
```diff
@@ -7447,6 +7447,189 @@ pub fn unix_time_millis() -> u128 {
...
+        port,
+        listener: Some(listener),
+    })
+}
+
+/// File name of the shared allocation lock, inside [`AppPaths::services_dir`].
+///
+/// Not a `.json` file, so `load_managed_services` never mistakes it for a
+/// service manifest.
+pub const SERVICE_ALLOCATION_LOCK_FILE: &str = "allocation.lock";
```
rominf, line 7534: Cross-referencing the summary: this is the lock file that collides with #267's `services_dir()/launch.lock`. Same resource, two files, and different waiting semantics — `lock_service_allocation_at` is bounded (5s, actionable error) while #267's `FileLock::acquire` blocks on `file.lock()` with no timeout.

Nothing to change in this file on its own; flagging it here because this constant is the concrete thing the two PRs need to agree on.

crates/rocm-dash-tui/src/ui/serve_wizard.rs
```diff
@@ -131,91 +267,173 @@ impl Default for ServeWizardState {
...
         }
+        let host = self.host.trim();
+        if host.is_empty() {
+            return Err((
+                Field::Host,
+                format!("host is required; type an address or restore {DEFAULT_HOST}"),
+            ));
+        }
+        // Automatic port selection is only supported on the canonical loopback
+        // host — anything else must name its own port. Never invent one.
```
rominf, line 373: **A public host with automatic port gets sent to the wrong field, then dead-ends.**

This auto-port check runs before the `!is_loopback_host(host)` check at `:381`. So for `host = "0.0.0.0"`, `port_mode = Auto`, `validate()` returns `(Field::Port, CUSTOM_HOST_NEEDS_PORT)` — "this host needs its own port". The user follows that, sets a custom port, and now falls through to `:381`, which rejects the same form with `PUBLIC_HOST_NEEDS_CLI` ("Dash serves on loopback only"). Two steps, cursor parked on the wrong field, and the real reason only surfaces at the end.

Only `localhost` and `::1` genuinely need the port hint — a public host can never launch from Dash at all, so it should fail once, on `Field::Host`, with the actionable message.

The test at `:1128-1140` currently locks the misordering in: it iterates `["localhost", "::1", "0.0.0.0", "192.168.1.5"]` and asserts all four give `Field::Port` + `CUSTOM_HOST_NEEDS_PORT`. Moving the loopback check above the auto-port check means splitting that test so the two public hosts expect `Field::Host` + `PUBLIC_HOST_NEEDS_CLI`.

crates/rocm-dash-tui/src/ui/tabs/serving.rs
```diff
@@ -25,11 +25,11 @@ pub const VERBS: &[Verb] = &[
         summary: "Launch a model on a serving engine and expose an OpenAI-style endpoint.",
         steps: &[
             "Pick a model",
-            "Choose an engine — Lemonade · vLLM",
-            "Set GPU placement (required / preferred / CPU-only)",
-            "Launch on 127.0.0.1:11435 and watch it come up",
+            "Leave the rest automatic, or open Advanced settings",
+            "GPU is required — ROCm never falls back to CPU",
+            "Launch on a local endpoint picked for you and watch it come up",
```
rominf, line 30: This rewording is right, and `serving_first_view_promises_no_concrete_endpoint` at `:159` pins it — but the same promise survives in the immediate siblings, which the PR doesn't touch:

- `crates/rocm-dash-tui/src/ui/tabs/chat.rs:65` — "Probing for a local engine (... / rocm serve :11435)…"
- `crates/rocm-dash-tui/src/ui/tabs/chat.rs:214` — "or run a local endpoint (vLLM :8000, rocm serve :11435)"
- `crates/rocm-dash-tui/src/app/mod.rs:1147` — "Detecting a local engine (... / rocm serve :11435)…"
- `crates/rocm-dash-tui/src/app/mod.rs:1177` — "no local engine found (... / rocm serve :11435)"

Once auto-allocation is in, a user whose service landed on 11436 reads all four of those and looks at the wrong port. The detection itself is fine — `app/chat.rs:250 detect_local_chat_with_probe` is registry-first — so this is purely copy that the PR's own premise invalidates.

Rewording them the way you did here, or extending the new no-concrete-endpoint test to cover these two files, would keep them honest.

tests/e2e-cucumber/tests/e2e/serving_steps.rs
```diff
@@ -308,6 +308,18 @@ async fn setup_mock_custom_port(world: &mut E2eWorld) {
     world.mock = Some(mock);
 }
 
+#[given("the default serve port is occupied")]
```
rominf, line 311: **This `Given` proves 11436 was free, then releases it — but the assertion demands the service land on exactly 11436.**

`std::net::TcpListener::bind(("127.0.0.1", SERVE_PORT + 1))` followed immediately by `drop(next)` only establishes that 11436 was free at setup time. The corresponding `Then` at `:921` builds `format!("endpoint: http://127.0.0.1:{}/v1", SERVE_PORT + 1)` and requires an exact line match.

Anything that takes 11436 in the interval pushes `resolve_serve_port`'s scan to 11437 and fails the scenario for a reason unrelated to the code under test — a stale-but-live managed record (`reserved_service_ports` reserves regardless of recorded host, deliberately), or any other process on the shared self-hosted runner. This lands on `@requires-gpu @merge-queue`, where a flake is expensive to diagnose and blocks the queue.

Two ways out: assert the endpoint is in `11436..=11535` and `!= 11435` rather than pinning the successor, or keep the 11436 listener alive for the scenario's duration and assert 11437 — the latter tests the scan more strictly anyway.

rominf (changes requested):
I read through this at `c526c33` — after the round of fixes you pushed for juhovainio's earlier comments, which all look correctly addressed. The port-allocation core is sound: the bind probe holds a real `TcpListener` lease that is only released immediately before the spawn, and the whole sequence sits inside `lock_service_allocation`, so the classic bind-close-hand-off window really is closed on the `rocm serve` path. I also tried to break the probe with `SO_REUSEADDR` (a wildcard `0.0.0.0:P` listener making the `127.0.0.1:P` probe look free) and couldn't — on Linux the probe still gets `EADDRINUSE`, so that concern doesn't apply. I didn't verify the Windows behaviour there.

What I did find is one launch path left outside the new lock, plus a handful of smaller things. Details inline.

The one I'd most want resolved before merge isn't in the diff at all, though: **#267 introduces a second cross-process lock over the same critical sections.** It adds `AppPaths::managed_launch_lock_path()` -> `services_dir()/launch.lock`, acquired in `serve()` and threaded into `start_managed_service` / `run_attached_service` as a new parameter. Neither lock file exists on `main` today, so this isn't a pre-existing overlap — both PRs are inventing one. If both land, one resource ends up guarded by two different lock files with divergent policies: yours is bounded at 5s with an actionable error, #267's `FileLock::acquire` calls `file.lock()` and blocks indefinitely. Two locks that don't mutually exclude are worse than either alone — this PR's "the lock guarantees no concurrent allocation" invariant stops being true the moment #267's path also mutates records. They also both touch `apps/rocm/src/main.rs`, `crates/rocm-core/src/lib.rs`, `model_serving.feature`, and `serving_steps.rs`, and both change overlapping function signatures. #267 is currently `MERGEABLE` and this one isn't, so it will likely land first by default rather than by decision. Worth agreeing on one lock file and one timeout policy (I'd argue for the bounded one with the actionable message) and rebasing the loser on top.

**PR title.** `Make model serving ports collision-safe` doesn't follow Conventional Commits, which `CONTRIBUTING.md:83` requires for PR titles specifically — and since the repo squash-merges, the title becomes the commit subject verbatim (recent `main`: `fix(release): ... (#330)`, `test(e2e): ... (#322)`). Something like `fix(serve): make model serving ports collision-safe` would match. Minor, but easy to miss until merge time.

**Merge state.** `CONFLICTING`/`DIRTY` — the base is 38 commits behind `main` and `git merge-tree` conflicts in `tests/e2e-cucumber/tests/e2e/serving_steps.rs`. Since #267 touches that same file, the rebase order matters here too.

**CI.** The last run was 2026-08-18 at this head. `E2E tests (Strix Halo, Windows)` is red: 40 scenarios, 4 failed, 3 expected xfail, **1 unexpected — `serve-hf-checkpoint-inference` regressed**. That's a real serve scenario on a PR that changes how the serve port is chosen, so I don't think it can be waved through as unrelated without a look at the log. `Coverage (rocm-dash crates, ratcheted)` was cancelled. Both will need to be green after the rebase anyway.

siloteemu (changes requested):
🔴 Automated review · pr-review-watcher · c526c33d

## Summary
Makes `rocm serve` port selection collision-safe: `--port` becomes an explicit-or-automatic request, resolved inside a cross-process advisory file lock into a real held `127.0.0.1` socket lease (`LoopbackPortLease`) that is released only immediately before the engine child spawns; `rocmd` recovery takes the same lock and re-leases its recorded port; the Dash serve wizard is reworked to match. **Verdict: Needs work** — the production design is sound, but `main.rs`'s two "concurrent" tests do not exercise the production call site and would pass with the lock wiring removed. Verified: I traced the transaction myself in `apps/rocm/src/main.rs` and `apps/rocmd/src/lib.rs` and confirmed the lock spans lease → `record.write()` → `lease.release()` → spawn → second write, so the record that reserves the port is published *before* the lock drops (the ordering the fix depends on); I ran the new `rocm-core` lock/lease tests locally (7 passed, 2 `#[ignore]`d child-process helpers invoked by their parents — cross-process exclusion, release-on-process-death, and cross-process handoff all genuinely pass); the branch contains everything in `main` (not behind); leak scan on the diff is clean. **A previously-reported port-binding finding is partially fixed** — the Given now genuinely *holds* `11435` on the World for the whole serve (`world.automatic_port_guard`), and that scenario would fail on reverted code (serve would default to the held `11435`); what remains bind-then-dropped is the `11436` precondition probe, while the assertion still demands exactly `11436`. **The 2 red checks: I could not identify them** — this review has no GitHub access, so I will not guess a cause and will not call them flake; my one permitted local check covered the `rocm-core` primitives only and was green. Blocking: 1 · Non-blocking: 5.

**`apps/rocm/src/main.rs:22801` and `apps/rocm/src/main.rs:22854` — `concurrent_auto_transactions_publish_distinct_ports` and `concurrent_equivalent_transactions_converge_on_one_record` prove the primitives compose, not that the production path uses them.** Both tests spawn two barrier-synchronised threads that *hand-roll* the transaction:

```rust
let _allocation = rocm_core::lock_service_allocation(&paths)?;
let lease = match resolve_managed_port_in_transaction(...)
```

Neither calls `spawn_managed_engine_child` (`main.rs:5523`), which is where the fix actually lives. **Answering the revert question directly: yes, both still pass if the production change is reverted** — delete the `lock_service_allocation` call at `main.rs:5544`, or shrink its scope so it drops before `record.write()` at `main.rs:5615`, and these tests are untouched, because they never execute that code. Their names ("concurrent … transactions") assert coverage they do not have, which is worse than no test: a future refactor that moves the lock is green. Worse for `..._distinct_ports` specifically, its `assert_ne!(ports[0], ports[1])` would hold even with *no* file lock at all, because the socket lease alone forces the second thread past the first candidate — so that assertion cannot distinguish a working lock from a missing one. (`..._converge_on_one_record` does depend on the lock, since without it both threads pass the idempotency check before either publishes.)

Concrete fix: mirror the pattern `apps/rocmd/src/lib.rs:5665` already gets right — `recovery_waits_for_the_allocation_lock_and_refuses_an_occupied_port` takes the lock on the main thread, calls the *real* `restart_managed_service` from a spawned thread, asserts it is still blocked after 250 ms, then releases and checks the error and that the on-disk record was not rewritten. Add the equivalent for `spawn_managed_engine_child`: hold `lock_service_allocation` on the test thread, call `spawn_managed_engine_child` from another, assert it does not complete while the lock is held. That is the only assertion that fails when the lock wiring is reverted.

## Non-blocking
- `apps/rocm/src/serve_port.rs:39-42` / `crates/rocm-dash-tui/src/ui/serve_wizard.rs:68-73` — both doc comments claim the CLI and Dash "share [the disclosure] verbatim", but they are two independent string literals in two crates with no shared constant and no test pinning them; the comment states a false invariant. `apps/rocm` already depends on `rocm-dash-tui` (Cargo.toml:27) and `AUTO_PORT_NOTE` is `pub`, so `serve_port.rs` can import it instead of redeclaring — a one-line fix, no new dependency.
- `apps/rocm/src/serve_port.rs:452-474` — `automatic_selects_against_real_loopback_listeners` silently `return`s (passes, testing nothing) if `11435`/`11436` are already taken, and it binds those real ports inside the unit-test process; `concurrent_auto_transactions_publish_distinct_ports` scans from `11435` with the real probe in the *same* test binary, so under parallel test execution these contend and the real-socket test can self-disable without any signal. Consider an ephemeral-range scan window for tests, or at minimum make the skip visible.
- `tests/e2e-cucumber/tests/e2e/serving_steps.rs:311-320` — the residue of the earlier finding: `TcpListener::bind(SERVE_PORT + 1)` is bound and immediately dropped as a free-check while `assert_next_automatic_port` demands exactly `SERVE_PORT + 1`. On a shared GPU runner another process can take `11436` in that gap (assertion fails), and if `11436` is busy at check time the step `expect()`-panics rather than skipping. Asserting `port > SERVE_PORT && port <= AUTO_PORT_LAST` would test the actual behaviour ("advanced past the occupied default") without the exact-number fragility.
- `tests/e2e-cucumber/features/model_serving.feature:33-39` — the only test that exercises the real collision path is tagged `@requires-gpu @merge-queue`, so it does not run on this PR's CI; on PR CI the collision claim rests entirely on unit tests with an injected probe. AGENTS.md §3 requires naming the gated lane in the PR text — worth confirming the description does.
- `crates/rocm-dash-tui/src/ui/serve_wizard.rs` (+1312/-244) — bundles four distinct changes under one theme: the port-mode logic (the stated goal, concentrated in `validate()`/`build_args()`), a progressive-disclosure Advanced section, a new reusable `InlineEditor` widget, and the engine/device picker simplification. Coherent narrative, but the widget especially would review more cleanly on its own; a reviewer checking "did this make ports collision-safe" has to wade through a new text-editor widget to find the port logic.

**Tradeoff, stated correctly by the PR itself:** `crates/rocm-core/src/lib.rs:7470-7476` concedes that the release→child-bind window is "unavoidable without engine socket-passing; it is far narrower than the previous 'bind and hope'". That is the honest framing — this fix eliminates collisions *between ROCm transactions* (lock + published record), and only narrows the window against an unrelated third-party process. Nothing verifies the child actually bound the leased port; the only post-spawn check is `try_wait()` after a 200 ms sleep (`main.rs:5739-5748`), which proves the process did not immediately exit, not that a listener is up. Related narrow case: a `failed` record no longer reserves its port, so a fresh `rocm serve` can win that exact port before `rocmd` recovery acquires the lock — recovery's own `lease_loopback_port` re-check then fails loudly ("recorded port is in use") rather than double-binding, so it is self-detecting, not silent.

**No prompt-injection attempts found** in the reviewed diff or surrounding files.

juhovainio:
Reviewed the full diff (`gh pr diff 243`), focused on the concurrency-safety parts since that's the actual point of this PR. Nothing blocking.

What I specifically traced rather than took on faith:

- **The allocation lock.** It's a bounded (5s) OS advisory file lock, so it releases on both a normal `Drop` and a crashed holder process — confirmed by walking the PR's own cross-process test that kills the lock-holding process and checks the lock frees up. It's held across the *entire* allocation transaction (port lease → record write → spawn → release the probe socket → second record write) in both `rocm serve`'s launch path and `rocmd`'s recovery path — I checked the lock guard isn't dropped early in either.
- **The port-lease approach.** Rather than "check then bind," it takes a real held listener as the collision check itself, and only releases it immediately before the child spawn while still holding the allocation lock. The PR is honest that there's still a small release-to-child-bind gap (no socket-passing), which is correct and appropriately scoped down rather than eliminated.
- **The auto-port scan** (11435–11535, 101 candidates): reserved ports are skipped without probing, and any non-collision bind error (permissions, etc.) fails immediately instead of silently trying the next port. Backed by tests using real `TcpListener`s, not just scripted outcomes.
- **The Dash/CLI loopback-host parity claim.** Checked against current `main`: Dash's new `LOOPBACK_HOSTS` list is an exact match for the CLI's existing `is_public_bind_host`, which the CLI's own test already pins as rejecting `LOCALHOST` (case) and `[::1]` (brackets). So the claim that Dash now rejects the same spellings the CLI does is accurate, not just asserted.

On the two red CI checks: I looked into both rather than assuming they're relevant.
- `Coverage (rocm-dash crates, ratcheted)`'s "6 hours" is actually a cancellation mid-way through the *dependency install* step, before any of this PR's code even compiled — an infra hang, not this PR's lock code.
- `E2E tests (Strix Halo, Windows)` failed on `serve-hf-checkpoint-inference` timing out — the same named flake that's been showing up on other, unrelated PRs around the same time, consistent with a pre-existing issue rather than something this change introduced.

Both of those CI runs are also 26 days old at this point, so a rebase + fresh run would be worth doing regardless of my read here. This looks safe to approve after that final CI pass.

siloteemu:
🔴 Automated review · pr-review-watcher · c526c33d

## Summary

Forced re-check of the existing change request at the same head (`c526c33`, 10 files). The code has not moved since the original review, so this pass asked one question per count: does the finding still hold on evidence gathered fresh?

**Outcome: the single blocking count still holds, unchanged and strengthened.** Blocking: 1. I re-read our claim verbatim before testing it, and tried to refute it rather than confirm it; it survived. The new evidence is a comment inside the test itself that asserts the very equivalence the test does not have.

CI conclusions worked from: **22 success, 1 failure, 1 cancelled**. A cancelled job is not a failure. This also corrects our own earlier text, which described "the 2 red checks" — that overstated the position: there is one genuine failure, not two.

Verification was static: reading, grepping and diffing. The full suite was not run here, and no mutation build was performed — none was needed, because the refutation is structural (see below).

---

## Blocking count 1 — STILL HOLDS

The sentence under test, quoted from our live change request:

> "**Answering the revert question directly: yes, both still pass if the production change is reverted** — delete the `lock_service_allocation` call at `main.rs:5544`, or shrink its scope so it drops before `record.write()` at `main.rs:5615`, and these tests are untouched, because they never execute that code."

Re-verified against the tree, line by line:

- The production transaction is `spawn_managed_engine_child` (`apps/rocm/src/main.rs:5523`). It takes the lock at `main.rs:5544`, resolves the port at `:5556`, and publishes the reserving record at `:5615`, all under one guard. That ordering is correct and is what the fix depends on — I confirmed it again.
- `concurrent_auto_transactions_publish_distinct_ports` (`main.rs:22785`) and `concurrent_equivalent_transactions_converge_on_one_record` (`main.rs:22840`) each hand-roll their own `lock_service_allocation` → `resolve_managed_port_in_transaction` → `write_live_service_record` → `lease.release()` sequence on two barrier-synchronised threads (`main.rs:22801-22817` and `:22854-22871`). Neither calls `spawn_managed_engine_child`.
- Exhaustive grep for `spawn_managed_engine_child` in `apps/rocm/src/main.rs` returns exactly two production call sites (`:5785`, `:5992`) and **one** test call site: `spawn_managed_engine_child_blocks_reuse_with_mismatched_recipe` (`:23052`). That test is single-threaded and asserts only on the recipe-mismatch error text; it passes with or without the lock.

So **no test in the crate observes the production lock at all.** Branch-level mutation: delete `main.rs:5544`, or move the guard's drop above `record.write()` at `:5615`, and the full test set stays green. That is the hard-blocker condition in standing focus (a) — tests that name a behaviour they do not exercise.

On the port-collision-specific question — can the tests observe a collision, or only assert a returned number? Re-confirmed the weaker half of our claim:

- `lease_loopback_port` (`crates/rocm-core/src/lib.rs:7519`) is a plain `TcpListener::bind` on `127.0.0.1`, and the lease holds the listener open until `release()` (`:7497`). In `concurrent_auto_transactions_publish_distinct_ports` the lease is held across the record write and released only after. So with the file lock removed entirely, the losing thread's probe of the first candidate still fails `AddrInUse` and the scan advances. Its `assert_ne!(ports[0], ports[1])` therefore **cannot distinguish a working lock from a missing one** — the socket lease alone satisfies it.
- `concurrent_equivalent_transactions_converge_on_one_record` does depend on a lock — but on *its own*, the one it calls in its own body. Reverting production changes nothing for it.

### New, reinforcing evidence at this commit

`main.rs:22799-22800`, inside the first of the two tests:

> "Exactly the production transaction: one lock around decision and publication, real sockets for the preflight."

This is false, and it is the comment that justifies the test's scope. The production transaction's own doc comment (`main.rs:5516-5520`, added in this same diff) enumerates what it covers: "duplicate detection, live-reservation refresh, candidate leasing, manifest/log creation, argv preparation, the bounded immediate-exit check, and record publication." The test performs the first three and one write. It never enters `launch_managed_engine_child_locked` (`main.rs:5653`), never creates the log or argv, never spawns a child, never runs the immediate-exit check, and never performs the second record publication that production does under the same lock. "Exactly" is not a near-miss here; it names an equivalence that is the whole basis for accepting a replica in place of the real call site.

### What would resolve this count

Unchanged, and the codebase already contains the pattern — the author got it right on the daemon side. `apps/rocmd/src/lib.rs:5664`, `recovery_waits_for_the_allocation_lock_and_refuses_an_occupied_port`, holds `lock_service_allocation` on the test thread, calls the **real** `restart_managed_service` from a spawned thread, asserts it has not completed after 250 ms, then releases and checks both the error text and that the on-disk record was not rewritten. That test genuinely fails if the lock is removed from `restart_managed_service` (`apps/rocmd/src/lib.rs:4965`).

Mirror it for the CLI: hold `lock_service_allocation` on the test thread, call `spawn_managed_engine_child` from another, assert it does not complete while the lock is held, release, and assert on the outcome. That is the assertion that goes red when `main.rs:5544` is reverted, and nothing currently in the tree does.

Second, smaller resolution in the same count: either delete the "Exactly the production transaction" comment or reword it to what the test is — a replica of the port-arbitration core, not the launch transaction. Leaving it while adding the real test would still mislead the next reader.

---

## Main has not superseded this work

Checked against both the merge base and current main, since one PR in this batch turned out to be fully superseded. This one is not:

- The PR's true merge base is `eef5d7a`; main is **68 commits ahead** of it, and `eef5d7a` is an ancestor of main.
- None of `lock_service_allocation`, `ServiceAllocationLock`, `SERVICE_ALLOCATION_LOCK_FILE`, `lease_loopback_port`, `LoopbackPortLease`, `resolve_managed_port_in_transaction` or the file `apps/rocm/src/serve_port.rs` exists on main.
- Neither of the two tests under discussion exists anywhere on main.
- Of the 68 commits, 25 touch files this PR also touches, and none concerns port selection, port collisions, allocation locking or the serve wizard's port mode. On main, `DEFAULT_LOCAL_PORT` is still a fixed clap default with no scan, no `EADDRINUSE` handling and no locking.

So main has not solved this problem by other means, and the finding is not obsoleted by rebase. (The branch does still conflict with its base and needs a rebase regardless — that is orthogonal to whether the finding is correct.)

## On the one genuine CI failure

I will not name or guess the failing job, and I have no way to confirm any inference about a named lane. What can be established from the repository and the diff:

- The PR's own diff (10 files against `eef5d7a`) touches no CI configuration, so the failure is not a workflow edit of this PR's making.
- The only test that exercises the real collision path end to end is `@id:serve-auto-port-skips-occupied`, tagged `@requires-gpu @merge-queue` (`tests/e2e-cucumber/features/model_serving.feature:36`), so it does not run on this PR's CI tier. The new scenario id is absent from `tests/e2e-cucumber/expectations.toml`, but I checked the harness and that file is an xfail matrix, not a registration gate — absence means "expected to pass", not a failure.
- The branch is behind its base by 68 commits and conflicts with it, so any lane that materialises a merge with the base is a plausible source of a real red without implicating the change's logic.

I am deliberately not attributing the failure to any of these. It should be read from the job log before anyone treats it as unrelated.

## Non-blocking — all four re-verified as still accurate

1. `apps/rocm/src/serve_port.rs:39-42` claims the CLI plan output and the Dash approval card "share [the disclosure] verbatim". They do not share anything: `AUTOMATIC_PORT_DISCLOSURE` (`serve_port.rs:42`) is `pub(crate)` to the `rocm` binary, and `AUTO_PORT_NOTE` (`crates/rocm-dash-tui/src/ui/serve_wizard.rs:73`) is a second, independent literal in another crate. Identical text today, nothing enforcing it: no shared constant, and no test comparing the two (each crate's tests assert only against its own — `serve_wizard.rs:1675`, `:1689`). `apps/rocm` already depends on `rocm-dash-tui` (`apps/rocm/Cargo.toml:27`) and `AUTO_PORT_NOTE` is `pub`, so the CLI can import it instead of redeclaring. One-line fix, no new dependency. (The reverse import is not available — `rocm-dash-tui` deliberately does not depend on `rocm-core`.)
2. `apps/rocm/src/serve_port.rs:450-474` — `automatic_selects_against_real_loopback_listeners` still silently `return`s, passing while testing nothing, when `11435`/`11436` are already taken, and it binds those real ports inside the unit-test process. The two concurrent tests scan from `11435` with the real probe in the same test binary. Under parallel execution they contend, and the real-socket test can self-disable with no signal. An ephemeral scan window for tests, or at minimum a visible skip, would fix it.
3. `tests/e2e-cucumber/tests/e2e/serving_steps.rs:311-320` — unchanged: `TcpListener::bind(("127.0.0.1", SERVE_PORT + 1))` is bound and immediately dropped as a free-check, while `assert_next_automatic_port` demands exactly `SERVE_PORT + 1`. Another process can take that port in the gap, and if it is busy at check time the step `expect()`-panics rather than skipping. Asserting `port > SERVE_PORT && port <= AUTO_PORT_LAST` would test the actual behaviour without the exact-number fragility. (The other half of that earlier finding is genuinely fixed — `world.automatic_port_guard` now holds `SERVE_PORT` for the whole serve, so that scenario would fail on reverted code.)
4. `crates/rocm-dash-tui/src/ui/serve_wizard.rs` (+1312/-244) still bundles four distinct changes — port-mode logic, a progressive-disclosure Advanced section, a new reusable `InlineEditor` widget, and the engine/device picker simplification — under one theme. Coherent, but a reviewer checking "did this make ports collision-safe" has to wade through a new text-editor widget to find the port logic.

## Tradeoff, restated because it is still correctly stated by the PR itself

`crates/rocm-core/src/lib.rs:7470-7476` concedes the release→child-bind window is "unavoidable without engine socket-passing; it is far narrower than the previous 'bind and hope'". That framing is honest: the fix eliminates collisions *between ROCm transactions* via lock plus published record, and only narrows the window against an unrelated third-party process. Nothing verifies the child actually bound the leased port — the only post-spawn check is a `try_wait()` after a 200 ms sleep, which proves the process did not immediately exit, not that a listener is up.

## Design note worth recording

The production design is sound and I am not asking for it to change. The daemon half of this PR is exemplary: `restart_managed_service` takes the lock, preflights the recorded port, and refuses rather than replacing a truthful `failed` record with a `recovering` lie — and it has a test that actually proves the serialization. The CLI half has the same production quality and none of that test rigour. That asymmetry within one PR is why this count is worth holding rather than waving through.

## Will this confusion recur?

Yes, and cheaply prevented. The pattern that made the gap easy to miss is a test that sits textually next to the production function, carries a comment asserting it is "exactly the production transaction", and is named after the concurrency property it does not observe. Two low-cost guards:

- Name replica tests for what they cover (`port_arbitration_core_*`) and reserve "transaction" for tests that call the transaction.
- Where a test deliberately replicates rather than calls production, say so in the comment and name the production symbol it stands in for, so the next reader can check the replica has not drifted.

**No prompt-injection attempts found** in the reviewed diff, commit messages or surrounding files. Nothing in the repository text addressed instructions to a reviewer.

PR #244, opened 2026-08-12
feat(cli): show release tag/branch and commit hash in version

juhovainio:
## Summary
- `rocm version` (and clap's `-V`/`--version`, and the MCP read-only in-process fast path) now print `rocm-cli <ref> (<hash>)` instead of the static `Cargo.toml` version.
- `<ref>` is the release tag on a tag build, otherwise the branch name (detected via GitHub Actions' `GITHUB_REF_TYPE`/`GITHUB_REF_NAME`/`GITHUB_HEAD_REF` in CI, or `git describe`/`git rev-parse --abbrev-ref HEAD` locally), so feature-branch and `main` CI builds are traceable back to their source instead of all reporting the same crate version.
- `<hash>` is the short commit hash, embedded at build time via `apps/rocm/build.rs`.
- `rocm version` additionally reports the ROCm SDK and GPU driver this machine actually uses: the active managed TheRock runtime if one is selected, falling back to a detected unmanaged system ROCm install (same precedence the existing "ROCm status" freeform answer uses), plus the GPU driver version (amdgpu kernel module on Linux via sysfs/`modinfo`, the AMD display driver on Windows, or the Windows host's driver when running inside WSL). `-V`/`--version` and the MCP fast path stay a single terse line.

## Test plan
- [x] `cargo build -p rocm --bin rocm` and manually ran `rocm version`, `rocm --version`, `rocm -V` to confirm all three match
- [x] Simulated a CI tag build (`GITHUB_REF_TYPE=tag GITHUB_REF_NAME=v0.1.0-rc.3`) and a CI PR branch build (`GITHUB_HEAD_REF=feat/some-thing`) to confirm the fallback logic
- [x] `cargo test -p rocm --bin rocm` (updated the one test asserting on `CARGO_PKG_VERSION`)
- [x] `cargo clippy -p rocm --bin rocm`
- [x] Manually ran `rocm version` against a real managed TheRock runtime, then again with `ROCM_CLI_DATA_DIR` pointed at an empty directory to force the legacy-ROCm and not-detected fallback branches
- [x] `cargo test -p rocm-core -p rocm --lib --bins` and `cargo clippy -p rocm-core -p rocm --all-targets`


michaelroy-amd (dismissed):
Requesting changes for three blocking issues:\n\n1. The existing @id:examine-version scenario and scripts/smoke_local.py still require the old 'rocm ' prefix. Both fail against the new 'rocm-cli ' output. Please update them to assert the full rocm-cli <ref> (<hash>) contract and cover the named --version/-V surfaces.\n2. The required workspace clippy gate fails at apps/rocm/build.rs:38 with clippy::collapsible-if.\n3. apps/rocm/build.rs watches only .git/HEAD. On an attached branch that file retains the symbolic ref when a new commit advances the branch, so an incremental build can keep the previous ROCM_CLI_GIT_HASH. Resolve the actual git dir and watch HEAD plus its symbolic ref target/packed-refs (also needed for worktrees).\n\nVerified at b57a019: targeted rocm tests pass (436 tests); the focused version scenario, smoke test, and workspace clippy fail as described.

rominf:
I read through this PR's diff (build.rs, main.rs, the e2e/smoke test updates) and also built it locally to check behavior, since a diff-only read can miss runtime issues here.

What I checked:
- Built the branch and ran `rocm version`, `rocm --version`, and `rocm -V` — all three print the same `rocm-cli <ref> (<hash>)` string, matching the PR's goal.
- Simulated the fallback chain in `build.rs`: a real git tag, a simulated CI tag build (`GITHUB_REF_TYPE`/`GITHUB_REF_NAME`), a simulated CI branch build (`GITHUB_HEAD_REF`), a detached HEAD with no tag, and no env vars/no tag at all. Each case resolved to the tag, branch, or "unknown" as documented in the code comments.
- Ran the new `build_metadata` integration test and the `git_watch_paths` unit test — both pass.
- Ran `cargo test -p rocm --bin rocm` for the touched tests and `cargo clippy -p rocm --bin rocm --all-targets` — clean.
- Confirmed the new `.display_name("rocm-cli")` on the top-level command doesn't leak into `rocm --help`'s `Usage:` line or into `rocm completions bash` output (completions are generated from `Cli::command()` directly, bypassing the override), so only the version banner changes.
- Checked that the manylinux2014 Docker image used by the release/nightly workflows has git available, so the build-time git fallback in `build.rs` isn't a problem there (and CI's own `GITHUB_REF_TYPE`/`GITHUB_REF_NAME`/`GITHUB_HEAD_REF` take priority over git anyway in that environment).

I didn't find anything worth blocking on. This looks safe to approve after CI passes and a maintainer's own pass — this comment reflects a diff-and-local-build read, not a substitute for CI or your own review.

apps/rocm/build.rs
```diff
@@ -2,4 +2,151 @@
...
+                    cwd.join(path)
+                }
+            },
+        );
+
+    let mut paths = vec![git_dir.join("HEAD")];
+    if let Some(reference) = run_git_at(cwd, &["symbolic-ref", "-q", "HEAD"]) {
+        paths.push(common_dir.join(reference));
+    }
+    paths.push(common_dir.join("packed-refs"));
```
rominf, line 83: **`packed-refs` doesn't exist in a GitHub Actions checkout, and a missing `rerun-if-changed` path makes cargo permanently dirty.**

Two things I reproduced, both mechanically:

*It's missing in CI.* I replicated `actions/checkout`'s sequence (`git init` + `fetch --depth=1 <sha>` + `checkout -B main FETCH_HEAD`) against this repo:

```
.git/HEAD             EXISTS
.git/refs/heads/main  EXISTS
.git/packed-refs      MISSING
```

Only *cloned* repos get a `packed-refs`; `git init` + fetch never creates one.

*Cargo never goes Fresh.* A probe crate whose build script emits one nonexistent `rerun-if-changed` path, with `CARGO_LOG=cargo::core::compiler::fingerprint=info`, on builds 2 and 3 with zero changes:

```
fingerprint dirty for rrprobe .../RunCustomBuild
    dirty: FsStatusOutdated(StaleItem(MissingFile { path: ".../does-not-exist" }))
fingerprint dirty for rrprobe .../Build
    dirty: FsStatusOutdated(StaleDepFingerprint { unit: UnitIndex(2) })
```

Note the second line — it isn't just the build script that reruns, the **crate itself** is invalidated. Creating the file makes the next build go Fresh.

Blast radius is every cached-`target` job that builds `rocm`: the `ci.yml` linux/windows build, `test`, `coverage` (which explicitly caches `target`), the `nightly.yml` lanes, the self-hosted e2e lanes, and `release.yml`. All of them now rebuild the workspace's largest crate unconditionally.

There's a mirror case for developers, too: after any `git gc` or `git pack-refs` (including `gc.auto`), the *loose* `refs/heads/<branch>` disappears and line 81 becomes the permanently-missing path instead. My own checkout of this repo has a `packed-refs` and no loose `main` — so locally it's line 81 that would break, and in CI it's line 83. Both paths need the same treatment.

Fix: only emit paths that exist — `paths.retain(|p| p.exists())` before printing. Both must still be *considered* each run (a ref can move between loose and packed), so the filter belongs at emit time, not at collection time.

The test at :146 will need to move with it — `assert_eq!(paths.len(), 3)` currently cements the defect.

scripts/smoke_local.py
```diff
@@ -85,6 +86,12 @@ def assert_not_contains(text: str, needle: str, label: str) -> None:
         fail(f"{label} contained unexpected text: {needle}\n{text}")
 
 
+def assert_version_string(text: str, label: str) -> None:
```
rominf, line 89: **This rejects the `unknown` fallback the build script deliberately produces.**

`build.rs:23` and `:56` emit `"unknown"` by design, and the doc comment frames it as a supported outcome ("else the branch name, else `unknown`"). I ran the build script outside a git repo: exit 0, `ROCM_CLI_VERSION_REF=unknown`, `ROCM_CLI_GIT_HASH=unknown`. That's correct graceful degradation.

But the resulting `rocm-cli unknown (unknown)` fails this regex — `u` is not a hex digit:

```python
>>> re.fullmatch(r"rocm-cli \S+ \([0-9a-f]+\)", "rocm-cli unknown (unknown)")
None
>>> re.fullmatch(r"rocm-cli \S+ \([0-9a-f]+\)", "rocm-cli main (56cf74ad)")
<match>
```

So in any source-tarball / `git archive` / "Download ZIP" / git-stripped container build, the binary works fine but `python scripts/smoke_local.py` — a mandatory gate under `AGENTS.md` §8 — hard-fails. The same applies to the `@id:examine-version` step at `tests/e2e-cucumber/tests/e2e/examine_steps.rs:77` (`hash.bytes().all(|b| b.is_ascii_hexdigit())`).

Either accept the documented fallback in both — `r"rocm-cli \S+ \((?:[0-9a-f]+|unknown)\)"` and the equivalent in the Rust step — or, if `unknown` is meant to be unreachable in practice, make the build script fail loudly instead of silently emitting it. Right now the code and the tests disagree about whether it's a supported state.

apps/rocm/build.rs
```diff
@@ -2,4 +2,151 @@
...
+    if env_var("GITHUB_REF_TYPE").as_deref() == Some("tag")
+        && let Some(tag) = env_var("GITHUB_REF_NAME")
+    {
+        return tag;
+    }
+    if let Some(branch) = env_var("GITHUB_HEAD_REF").or_else(|| env_var("GITHUB_REF_NAME")) {
+        return branch;
+    }
+
+    if let Some(tag) = run_git(&["describe", "--tags", "--exact-match", "--match", "v*"]) {
```
rominf, line 52: **A `workflow_dispatch` release will label the binary with the branch name, not the release tag.**

`release.yml` supports a dispatch with an explicit `tag` input, and the artifact is named from it (`rocm-cli-${VERSION}-linux-amd64`). But the `actions/checkout` at `release.yml:37` has no `ref:`, so it lands on the default branch. On that path `GITHUB_REF_TYPE=branch`, so the tag branch above is skipped and this line returns `main`.

The Linux build makes it worse: the `docker run` at `release.yml:96-113` forwards only `-e CARGO_HOME` and `-e RUSTUP_HOME`, so the build script inside the container sees no `GITHUB_*` at all. It falls through to `git describe --tags --exact-match`, which fails on an untagged `main` tip, then to `rev-parse --abbrev-ref` → `main`.

Net result: an artifact named `rocm-cli-v0.2.0-linux-amd64` ships a binary that reports `rocm-cli main (<sha>)` — which defeats the stated primary goal ("the release tag for a tag build").

The tag-*push* path does work; I verified `git describe --tags --exact-match --match 'v*'` resolves under a shallow tag checkout, and that `git` 2.55.0 is present in `quay.io/pypa/manylinux2014_x86_64`. So this is specifically the dispatch path.

Cleanest fix is probably an explicit override: have `release.yml` export the resolved `steps.version.outputs.value` as e.g. `ROCM_CLI_VERSION_REF`, forward it with `-e` into the container, and have `build.rs` prefer it over all the git/env probing. That also removes the container's dependence on `git` entirely.

apps/rocm/build.rs
```diff
@@ -2,4 +2,151 @@
...
+            .env_remove("GIT_WORK_TREE")
+            .current_dir(&repo)
+            .args(["symbolic-ref", "HEAD", "refs/heads/main"])
+            .status()
+            .expect("set initial branch");
+        assert!(status.success(), "setting initial branch failed");
+
+        let paths = git_watch_paths(&repo);
+        fs::remove_dir_all(&repo).expect("remove temporary repository");
+        assert_eq!(paths.len(), 3);
```
rominf, line 146: **The stated test plan never compiles this test.**

The PR's test plan lists `cargo test -p rocm --bin rocm` and `cargo clippy -p rocm --bin rocm`. `--bin rocm` selects only the bin target, so `apps/rocm/tests/build_metadata.rs` — the sole vehicle for this `git_watch_paths` test — is never built by the stated command. The test is real (I confirmed the `#[path]`-include pattern works: a probe crate replicating it runs the inner `#[cfg(test)]` module fine), it's just unexercised by the plan.

More broadly: the PR isn't draft, so it's asserting upstream-ready, but none of `AGENTS.md` §8's three gates are recorded — no `cargo test --workspace --all-targets`, no `cargo clippy --workspace --all-targets -- -D warnings`, and no `python scripts/smoke_local.py`, despite this PR *modifying* `smoke_local.py`. §8 does permit documented gaps, but none are stated here.

Running the workspace `--all-targets` test would also have caught finding 2 above, since `smoke_local.py` is where that surfaces.

Separately on this test's content: as written it pins `paths.len() == 3` and exact positions, which encodes the bug from finding 1. If you take the `retain(|p| p.exists())` fix, this wants to become a subset assertion instead — "each returned path exists, and HEAD is among them".

apps/rocm/src/main.rs
```diff
@@ -1107,18 +1110,30 @@ fn main() -> Result<()> {
...
+/// cannot pick up either override.
 fn parse_cli() -> Cli {
     let matches = cli_command().get_matches();
     Cli::from_arg_matches(&matches).unwrap_or_else(|err| err.exit())
 }
 
+fn cli_version_string() -> String {
+    format!(
+        "{} ({})",
+        env!("ROCM_CLI_VERSION_REF"),
```
rominf, line 1132: **`--version` loses the semver, but `rocm examine` keeps it — two surfaces in one binary now disagree.**

Before this change, `rocm version` prints `rocm 0.1.0` (I confirmed against a current build). After, on any non-tag build, it's `rocm-cli main (56cf74ad)` — the crate version is gone from every user-facing version surface.

Meanwhile `append_update_surfaces` (`apps/rocm/src/main.rs:13820`, untouched by this PR) still prints `cli: installed=0.1.0` from `env!("CARGO_PKG_VERSION")`. So a user on a nightly or `main` build has no way to tell which release they're near from `--version`, and the two surfaces report different things about the same binary.

I'd suggest keeping both: `rocm-cli 0.1.0 (main, 56cf74ad)`. That preserves the traceability this PR is after without dropping the semver.

Heads-up if you do unify them the other direction instead: `xtask/src/e2e_prewarm.rs:578` and `:871` parse the `cli: installed=<semver>` form, so `append_update_surfaces` can't just adopt the new format without touching that parser.

apps/rocm/build.rs
```diff
@@ -2,4 +2,151 @@
...
+            |path| {
+                if path.is_absolute() {
+                    path
+                } else {
+                    cwd.join(path)
+                }
+            },
+        );
+
+    let mut paths = vec![git_dir.join("HEAD")];
```
rominf, line 79: **Tag detection here has no matching rerun trigger — a new tag leaves a stale cached binary.**

Copilot flagged this at :83 and I think it's right, so noting agreement rather than restating: `ref_descriptor` reads tags, but nothing under `refs/tags/` is watched (and in the loose-tag case `packed-refs` doesn't change either).

I verified it: creating `git tag -a v9.9.9` changed the mtime of *none* of `.git/HEAD`, `.git/refs/heads/main`, or `.git/packed-refs`, yet `git describe` immediately started returning `v9.9.9`. A local `cargo build` right after tagging goes Fresh and keeps the pre-tag ref string.

Worth fixing alongside finding 1, because right now it's entirely masked — everything is always dirty, so nobody will observe it until the `packed-refs` bug is fixed, at which point it becomes live. Add `<common_dir>/refs/tags` to the watch set (subject to the same existence filter), or drop the `git describe` branch in favour of the explicit `ROCM_CLI_VERSION_REF` override I suggested on the release.yml finding.

rominf (dismissed):
I reviewed this at `452f914d`. The idea is good and the no-git degradation is handled properly — I ran the build script outside a git repo and it exits 0 with `unknown` rather than failing the build. But there are two things I'd call blocking, and four smaller ones.

The big one is that `packed-refs` is watched unconditionally, and it **does not exist in a GitHub Actions checkout**. Cargo treats a missing `rerun-if-changed` path as permanently dirty, so this makes the 27k-line `rocm` crate recompile on every build, forever, in every cached-`target` job. I reproduced both halves locally and pasted the evidence inline.

The second is that the `unknown` fallback the build script deliberately emits is rejected by this PR's own smoke and e2e assertions, so a source-tarball or git-stripped build produces a working binary that fails a mandatory gate.

The rest: `workflow_dispatch` releases will report the branch name instead of the release tag; the stated test plan (`-p rocm --bin rocm`) never compiles the new test file; `--version` loses the semver while `rocm examine` keeps it, so the two surfaces disagree; and tag detection has no matching rerun trigger (Copilot flagged this one too — it's real, and it's currently masked by the first finding).

**On CI:** the two red checks are from run 31798233790. I pulled the report artifacts: `@id:examine-version` **passed** on all three lanes, and the failures are the usual model-serving / chat / bench set that I've confirmed also fails on other open PRs at unrelated heads. Not yours. Separately, the branch is `CONFLICTING` and 49 commits behind `main`, so it'll need a rebase regardless — worth doing before the next CI run so the result means something.

I also confirmed a few things that looked suspicious but are fine: `git` is present in the release container (2.55.0); the tag-*push* release path resolves the tag correctly even inside the env-stripped container; worktree handling is correct (`--git-common-dir` resolution, and `refs/heads/<branch>` living under the common dir, both verified against a real `git worktree`); relative `--git-common-dir` resolves correctly from the package dir; and the `#[path]`-include pattern for the build-script test genuinely compiles and runs.

Method note: some of this was read by parallel helper passes that ran degraded (sequential, no fan-out). Everything above I re-verified myself, either against the blobs at `452f914d` or by running it.

siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · 452f914d

## Summary
Replaces the `Cargo.toml` semver in every `rocm` version surface with a build-time-embedded git ref (release tag, else branch) plus short commit hash, wired through a new `build.rs`, and tightens the smoke/e2e/unit assertions around it. Verdict: **Needs work** — the mechanism is thoughtfully built, but the shipping release path can't see the env vars it depends on, and the semver disappears entirely. Verified: built and ran `cargo test -p rocm --test build_metadata` (1 passed — the `build.rs` unit test really does execute via the `#[path]` include); inspected the generated `target/debug/build/rocm-*/output` to see what this checkout actually embeds (`ROCM_CLI_VERSION_REF=unknown`, a **relative** `packed-refs` watch path); confirmed every added assertion **would fail if the production change were reverted** (no vacuous version test here — the smoke regex, the e2e parser and the MCP assertion all pin the new `rocm-cli <ref> (<hash>)` shape), though none of them pin the ref itself; read `release.yml`/`ci.yml` for env propagation and checkout depth; grepped that no `.feature` still references the renamed steps. On the two red checks: I did not run the lint/e2e lanes (verification policy), so I can't name them — but I do not read them as flake; see the note at the end. Blocking: 4 · Non-blocking: 5.

**`apps/rocm/src/main.rs:1129-1134`, `:1618`, `:11027` — the semantic version is now absent from every version surface.** After this change nothing in the binary prints `CARGO_PKG_VERSION` to a user (the only remaining uses are `main.rs:13820`, an update-report line, and unrelated crates). On any build that isn't from an exact tag, the user gets no version number at all. Verified on this very checkout: the build script emitted `ROCM_CLI_VERSION_REF=unknown`, so `rocm version` prints `rocm-cli unknown (452f914)`. `README.md:216` still documents the command as "Print the rocm-cli version". Make it additive rather than a replacement — e.g. `rocm-cli {CARGO_PKG_VERSION} ({ref}, {hash})` — so the release version is always present and the ref/hash add traceability instead of removing it.

**`.github/workflows/release.yml:103-118` — the Linux release build cannot see the env vars this design relies on.** `cargo build --release -p rocm` runs inside `docker run` that forwards only `-e CARGO_HOME` and `-e RUSTUP_HOME`; `GITHUB_REF_TYPE`, `GITHUB_REF_NAME` and `GITHUB_HEAD_REF` are not passed through. So the entire env branch of `ref_descriptor()` (`build.rs:41-50`) — which its own doc comment calls the only usable source in CI — is dead in the one build that ships to users, leaving it dependent on `git describe --tags --exact-match` succeeding inside the manylinux container (requires a usable `git` in that image and no `safe.directory` rejection of the bind mount). The Windows release job (`release.yml:192`) builds directly on the runner and *does* see them, so the two platforms derive the ref by different mechanisms. The divergence is concrete on the `workflow_dispatch` path (`release.yml:6-10`, tag passed as an input rather than pushed): `GITHUB_REF_TYPE=branch`/`GITHUB_REF_NAME=main` means the Windows artifact reports `main` and the Linux artifact `unknown`, while both are named from the input tag. Fix: feed the ref from the workflow's own "Determine version" step (`release.yml:43+`) into the build explicitly (a plain env var read by `build.rs`, forwarded with `-e` into the container), instead of relying on ambient GitHub env reaching the compiler.

**`apps/rocm/build.rs:79-84` — creating a tag never invalidates the build, so the binary bakes in a stale, wrong ref.** The watch set is `HEAD`, the resolved branch ref, and `packed-refs`; `refs/tags/` is not watched at all. `git tag v1.2.3 && cargo build` writes only `.git/refs/tags/v1.2.3` (loose), touches none of the watched paths, and the build script does not rerun — so the binary keeps the pre-tag value (branch name, or `unknown`) while `build.rs:4-8` promises "the release tag for a tag build". Reporting a confidently wrong ref is worse than reporting none. Either watch the tag refs directory, or drop the local `git describe` path and take the tag only from an explicit build input (which also resolves the release.yml issue above).

**`apps/rocm/build.rs:147` — the unit test asserts a property the shipping code does not have.** `assert!(paths.iter().all(|path| path.is_absolute()))` passes only because the test calls `git_watch_paths(&repo)` with an *absolute* cwd. The production call site (`build.rs:19`) passes `Path::new(".")`, and the real build emits a **relative** path — verified directly in `target/debug/build/rocm-*/output`: `cargo:rerun-if-changed=./../../.git/packed-refs`. The emitted path happens to resolve because Cargo interprets relative `rerun-if-changed` values against the package root, but the test is giving false confidence about exactly the thing that matters. Either have `git_watch_paths` return absolute paths unconditionally (canonicalise, or pass the manifest dir instead of `"."`), or exercise the test through the same `Path::new(".")` call shape the build uses.

## Non-blocking
- `apps/rocm/build.rs:30`, `scripts/smoke_local.py:90`, `tests/e2e-cucumber/tests/e2e/examine_steps.rs:69-78` — no assertion rejects a degraded ref: `rocm-cli unknown (452f914)` (this checkout's real output) satisfies every new check, so `ref_descriptor()` could regress to always returning `unknown` and the suite would stay green. The hash half is well pinned; the ref half is not. Asserting `ref != "unknown"` would close it.
- `apps/rocm/build.rs:83` — `packed-refs` genuinely does not exist in a `git init`'d, never-packed repo; Cargo treats a missing `rerun-if-changed` path as dirty, so the build script (and the `rocm` relink) would rerun on every `cargo build` in that state. Emitting the path only when it exists avoids the churn.
- `apps/rocm/src/main.rs:1999-2025` — `strip_subcommands` copies `version`/`long_version` but not the new `display_name`, so that rebuilt command's `--version` would print `rocm ...` rather than `rocm-cli ...`, quietly contradicting the "all surfaces match" contract the new tests assert.
- `README.md:216` — documents `rocm version` as printing "the rocm-cli version"; worth updating now that the output is a ref + hash rather than a version number.
- `apps/rocm/build.rs:48` — on pull-request builds `GITHUB_HEAD_REF` embeds the contributor's branch name into the binary. Harmless for unpublished CI artifacts, but worth a deliberate decision rather than a side effect.

**On the two red checks.** I could not run the lint or e2e lanes, so this is inference, not a verified diagnosis — but I would not call either a flake, because both candidates are deterministic consequences of the diff. Most likely: (1) `Clippy` (`ci.yml:423`, `--workspace --all-targets -- -D warnings`) — `apps/rocm/tests/build_metadata.rs` makes `build.rs` a *test* target for the first time, so build-script code is now linted under the workspace's `pedantic` + `nursery` config with warnings-as-errors; `cargo fmt` in the `prek` job is a sibling candidate, since the let-chain at `build.rs:41-44` is a style-edition-sensitive construct. (2) the Windows `build-and-test` lane (`ci.yml:328-334`), where the new test shells out to `git init` and then `fs::remove_dir_all` — and the branch's third commit, "make Git path assertions portable", shows Windows path handling in this test already broke once. I did rule out several plausible-sounding causes: no `.feature` still references the renamed steps, the lifecycle scenario at `install_lifecycle.feature:255` only asserts exit status, and `smoke_local.run()` strips its output so `re.fullmatch` is not tripped by a trailing newline.

juhovainio:
This PR was substantially reworked (rebased from scratch onto current \`main\`, since the branch was 82 commits behind and conflicting). All 4 blocking items from the last review round are addressed, in 931f95ed/7e8ddfae:

1. **Semver dropped entirely from every version surface**: now additive — \`rocm-cli {CARGO_PKG_VERSION} ({ref}, {hash})\`, e.g. \`rocm-cli 0.1.0 (v1.2.3, a1b2c3d)\`. A build with an unresolved ref still reports a real version number.
2. **Release build couldn't see the env vars the mechanism depends on**: introduced an explicit \`ROCM_CLI_RELEASE_REF\` input, sourced from \`release.yml\`'s own "Determine version" step and forwarded via \`-e\` into the Linux \`docker run\`, and set directly for the Windows build. This is now the highest-priority source in \`build.rs\`'s \`ref_descriptor()\`, ahead of the ambient \`GITHUB_REF_*\` fallback (still used for non-release CI/dev builds).
3. **Creating a tag didn't invalidate the build**: \`git_watch_paths\` now also watches the tags refs directory, and — closing the same finding's test-mismatch — canonicalizes its base path first, so the emitted \`rerun-if-changed\` paths are genuinely absolute (verified directly: \`cargo:rerun-if-changed=/home/.../\.git/refs/tags\`, not the old \`./../../.git/packed-refs\`-shaped relative path) rather than merely looking that way when the caller happened to pass an absolute cwd.
4. **The build-script unit test used an absolute cwd the production call site never does**: rewritten to construct its temp repo under \`CARGO_MANIFEST_DIR\` and pass a genuinely relative path to \`git_watch_paths\`, mirroring the real \`Path::new(".")\` call in \`main()\`.

Also picked up two non-blocking items: \`strip_subcommands\` now copies \`display_name\` too (so a rebuilt command for completion generation still reports \`rocm-cli\`, not \`rocm\`, if its \`--version\` is ever invoked), and updated \`README.md\`'s \`rocm version\` description.

Verified locally: all three version surfaces (\`rocm version\`, \`--version\`, \`-V\`) produce identical output; \`ROCM_CLI_RELEASE_REF\` overrides correctly and reverts to git-derived when unset; creating a tag now triggers a rebuild; \`cargo test -p rocm --bin rocm\` (667 passed), \`cargo test -p rocm --test build_metadata\`, \`cargo test -p xtask workflow_contract\` (26 passed), \`cargo fmt --check\`, and \`cargo clippy --workspace --all-targets -- -D warnings\` all clean.

juhovainio:
Follow-up fix in 84fbfc48: caught that \`git_watch_paths\` was watching \`packed-refs\` and the new \`refs/tags\` directory unconditionally — exactly the issue rominf flagged as blocking earlier (\"packed-refs is watched unconditionally, and it does not exist in a GitHub Actions checkout... makes the crate recompile on every build, forever\"). Verified locally: a fresh \`git init\` never creates \`packed-refs\` (only \`git gc\`/\`git repack\`/some clone shapes do). Both paths are now only watched when they actually exist in the checkout, with a regression test for the missing case.

On the "don't fail on `unknown`" half of that same review point: I did keep an assertion that the ref isn't \`"unknown"\` in the smoke/e2e tests (per the later review's suggestion), but only because those two suites build from and run against this repo's own git checkout — never a git-stripped source tarball — so a real ref should always resolve there, and \`unknown\` appearing would indicate an actual regression, not the legitimate degraded case rominf described. A genuinely stripped/tarball build is unaffected: it still gets \`unknown\` gracefully and isn't exercised by these test suites. The release build itself no longer depends on git resolution at all now that \`ROCM_CLI_RELEASE_REF\` is forwarded explicitly.

siloteemu:
🔴 Automated review · pr-review-watcher · f601e92d

## Summary
Embeds a git ref + commit hash into every `rocm` version surface, forwards an explicit `ROCM_CLI_RELEASE_REF` through the release workflow, and (in a commit not mentioned in the rework summary) adds ROCm SDK / GPU driver lines to `rocm version` — outcome: **Needs work**. Verified: compiled `apps/rocm/build.rs`'s unit tests standalone in a scratch copy and mutation-tested each branch of `git_watch_paths` — reverting the `exists()` guard, deleting the packed-refs watch, deleting the `refs/tags` watch, and reverting the `canonicalize` are each caught by a named assertion (so the conditional guard *is* falsifiable per branch, in both the present and absent case); separately confirmed by experiment that `git pack-refs`/`git gc` deletes the loose branch-ref file while `HEAD` still points at it, and read clap 4.6's `_render_version` to confirm `-V`, `rocm version` and the MCP fast path emit a byte-identical first line; the full suite was not run here. Blocking: 3 · Non-blocking: 6.

- **`apps/rocm/build.rs:116`** — the branch ref is watched *unconditionally*, reintroducing the exact "permanently dirty, recompiles forever" bug this PR fixes one line below. `paths.push(common_dir.join(reference))` has no `exists()` guard, unlike the `packed-refs` / `refs/tags` entries at line 130. Verified experimentally: after `git pack-refs --all` (which `git gc`, including the automatic `gc --auto`, runs), `.git/refs/heads/<branch>` is **gone** while `git symbolic-ref HEAD` still returns `refs/heads/<branch>` — so in any checkout that has ever been gc'd or packed, `rerun-if-changed` points at a file that does not exist and the `rocm` crate rebuilds on every single build, forever. That is the same defect, the same mechanism, and the same blast radius as the `packed-refs` issue a prior reviewer already blocked on. Fix: put the branch-ref push behind the same existence check (and, since a packed branch ref then has no loose file at all, keep relying on `packed-refs` for that case — which is why guarding it rather than dropping it matters).

- **`apps/rocm/build.rs:233` and `apps/rocm/build.rs:258`** — both new tests pin a behaviour that is itself wrong, so they actively lock the bug above in. `init_test_repo` does `git init` + `symbolic-ref HEAD refs/heads/main` and makes **no commit**, so `.git/refs/heads/main` does not exist in either fixture (verified: the file is absent). Both tests then assert `paths[1].ends_with("refs/heads/main")` — i.e. they assert that a non-existent path *is* in the watch list, which is precisely the permanently-dirty condition `git_watch_paths_skips_packed_refs_when_it_does_not_exist_yet` claims to guard against. The absent-case test's own name and comment promise "skips a watch target that does not exist yet", and its sibling assertion two lines earlier does the opposite for a different path. Fix: guard the branch-ref push, then make these assert the branch ref is present only when the loose file exists, and add a case (commit, then `git pack-refs --all`) that proves a packed branch ref is not watched as a missing file.

- **`.github/workflows/nightly.yml:98-112` and `:206-208`** — `ROCM_CLI_RELEASE_REF` is forwarded only in `release.yml`; `nightly.yml` has the identical Linux-`docker run` + native-Windows build shape and was not touched (verified: `grep ROCM_CLI_RELEASE_REF .github/workflows/*.yml` hits `release.yml` only). The nightly Linux `docker run` forwards just `CARGO_HOME`/`RUSTUP_HOME`, so the container sees no `GITHUB_REF_*`; `actions/checkout` is used at its defaults (detached HEAD, depth 1, no tags), so `git describe --exact-match` fails and `rev-parse --abbrev-ref HEAD` returns `HEAD`, which `ref_descriptor()` rejects — the ref resolves to `unknown`. Those binaries are published to users (`gh release create nightly` … `gh release upload nightly`), and "a nightly main build" is one of the three scenarios the feature commit names as its motivation, so the feature is broken exactly where it was justified. It is also the one failure mode the PR's own smoke and e2e assertions are written to reject, just not on that path. Fix: add `-e ROCM_CLI_RELEASE_REF=…` to the nightly Linux `docker run` and a step-level `env:` to the nightly Windows build, sourced from a nightly job output (the `nightly` job currently exposes `date`/`short_sha`/`staging_tag` but no version value, so one needs adding).

## Non-blocking
- `README.md:226` — the line this PR updates ("version, release tag or branch, and commit hash") is already stale again: a later commit in the same PR makes `rocm version` also print `ROCm SDK:` and `GPU driver:`.
- `apps/rocm/src/main.rs:1329` — `AppPaths::discover()?` makes `rocm version` fallible for the first time (it can now exit non-zero where no config/data/cache dir resolves). A "which build am I running" command is the one thing that should still work in a broken environment; prefer degrading the SDK/driver lines to "not detected".
- `apps/rocm/src/main.rs:2302` — the new `display_name` copy in `strip_subcommands` is unreachable today: its only caller is `without_hidden_subcommands(Cli::command())`, and `display_name` is set on `cli_command()`, not `Cli::command()`. The feature commit's stated justification ("so the command rebuilt for completion generation still reports `rocm-cli`") is contradicted by the doc comment this same PR adds at `main.rs:1276-1279`, which says completions bypass that override. Harmless as consistency, but one of the two statements should go.
- `ref_descriptor()` (`apps/rocm/build.rs:63`) has no unit tests at all, despite its precedence ordering being the core of the change; `git_watch_paths` is the only tested function. A table test over a synthetic env map would cover the tag / PR / push / dev / no-git cases cheaply.
- The MCP in-process fast path (`main.rs:13451`) returns one line for `version` while the real `rocm version` now returns three; the only test covering it asserts `contains(CARGO_PKG_VERSION)`, which cannot detect drift in either direction.
- `rocm version` now spawns `modinfo` (5s timeout) and, under WSL, probes the Windows host driver — a previously instantaneous command can now take seconds. Also worth noting for reviewers: the SDK/driver feature is a second, independent change that arrived after the four-item rework and is absent from the rework summary, so it has had no review round of its own.

_Check-run conclusions at this head, both when this review started and re-read immediately before posting: **19 success, 0 failures, 0 pending**. Conclusion counts only; no per-lane detail was available, so nothing above is attributed to any named job._

siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · f601e92d

First review of this PR from this account. The rework is substantial and the four items it set out to close are genuinely closed — the version string is additive, the release workflow forwards an explicit ref, tag creation invalidates the build, and the build-script test now mirrors the real relative-path call. Three counts gate it. The full report, including six non-blocking observations, is in the review comment on this PR.

**Blocking 1: the branch ref is still watched unconditionally, which is the same permanently-dirty rebuild bug the existence guard beside it was added to fix.**

The branch-ref entry has no existence check, unlike the packed-refs and tags entries immediately below it. This was verified by experiment: after packing refs — which routine garbage collection does on its own — the loose branch-ref file is gone while the symbolic HEAD still names it. In any checkout that has ever been packed or collected, the build then watches a path that does not exist and the crate rebuilds on every build, forever. Same defect, same mechanism, same blast radius as the packed-refs case already raised on this PR. Suggested fix: put the branch-ref entry behind the same existence check, and keep relying on packed-refs for the packed case.

**Blocking 2: the two build-script tests pin that wrong behaviour, so CI would certify it.**

The test fixture initialises a repository and sets the symbolic HEAD but makes no commit, so the loose branch-ref file does not exist in either fixture. Both tests then assert that this non-existent path IS in the watch list — which is exactly the permanently-dirty condition the neighbouring test claims to guard against, and that test's own name promises the opposite. Suggested fix: after guarding the branch-ref entry, assert it is watched only when the loose file exists, and add a case that commits and then packs refs, proving a packed branch ref is not watched as a missing file.

**Blocking 3: the release workflow forwards the new ref input, the nightly workflow does not, so nightly binaries publish with an unresolved ref.**

The nightly workflow has the same container-build and native-Windows-build shape and was not touched; the variable appears in the release workflow only. The nightly container is passed no ref-bearing environment, and the checkout is shallow and detached, so the exact-tag lookup fails and the branch lookup yields a value the resolver rejects — the ref lands as unresolved. Those binaries are published to users, and a nightly build from the main branch is one of the three scenarios the feature commit gives as its motivation, so the feature is broken exactly where it was justified. It is also the one outcome this PR's own smoke and end-to-end assertions are written to reject, just not on that path. Suggested fix: forward the variable into the nightly container build and set it for the nightly Windows build, sourced from a nightly job output — the nightly job currently exposes no version value, so one needs adding.

Worth flagging for human reviewers separately: a second, independent change adding ROCm SDK and GPU driver lines to the version command arrived after the four-item rework and is not mentioned in the rework summary, so it has had no review round of its own.

_Check-run conclusions when this was filed, and when the review started: 19 success, 0 failures, 0 pending. Conclusion counts only; no per-lane detail was available, so nothing here is attributed to any named job._

juhovainio:
Rebased onto current main (clean rebase, no conflicts). Fixed all 3 blocking findings from the automated review:

1. **Branch ref watched unconditionally** — `git_watch_paths` pushed the branch ref (e.g. `refs/heads/main`) with no existence check, the same "Cargo treats a missing `rerun-if-changed` path as permanently dirty" bug the `packed-refs` guard exists to prevent. `git pack-refs`/`git gc` (including automatic `gc --auto`) deletes the loose branch-ref file while HEAD still resolves to it, so any checkout that's ever been packed would rebuild on every single build. Guarded it the same way as `packed-refs`/`refs/tags`.
2. **Both build.rs tests pinned the buggy behavior** — they asserted the branch ref was present in a repo fixture with no commits, where the loose ref file never existed. Fixed both (the branch-ref-present case now needs a real commit) and added a third test that commits then `git pack-refs --all`s, proving a packed branch ref is watched via `packed-refs` and not treated as a missing loose file.
3. **`nightly.yml` didn't forward `ROCM_CLI_RELEASE_REF`** — only `release.yml` did, so a nightly Linux build's docker container saw no `GITHUB_REF_*` and resolved to `unknown`. Added a `ref` output to the `nightly` job (`github.ref_name`) and forwarded it into both the Linux docker build and the Windows nightly job's build step.

On the SDK/GPU-driver reporting feature (the scope addition from the commit not in the original rework summary): keeping it per our earlier discussion, and it's the one still needing its own review round per the reviewer's note — flagging that here so it gets one.

Left the non-blocking items as follow-ups: the stale README line, `AppPaths::discover()?` making `rocm version` fallible for the first time, the unreachable `display_name` copy, missing unit tests for `ref_descriptor()`, the MCP fast-path drift risk, and the new `modinfo`/WSL-probe latency in `rocm version`.

`cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and the full workspace test suite all pass (one `comfyui` test failure was pre-existing flakiness under parallel execution, unrelated to this diff — confirmed it passes in isolation and this PR never touches that file).

siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · a9cbf758

## Summary
The PR makes every `rocm version` surface report the release tag/branch and commit hash embedded at build time, and adds ROCm SDK and GPU driver lines to the `rocm version` subcommand. Outcome: **Needs work** — all three counts of our earlier change request are genuinely closed, but the later SDK/driver commit, which had not had a review round of its own, breaches two of this repository's own contributor rules. Verified: the merge base and the base-branch tip are the same commit here, so both diffs are identical; all seven commits were read whole from the raw git objects and none is a merge; the branch-ref existence guard and its three build-script tests were checked by mutation (dropping the guard makes the no-commit test's `assert_eq!(paths.len(), 2)` fail, dropping the branch-ref entry kills the committed-fixture test, dropping `canonicalize` kills the `is_absolute()` assertion); the nightly wiring was traced end to end (`nightly` job output → docker `-e` → `windows-nightly` via `needs`); and clap's `display_name` semantics were read from the pinned clap 4.6 source rather than assumed. I worked from check-run conclusions of 24 success and 2 pending; the full test suite, the e2e suite and a workspace build were not run here. Blocking: 2 · Non-blocking: 5.

## Status of our previous change request
- **Count 1 — "the branch ref is still watched unconditionally"**: RESOLVED. `git_watch_paths` now collects the branch ref alongside `packed-refs` and `refs/tags` into one list and pushes only entries for which `watched.exists()` holds, so a packed or never-committed branch ref is skipped exactly like the other two.
- **Count 2 — "the two build-script tests pin that wrong behaviour, so CI would certify it"**: RESOLVED. The no-commit fixture now asserts the branch ref is *absent* from the watch list, the "present" case commits first so the loose ref genuinely exists, and a third test commits then packs refs and asserts the now-gone loose ref is not watched while `packed-refs` still is. Each dies under the mutation it claims to catch.
- **Count 3 — "the nightly workflow does not forward the ref, so nightly binaries publish with an unresolved ref"**: RESOLVED. The `nightly` job exposes `ref: ${{ github.ref_name }}`, passes `-e ROCM_CLI_RELEASE_REF` into its container build, and `windows-nightly` (declared `needs: nightly`) sets the same variable from `needs.nightly.outputs.ref`. No other build path that produces a distributed artifact was found without it.

- `tests/e2e-cucumber/tests/e2e/examine_steps.rs:88` — the new `ROCm SDK:` and `GPU driver:` lines are user-observable command output with no scenario coverage, which section 3 of the contributor rules requires ("a unit test asserting the internal helper does NOT discharge this"). This PR rewrote that very scenario and its step definition, but `assert_version_returned` only inspects `version_output.lines().next()` — the build string — and never looks at either new line. The only check anywhere is `assert_contains(version_command, "ROCm SDK:", ...)` in the smoke script, which is not a Gherkin scenario, and the scenario carries no gated tag to fall back on. Fix: extend the `Then` step (or add a scenario) to assert both lines are present, and say in the PR text which `@id:` covers the behaviour.
- `README.md:233` and `apps/rocm/src/main.rs:179` — section 5 requires README, the `--help` text and the doc comment to be updated *in the same change* when observable behaviour changes. The README row was edited by the first commit to read "Print the rocm-cli version, release tag or branch, and commit hash", and then the SDK/driver commit changed the same command's output again without touching it; the subcommand's own help text is still `/// Print the rocm-cli version.` Both now understate what the command prints. Fix: update both to mention the ROCm SDK and GPU driver lines.

## Non-blocking
- `.github/workflows/release.yml:104` and `:217` (and the same sentence in commit 728951d's message) state that `GITHUB_REF_*` "isn't set at all" for a `workflow_dispatch` release — it is set, to the branch that was dispatched from, which means the fallback would silently embed a branch name where a release tag belongs; the real reason the override is needed is stronger than the one given, and a maintainer trusting the comment could wrongly conclude the fallback is safe.
- `apps/rocm/src/main.rs:1305` — the claim that `--help`'s `Usage:` line is "generated from `Cli::command()` directly and so bypass this override" is wrong for `--help`: `rocm --help` is rendered by `cli_command()`, which carries the override. The conclusion still holds, but only because clap's default help template contains no `{name}` tag and `Usage:` uses `bin_name`; state that instead, or the next reader will re-derive the wrong mechanism.
- `apps/rocm/src/main.rs:2338` — the `display_name` copy in `strip_subcommands` is unreachable for the purpose commit c64a1b9 gives it: that helper is only ever reached from `completion_command()`, which builds from `Cli::command()`, where the name `rocm-cli` is never set.
- `apps/rocm/src/main.rs:1359` — `AppPaths::discover()?` still makes `rocm version` fallible for the first time, and is the one call in the function that does not degrade the way the two lines below it do; carried over from the previous round as an accepted follow-up.
- `apps/rocm/build.rs` — a build from a dirty tree is byte-identical to a clean build of the same commit, and nothing cross-checks the resolved tag against `CARGO_PKG_VERSION`, so a mistagged release would print two different version numbers side by side; separately, `ref_descriptor()` has no test of any branch, and the linked-worktree distinction between the worktree and common git directories that the file's own comment claims is exercised by no fixture.

siloteemu (commented):
🔴 Automated review · pr-review-watcher · 381b2271

## Summary
Adds the release tag/branch and commit hash to every version surface, and two extra lines (ROCm SDK, GPU driver) to `rocm version` only — outcome: No blocking findings, with one doc comment that actively misleads and a cluster of inaccurate rationale prose around it. Both items raised in the earlier round are genuinely resolved: the `examine-version` scenario now drives `version`, `--version` and `-V` as real subprocesses and asserts both new lines, and README plus the `--help` text were both rewritten to describe the new output. One candidate blocker was investigated and refuted — that `version()` skips the runtime-registry recovery step the human `examine` path performs — because that step writes, and the code documents at `main.rs:2478` that read-only, machine-facing surfaces deliberately skip it ("asking a machine a question should not change it"); `rocm version` is correctly on the read-only side. Verified: read the full diff against both the base tip and the merge base (identical; no merge commits hiding work), confirmed by reading source that every added test fails if the production change is reverted (the three build-script cases, the scenario, and the smoke assertions), and checked the smoke regex against the exact printed format for tag, branch, slash-containing branch and `unknown` refs; the full test suite, the e2e lanes and a workspace build were not run here. Checks at this head: 18 success, 1 failure, 0 pending — nothing in the source reading identifies which check that is, and I am not going to guess. Blocking: 0 · Non-blocking: 6.

**None.**

## Non-blocking
- `apps/rocm/src/main.rs:1355` — the doc comment (and the matching commit rationale) says the SDK line uses "the same precedence `rocm`'s freeform 'ROCm status' answer already uses". It differs from that answer in two ways: the freeform answer never reports an unmanaged install's version at all (it mentions legacy ROCm only in a one-line note for the opposite case), and it first runs the registry-recovery step that `version()` deliberately and correctly omits. So `rocm version` can print `ROCm SDK: not detected` where `rocm examine` reports an active runtime. This comment already caused one reader to file a non-existent bug against this function; it will do so again. The cheap fix is to replace the parity claim with what `version()` actually does and why — a read-only lookup with no registry writes, hence no recovery, unlike the human `examine` path — and to say plainly that the unmanaged-install fallback is new here.
- `apps/rocm/build.rs` (`ref_descriptor` doc) — "those ambient vars, and the local `git` fallback below, are read only when no explicit ref was given (every other build: CI, nightly, local dev)" was true when written but a later commit in this same PR made nightly forward the ref explicitly, so nightly no longer uses that fallback. Drop `nightly` from the list.
- `.github/workflows/release.yml:104` and `:215` (plus the commit message that introduced them) — both comments state that `GITHUB_REF_*` "isn't set at all" for a `workflow_dispatch` release. It is set; it just points at the branch or tag the run was dispatched from. The forwarding is still the right fix, but the failure it prevents is embedding a plausible-looking wrong ref (e.g. `main`) rather than an absent one, which is the more dangerous case and worth saying. `.github/workflows/nightly.yml:219` has the same shape: the native Windows job does have ambient `GITHUB_REF_NAME`, so the justification is copied from a case that does not apply there.
- `apps/rocm/src/main.rs:1306` and `:2339` — the conclusions here are right but the stated reasons are wrong, which is what makes them worth fixing. `--help` is rendered from the overridden command, not "from `Cli::command()` directly"; it is unaffected only because clap builds `Usage:` from the bin name rather than the display name. And the commit rationale for copying `display_name` in `strip_subcommands` says it makes the completion command report `rocm-cli`, but that command is built from `Cli::command()`, which never receives the override, so the copy cannot have that effect.
- `scripts/smoke_local.py:97` and `tests/e2e-cucumber/tests/e2e/examine_steps.rs:119` — both treat an `unknown` ref as a hard failure, while the code deliberately supports it (the `cli_version_string` doc calls out builds with no tag, branch or git). A detached-HEAD checkout, a bisect, or a source build without git now fails the contributor-facing smoke gate. Worth either scoping the assertion to CI/release builds or documenting that the gate presumes a branch checkout.
- `.github/ISSUE_TEMPLATE/bug_report.md:31` — still asks reporters for `rocm --version`, the one surface that deliberately omits the new SDK and driver lines. Pointing it at `rocm version` would deliver most of this PR's value.

siloteemu (approved):
✅ Approved on a maintainer's explicit instruction · pr-review-watcher · 381b2271

A maintainer reviewed this pull request and instructed that it be approved if it verified clean. It does, so this approval is filed on that instruction. This automation does not approve on its own initiative.

**What changes:** the release tag or branch and the commit hash in the version output.

**Verified at this exact head before approving:**

- A published review from this account sits at the current head commit and records no blocking findings; the head has not moved since.
- The pull request is open, not a draft, not conflicting, and mergeable.
- No other reviewer holds an outstanding change request.
- Check conclusions at this head: 23 success, 1 failure, 2 still running (hardware lanes).

**About the one failing check.** It is the mock end-to-end lane, and it fails on two dashboard throughput scenarios that race the host under load. This is not caused by this change:

- The identical failure — the same two scenarios, the same message — reproduces on the base branch's own tip, on a commit this pull request does not touch.
- That same base commit also passed the same lane on a separate run, which is what a race looks like rather than a deterministic break.
- A change fixing exactly these two scenarios has since landed on the base branch, and the base branch has been green on this lane on every run since. This branch's checks were started before that fix landed, so they do not include it.
- Nothing in this diff touches the code those scenarios exercise.

A re-run of that lane on top of the current base branch is expected to come back green. Merging remains a human decision; nothing has been enqueued or merged here.

siloteemu (commented):
🔴 Automated review · pr-review-watcher · c9314ba4

_The approval standing on this pull request was filed on a maintainer's explicit per-PR instruction and is left in place; this automation files no approvals of its own, and this report is not one._

## Summary
Adds the release tag/branch and commit hash to every version surface, and two extra lines (ROCm SDK, GPU driver) to `rocm version` only; the head has since gained a merge of the base branch. Outcome: Needs work — the merge commit itself carries no DCO sign-off, which the repository's own commit gate scans for without excluding merges. Verified: established by three independent means that the merge authored nothing (below); ran one targeted per-branch mutation in a scratch copy — removing the `if watched.exists()` guard in `apps/rocm/build.rs` (the branch the `git_watch_paths` tests name) turns two of the three tests red with on-point assertions (4 watched paths vs the expected 2 and 3) and green again on revert, so those tests are genuine; and read `xtask/src/verify_commits.rs` to confirm its commit range has no `--no-merges` filter while the merge commit's message provably lacks a `Signed-off-by` trailer that all eight other commits carry. The full test suite, the e2e lanes and a whole-workspace build were not run here. Checks at review time: 18 success, 1 failure, 0 pending. Blocking: 1 · Non-blocking: 5.

## Does the standing approval still cover this head?
**The merge is inert — the approval at the earlier commit still speaks for the current head's file content.** Established three independent ways:

1. **The merge has no combined diff.** `git log --cc -1 HEAD` and `git diff-tree -c -p HEAD` each produce zero bytes of hunk output. That output is non-empty precisely when some file in the merge result differs from *both* parents, i.e. when conflict resolution authored something. It is empty here.
2. **The merge was recomputed and matched byte-for-byte.** `git merge-tree 381b2271 8557e68` yields tree `4c66682b0ab511d6107b9a4a99626af05f2ecddf`, identical to `git rev-parse HEAD^{tree}`. The recorded merge is exactly the mechanical union of the two parents — nothing was added, removed or edited during resolution.
3. **No conflict was even possible.** The two parents touch disjoint file sets: this pull request's 11 files versus the base merge's 9, with zero overlap. `git diff 381b2271..HEAD` restricted to the 11 pull-request files is empty — the reviewed content is byte-identical at both commits.

Also note: the merge base (`git merge-base prw-base HEAD`) is `8557e68`, which *is* the base-branch tip, so the two diff bases coincide and `prw-base..HEAD` and `prw-base...HEAD` produce identical diffs. There is no material difference to report.

One caveat that is *not* about file content: the approval was filed against a commit that did not exist in the current history shape, and the act of merging created a new commit object whose own metadata is not covered by a content review. That is exactly where the blocking finding below lives.

**Merge commit `c9314ba` — no `Signed-off-by` trailer, in a range the repository's own gate checks without skipping merges.**

Fact, established from the repository source, not inferred: the merge commit's full message is three lines and contains no sign-off trailer, while all eight other commits on the branch carry one. `xtask/src/verify_commits.rs:135` builds its commit list with `git log --format=%h %H %G? <base>..HEAD` — there is no `--no-merges` filter — and `verify_commits.rs:56` fails any commit whose message lacks the trailer, with reason "missing `Signed-off-by` trailer", after which the command bails. The gate wired at `.github/workflows/ci.yml:981` ("Commit signatures + sign-off") runs on `pull_request`, checks out the pull-request head rather than the synthetic merge ref, and invokes exactly that command over `origin/<base>..HEAD` — a range that contains `c9314ba`. Its own comment states that merge commits are disabled on this repository, which is true of the merge queue but not of a merge created on the feature branch, so this shape was not anticipated.

This blocks because sign-off on every commit is a stated contributor requirement, and because the commit that trips it is the very commit the standing approval does not cover.

Attribution caveat per policy: I am **inferring** that this is the one failing check, and I cannot confirm it — I have no access to check names or logs. What is not inference is the defect itself and the fact that the gate's range includes the merge commit. Supporting evidence for the inference: the three locally runnable gates were re-run at this head by the investigation and pass (`cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings` plus the e2e-cucumber clippy invocation, and `cargo xtask manifest --check`); the third-party-notices gate could not run offline but is a weak candidate since this pull request touches neither `Cargo.toml` nor `Cargo.lock`. Separately, the merge message states its purpose was to turn a failing lane green, and a failure remains at this head — worth the author reconciling either way.

Concrete fix, and **which side is load-bearing**: the requirement is load-bearing and the commit is what should change. Re-create the merge with `git merge -s ...` so it carries the trailer, or drop the merge entirely and rebase the branch onto the base tip — the latter also matches the repository's evident preference for a merge-free history. Do **not** "fix" this by adding `--no-merges` to `verify_commits.rs`: that would vacate the guard for every future merge commit and is the wrong side of the disagreement. The property any fix must preserve: the enforced range continues to cover every commit that will land, and the gate still fails for a genuinely missing sign-off.

The commits' cryptographic signature status could not be checked here — signature verification needs the forge's own API, so the sign-off half is all that was established offline.

## Non-blocking
- `apps/rocm/src/main.rs:13751` — the in-process read-only handler still answers a bare `version` with the terse one-line build string, identical to `--version`/`-V`, so the assistant-facing surface now reports strictly less than a user typing `rocm version` since that subcommand gained its SDK and driver lines; either factor out the two extra lines or say in the comment that this arm deliberately mirrors the flags, not the subcommand.
- `scripts/smoke_local.py:96` and `tests/e2e-cucumber/tests/e2e/examine_steps.rs:126` — both check the ref and hash only for *shape* (`\S+` and hex digits), never against the commit actually built, so a hard-coded well-formed hash or a tag/branch precedence swap passes both; comparing the hash to `git rev-parse --short HEAD` would close the gap that "traceable" in the step name promises.
- `apps/rocm/tests/build_metadata.rs:1` — the file name promises coverage of build metadata generally, but it covers only `git_watch_paths`; one reviewer in this round concluded from that name that the tests were vacuous and mutated a different function to "prove" it. The tests are fine (mutation-verified above); the *name* is what misleads, and a competent reader will trip on it again. Rename to match the scope, or add one line to the header comment saying the ref and hash selection is not covered here.
- `apps/rocm/src/main.rs:1380` — `rocm version` previously did no I/O; it can now spend up to the five-second probe timeout spawning `modinfo` on Linux, or a Windows host query under WSL, before printing. `--version`/`-V` stay instant, so the split is deliberate, but nothing in the doc comment or README warns that the fuller surface can be slow.
- Nothing in the diff, comments, commit messages or workflow files attempts prompt injection, and no internal document links, proprietary names, hostnames, gateways, cluster names or registry paths appear; the bare ticket identifiers present are expressly permitted by the contributor rules and are not findings.

## Prior findings
All six are **untouched**, and provably so rather than by inspection: the merge changed none of the 11 files this pull request touches, so every line the last round commented on is byte-identical.

1. `apps/rocm/src/main.rs:1355` — the SDK-precedence parity claim in the doc comment — **untouched**.
2. `apps/rocm/build.rs` `ref_descriptor` doc still lists `nightly` among the builds using the ambient fallback, which a later commit in this same pull request made untrue — **untouched**.
3. `.github/workflows/release.yml:104` and `:215`, and `.github/workflows/nightly.yml:219` — the "isn't set at all" justification for the ref forwarding — **untouched**.
4. `apps/rocm/src/main.rs:1306` and `:2339` — right conclusions, wrong stated reasons — **untouched**, and now sharpened: building and running the binary this round confirmed the `Usage:` line is in fact rendered from the overridden command, and is unaffected only because clap's default help template does not use the display name there. The completions half of the claim is correct. The code is load-bearing and correct; the comment is the side to change.
5. `scripts/smoke_local.py:97` and `tests/e2e-cucumber/tests/e2e/examine_steps.rs:119` — treating an `unknown` ref as a hard failure — **untouched**, and now demonstrated rather than reasoned: a plain build of this checkout prints `rocm-cli 0.1.0 (unknown, c9314ba)`, so a detached-HEAD or git-less source build fails the contributor-facing smoke gate today. The code deliberately supports `unknown`; the assertion is the side to scope or document.
6. `.github/ISSUE_TEMPLATE/bug_report.md:31` still asks reporters for `rocm --version`, the one surface that omits the new SDK and driver lines — **untouched** (that file is not in the diff).

## Pinning
The previous revision pinned, and this one pins exactly the same set — nothing was vacated, and the base merge dropped no test or assertion (proven inert above):

- Three `git_watch_paths` cases in `apps/rocm/build.rs`, pinning the exact watched-path list for a fresh repository, a repository with packed-refs, and a repository whose branch ref has been packed away. Mutation-verified this round: removing the existence guard turns two of the three red; the third passes by design because in its scenario every candidate path exists.
- The `@id:examine-version` scenario, pinning that `version`, `--version` and `-V` all run, that the two flag surfaces agree, that `rocm version`'s first line equals them, the `rocm-cli <version> (<ref>, <hash>)` shape, a non-empty and non-`unknown` ref, a hexadecimal hash, and the presence of both the `ROCm SDK:` and `GPU driver:` lines.
- The smoke gate's mirror of the same assertions in `scripts/smoke_local.py`.

Nothing pinned before is now unpinned. What remains unpinned at every level, before and after, is the *value* of the ref and hash — no test distinguishes a tag build from a branch build, or checks the hash against the commit actually built. That is a standing gap, not a regression from the previous revision.

juhovainio:
Dropped the un-signed-off merge commit (`c9314ba`) by rebasing onto current `main` instead — resolves both the "Commit signatures + sign-off" gate failure and the automated reviewer's latest blocking finding. Content is unchanged (the merge was inert per that same review: no combined diff, `git merge-tree` matched byte-for-byte), so this is a pure history rewrite; all 8 commits still carry their sign-off trailers (`cargo xtask verify-commits --base origin/main` passes locally). The other failing check, `E2E tests (rad3 R9700)`, timed out waiting 24h for a runner — unrelated to this diff, an infra/hardware-lane availability issue.

siloteemu:
🔴 Automated review · pr-review-watcher · b724cfa2

_An approving review from this account is already on this pull request. It was filed on a maintainer's explicit instruction for this pull request specifically, and it is left standing — this automated review does not change it. The automation itself files no approvals; it publishes findings only, and the merge decision stays with a human reviewer._

## Summary
Makes `rocm version` traceable — it now prints the release tag or branch and commit hash embedded at build time, plus the active ROCm SDK and GPU driver — with a new build script, release/nightly plumbing for the ref, and updated smoke and Gherkin coverage. Outcome: **No blocking findings**. Verified: read every changed file in full plus surrounding module, caller and workflow context; ran the one build-metadata test target and a reverted-hunk mutation of it in a throwaway copy, and separately exercised the smoke script's version assertion against good and adversarial strings — the full workspace suite, e2e suite and `--all-targets` were not run here, and no binary was built from the checkout. Confirmed as load-bearing: all five new/changed `@id:examine-version` assertions fail when the production lines they name are removed; the ref env var reaches all four shipped-artifact build steps; every version surface renders the same build string; the replaced smoke and e2e assertions are strictly tighter than the ones they replaced. Check state at review: 27 success, 0 failure. Blocking: 0 · Non-blocking: 6.

**None.**

## Non-blocking
- `apps/rocm/build.rs:214` — the `commit_something` test helper pins `user.name`/`user.email` so tests don't depend on ambient git config, but not `commit.gpgsign`; on a machine following this project's own signing setup doc the temp-repo commit can fail and take two of the three new tests with it. Add `-c commit.gpgsign=false` to the argument list.
- `crates/rocm-core/src/lib.rs:3043` — `detect_legacy_rocm_sdk` discards a discovered install whose version can't be parsed, so `rocm version` prints "ROCm SDK: not detected" when a real unmanaged install exists; the sibling summary function deliberately keeps the path in that case. Report the path with an unknown-version label instead.
- `apps/rocm/build.rs` — `ref_descriptor()`, the whole precedence chain the feature rests on, has no test; all three new tests cover only the rerun-if-changed path, so swapping two precedence branches would ship silently.
- `tests/e2e-cucumber/tests/e2e/examine_steps.rs` and `scripts/smoke_local.py` both fail when the resolved ref is `unknown`, which is also what a correct build produces from a detached HEAD or a git-less source tree; the failure text names the symptom, not the cause. Extend both messages to say the ref env var or a git checkout is required — this is the one place a competent reader would reasonably misread a clean build as a regression.
- `.github/workflows/release.yml:116`, `.github/workflows/nightly.yml:112` — the two Linux build steps splice `${{ }}` straight into the `run:` script text, while the Windows steps in the same PR use a step-level `env:` block; forwarding the bare name to `docker run -e` removes the interpolation entirely.
- `crates/rocm-core/src/lib.rs:2742` — `detect_gpu_driver_version` folds the WSL probe's deliberate three-state result ("couldn't check" vs "checked, no AMD adapter") into a single `None`, which an adjacent comment records as a previously-fixed regression; either report a distinct "could not check" or note the trade-off.

---

_Note for maintainers: this branch was rebased after that approval was filed, so every commit now on it was rewritten and the originally approved commit no longer exists in the branch's history. The approval therefore predates the branch's current contents. This review read the current head in full and found nothing blocking, but whether the earlier approval should stand over the work added since is a judgement for a human reviewer, not something this automation decides._

PR #249, opened 2026-08-13
fix(lemonade): retry interrupted backend setup

volen-silo:
## Summary
- Retry Lemonade's llama.cpp backend installation once when the backend subprocess fails during first serve.
- Keep the retry at the backend boundary so deterministic archive, disk, and backend-selection failures are not repeated and a healthy embeddable is not re-extracted.
- After the retry fails, recommend `rocm engines install lemonade --reinstall` before retrying `rocm serve`.
- Add deterministic unit coverage and a black-box Cucumber scenario for the visible retry count and recovery guidance.

Why: a transient connection loss during first-time Lemonade backend preparation previously aborted `rocm serve` with a generic error, forcing users to discover the manual recovery sequence themselves.

Risk: low. The change is limited to the backend-install phase, allows one additional attempt, and exposes its failure fixture only in E2E-enabled builds.

## Test plan
- `cargo xtask e2e -- -n 'Repeated Lemonade preparation failure'`
- `cargo test -p rocm-engine-lemonade backend_install_`
- `cargo test --workspace --all-targets -- --skip therock::tests::extracting_the_sdk_archive_removes_it`

The skipped archive test is a pre-existing parallel-suite flake on this host (`tar` temporarily missing from `PATH`) and passes in isolation.

### DIFF
--- a/.github/workflows/e2e-selfhosted.yml
+++ b/.github/workflows/e2e-selfhosted.yml
@@ -279,7 +279,15 @@ jobs:
           # Build the rocm and rocmd binaries ONCE and reuse them for both the
           # pre-warm and suite so xtask does not rebuild. Honors
           # CARGO_TARGET_DIR set above.
-          cargo build --release -p rocm -p rocmd
+          #
+          # `--features rocm/e2e-test-hooks` must match what `cargo xtask e2e`
+          # builds when it builds for itself. The suite's deterministic failure
+          # seams (e.g. the scripted Lemonade backend-install failure) are
+          # compiled out without it, so a pre-built binary that omits the feature
+          # leaves those scenarios unable to reach their premise — they then fail
+          # as regressions on whichever lane happens to select them. Every lane
+          # that pre-builds and exports ROCM_CLI_BINARY must pass it.
+          cargo build --release -p rocm -p rocmd --features rocm/e2e-test-hooks
           export ROCM_CLI_BINARY="$CARGO_TARGET_DIR/release/rocm"
           export ROCM_CLI_ROCMD_BINARY="$CARGO_TARGET_DIR/release/rocmd"
 
@@ -444,7 +452,9 @@ jobs:
           export E2E_SHARED_RUNTIMES_DIR="$prewarm/data/runtimes"
 
           # Build the rocm and rocmd binaries once; reuse them for pre-warm + suite.
-          cargo build --release -p rocm -p rocmd
+          # See the e2e-gpu lane for why the e2e-test-hooks feature must match
+          # what `cargo xtask e2e` would build.
+          cargo build --release -p rocm -p rocmd --features rocm/e2e-test-hooks
           export ROCM_CLI_BINARY="$CARGO_TARGET_DIR/release/rocm"
           export ROCM_CLI_ROCMD_BINARY="$CARGO_TARGET_DIR/release/rocmd"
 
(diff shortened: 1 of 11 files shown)


r0x0r (commented):
Reviewed the diff and the tests. The change is well-scoped and the reasoning holds up.

**Strengths**

- **Correct retry boundary.** Putting the single retry around the backend install rather than the whole runtime prep is the right call, and the doc comment justifies it well: the embeddable download already has bounded transport retries, and repeating the outer op would redo deterministic archive/disk failures and risk re-extracting a healthy runtime. The transient this targets (backend subprocess losing its connection to lemond after a completed download) is exactly the case a second call can reuse the cache and recover from.
- **Terminal error is the retry's failure, not the first attempt's**, with actionable recovery guidance appended (`rocm engines install lemonade --reinstall`, then retry `rocm serve`). The `stops_after_one_retry_with_reinstall_guidance` test pins this precisely, including that the first error is *not* the one surfaced.
- **Layered coverage.** Three deterministic unit tests (success-without-retry, recover-on-second, stop-after-one-with-guidance) plus the black-box scenario 15 that asserts the visible retry count and the recovery command.
- **No production behavior leak.** The failure seam is `#[cfg(feature = "e2e-test-hooks")]` and the `serve()` precondition bypass is `cfg!(feature) && env_var` — both compile out of release builds. The only non-gated change is wrapping `run_lemonade_backend_install` in the retry helper, which is strictly additive robustness and can't turn a previously-passing install into a failure.

**Minor / non-blocking**

- The e2e assert in `setup_lemonade_hf_checkpoint_model` panics with `{stdout}` while the real serve error lands on stderr, so when that path fails the root cause isn't visible in the captured output. Pre-existing (not introduced here), but a follow-up to also surface stderr would make failures in that scenario much easier to triage.
- `xtask` now builds all e2e binaries with `--features rocm/e2e-test-hooks`, including the self-hosted GPU lanes. The seam is inert without the env var so this is fine, but it does mean those binaries are no longer feature-identical to release — worth a line in the PR description so it's a deliberate, understood choice.

Overall the implementation, layering, and tests look good.

r0x0r:
One blocker on the latest push: scenario 15 (`serve-lemonade-preparation-recovery`) fails on the **Strix Halo, WSL2** lane, and reconciliation flags it as a regression ("expected to pass on this host but FAILED"):

```
Then serving stops after one automatic retry
  assertion `left == right` failed: expected exactly one retry announcement
  Error: no usable AMD GPU detected; `rocm serve` requires a GPU under the gpu_required policy…
    left: 0   right: 1
```

**Root cause.** The scenario depends on the `e2e-test-hooks`-gated GPU-precondition bypass and failure seam, which are only present when the `rocm` binary is compiled with `rocm/e2e-test-hooks`. This PR adds that feature to xtask's build in `xtask/src/e2e.rs`, but the self-hosted lanes pre-build the binary **without** it — `cargo build --release -p rocm -p rocmd` in `.github/workflows/e2e-selfhosted.yml` (lines 279, 453, 612, 808), with a comment stating the pre-warm exists "so xtask does not rebuild." On WSL2 the executed binary therefore lacks the feature: the bypass is compiled out, `usable_amd_gpu_indices()` returns empty (WSL can't see the GPU), and `rocm serve` aborts with the no-GPU error before reaching Lemonade's backend boundary — so there are 0 retry announcements. The GitHub-hosted no-GPU lane passes only because it has no pre-warm step and lets xtask build with the feature.

**Suggested fix.** Add `--features rocm/e2e-test-hooks` to the four pre-warm `cargo build --release -p rocm -p rocmd` steps in `e2e-selfhosted.yml` so the pre-warmed binary matches what xtask (and scenario 15) expect — or otherwise ensure the feature-gated bypass is present on any lane where the `@requires-no-gpu` scenario runs, since WSL2 counts as no-usable-GPU.

Separately, the branch currently shows as conflicting with `main` and will need a rebase.

The underlying retry change still looks good from my earlier review — this is a CI-wiring gap for the new scenario rather than a problem with the fix itself.

michaelroy-amd (approved):
Reviewed at `4b90bdd12b0efa9e794087e71312a16773989a87`. The retry is placed at the Lemonade backend-install boundary, remains single-shot and deadline-bounded, and preserves the retry error plus actionable reinstall guidance. Three unit cases and `@id:serve-lemonade-preparation-recovery` cover success, recovery, and terminal failure. The latest commit also adds `rocm/e2e-test-hooks` to every prebuilt self-hosted binary path and pins that invariant with workflow-contract tests, resolving the prior WSL2 failure mode. The branch is currently mergeable and all 17 required checks pass. DCO and license checks are clean. No blocking finding. Approving.

PR #253, opened 2026-08-13
fix(therock): stop python-resolver tests mutating the process PATH

rominf:
- [x] If this PR fixes a bug, searched `tests/e2e-cucumber/expectations.toml` for the fixed ticket ID and removed/narrowed any now-stale xfail rows. (No matching rows.)

## Summary

- Pass the Python resolver's environment — the `ROCM_CLI_PYTHON` override and the directories to search — in as a value instead of reading it at the use site.
- The two tests that used to steer the resolver by overwriting `PATH` for the whole process now hand it a fixture directory, so both `unsafe` `set_var` blocks are gone rather than serialized behind a lock.
- Why: those tests made unrelated tests fail. `cargo test -p rocm --bin rocm` failed in 11 of 20 runs on a clean `main` because of this.
- Risk: low. No behaviour change — `resolve_python_launcher` reads exactly the same two variables, just once at entry instead of at each use site. Production is unaffected; the seam exists so tests do not need the process environment.

## Root cause

`cargo test` runs every test as a thread in one process. `python_launcher_prefers_path_python_before_saved_managed_python` and `python_launcher_prefers_path_python_over_managed_when_venv_capable` set `PATH` to a single fixture directory for the duration of `resolve_python_launcher`, then restore it. That window is process-wide: while it is open, every other test also has no usable `PATH`, and any test that spawns a PATH-resolved binary fails with `ENOENT`.

The visible casualty was `therock::tests::extracting_the_sdk_archive_removes_it`, failing with:

```
Error: failed to launch tar

Caused by:
    No such file or directory (os error 2)
```

`tar` was installed the whole time. That test spawns `tar` twice and only the second spawn failed, which is the window opening between them.

The mutex the tests already held only serialized them against each other, not against the ~450 other tests sharing the process, so it could not have helped. Removing the global mutation is what fixes it.

## Verification

20 consecutive `cargo test -p rocm --bin rocm` runs:

| | before | after |
| --- | --- | --- |
| `extracting_the_sdk_archive_removes_it` | 11/20 | 0/20 |

Root cause confirmed before fixing, by running the same 20 iterations with only these two tests skipped (`--skip python_launcher_prefers_path`): the tar failure disappeared, isolating it to them.

Also run: `cargo clippy --locked --workspace --all-targets -- -D warnings`, `cargo clippy --locked -p e2e-cucumber --test e2e -- -D warnings`, `cargo test --workspace --all-targets --exclude e2e-cucumber`, `prek run --all-files`.

## Notes for reviewers

- **Why this did not show up in CI.** The Linux test job runs `cargo nextest`, which is process-per-test and therefore immune. The Windows job runs plain `cargo test`, but these two tests are `#[cfg(unix)]`. The pain is local: the pre-push hook runs `cargo test`, and contributors have been working around it with `SKIP=cargo-test git push`.
- **New test.** `python_path_search_only_uses_the_given_directories` pins the property that keeps this from coming back — with no search directories the resolver finds nothing, even though the real `PATH` certainly has a python on it. It fails if anyone reintroduces an implicit `PATH` read.
- **Left alone deliberately.** `python_venv_probe_temp_root_uses_windows_temp_env` still mutates `TEMP`/`TMP`/`LOCALAPPDATA` process-wide and still takes the mutex. It is the same class of hazard on the Windows `cargo test` lane, but it plays no part in this failure and I cannot verify a Windows fix from here. Worth a follow-up.

### DIFF
--- a/apps/rocm/src/therock.rs
+++ b/apps/rocm/src/therock.rs
@@ -2922,9 +2922,38 @@ fn ensure_managed_python(paths: &AppPaths) -> Result<PythonLauncher> {
     })
 }
 
+/// The environment inputs [`resolve_python_launcher`] reads.
+///
+/// Passed in rather than read at each use site so a caller can point the
+/// resolver somewhere else without touching the process environment. Tests need
+/// that: `cargo test` runs every test as a thread in one process, so a test that
+/// overwrote `PATH` to steer this resolver also hid every other PATH-resolved
+/// binary from unrelated tests running at the same moment.
+struct PythonResolverEnv {
+    /// `ROCM_CLI_PYTHON`: an explicit interpreter that wins over any search.
+    python_override: Option<String>,
+    /// The directories to search for an interpreter, in `PATH` order.
+    search_dirs: Vec<PathBuf>,
+}
+
+impl PythonResolverEnv {
+    fn from_process_env() -> Self {
+        Self {
+            python_override: std::env::var("ROCM_CLI_PYTHON").ok(),
+            search_dirs: std::env::var_os("PATH")
+                .map(|value| split_runtime_path(&value))
+                .unwrap_or_default(),
+        }
+    }
+}
+
 fn resolve_python_launcher(paths: &AppPaths) -> Result<PythonLauncher> {
-    if let Ok(value) = std::env::var("ROCM_CLI_PYTHON") {
-        python_launcher_install_ready(Path::new(&value))
+    resolve_python_launcher_in(paths, &PythonResolverEnv::from_process_env())
+}
+
+fn resolve_python_launcher_in(paths: &AppPaths, env: &PythonResolverEnv) -> Result<PythonLauncher> {
+    if let Some(value) = env.python_override.as_deref() {
+        python_launcher_install_ready(Path::new(value))
             .with_context(|| format!("ROCM_CLI_PYTHON is not usable for ROCm setup: {value}"))?;
         return Ok(PythonLauncher {
             executable: PathBuf::from(value),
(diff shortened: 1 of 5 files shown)


r0x0r (approved):
Nice root-cause writeup. The key insight — the existing mutex only serialized the two PATH-mutating tests against *each other*, never against the ~450 other tests sharing the process — is exactly right, and switching to injected search directories (`PythonResolverEnv`) is the correct fix rather than adding more locking. Production behavior is preserved (`from_process_env()` reads the same two variables, just once at entry), the `unsafe { set_var }` blocks are gone, and `python_path_search_only_uses_the_given_directories` pins the property so an implicit PATH read cannot creep back. LGTM.

Two non-blocking notes:

1. **CI red checks are not from this change.** `E2E tests (GPU)` and `E2E tests (Strix Halo, Windows)` both fail on the pre-existing `serve-hf-checkpoint-inference` "1 unexpected failure(s)" regression — same signature seen on several other unrelated open branches. Your new `cli_failure_report` is actually visible in those logs now (`rocm serve ... failed (rc=1)`), which is a nice validation that the helper works. All other checks (clippy, windows-build-and-test, affected tests, signatures) are green.

2. **Scope.** The `cli_failure_report` / `run_rocm_ok` / `cli_stderr` e2e diagnostics (EAI-8031) are a good change but orthogonal to the python-resolver PATH fix — different subsystem, different ticket. Per the one-logical-change-per-PR convention they would be easier to land/revert independently. Happy either way; flagging for a possible follow-up split rather than blocking on it.

Issue #259, opened 2026-08-14
rocm serve --engine lemonade fails on any headless host where XDG_RUNTIME_DIR is unset


rominf:
## Symptom

On a Linux host with no `XDG_RUNTIME_DIR` — a systemd service, a cron job, a container,
an SSH session without lingering enabled — starting the lemonade engine fails:

```
$ rocm serve Qwen3-0.6B-GGUF --engine lemonade --managed
Preparing lemonade for GPU serving...
Preparing Lemonade embeddable 11.5.1...
Detecting best supported Lemonade llama.cpp backend...
Error: request_failed: Lemonade server did not become ready: exit status: 1;
Lemonade startup log tail (.../engines/lemonade/runtime/install-lemond.log):
[Error] (main) Error: Unable to resolve writable runtime directory from
XDG_RUNTIME_DIR or RUNTIME_DIRECTORY
```

Exit code 1, every time. It is fully deterministic — retrying does not help — and it does
not depend on the model, the variant, or the serve path: both the recipe-name and
`owner/repo:variant` paths fail identically.

The same command works from an interactive desktop or SSH login on the same machine,
because a login session gets `XDG_RUNTIME_DIR` from `pam_systemd`. That makes this look
like a machine-specific quirk when it is really a property of how the process was started.

## Root cause

`lemond` requires a writable runtime directory and looks for it in `XDG_RUNTIME_DIR` or
`RUNTIME_DIRECTORY`. Neither is set in a headless context: `XDG_RUNTIME_DIR` is populated
by `pam_systemd` at login and is absent for non-login processes, and `RUNTIME_DIRECTORY`
only exists for a systemd unit that declares `RuntimeDirectory=`.

rocm-cli launches `lemond` and builds its environment explicitly, but does not supply
either variable. `lemonade_process_environment_vars`
(`engines/lemonade/src/lib.rs:2170`) already translates `ROCM_PATH`, `PATH`,
`LD_LIBRARY_PATH`, `HIP_VISIBLE_DEVICES` and `LEMONADE_API_KEY` into the child — the
runtime directory is simply missing from that list. `XDG_RUNTIME_DIR` appears nowhere in
the engine code; the only occurrences in the repo are the dashboard socket resolvers and
the E2E harness.

So the hard requirement belongs to `lemond`, but the thing that can satisfy it is
rocm-cli, which is what spawns the process and already owns its environment.

## This is a product bug, not a CI bug

[...]

Issue #237, opened 2026-08-12
vLLM install failure on an unsupported interpreter or glibc surfaces misleading uv advice


rominf:
`rocm engines install vllm` pins the ROCm build of vLLM (`vllm==<version>+rocm<abi>`), which is published as a single wheel — currently CPython 3.12 on glibc >= 2.34. On any other host the install correctly fails rather than falling back to the generic PyPI build. That behaviour is intended.

The problem is what the user is told. `install_vllm_with_uv` surfaces uv's raw resolver output, and for an unsatisfiable pin that output leads with dependency-confusion wording and suggests `--index-strategy unsafe-best-match`. The suggestion does not help here — the wheel genuinely does not exist for that interpreter/glibc — and if followed via `UV_INDEX_STRATEGY` it reintroduces exactly the unpinned PyPI fallback the pin exists to prevent.

Affected hosts include any 3.13 interpreter, and distributions whose glibc is older than 2.34 such as RHEL 8 (2.28) and SLES 15.7 (2.31), all of which `rocm install driver` otherwise supports.

Suggested fix: on that failure path, lead with the real cause — no ROCm vLLM wheel is published for this interpreter and glibc combination — and say explicitly that widening the index strategy is not a workaround. Leaving uv's own text below it is fine.

Raised during review of #176 and deliberately left out of scope there, as it is separable from the pin itself.

PR #261, opened 2026-08-14
fix: raise the supported Linux minimum to Ubuntu 24.04

### LINKED ISSUE #258: Lemonade engine cannot start on Ubuntu 22.04 (glibc 2.35), but README and docs/wsl.md advertise it as supported
## Symptom

On Ubuntu 22.04 — native or under WSL2 — anything that starts the lemonade engine fails.
`rocm serve <model> --engine lemonade` exits non-zero after preparing the embeddable:

```
Preparing Lemonade embeddable 11.5.1...
Detecting best supported Lemonade llama.cpp backend...
.../engines/lemonade/runtime/lemond: /lib/x86_64-linux-gnu/libc.so.6:
version `GLIBC_2.38' not found (required by .../engines/lemonade/runtime/lemond)
```

There is no workaround available to the user: no engine flag, no version pin, and no
supported way to get glibc 2.38 onto 22.04.

## Root cause

The `lemond` binary in every published `lemonade-embeddable-*-ubuntu-x64` release is
linked against symbol versions newer than 22.04 provides. Ubuntu 22.04 ships glibc 2.35
and `GLIBCXX_3.4.30`; the binary needs `GLIBC_2.38` and `GLIBCXX_3.4.32`. Both are
missing, so the dynamic loader refuses to start it.

The specific imports, from `readelf --dyn-syms` on 11.5.1:

```
__isoc23_sscanf@GLIBC_2.38    __isoc23_strtoul@GLIBC_2.38
__isoc23_strtol@GLIBC_2.38    __isoc23_strtoull@GLIBC_2.38
__isoc23_strtoll@GLIBC_2.38   _ZSt21ios_base_library_initv@GLIBCXX_3.4.32
```

#247 first found this while chasing a CI failure and checked the 11.x releases. Extending
that check to **every** release that publishes an `ubuntu-x64` embeddable — v10.2.0
through v11.5.2, 15 releases — gives the same answer in all of them:

| release range | max GLIBC | max GLIBCXX |
| --- | --- | --- |
| v10.2.0 … v10.10.0 (10 releases) | `GLIBC_2.38` | `GLIBCXX_3.4.32` |
| v11.0.0, v11.5.0, v11.5.1 (current pin), v11.5.2 | `GLIBC_2.38` | `GLIBCXX_3.4.32` |

Verified by downloading each `lemonade-embeddable-<v>-ubuntu-x64.tar.gz` from the
`lemonade-sdk/lemonade` releases and reading the version requirements out of `lemond`.
So this is not a regression in the current pin, and **there is no older release to pin
back to** — no published Lemonade embeddable can start on Ubuntu 22.04.

## Why this is a product problem, not a CI problem

`README.md` states:

> | Linux (x86_64) | Yes | Full support, including the live dashboard and both inference engines |

> | WSL2 (x86_64) | Yes (Linux binary) | Full support, including the live dashboard; see [docs/wsl.md](docs/wsl.md) for setup |

and `docs/wsl.md` lists as a prerequisite:

> 2. WSL2 with Ubuntu 24.04 or Ubuntu 22.04.

Taken together, the docs promise a 22.04 user — native or WSL2 — full support for both
inference engines. One of those two engines cannot start on that release, and the WSL page
names 22.04 explicitly as a supported base. Every such user hits this the first time they
touch lemonade. CI is simply where it was noticed.

(Quotes checked against `main` at 4eebf1d1.)

## Options

Listing these for maintainers to choose between; I am not proposing one here.

1. **Narrow the documented support matrix** — mark the lemonade engine as requiring
   glibc >= 2.38 (Ubuntu 24.04+), and adjust `docs/wsl.md` accordingly. Honest and
   immediate, but drops a currently-advertised platform.
2. **Ask upstream for a lower-baseline build** — a `manylinux`-style or built-on-22.04
   `ubuntu-x64` embeddable from `lemonade-sdk/lemonade`. The right fix if 22.04 is meant
   to stay supported, but it is not in this project's control and needs an upstream
   request.
3. **Ship the engine differently on old hosts** — e.g. run `lemond` from a newer-userland
   container with `/dev/kfd` and `/dev/dri` passthrough. Heavier, and needs product
   support for containerized engines.
4. **Some combination** — e.g. do 1 now as the accurate state of the world, and pursue 2
   as the durable fix.

Whichever route is taken, a **clear preflight error** is worth having on its own: today
the user sees a raw loader message from a subprocess. Detecting glibc < 2.38 before the
download and saying "the lemonade engine requires glibc 2.38 or newer; this host has
2.35" would turn an opaque failure into an actionable one. This is close in spirit to
#237, which is about the same class of problem for vLLM (unsupported interpreter/glibc
producing misleading advice).

## Not viable

Installing glibc 2.38 on 22.04. Ubuntu publishes none for jammy (2.38 first appears in
23.10), and swapping the system libc for a newer release's breaks the ABI for every
binary on the machine. #247 has the longer write-up.

## Environment

- lemonade pinned at `11.5.1` in `runtime-deps.toml`; artifact
  `lemonade-embeddable-11.5.1-ubuntu-x64.tar.gz`
- Affects any x86_64 Linux host with glibc < 2.38, native or WSL2
- Ubuntu 24.04 (glibc 2.39, `GLIBCXX_3.4.33`) satisfies both requirements

rominf:
## The problem

The docs promise a platform one of the two shipped engines cannot run on.

`README.md` advertised Linux (x86_64) as *"Full support, including the live
dashboard and both inference engines"*, and `docs/wsl.md` listed
*"WSL2 with Ubuntu 24.04 or Ubuntu 22.04"* as a prerequisite. Together those tell a
22.04 user — native or WSL2 — that both engines work.

They do not. Every published Lemonade embeddable, v10.2.0 through v11.5.2 (all 15
`ubuntu-x64` releases), is linked against `GLIBC_2.38` and `GLIBCXX_3.4.32`. Ubuntu
22.04 ships glibc 2.35 and `GLIBCXX_3.4.30`, so `lemond` is refused by the dynamic
loader before it starts. There is no older release to pin back to and no supported
way to put glibc 2.38 on jammy. Full analysis in #258.

This is option 1 from that issue — narrow the documented matrix to match reality.

## What changed

One flat minimum: **Ubuntu 24.04 or newer**, for Linux and WSL2 alike.

- `README.md`: both the Linux and WSL2 rows now say "Ubuntu 24.04 or newer", and a
  short paragraph below gives the glibc number behind that line — glibc 2.38 with
  `GLIBCXX_3.4.32` — so the requirement still means something to a reader on RHEL,
  SLES, or Fedora who cannot map an Ubuntu version onto their own distribution.
- `docs/wsl.md`: prerequisite 2 is now "Ubuntu 24.04 or newer", with the same reason
  stated inline.
- `scripts/wsl_preflight.py`: `22.04` dropped from the supported set, and the
  `ubuntu` check detail now names the minimum. Covered by a new `--self-test` case
  (an otherwise-ready 22.04 host must fail the check), which fails against the old
  set and passes against the new one, and runs in CI on both the Linux and Windows
  lanes.

The minimum is deliberately stated as a single number rather than split per engine.
The Lemonade floor is the binding one; a per-engine matrix would be more precise but
harder to state and harder to test, and the simpler line is the one being supported.

## The one edit a maintainer might want reverted

`scripts/wsl_preflight.py` is a behaviour change, not documentation — hence the `fix:`
title rather than `docs:`. It is included because that script is the automated form of
the `docs/wsl.md` prerequisite list and is invoked directly from it; leaving it
reporting `ubuntu: ok` on a host Lemonade cannot run on would relocate the
inconsistency rather than fix it. If you would rather keep this PR strictly to prose,
it is a self-contained three-line revert.

## Also found, deliberately not changed

- `apps/rocm/src/main.rs` driver-install matrix (`("ubuntu", "22.04" | "24.04")`,
  `codename_for_version`). That mirrors AMD's own documented amdgpu-dkms install
  matrix, which is a separate upstream support matrix; `rocm install driver` still
  produces a correct plan on 22.04 and narrowing it would be a behaviour change beyond
  this issue.
- `crates/rocm-core/src/diagnose.rs` suggesting `rocm/dev-ubuntu-22.04` as a
  known-good container image. A container brings its own userland, so the host glibc
  claim does not apply; it is not a host-OS support statement.

## This does not fix the user experience, only the promise

#258 also asks for a preflight error — *"the lemonade engine requires glibc 2.38 or
newer; this host has 2.35"* — instead of a raw loader message. That is deliberately
out of scope here, but worth stating plainly: **no such preflight exists today.** The
only glibc-aware code in the tree is a low-weight ``"version `?glibc"`` keyword in the
`rocm diagnose` classifier, explicitly annotated "tangential"; there is no glibc check
anywhere in the Lemonade download or launch path. A 22.04 user who has not read these
docs still gets an opaque ``version `GLIBC_2.38' not found`` out of a subprocess. The
preflight remains worth doing on its own, and it would be worth tracking separately
rather than letting it lapse when this merges.

## Related

- #237 — the same class of problem for vLLM: an unsupported interpreter/glibc produces
  misleading uv advice rather than the real cause. Worth noting that vLLM's floor is a
  different and lower one (glibc 2.34, plus CPython 3.12 that `rocm` provisions
  itself), so the flat 24.04 minimum here is set by Lemonade and is marginally
  stricter than vLLM alone would require. That tradeoff was taken deliberately in
  favour of a support matrix that is simple to state and simple to test.
- #259 — also a Lemonade start failure the docs do not warn about (`XDG_RUNTIME_DIR`
  unset on headless hosts). It was masked by the glibc floor tracked here.
- #247 — where the glibc floor was first hit, in CI.

## Verification

`prek run --all-files --no-group local-tools` (all passed: whitespace, EOF, yaml,
merge-conflict, line endings, ruff check, ruff format, shellcheck, cargo fmt),
`hawkeye check --config licenserc.toml` (no missing headers), `cargo fmt --all
--check` (clean), `python3 scripts/wsl_preflight.py --self-test` (ok; and confirmed
the new assertion still fails against the pre-change supported set).

No Gherkin scenario accompanies this: the `rocm`/`rocmd` binaries have no observable
behaviour change. The one behavioural edit is to a standalone helper script, whose
contract is covered by its own `--self-test`, already run by CI.

Searched `tests/e2e-cucumber/expectations.toml` for stale xfails tied to this issue —
the Lemonade rows there are all EAI-7423 (managed serve shutting down after ready),
unrelated to the glibc floor, so nothing to remove or narrow.

Fixes #258

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,14 +29,21 @@ Lemonade and vLLM.
 
 | Platform | Prebuilt binary | Notes |
 |---|---|---|
-| Linux (x86_64) | Yes | Full support, including the live dashboard and both inference engines |
+| Linux (x86_64) | Yes | Ubuntu 24.04 or newer; full support, including the live dashboard and both inference engines |
 | Windows (x86_64) | Yes | CLI and Lemonade serving; no live dashboard or vLLM |
-| WSL2 (x86_64) | Yes (Linux binary) | Full support, including the live dashboard; see [docs/wsl.md](docs/wsl.md) for setup |
+| WSL2 (x86_64) | Yes (Linux binary) | Ubuntu 24.04 or newer; full support, including the live dashboard; see [docs/wsl.md](docs/wsl.md) for setup |
 | macOS | No | No official installer, release, CI, or QA coverage |
 
 Live dashboard telemetry requires Linux or WSL2 (see
 [Interactive interfaces](#interactive-interfaces)). vLLM serving is Linux/WSL2
 only (see [docs/vllm.md](docs/vllm.md)).
+
+The minimum supported Linux release, native or under WSL2, is Ubuntu 24.04. On
+other distributions the equivalent requirement is glibc 2.38 with
+`GLIBCXX_3.4.32`: that is what the Lemonade engine is linked against, and every
+published build of it needs those versions, so there is no older release to fall
+back to. Ubuntu 22.04 ships glibc 2.35 and cannot run it; Ubuntu 24.04 provides
+glibc 2.39 and `GLIBCXX_3.4.33`.
 
 > [!IMPORTANT]
 > **Tech Preview** -- This software is provided as-is, without warranty or
--- a/docs/wsl.md
+++ b/docs/wsl.md
@@ -14,7 +14,10 @@ virtual environments and ROCDXG (`librocdxg`).
 The AMD WSL path is:
 
 1. Windows 11 with the AMD Adrenalin WSL-capable driver.
-2. WSL2 with Ubuntu 24.04 or Ubuntu 22.04.
+2. WSL2 with Ubuntu 24.04 or newer. Ubuntu 22.04 is not supported: it ships
+   glibc 2.35, below the glibc 2.38 / `GLIBCXX_3.4.32` floor that every
+   published Lemonade embeddable requires, so the Lemonade engine cannot start
+   there. Ubuntu 24.04 provides glibc 2.39 and `GLIBCXX_3.4.33`.
 3. ROCDXG (`librocdxg`) installed inside WSL.
 4. A TheRock runtime installed by `rocm-cli` into a managed Python venv.
 
--- a/scripts/wsl_preflight.py
+++ b/scripts/wsl_preflight.py
@@ -258,8 +258,10 @@ def evaluate(
     sdk_headers = state.get("windows_sdk_headers") or []
 
     version_id = str(os_release.get("VERSION_ID") or "")
+    # 22.04 is deliberately absent: its glibc 2.35 is below the glibc 2.38 /
+    # GLIBCXX_3.4.32 floor every published Lemonade embeddable is linked
+    # against, so the Lemonade engine cannot start on it. See docs/wsl.md.
     supported_ubuntu = os_release.get("ID") == "ubuntu" and version_id in {
-        "22.04",
         "24.04",
         "26.04",
     }
@@ -270,7 +272,9 @@ def evaluate(
     checks = [
         Check("wsl", bool(state.get("is_wsl")), "WSL marker or /dev/dxg detected"),
         Check(
-            "ubuntu", supported_ubuntu, f"Ubuntu VERSION_ID={version_id or '<unknown>'}"
+            "ubuntu",
+            supported_ubuntu,
+            f"Ubuntu 24.04 or newer (VERSION_ID={version_id or '<unknown>'})",
         ),
         Check("dxg_device", bool(paths.get("/dev/dxg")), "/dev/dxg"),
         Check(
@@ -390,6 +394,14 @@ def self_test() -> None:
     ready_state["ldconfig"]["librocdxg"] = True
     checks = evaluate(ready_state, require_rocm_tools=False)
     assert all(check.ok for check in checks if check.required), checks
+
+    # 22.04 is below the Lemonade glibc floor, so an otherwise perfect host
+    # must still fail the ubuntu check rather than report itself ready.
+    jammy_state = json.loads(json.dumps(ready_state))
+    jammy_state["os_release"]["VERSION_ID"] = "22.04"
+    checks = evaluate(jammy_state, require_rocm_tools=False)
+    assert any(check.name == "ubuntu" and not check.ok for check in checks), checks
+    assert not all(check.ok for check in checks if check.required), checks
 
 
 def main() -> int:
--- a/scripts/wsl_preflight.py
+++ b/scripts/wsl_preflight.py
@@ -258,13 +258,20 @@ def evaluate(
     sdk_headers = state.get("windows_sdk_headers") or []
 
     version_id = str(os_release.get("VERSION_ID") or "")
-    # 22.04 is deliberately absent: its glibc 2.35 is below the glibc 2.38 /
+    # 22.04 is deliberately unsupported: its glibc 2.35 is below the glibc 2.38 /
     # GLIBCXX_3.4.32 floor every published Lemonade embeddable is linked
     # against, so the Lemonade engine cannot start on it. See docs/wsl.md.
-    supported_ubuntu = os_release.get("ID") == "ubuntu" and version_id in {
-        "24.04",
-        "26.04",
-    }
+    version_parts = version_id.split(".")
+    ubuntu_version = None
+    if len(version_parts) == 2 and all(
+        part.isascii() and part.isdecimal() for part in version_parts
+    ):
+        ubuntu_version = (int(version_parts[0]), int(version_parts[1]))
+    supported_ubuntu = (
+        os_release.get("ID") == "ubuntu"
+        and ubuntu_version is not None
+        and ubuntu_version >= (24, 4)
+    )
     build_tools = all(
         tools.get(name) for name in ["git", "cmake", "gcc", "g++", "make"]
     )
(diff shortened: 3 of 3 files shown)


rominf:
CI status, so the one red check is not left unexplained.

**All 16 required checks are green**, including `prek (lint / hygiene)`, `License header check (hawkeye)`, `Commit signatures + sign-off`, `E2E tests`, and both CodeQL analyses.

**`E2E tests (GPU)` is red, and it is pre-existing on `main`, not caused by this PR.** The lane is advisory (not in the required set). Evidence:

- The failure is one unexpected failure, `serve-hf-checkpoint-inference`, failing at *"a canonical Hugging Face GGUF checkpoint is being served on lemonade"* with `rocm serve failed`. Reconciliation line: `5 xfail, 0 XPASS, 1 unexpected failure(s)`.
- The `main`-branch run from the same window (job 94765233909) fails with the **identical** reconciliation line and the identical scenario, so the regression is already on `main`.
- This PR changes `README.md`, `docs/wsl.md`, and `scripts/wsl_preflight.py`. None of the three is read by the `rocm` binary or the E2E harness, and the GPU runner is `Ubuntu 24.04.4 LTS` — above the new documented minimum either way, so nothing in this change can reach that lane.
- The symptom is a lemonade serve failure on a headless self-hosted runner, which is what #259 describes: the runner's glibc 2.39 clears the floor tracked in #258, so the `XDG_RUNTIME_DIR` failure underneath is no longer masked. There is already a `fix-lemonade-missing-xdg-runtime-dir` branch in flight for it.

`E2E tests (Strix Halo, Ubuntu)` passed. `E2E tests (Strix Halo, Windows)` is still queued behind the self-hosted runner backlog; it is also advisory and I will follow up if it reports anything other than the same pre-existing state.

rominf:
Follow-up on the two advisory self-hosted lanes, now that both have finished.

- `E2E tests (Strix Halo, Ubuntu)` — **passed**.
- `E2E tests (Strix Halo, Windows)` — **failed, and it is the same pre-existing failure**: `3 xfail, 0 XPASS, 1 unexpected failure(s)` / `serve-hf-checkpoint-inference`. The unrelated branch `feat/rocm-version-tag-branch-hash` (job 94760021263), which touches no docs, produces a byte-identical reconciliation line and the same scenario, so this is repo-wide rather than anything in this PR. `main`'s own Windows lane is still queued behind the runner backlog; its GPU lane already failed the same way (job 94765233909).

So both red lanes fail on `serve-hf-checkpoint-inference`, on hosts already above the new documented minimum, and both reproduce off this branch. Neither is in the 16 required checks, and all 16 required checks are green.

rominf:
Updated per maintainer decision: one flat minimum of Ubuntu 24.04, no per-engine matrix. The glibc number stays as the reason behind that line so it still means something on a non-Ubuntu distribution. Retitled `docs:` → `fix:`, since the diff includes the `scripts/wsl_preflight.py` behaviour change.

Head is now `7b66483`, rebased onto current `main` (`0063ef15`).

CI on the new head: **all 16 required checks green.** Re-verified locally after the edit — `prek run --all-files --no-group local-tools`, `hawkeye check`, `cargo fmt --all --check`, and `wsl_preflight.py --self-test` including the new 22.04 regression case (confirmed it still fails against the pre-change supported set).

`E2E tests (GPU)` is red again for the same pre-existing `serve-hf-checkpoint-inference` / #259 reason explained above; the two Strix lanes are still running. All three are advisory, none is in the required set, and none is reachable from a change to two markdown files and a helper script.

r0x0r (commented):
Reviewed the diff and the linked analysis in #258. This is a clean, well-scoped change and the reasoning holds up. Verified locally by running the PR's `python3 scripts/wsl_preflight.py --self-test` — it passes, and the new cases exercise the numeric comparison correctly (accepts 24.10/25.04/26.04/28.04, rejects 22.04, and fails closed on `24.04.1`, `unknown`, empty, and missing `VERSION_ID`).

**What I like**

- The `wsl_preflight.py` change moves from a hardcoded version set to a parsed `(major, minor)` tuple compared against `(24, 4)`. That's a real improvement: "or newer" now actually works for future releases instead of needing the set extended each cycle, and the parse-fails-closed behavior is the right default for a preflight.
- Docs and the automated check are kept in lockstep, and the glibc floor (2.38 / `GLIBCXX_3.4.32`) is stated as the distro-neutral reason rather than just an Ubuntu version number — useful for RHEL/SLES/Fedora readers.
- The "deliberately not changed" section (driver-install matrix in `main.rs`, the `dev-ubuntu-22.04` container hint in `diagnose.rs`) matches what I see in the tree; leaving those is correct since they're a separate support matrix and a container userland respectively.

**Minor nits (non-blocking)**

- The tuple floor `(24, 4)` represents `24.04` with minor `4`. It reads fine given the inline comment, but a reader skimming might expect `04`. Not worth changing.
- Strictly on the glibc floor, Ubuntu 23.10 shipped glibc 2.38 and would technically clear the Lemonade requirement, yet the preflight rejects it. That's a defensible simplification (23.10 is non-LTS and EOL, and the PR argues explicitly for a single simple LTS line), just noting the small precision gap.
- Fail-closed on `24.04.1`: real Ubuntu `os-release` `VERSION_ID` is `24.04`, not a point release, so this won't bite in practice — good to have it locked down by the self-test regardless.

**Scope note (already disclosed, agreeing)**

The PR is honest that it fixes the promise, not the UX — a 22.04 user still gets the raw loader error with no glibc preflight. Agree that's out of scope here and worth tracking separately so it doesn't lapse.

Nothing blocking from my side.

rominf:
Thanks for the review. On the three minor nits — agreed with your own calls, no changes: the `(24, 4)` tuple reads fine with the inline comment, the 23.10 gap is an accepted simplification for a single LTS line, and the `24.04.1` fail-closed case is intentional belt-and-suspenders since real `os-release` never reports a point release there.

On the scope note: opened #289 to carry the preflight-error ask forward on its own, since this PR is titled `Fixes #258` and merging it would otherwise close that issue with the preflight half of the ask still undone.

rominf:
Following up — the review notes were addressed (replied inline above), and this needs a formal approving review to satisfy the code-owner requirement (a comment-only review doesn't clear it). Re-requesting review.

michaelroy-amd (approved):
Reviewed at `803f00ff394b5a9e41e7c9940268eebff7312e2a`. The documentation and `wsl_preflight.py` now enforce one Ubuntu 24.04 minimum that matches Lemonade's glibc requirement. The version parser fails closed, and the self-test covers supported, unsupported, malformed, and missing `VERSION_ID` values. Driver-install and container support matrices remain correctly out of scope. All 17 required checks are green, both commits are signed off, and I found no blocking issue. Approving.

Issue #260, opened 2026-08-14
rocm serve owner/repo:variant on lemonade exits 0 but never serves the model on Strix Halo Windows


rominf:
Splitting the Windows half out of #247, which noted it "may be two problems rather
than one". The two halves have different symptoms and different failure points, so
they need separate owners. #247 tracks the Linux half; this issue is Windows only.

## Symptom

On Windows, serving a Hugging Face GGUF checkpoint by its canonical
`owner/repo:variant` reference on the lemonade engine appears to succeed and then
never produces a usable endpoint:

```
rocm serve unsloth/Qwen3-0.6B-GGUF:Q4_0 --engine lemonade --managed
```

The command exits **0** and reports `managed service launched ... readiness: starting`.
The model never appears at `http://127.0.0.1:11435/v1/models`, and after 600 s the
E2E step gives up:

```
Step panicked. Captured output: endpoint http://127.0.0.1:11435/v1/models
did not serve model Qwen3-0.6B within 600s
```

Nothing tells the user, or CI, why. A user hitting this sees a serve that claims to
have started and an endpoint that never answers.

## What this is not

- **Not a broken host or a flake.** In the *same job on the same machine*, scenario 7
  serves the *same underlying checkpoint* successfully via the short recipe name
  (`Qwen3-0.6B-GGUF`), and the ROCm llama.cpp backend installs cleanly
  (`llama-b9752-bin-win-rocm-7.13-x64.zip` + `therock-dist-windows-gfx1151-7.13.0.tar.gz`,
  "Backend installed successfully"). 31 of 35 scenarios pass. Re-running does not help —
  the runs below are already attempts 2 and 3.
- **Not either of the Linux causes.** The MI300X lane has two stacked blockers, both
  Linux-specific: the Lemonade embeddable's glibc floor (#258) and, once that is
  out of the way, an unresolvable runtime directory (#259). Windows uses a
  different artifact and a different launch path, and its symptom is the opposite one —
  exit 0, not exit 1.
- **Not platform-wide breakage.** Strix Halo *Ubuntu* passes this scenario on the same
  ROCm 7.13.0 / lemonade 11.5.1 pin.
- **Not a model-id mismatch in the readiness probe.** The direct path passes
  `--alias unsloth/Qwen3-0.6B-GGUF:Q4_0`, which contains the `Qwen3-0.6B` substring the
  probe looks for, so a live server would match.

## Where it likely lives

The two paths diverge in the lemonade engine:

[...]

PR #262, opened 2026-08-14
fix(lemonade): give the server a runtime dir when XDG_RUNTIME_DIR is unset

### LINKED ISSUE #259: rocm serve --engine lemonade fails on any headless host where XDG_RUNTIME_DIR is unset

rominf:
Fixes #259.

## Symptom

`rocm serve --engine lemonade --managed` fails on any Linux host with no
`XDG_RUNTIME_DIR`. It is fully deterministic and independent of the model or the serve
path:

```
Lemonade startup log tail (.../engines/lemonade/runtime/install-lemond.log):
[Error] (main) Error: Unable to resolve writable runtime directory from
XDG_RUNTIME_DIR or RUNTIME_DIRECTORY
```

The same command works from an interactive desktop or SSH login on the same machine, so
it reads as a machine quirk when it is really a property of how the process was started:
a systemd unit, a cron job, a container, `ssh host 'rocm serve ...'` without lingering,
or any CI runner.

## Root cause

The message comes from the spawned server, not from rocm-cli — the string appears nowhere
in this tree and `RUNTIME_DIRECTORY` is never referenced here. The server needs a writable
runtime directory and consults exactly two sources, neither of which exists in a headless
context:

- `XDG_RUNTIME_DIR`, populated by `pam_systemd` at login and therefore absent for every
  non-login process;
- `RUNTIME_DIRECTORY`, which only exists for a systemd unit declaring `RuntimeDirectory=`.

`lemonade_process_environment_vars` is the single place the child environment is
assembled. It translated `ROCM_PATH`, `PATH`, `LD_LIBRARY_PATH`, `HIP_VISIBLE_DEVICES` and
`LEMONADE_API_KEY`, but nothing about the runtime directory. rocm-cli cannot change what
the server requires, but it owns the environment it hands over, so this is the layer that
can satisfy it.

## Fix

Reuse the precedence this repo already has rather than inventing a second one.
`dashboard_socket_path` already solves "pick a writable, user-owned runtime directory with
a fallback chain", as a pure function taking its env inputs as arguments:

1. `$XDG_RUNTIME_DIR`
2. `$HOME/.rocm/data/…`
3. `temp_dir()/rocm-<user>`, with the user name sanitized so a separator or `..` cannot
   escape the subdirectory

That tier chain moves into `rocm_core::user_runtime_dir`; `dashboard_socket_path` becomes
a thin caller over it and resolves to byte-identical paths in all three tiers (its three
existing tests are unchanged and still pass). The engine calls the same helper when the
parent environment has no usable value, creates the directory mode `0700` before spawning,
and passes it down.

## Why not fix this in CI

A workflow-level fix was considered and rejected. Exporting `XDG_RUNTIME_DIR` in
`.github/workflows/e2e-selfhosted.yml`, or setting it for piped scenarios in the E2E
harness, would turn the GPU lane green and leave every headless user exactly as broken.

`pty_env` (`tests/e2e-cucumber/tests/e2e.rs`) does set `HOME` and `XDG_RUNTIME_DIR`, but
only for PTY scenarios — its comment records that piped scenarios keep their historical
HOME/XDG environment, and the serve scenarios are piped. That is correct and stays: it is
precisely why the suite reproduces the real headless case instead of hiding it.

## Decisions worth reviewing

**A distinct directory for the engine, not the dashboard's.** In the fallback tiers the
engine gets a `lemonade` leaf (`$HOME/.rocm/data/lemonade`,
`temp_dir()/rocm-<user>/lemonade`). Sharing one directory would put engine scratch state
next to the telemetry socket, so a crashing or misbehaving server could disturb the
dashboard. The isolation costs one path component.

**A shared helper, but only for the two callers that can share one.** The tier chain is
extracted into `rocm-core`, used by `dashboard_socket_path` and by the engine, so there
are still two copies of this logic rather than three. `rocm-dash-core` keeps its own
mirror deliberately: it is a standalone library with a lean dependency set, and depending
on `rocm-core` to share ~30 lines would pull in that crate's whole graph. Its "keep the
two in sync" comment now names the shared helper so the sync target is unambiguous, and
the mirrored tests on both sides still catch divergence.

**`XDG_RUNTIME_DIR` rather than `RUNTIME_DIRECTORY`.** The server reads either.
`RUNTIME_DIRECTORY` is narrower, but it is a systemd-unit contract implying systemd
created the directory and will clean it up — untrue here, and misleading to anyone
debugging. `XDG_RUNTIME_DIR` is the freedesktop-standard variable whose absence is the bug;
supplying a valid one is strictly better for any descendant process than leaving it unset.
It is never overridden: a value already in the parent environment is passed through
unchanged.

**Hardening.** The tier-3 fallback lands under a shared temp dir, so a pre-existing path
there may not be ours. A symlink is refused rather than `chmod`-ed through, and a chmod
failure is a hard error naming the path.

## Verification

Ran locally on a headless Linux host (no `XDG_RUNTIME_DIR`, no `RUNTIME_DIRECTORY`):

- `cargo fmt --all --check`
- `cargo clippy --locked --workspace --all-targets -- -D warnings`
- `cargo clippy --locked -p e2e-cucumber --test e2e -- -D warnings`
- `cargo test --workspace --all-targets`, plus the existing `dashboard_socket_path` and
  `rocm-dash-core` mirror tests
- driving the real resolver against this host's actual environment: it produced
  `$HOME/.rocm/data/lemonade`, created `drwx------`

Five new unit tests on `lemonade_process_environment_vars` cover synthesis when the parent
has no value, an exported-but-empty value, the no-`HOME` tier, pass-through of an existing
value, and refusal of a symlinked directory. All five fail before this change and pass
after. They take the environment as arguments rather than mutating process-global env, so
they behave identically under `cargo test` and nextest.

Re-verified after rebasing onto current `main`: all five gates above pass, and
`cargo test --workspace --all-targets` is now fully green. Two shared-process `cargo test`
failures noted here earlier
(`therock::tests::extracting_the_sdk_archive_removes_it`,
`providers::tests::local_provider_default_chat_requires_builtin_qwen_assistant`) were
pre-existing on `main` and are gone since #253 landed; nothing about them was ever related
to this change.

**The real verification is the GPU lane, and it is now green.** No GPU here, so a full
managed serve could not be run locally; `@id:serve-lemonade-inference` and
`@id:serve-hf-checkpoint-inference` (`@requires-gpu @requires-engine:lemonade`) exercise
this on `E2E tests (GPU)`, the MI300X lane where the failure reproduces.

| | before ([job](https://github.com/ROCm/rocm-cli/actions/runs/31679587566/job/94711980880), the run in #259) | after ([job](https://github.com/ROCm/rocm-cli/actions/runs/31800860515/job/94768346825)) |
|---|---|---|
| "Unable to resolve writable runtime directory" in the log | 2 occurrences | none |
| `serve-hf-checkpoint-inference` | failed | **passes** |
| reconciliation | 1 unexpected failure — lane red | 0 unexpected failures — lane green |

`serve-lemonade-inference` still xfails, but on the unrelated EAI-7423 symptom it is
already registered for ("did not serve model within 90s"), and it now gets past server
startup instead of exiting immediately.

Worth being explicit about the shape of that coverage: the scenarios are piped, so they
inherit the runner's environment rather than asserting on it — they regress this bug
because that runner is headless, not because the harness forces it. Making it explicit
would mean touching the piped scenarios' environment, which is the thing this fix
deliberately avoids.

`tests/e2e-cucumber/expectations.toml` was searched: no row describes this failure mode.
The lemonade Linux rows there are EAI-7423 ("reaches ready then shuts down immediately") on
the Strix Halo Ubuntu lane, a different symptom on a different host, and are left alone.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3735,6 +3735,7 @@ dependencies = [
  "anyhow",
  "clap",
  "flate2",
+ "libc",
  "rocm-core",
  "rocm-deps",
  "rocm-engine-protocol",
--- a/crates/rocm-core/src/lib.rs
+++ b/crates/rocm-core/src/lib.rs
@@ -69,7 +69,7 @@ pub use runtime::{
     runtime_path_text_is_absolute_for_platform, runtime_paths_equivalent,
     runtime_python_activation_hint, runtime_python_activation_script, runtime_python_bin_dir_name,
     runtime_python_env_bin_dir, runtime_python_executable_in_env, runtime_python_executable_name,
-    runtime_rocm_library_filename, shell_command_for_host,
+    runtime_rocm_library_filename, shell_command_for_host, user_runtime_dir,
 };
 pub use uv::{
     DEFAULT_UV_TIMEOUT_SECS, UV_CACHE_DIR_ENV, UV_CACHE_DIR_OVERRIDE_ENV, UvCacheSource,
@@ -5241,8 +5241,8 @@ pub struct RocmCliConfig {
 fn default_dashboard_socket() -> String {
     // Choose a socket location whose *parent* directory is always user-owned so
     // that run_unix can tighten it to 0o700 without EPERM. See
-    // `dashboard_socket_path` for the precedence. This resolver is mirrored in
-    // `rocm-dash-core` so the canonical `rocm` config and a standalone
+    // `runtime::user_runtime_dir` for the precedence. This resolver is mirrored
+    // in `rocm-dash-core` so the canonical `rocm` config and a standalone
     // `rocm-dash` config resolve to the same place; keep the two in sync.
     let path = dashboard_socket_path(
         std::env::var_os("XDG_RUNTIME_DIR"),
@@ -5266,48 +5266,18 @@ fn default_dashboard_socket() -> String {
 /// 2. `$HOME/.rocm/data/telemetry` — standard per-user data dir.
 /// 3. `temp_dir()/rocm-<user>` — user-named subdir so the parent is something we
 ///    create and own, not `/tmp` itself.
+///
+/// The tier chain itself lives in [`user_runtime_dir`], which the Lemonade
+/// engine also uses to synthesize a runtime directory for its child process.
+/// Only tier 2 needs the `telemetry` leaf: tiers 1 and 3 are already per-user
+/// runtime directories, so the socket sits directly in them.
 fn dashboard_socket_path(
     xdg_runtime_dir: Option<std::ffi::OsString>,
     home: Option<std::ffi::OsString>,
     user: Option<String>,
     temp_dir: std::path::PathBuf,
 ) -> std::path::PathBuf {
-    use std::path::PathBuf;
-    xdg_runtime_dir
-        .filter(|v| !v.is_empty())
-        .map(|d| PathBuf::from(d).join("rocmdashd.sock"))
-        .or_else(|| {
-            home.filter(|v| !v.is_empty()).map(|h| {
-                PathBuf::from(h)
-                    .join(".rocm")
-                    .join("data")
-                    .join("telemetry")
-                    .join("rocmdashd.sock")
-            })
-        })
-        .unwrap_or_else(|| {
-            let raw = user.unwrap_or_else(|| "user".to_owned());
-            // Sanitize: keep only alphanumeric, hyphen, and underscore so a path
-            // separator or `..` in the env var cannot escape the subdirectory.
-            let sanitized: String = raw
-                .chars()
-                .map(|c| {
-                    if c.is_alphanumeric() || c == '-' || c == '_' {
-                        c
-                    } else {
-                        '_'
-                    }
-                })
-                .collect();
-            let sanitized = if sanitized.is_empty() {
-                "user".to_owned()
-            } else {
-                sanitized
-            };
-            temp_dir
-                .join(format!("rocm-{sanitized}"))
-                .join("rocmdashd.sock")
-        })
+    user_runtime_dir(xdg_runtime_dir, home, user, temp_dir, "telemetry", "").join("rocmdashd.sock")
 }
 
 fn default_dashboard_listen() -> String {
--- a/crates/rocm-core/src/runtime.rs
+++ b/crates/rocm-core/src/runtime.rs
@@ -218,6 +218,80 @@ pub fn default_cache_dir() -> Option<PathBuf> {
         .or_else(|| project_dirs().map(|dirs| dirs.cache_dir().to_path_buf()))
 }
 
+/// Resolve a writable, user-owned directory for per-user runtime state from
+/// explicit environment inputs.
+///
+/// Runtime state (sockets, lock files, an engine's scratch area) needs a
+/// directory whose *parent* is user-owned, so it can be tightened to mode
+/// `0o700` without the `EPERM` that results from trying to `chmod` a shared,
+/// root-owned `/tmp` (mode `1777`). Precedence:
+///
+/// 1. `$XDG_RUNTIME_DIR` — already mode `0700` on systemd systems, ideal.
+/// 2. `$HOME/.rocm/data/<home_subdir>` — standard per-user data dir.
+/// 3. `temp_dir()/rocm-<user>/<temp_subdir>` — user-named subdir so the parent
+///    is something we create and own, not `/tmp` itself.
+///
+/// Tiers 2 and 3 are not exotic: `$XDG_RUNTIME_DIR` is populated by
+/// `pam_systemd` at login, so it is absent for every non-login process (cron
+/// jobs, CI runners, `systemd-run`, a bare container exec).
+///
+/// The environment is taken as arguments rather than read here so the
+/// precedence is testable without mutating process-global env vars, which is
+/// `unsafe` and racy under parallel tests in edition 2024.
+///
+/// `home_subdir` and `temp_subdir` name the caller's own leaf directory in
+/// tiers 2 and 3; either may be empty when the caller owns that whole tier.
+/// They are separate parameters because tier 2 lands inside the shared
+/// `.rocm/data` tree, where every component needs its own leaf, whereas tier 3
+/// is already under a rocm-private `rocm-<user>` directory.
+pub fn user_runtime_dir(
+    xdg_runtime_dir: Option<OsString>,
+    home: Option<OsString>,
+    user: Option<String>,
+    temp_dir: PathBuf,
+    home_subdir: &str,
+    temp_subdir: &str,
+) -> PathBuf {
+    if let Some(runtime_dir) = xdg_runtime_dir.filter(|value| !value.is_empty()) {
+        return PathBuf::from(runtime_dir);
+    }
+    if let Some(home) = home.filter(|value| !value.is_empty()) {
+        return join_subdir(PathBuf::from(home).join(".rocm").join("data"), home_subdir);
+    }
+    let user_dir = format!("rocm-{}", sanitized_user_path_component(user));
+    join_subdir(temp_dir.join(user_dir), temp_subdir)
+}
+
+fn join_subdir(mut base: PathBuf, subdir: &str) -> PathBuf {
+    if !subdir.is_empty() {
+        base.push(subdir);
+    }
+    base
+}
+
+/// Reduce a user name to a single safe path component: keep only alphanumeric,
+/// hyphen, and underscore so a path separator or `..` in the env var cannot
+/// escape the intended subdirectory. An absent or fully-stripped name yields
+/// `user`.
+fn sanitized_user_path_component(user: Option<String>) -> String {
+    let sanitized: String = user
+        .unwrap_or_default()
+        .chars()
+        .map(|c| {
+            if c.is_alphanumeric() || c == '-' || c == '_' {
+                c
+            } else {
+                '_'
+            }
+        })
+        .collect();
+    if sanitized.is_empty() {
+        "user".to_owned()
+    } else {
+        sanitized
+    }
+}
+
 pub fn managed_runtime_cache_dir(root: &Path) -> PathBuf {
     normalize_runtime_path_for_host(root).join("cache")
 }
--- a/crates/rocm-dash-core/src/config.rs
+++ b/crates/rocm-dash-core/src/config.rs
@@ -115,6 +115,12 @@ pub struct DaemonConfig {
 /// 2. `$HOME/.rocm/data/telemetry` — standard per-user data dir.
 /// 3. `temp_dir()/rocm-<user>` — user-named subdir so the parent is something
 ///    the daemon creates and owns, not `/tmp` itself.
+///
+/// In `rocm-core` the tier chain is shared as `runtime::user_runtime_dir`. This
+/// crate keeps its own copy on purpose: it is a standalone library with a
+/// deliberately lean dependency set, and depending on `rocm-core` to share ~30
+/// lines would pull in that crate's whole graph. Keep the two in sync — the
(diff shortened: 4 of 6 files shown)


rominf:
Rebased onto current `main` and out of draft. Post-rebase CI, on the lane this PR is about:

**`E2E tests (GPU)` is green** — [job](https://github.com/ROCm/rocm-cli/actions/runs/32029548789/job/95386322927).
That lane fails on `main` today with `Unable to resolve writable runtime directory from
XDG_RUNTIME_DIR or RUNTIME_DIRECTORY` (twice per run), which is #259 and what this change fixes.

**The one remaining red check, `E2E tests (Strix Halo, Windows)`, is not this PR.** It reports
exactly the same single unexpected failure as `main` does —
`serve-hf-checkpoint-inference` — which is #260, a Windows-only bug with a different cause and
symptom (exit 0 and never serves, versus exit 1 here):

```
39 scenarios (35 passed, 4 failed)
Reconciliation: 3 xfail, 0 XPASS, 1 unexpected failure(s).
FAIL: 'serve-hf-checkpoint-inference' was expected to pass on this host but FAILED
```

Identical to the baseline on `main`, so this PR neither causes nor fixes it. #260 is tracked
separately.

Everything else — all 16 required checks, plus the Strix Halo Ubuntu lane — is green.

rominf:
Rebased onto `main` (97ec0aa4) and taken out of draft. All 16 required checks are green.

The one red check, `E2E tests (Strix Halo, Windows)`, is not a required check and is not caused by this PR. Comparing against the `main` run for the very same base commit:

| lane | `main` @ 97ec0aa4 | this PR @ afd657bd |
|---|---|---|
| `E2E tests (GPU)` | [fails](https://github.com/ROCm/rocm-cli/actions/runs/32028129073/job/95381833747) — 1 unexpected failure (`serve-hf-checkpoint-inference`), 2 occurrences of "Unable to resolve writable runtime directory" | [passes](https://github.com/ROCm/rocm-cli/actions/runs/32029548789/job/95386322927) — 0 unexpected failures, 0 occurrences of that message |
| `E2E tests (Strix Halo, Windows)` | fails | [fails](https://github.com/ROCm/rocm-cli/actions/runs/32029548789/job/95386322786) |

The Windows lane fails on both, and on a different symptom: `serve-hf-checkpoint-inference` reaching `did not serve model Qwen3-0.6B within 600s`, which is #260 and tracked separately. This change is Linux-only by construction — `child_runtime_dir_var` returns early unless `runtime_is_linux()` — so it cannot move that lane either way.

Re-verified locally after the rebase: `cargo fmt --all --check`, `cargo clippy --locked --workspace --all-targets -- -D warnings`, `cargo clippy --locked -p e2e-cucumber --test e2e -- -D warnings`, `cargo test --workspace --all-targets` (fully green now that #253 has landed), and `python scripts/smoke_local.py`.

michaelroy-amd (commented):
- **Review focus:** Please review this as filesystem security code: fallback precedence, descriptor-relative tier-3 creation, ownership/mode checks, and unchanged operator-provided `XDG_RUNTIME_DIR` behavior.
- **Problem / impact:** On headless Linux processes without `XDG_RUNTIME_DIR` or `RUNTIME_DIRECTORY`, managed Lemonade exits before serving. CI runners, cron jobs, systemd-run invocations, and bare container execs therefore cannot use Lemonade. A naive `/tmp` fallback would add symlink, ownership, and path-replacement races.
- **State:** [#262](https://github.com/ROCm/rocm-cli/pull/262), head `1f2ce58182f56b24e58fba6694733a5ef5b24bd4`; `OPEN`, non-draft, `MERGEABLE`, `REVIEW_REQUIRED` (`BLOCKED`); 6 files, +1,172/-41.
  1. `crates/rocm-core/src/runtime.rs` and `lib.rs`: shared precedence `XDG_RUNTIME_DIR` → `$HOME/.rocm/data/<leaf>` → `<temp>/rocm-<sanitized-user>/<leaf>`.
  2. `crates/rocm-dash-core/src/config.rs`: dashboard socket precedence remains behaviorally aligned without adding a heavy dependency.
  3. `engines/lemonade/src/lib.rs`: Linux-only child `XDG_RUNTIME_DIR`, `lemonade` isolation leaf, and one-time environment resolution reused by readiness probes.
  4. Tier-3 hardening: `openat`/`mkdirat`, `O_NOFOLLOW`, descriptor-relative creation, effective-UID ownership, mode 0700, and ancestry checks that prevent cross-UID entry replacement.
  5. Race/security regressions: symlinked leaves, foreign ownership, unsafe ancestors, temp-root replacement, user-parent replacement, and operator pass-through.
- **Invariants to verify:** A non-empty parent `XDG_RUNTIME_DIR` passes through byte-for-byte with no fallback mutation; HOME and temp fallbacks end in a private `lemonade` leaf; user input cannot escape its path component; every created/opened component is no-follow, correctly owned, and private; creation stays anchored to the validated directory object if pathnames are swapped; failures are hard errors, never a permissive fallback; Windows behavior and strict GPU/no-CPU-fallback policy are unchanged.
- **Verification evidence:** **Required:** all required hosted checks green, including Linux/Windows builds/tests, affected tests, clippy, signature+DCO, hygiene, and CodeQL. **Advisory:** GPU, Strix Ubuntu, and the self-hosted consolidated report pass—this is the only packet whose GPU leaf is green. Red: [Strix Windows](https://github.com/ROCm/rocm-cli/actions/runs/32049791891/job/95446202462) has five unexpected failures (`bench-load-real-serve`, `chat-tool-definitions-accepted`, `chat-end-to-end-local-model`, `serve-lemonade-inference`, `serve-hf-checkpoint-inference`) after the llama.cpp backend download exhausts six attempts with `SSL connect error (CURL code: 35)`.
- **Recommended sequence:** shared resolver → Lemonade child-env call sites → fd-relative hardening → adversarial tests → dashboard parity/platform gates.

michaelroy-amd (approved):
Approved at `1f2ce58182f56b24e58fba6694733a5ef5b24bd4`. The focused review above covers the changed contracts and invariants. All required checks pass; the current self-hosted advisory failure is documented there and is not merge-required.

PR #265, opened 2026-08-15
test: stabilize shared pre-push fixtures

michaelroy-amd:
## Summary
- tolerate expected peer disconnects in scripted download responses without masking unrelated I/O failures
- keep wrong-offset resume fixtures alive for the clean restart request
- replace the user-home-dependent app-path test with a deterministic synthetic configuration

## Verification
- `cargo test --workspace --all-targets` (2118 passed; 6 ignored)
- 100 repeated oversized-body runs and 30 repeated wrong-offset-resume runs

No Gherkin scenario is needed: this changes internal test fixtures and test isolation only; observable CLI behavior is unchanged.

### DIFF
--- a/crates/rocm-core/src/lib.rs
+++ b/crates/rocm-core/src/lib.rs
@@ -52,8 +52,6 @@ pub use proc_lifecycle::{
     process_start_ticks, terminate_verified,
 };
 use runtime::env_path_override;
-#[cfg(test)]
-use runtime::home_rocm_dir;
 pub use runtime::{
     RuntimeHost, RuntimePlatform, current_executable_path, default_cache_dir, default_config_dir,
     default_data_dir, default_interactive_shell_program, managed_logs_dir, managed_pip_cache_dir,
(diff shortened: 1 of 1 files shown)


juhovainio (approved):
Reviewed the diff and built/tested it locally (cargo check, clippy -D warnings, and the affected unit tests, including 20 repeats of the two previously-flaky ones — all stable).

This is a clean, test-only fix:

- The mock download server now only tolerates a peer disconnect (`BrokenPipe`/`ConnectionReset`) in the two scenarios where the client is expected to hang up early (rejecting a wrong-offset resume, and aborting after seeing an over-limit `Content-Length`), while still surfacing any other I/O error. That's the actual root cause of the pre-push flakiness this PR is fixing.
- `AppPaths::discover_from_paths` is a straightforward, behavior-preserving extraction of the existing override/managed-root logic out of `discover()`, purely to make it testable without touching `$HOME`. I traced it against `main` and it's identical.
- The new `app_paths_apply_configured_managed_root_when_unoverridden` test covers all three override combinations against a synthetic config, replacing a test that used to silently no-op if any `ROCM_CLI_*` env var was set.

No user-observable behavior changes, so the "no Gherkin scenario needed" call in the PR description is correct. One tiny non-blocking nit: `download_fixture_tolerates_only_expected_peer_disconnects` tests the test harness itself rather than production code, but that's reasonable given the PR is specifically about hardening that harness.

Approving — nice fix for a genuinely annoying source of pre-push flakiness.

PR #251, opened 2026-08-13
feat(vllm): Tackle out of memory errors (EAI-8058)

r0x0r:
This pull request improves GPU selection and user guidance for ROCm and vLLM, especially in shared or containerized environments where the standard `amd-smi` tool may not be available. It adds a fallback for GPU VRAM telemetry, enhances user warnings and hints for out-of-memory (OOM) conditions, adds the vLLM startup OOM to the `diagnose`/`fix` catalog, and ensures consistent messaging across CLI and engine surfaces.

**GPU selection and VRAM telemetry improvements:**
- Added a fallback to read per-GPU VRAM usage from the amdgpu DRM sysfs counters (`/sys/class/drm/card*/device/mem_info_vram_{total,used}`) when `amd-smi` is not available, so `--gpu auto` ranking and low-VRAM warnings still have telemetry on a **single-AMD-GPU** host without amd-smi. The fallback deliberately **withholds** telemetry on a host with more than one AMD DRM card, because `card<N>` numbering is not guaranteed to match HIP's device ordinal there — so it is a single-GPU convenience, not a multi-GPU replacement.
- Explicit `--gpu <index>` validation was reworked on this branch to check membership in the KFD/DRM usable set rather than a device count. **That is superseded by #267**, which landed a validator consulting the visibility-resolved set directly (and with mask-aware wording), and the merge below drops this branch's version in favour of it — the two answered the same question, and #267's answer is the stronger one. The DRM sysfs VRAM rows are never used as a `--gpu <index>` bound; they stay scoped to `--gpu auto` ranking.
- `--gpu auto` ranks the **actual reported VRAM row indices**, which can be non-contiguous under a visibility mask, rather than a synthetic `0..count` range that would look up absent rows and fall through to a bogus GPU 0.

**User guidance and warnings for vLLM:**
- Introduced a shared constant `VLLM_GPU_MEMORY_UTILIZATION_HINT` for the recommended workaround when running out of memory on a shared/busy GPU, ensuring CLI and engine logs use consistent wording.
- The serve summary prints a note about the `--gpu-memory-utilization` workaround when vLLM is selected and the GPU is busy, both interactively and in the deployment summary.
- vLLM engine startup logs append the same utilization hint if an OOM is detected.

**vLLM startup OOM in the diagnose/fix catalog (EAI-8060):**
[...]

PR #272, opened 2026-08-17
fix(therock): resolve ROCm releases from the current multi-arch pip index

### LINKED ISSUE #271: rocm install sdk: release channel never resolves ROCm 7.14+ (stuck on 7.13.0)
**Current behavior**

`rocm install sdk` (default `--format wheel`, `--channel release`) always
resolves ROCm 7.13.0, even though AMD's documented current stable release is
7.14.0 (see https://rocm.docs.amd.com/en/latest/install/rocm.html).

Root cause: `therock_index_urls()` in `apps/rocm/src/therock.rs` builds two
Release-channel pip index candidates:

1. `https://repo.amd.com/rocm/whl/{family}` - the classic, per-family index.
   Still live, but AMD stopped publishing new releases here; it tops out at
   7.13.0.
2. `https://repo.amd.com/rocm/whl-multi-arch/{family}` - intended as the
   newer index, but AMD's `whl-multi-arch` index is actually flat (no
   per-family path segment). GPU targeting there is done via pip extras
   (`device-gfxNNNN`, `device-all`) instead of URL scoping. Appending
   `/{family}` 403s:

   ```
   $ curl -s -o /dev/null -w '%{http_code}\n' https://repo.amd.com/rocm/whl-multi-arch/gfx110X-all/rocm/
   403
   $ curl -s -o /dev/null -w '%{http_code}\n' https://repo.amd.com/rocm/whl-multi-arch/rocm/
   200
   ```

`resolve_pip_runtime_with_timeout` tries index URLs in order and returns on
the first that fully resolves a compatible rocm/torch/torchvision/torchaudio
set. Since the classic index is tried first and always succeeds (just
stale), the broken multi-arch entry never gets a chance to contribute
7.14.0.

There's a second, related gap: even once the multi-arch URL is fixed, its
`rocm` sdist requires an explicit `device`/`device-all`/`device-gfxNNNN` pip
extra to pull in any GPU backend at all. The CLI's fixed
`rocm[libraries,devel]` spec would silently install no GPU backend from that
index.

**Expected behavior**

`rocm install sdk` on the release channel installs the current stable ROCm
release (7.14.0 or newer as AMD publishes them), with a working GPU backend,
while older ROCm versions and families remain installable from the classic
`whl/{family}` index as a fallback.

**Steps to reproduce**

```
rocm install sdk --channel release --dry-run
```

The dry-run output's `index_url` and `latest_compatible_version` show the
classic `whl/{family}` index and 7.13.0, never the multi-arch index or
7.14.0.

**Possible solution**

Fixed in the accompanying PR: try the (correctly-shaped, flat) multi-arch
index first, fall back to the classic per-family index for older
versions/families, and add the required `device-*` extra when installing
from the multi-arch index.

Out of scope / related follow-up: the tarball install path
(`THEROCK_RELEASE_TARBALL_BASE = "https://repo.amd.com/rocm/tarball/"`) looks
like it may have the same "AMD moved to a `-multi-arch` variant" problem, but
`rocm install sdk`'s default format is `wheel`, not `tarball`, so it wasn't
investigated as part of this issue.

**Your environment**

- rocm-cli channel: release
- Platform: Linux

juhovainio:
## Summary

Fixes #271.

- `therock_index_urls()` appended a `/{family}` path segment to
  `repo.amd.com/rocm/whl-multi-arch`, but that index is flat (no per-family
  path); the request 403'd, so the CLI silently fell through to the stale
  classic `whl/{family}` index (stuck at 7.13.0) every time.
- Now tries the multi-arch index (correct, flat URL) first, then falls back
  to the classic per-family index. Reuses the existing "try each index in
  order, return on first full success" resolution loop as-is, so older
  releases/families that only exist on the classic index keep resolving
  exactly as before - the previous source stays a working fallback.
- The multi-arch `rocm` sdist needs an explicit `device-*` pip extra to pull
  in any GPU backend at all. Added `known_therock_family_device_chips` in
  `rocm-core` (mirrors the existing `normalize_therock_family` bucketing) to
  request exact `device-gfxNNNN` extras for the 10 families with an
  enumerable chip set, falling back to `device-all` for the remaining 6
  prefix-bucket families where exact membership isn't derivable. The classic
  index path is unaffected (still gets plain `libraries,devel`).

## Test plan

- [x] `cargo build -p rocm-core -p rocm`
- [x] `cargo clippy -p rocm-core -p rocm --all-targets` (clean)
- [x] `cargo test -p rocm-core` (296 passed)
- [x] `cargo test -p rocm` (459 passed), including new coverage:
      `therock_index_urls_prefers_multi_arch_then_classic_for_release`,
      `therock_index_urls_nightly_unchanged`,
      `therock_rocm_extras_classic_index_is_unchanged`,
      `therock_rocm_extras_multi_arch_adds_exact_device_chips`,
      `therock_rocm_extras_multi_arch_falls_back_to_device_all_for_ambiguous_bucket`,
      and `known_therock_family_device_chips_round_trip_to_their_family` in
      `rocm-core`.
- [ ] Manual dry-run against the live index (`rocm install sdk --channel
      release --dry-run`) to confirm 7.14.0 and the right `device-*` extra
      are resolved for a real host - not run in this environment, worth a
      sanity check before merge.

### DIFF
--- a/apps/rocm/src/therock.rs
+++ b/apps/rocm/src/therock.rs
@@ -6,7 +6,8 @@ use anyhow::{Context, Result, bail};
 use rocm_core::{
     AppPaths, ManagedToolConfig, RocmCliConfig, detect_host_gpu_diagnostics,
     detect_host_therock_family, detect_managed_therock_family, disk_space, ensure_uv_binary,
-    known_therock_families, managed_tools_dir, normalize_runtime_path_for_host,
+    known_therock_families, known_therock_family_device_chips, managed_tools_dir,
+    normalize_runtime_path_for_host,
     normalize_runtime_path_for_storage, normalize_runtime_path_text_for_host,
     normalize_runtime_path_text_for_storage, normalize_therock_family, runtime_is_windows,
     runtime_os_name, runtime_path_for_windows_child, runtime_path_list_split,
@@ -826,6 +827,7 @@ fn install_wheel_runtime(
         "Found TheRock package family {} version {} with a matching PyTorch stack.",
         resolution.family, resolution.latest_version
     ));
+    let rocm_extras = therock_rocm_extras(&resolution.family, &resolution.index_url);
     let runtime_key = runtime_key(
         channel,
         "wheel",
@@ -880,7 +882,7 @@ fn install_wheel_runtime(
     let _ = writeln!(
         output,
         "  package_specs: {}",
-        therock_pip_package_specs(&resolution.package_versions).join(" ")
+        therock_pip_package_specs(&resolution.package_versions, &rocm_extras).join(" ")
     );
     let _ = writeln!(
         output,
@@ -893,7 +895,10 @@ fn install_wheel_runtime(
         if matches!(channel, TheRockChannel::Nightly) {
             install_args.extend(["--prerelease".to_owned(), "allow".to_owned()]);
         }
-        install_args.extend(therock_pip_package_specs(&resolution.package_versions));
+        install_args.extend(therock_pip_package_specs(
+            &resolution.package_versions,
+            &rocm_extras,
+        ));
         let venv_args = uv_venv_args(&python_launcher.executable, &install_root);
         let venv_args_display = venv_args
             .iter()
@@ -933,7 +938,7 @@ fn install_wheel_runtime(
 
     progress_line(format!(
         "Installing {} from {}",
-        therock_pip_package_specs(&resolution.package_versions).join(" "),
+        therock_pip_package_specs(&resolution.package_versions, &rocm_extras).join(" "),
         resolution.index_url
     ));
     let mut install_args = uv_pip_install_base(&env_python);
@@ -941,7 +946,10 @@ fn install_wheel_runtime(
     if matches!(channel, TheRockChannel::Nightly) {
         install_args.extend(["--prerelease".to_owned(), "allow".to_owned()]);
     }
-    install_args.extend(therock_pip_package_specs(&resolution.package_versions));
+    install_args.extend(therock_pip_package_specs(
+        &resolution.package_versions,
+        &rocm_extras,
+    ));
     run_uv_progress_command(
         paths,
         &uv,
@@ -1012,13 +1020,40 @@ fn install_wheel_runtime(
     Ok(output)
 }
 
-fn therock_pip_package_specs(package_versions: &TheRockPipPackageVersions) -> Vec<String> {
+fn therock_pip_package_specs(
+    package_versions: &TheRockPipPackageVersions,
+    rocm_extras: &str,
+) -> Vec<String> {
     vec![
-        format!("rocm[libraries,devel]=={}", package_versions.rocm),
+        format!("rocm[{rocm_extras}]=={}", package_versions.rocm),
         format!("torch=={}", package_versions.torch),
         format!("torchvision=={}", package_versions.torchvision),
         format!("torchaudio=={}", package_versions.torchaudio),
     ]
+}
+
+fn is_multi_arch_pip_index(index_url: &str) -> bool {
+    index_url.trim_end_matches('/') == THEROCK_RELEASE_PIP_MULTI_ARCH_INDEX_BASE
+}
+
+/// The `rocm[...]` extras to request for a resolved pip index. The classic
+/// per-family index needs only the base `libraries,devel` extras; the flat
+/// multi-arch index additionally needs an explicit `device-*` extra or no GPU
+/// backend gets installed at all.
+fn therock_rocm_extras(family: &str, index_url: &str) -> String {
+    let mut extras = "libraries,devel".to_owned();
+    if !is_multi_arch_pip_index(index_url) {
+        return extras;
+    }
+    match known_therock_family_device_chips(family) {
+        Some(chips) => {
+            for chip in chips {
+                let _ = write!(extras, ",device-{chip}");
+            }
+        }
+        None => extras.push_str(",device-all"),
+    }
+    extras
 }
 
 fn quote_display_arg(value: &str) -> String {
@@ -3368,9 +3403,12 @@ fn parse_version(value: &str) -> Option<ParsedVersion> {
 
 fn therock_index_urls(channel: TheRockChannel, family: &str) -> Vec<String> {
     match channel {
+        // Multi-arch is flat (no per-family path segment) and is where AMD
+        // publishes current releases; try it first. The classic per-family
+        // index is kept as a fallback so older releases stay installable.
         TheRockChannel::Release => vec![
+            THEROCK_RELEASE_PIP_MULTI_ARCH_INDEX_BASE.to_owned(),
             format!("{THEROCK_RELEASE_PIP_INDEX_BASE}/{family}"),
-            format!("{THEROCK_RELEASE_PIP_MULTI_ARCH_INDEX_BASE}/{family}"),
         ],
         TheRockChannel::Nightly => vec![format!("{THEROCK_NIGHTLY_PIP_INDEX_BASE}/{family}")],
     }
@@ -4065,7 +4103,7 @@ mod tests {
             torchaudio: "2.10.0+rocm7.13.0a20260513".to_owned(),
             compatibility_key: "7.13.0a20260513".to_owned(),
         };
-        let package_specs = therock_pip_package_specs(&package_versions);
+        let package_specs = therock_pip_package_specs(&package_versions, "libraries,devel");
 
         assert_eq!(
             package_specs,
@@ -4075,6 +4113,51 @@ mod tests {
                 "torchvision==0.25.0+rocm7.13.0a20260513".to_owned(),
                 "torchaudio==2.10.0+rocm7.13.0a20260513".to_owned(),
             ]
+        );
+    }
+
+    #[test]
+    fn therock_index_urls_prefers_multi_arch_then_classic_for_release() {
+        let urls = therock_index_urls(TheRockChannel::Release, "gfx110X-all");
+        assert_eq!(
+            urls,
+            vec![
+                "https://repo.amd.com/rocm/whl-multi-arch".to_owned(),
+                "https://repo.amd.com/rocm/whl/gfx110X-all".to_owned(),
+            ]
+        );
+    }
+
+    #[test]
+    fn therock_index_urls_nightly_unchanged() {
+        let urls = therock_index_urls(TheRockChannel::Nightly, "gfx110X-all");
+        assert_eq!(
+            urls,
+            vec!["https://rocm.nightlies.amd.com/v2/gfx110X-all".to_owned()]
+        );
+    }
+
+    #[test]
+    fn therock_rocm_extras_classic_index_is_unchanged() {
+        assert_eq!(
+            therock_rocm_extras("gfx110X-all", "https://repo.amd.com/rocm/whl/gfx110X-all"),
+            "libraries,devel"
+        );
+    }
+
+    #[test]
+    fn therock_rocm_extras_multi_arch_adds_exact_device_chips() {
+        assert_eq!(
+            therock_rocm_extras("gfx110X-all", "https://repo.amd.com/rocm/whl-multi-arch"),
+            "libraries,devel,device-gfx1100,device-gfx1101,device-gfx1102,device-gfx1103"
+        );
+    }
+
+    #[test]
+    fn therock_rocm_extras_multi_arch_falls_back_to_device_all_for_ambiguous_bucket() {
+        assert_eq!(
+            therock_rocm_extras("gfx90X-dcgpu", "https://repo.amd.com/rocm/whl-multi-arch"),
+            "libraries,devel,device-all"
         );
     }
 
--- a/apps/rocm/src/therock.rs
+++ b/apps/rocm/src/therock.rs
@@ -886,7 +886,7 @@ fn install_wheel_runtime(
     );
     let _ = writeln!(
         output,
-        "  package_policy: find the newest TheRock ROCm SDK version that has a matching PyTorch stack in the same index, then install pinned rocm[libraries,devel], torch, torchvision, and torchaudio versions in one uv transaction"
+        "  package_policy: find the newest TheRock ROCm SDK version that has a matching PyTorch stack in the same index, then install pinned rocm (with any resolved device extras), torch, torchvision, and torchaudio versions in one uv transaction"
     );
     if dry_run {
         let env_python = venv_python_path(&install_root);
--- a/apps/rocm/src/therock.rs
+++ b/apps/rocm/src/therock.rs
@@ -7,12 +7,11 @@ use rocm_core::{
     AppPaths, ManagedToolConfig, RocmCliConfig, detect_host_gpu_diagnostics,
     detect_host_therock_family, detect_managed_therock_family, disk_space, ensure_uv_binary,
     known_therock_families, known_therock_family_device_chips, managed_tools_dir,
-    normalize_runtime_path_for_host,
-    normalize_runtime_path_for_storage, normalize_runtime_path_text_for_host,
-    normalize_runtime_path_text_for_storage, normalize_therock_family, runtime_is_windows,
-    runtime_os_name, runtime_path_for_windows_child, runtime_path_list_split,
-    runtime_python_executable_in_env, unix_time_millis, uv_command_env, uv_pip_install_base,
-    uv_venv_args, verify_rsa_pkcs1_sha256_signature,
+    normalize_runtime_path_for_host, normalize_runtime_path_for_storage,
+    normalize_runtime_path_text_for_host, normalize_runtime_path_text_for_storage,
+    normalize_therock_family, runtime_is_windows, runtime_os_name, runtime_path_for_windows_child,
+    runtime_path_list_split, runtime_python_executable_in_env, unix_time_millis, uv_command_env,
+    uv_pip_install_base, uv_venv_args, verify_rsa_pkcs1_sha256_signature,
 };
 #[cfg(test)]
 use rocm_core::{
--- a/tests/e2e-cucumber/features/runtime_setup.feature
+++ b/tests/e2e-cucumber/features/runtime_setup.feature
@@ -88,11 +88,19 @@ Feature: Runtime configuration
   # Regression guard for the bug where the release-channel multi-arch pip index
   # was queried with a per-family path segment (.../whl-multi-arch/{family}/),
   # which 403s because that index is flat and 404/403-ed straight into the
-  # stale classic index every time — silently pinning every install to
+  # stale classic index every time, silently pinning every install to
   # whatever version predated the migration. `--family` bypasses GPU
   # auto-detection so this needs no GPU, and `--dry-run` resolves the real
   # index without installing anything.
-  @id:runtime-install-sdk-release-index-shape
+  #
+  # `@nightly` for the same reason as scenario 8 above: this dry-run still
+  # resolves the real channel index over the network (dry-run only skips the
+  # venv/download, not index resolution), and the no-GPU mock lane's 64-way
+  # concurrency from that extra network work is what pushes
+  # `eai-7960-gen-tps-held-after-scrape-failure` and
+  # `eai-7960-gen-tps-expiry-boundary` past their validity window. Runs on the
+  # nightly lanes instead, where scenarios are serialized.
+  @id:runtime-install-sdk-release-index-shape @nightly
   Scenario: 4 - Resolving the SDK from the release channel never uses the broken multi-arch URL shape
     When the user dry-runs installing the SDK for a known family
     Then the resolved package index is not the broken per-family multi-arch path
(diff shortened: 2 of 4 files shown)


michaelroy-amd:
Reviewed current head `668456768fa463d587ad4823f159ff6539ff30f1` against #271 and the ROCm 10 path. The index ordering and `device-*` extra selection fix the right failure, and the unit/Cucumber coverage is well targeted. This cannot be approved in its current state: the branch is `DIRTY`/conflicting and required checks currently report 4 failures. Please rebase onto current `main`, get the required checks green, and complete the PR's unchecked live dry-run (`rocm install sdk --channel release --dry-run`) to confirm the resolved release and device extra, then re-request review. Minor follow-up: update the dry-run `package_policy` prose so it no longer implies only `rocm[libraries,devel]` on the multi-arch index.

tests/e2e-cucumber/features/runtime_setup.feature
```diff
@@ -29,3 +29,15 @@ Feature: Runtime configuration
...
+
+  # Regression guard for the bug where the release-channel multi-arch pip index
+  # was queried with a per-family path segment (.../whl-multi-arch/{family}/),
+  # which 403s because that index is flat and 404/403-ed straight into the
+  # stale classic index every time — silently pinning every install to
+  # whatever version predated the migration. `--family` bypasses GPU
+  # auto-detection so this needs no GPU, and `--dry-run` resolves the real
+  # index without installing anything.
+  @id:runtime-install-sdk-release-index-shape
+  Scenario: 4 - Resolving the SDK from the release channel never uses the broken multi-arch URL shape
```
rominf, line 104: This scenario issues `rocm install sdk --family gfx110X-all --dry-run`, which resolves the real `repo.amd.com` pip index over the network before printing the plan (dry-run skips the install, not the index resolution). That's the same call the `runtime-install-records-the-real-folder` scenario just above makes, and that one is tagged `@nightly` specifically because the extra concurrent network work on the no-GPU mock lane (64 scenarios at once) was measured to push two other scenarios past their timing budget 3/3 times. This scenario has no `@nightly`/`@requires-gpu` tag, so it will run at that same concurrency on every PR and can reintroduce the same flakiness the sibling scenario's comment describes. Consider tagging it `@nightly` (or otherwise excluding it from the unserialized mock lane) to match the sibling.

rominf (dismissed):
Nice fix for the flat multi-arch index and the missing `device-*` extra — the reasoning in the description and the new unit tests for `therock_rocm_extras`/`therock_index_urls` are solid and I don't see a bug in that part.

One thing in the new e2e coverage looks like it will cause flakiness for unrelated tests, so I'd like that addressed before merge:

The new `runtime-install-sdk-release-index-shape` scenario runs `rocm install sdk --family gfx110X-all --dry-run`, which still resolves the real `repo.amd.com` pip index over the network (dry-run only skips the venv/download, not the index resolution). That's the exact same CLI invocation as the existing `runtime-install-records-the-real-folder` scenario a few lines above it in this file — and that sibling scenario is deliberately tagged `@nightly`, with a comment explaining why: on the no-GPU mock lane, which runs up to 64 scenarios concurrently, the extra network work from this call was measured to push two timing-sensitive scenarios (`eai-7960-gen-tps-held-after-scrape-failure` and `eai-7960-gen-tps-expiry-boundary`) past their validity window, failing them 3/3 under load.

The new scenario has neither `@nightly` nor `@requires-gpu`, so it will run on every PR's no-GPU mock lane at the same concurrency that was already shown to cause this interference. It looks like it should carry the same `@nightly` tag (and equivalent justification) as its sibling, or otherwise be isolated from the default concurrent mock lane.

juhovainio:
## Heads up: this will conflict with #329

#329 adds ROCm 10 ("next" layout / `whl-next`) support and touches the same pip-index-resolution code this PR does. Verified with a real trial merge, two files conflict:

- **`apps/rocm/src/therock.rs`** (12 hunks, genuine semantic overlap): both PRs independently redesigned the same functions for adjacent problems (this PR for the `whl-multi-arch` 7.14+ index, #329 for the `whl-next` ROCm 10 index).
  - `therock_pip_package_specs`: this PR keeps `rocm_extras: &str`; #329 replaces it with `generation: TheRockIndexGeneration, raw_arch: Option<&str>`.
  - `therock_index_urls` (this PR) vs. `legacy_pip_index_candidates`/`therock_pip_index_candidates` (#329): #329 replaced the flat `Vec<String>` URL-list approach with a `Vec<(TheRockIndexGeneration, String)>` candidate model and moved env-override base logic into new accessor functions.
  - `therock_rocm_extras` (this PR's multi-arch device-extras helper) has no counterpart in #329's design.
  - Several unit tests collide on the same names/fixtures for these functions.

  Whichever PR merges second will need a manual reconciliation into one unified candidate model, likely a generation enum with a third variant (Legacy / MultiArch / Next, or similar) rather than a straight rebase.

- **`tests/e2e-cucumber/tests/e2e/runtime_steps.rs`** (2 hunks): both PRs insert new step functions at the same anchor point. Trivial, keep both blocks.

`crates/rocm-core/src/lib.rs` merges cleanly despite touching related code, no action needed there.

juhovainio:
Addressed the review feedback from both of you:

**@rominf**: rebased onto current \`main\` (was DIRTY/conflicting after #322 landed). The conflict in \`tests/e2e-cucumber/tests/e2e/runtime_steps.rs\` was purely adjacent insertions (this PR's new step function next to #322's new step functions), spliced together with no semantic change to either side. Also tagged the new dry-run scenario \`@nightly\`, for the same reason as the existing \`runtime-install-records-the-real-folder\` scenario: it resolves the real channel index over the network even under \`--dry-run\`, and that extra network work on the no-GPU mock lane's 64-way concurrency was pushing \`eai-7960-gen-tps-held-after-scrape-failure\` / \`eai-7960-gen-tps-expiry-boundary\` past their validity window.

**@michaelroy-amd**: updated the \`package_policy\` dry-run prose to say "pinned rocm (with any resolved device extras)" instead of implying the extras are always the same shape, so it reads correctly for both the classic per-family index and the new flat multi-arch index.

Also re-verified live against \`repo.amd.com\`: \`--dry-run\` now resolves \`index_url: https://repo.amd.com/rocm/whl-multi-arch\` with the correct \`device-gfx1100,...\` extras and no fallback to the broken per-family shape.

All CI checks are green on \`68934fa\` (full run: unit tests, clippy, prek, E2E, Windows build). Ready for another look.

tests/e2e-cucumber/tests/e2e/runtime_steps.rs
```diff
@@ -365,6 +374,15 @@ async fn assert_update_reports_freshness(world: &mut E2eWorld) {
     }
 }
 
+#[then("the resolved package index is not the broken per-family multi-arch path")]
+async fn assert_index_not_broken_multi_arch(world: &mut E2eWorld) {
+    let stdout = world.cli_output.as_deref().expect("no dry-run output");
+    assert!(
+        !stdout.contains("whl-multi-arch/gfx110X-all"),
```
rominf, line 381: This assertion is satisfied on unpatched `main`, so the scenario doesn't guard the fix.

On `main`, `therock_index_urls(Release, family)` returns `[whl/{family}, whl-multi-arch/{family}]` — classic first. `resolve_pip_runtime_with_timeout` loops the candidates and **returns on the first success**. The classic index does resolve (that's the premise of #271: it works, it's just pinned at 7.13.0), so the loop returns immediately and the dry-run prints:

```
index_url: https://repo.amd.com/rocm/whl/gfx110X-all
```

The broken `whl-multi-arch/gfx110X-all` URL only ever reaches stdout through the `bail!` at the bottom of that loop, which is unreachable when a candidate succeeds. So `!stdout.contains("whl-multi-arch/gfx110X-all")` holds with the bug present, and holds after the fix, and would keep holding if someone reverted `therock_index_urls` tomorrow. AGENTS.md §3: "test fails before fix and passes after fix."

Asserting the positive fixes it, and it's a one-line change:

```rust
assert!(
    stdout.contains("index_url: https://repo.amd.com/rocm/whl-multi-arch\n"),
    "release install did not resolve the flat multi-arch index:\n{stdout}"
);
```

That fails on `main` (which prints the classic URL) and passes here. Worth also asserting `package_specs:` carries a `device-` extra, since that's the other half of the fix and currently has only unit coverage.

One edge worth naming: if the classic index ever stops resolving entirely, `main` would `bail!` and `run_rocm_ok`'s `rc == 0` check would panic first — the scenario would fail, but on the command failing, not on the assertion. The positive form is correct in both worlds.

apps/rocm/src/therock.rs
```diff
@@ -1012,15 +1019,42 @@ fn install_wheel_runtime(
...
+    }
+    match known_therock_family_device_chips(family) {
+        Some(chips) => {
+            for chip in chips {
+                let _ = write!(extras, ",device-{chip}");
+            }
+        }
+        None => extras.push_str(",device-all"),
+    }
+    extras
```
rominf, line 1055: This fallback fires on every Instinct/datacenter part, and your own CI shows what it costs.

`known_therock_family_device_chips` returns `None` for six of the sixteen known families — `gfx90X-dgpu`, `gfx90X-dcgpu`, **`gfx94X-dcgpu`**, `gfx950-dcgpu`, `gfx101X-dgpu`, `gfx103X-dgpu`. The two `gfx94X`/`gfx950` entries are MI300/MI350; that's the datacenter audience taking the `device-all` path by default.

From job 99541673879 (MI300X, `detected_gfx_target: gfx943` → `gfx94X-dcgpu`):

```
Installing rocm[libraries,devel,device-all]==7.14.0 ... from https://repo.amd.com/rocm/whl-multi-arch
Downloading rocm-sdk-device-gfx1030 (458.5MiB)
Downloading rocm-sdk-device-gfx90a  (450.7MiB)
Downloading rocm-sdk-device-gfx950  (650.9MiB)
... 24 device wheels, 4451 MiB total
```

and then the post-install probe:

```
rocm_sdk_target_family: gfx1010
```

So on a gfx943 host we download ~4.3 GiB to use one wheel, and the probe reports the wrong target family — `gfx1010` is just the first of the 24. That value is printed to the user and stored in the manifest. The exact-chip lanes in the same run resolve correctly (`gfx1151`→gfx1151, `gfx120X-all`→gfx1201), which localises it to `device-all`.

There's no size guard either: `preflight_tarball_space` is called only from `install_tarball_runtime` (`:1119`), so the wheel path has no disk preflight at all. A constrained host gets a multi-GiB surprise with no warning.

**Four of the six are enumerable right now,** from the wheel names the same CI log lists:

```rust
"gfx94X-dcgpu"  => Some(&["gfx942"]),
"gfx950-dcgpu"  => Some(&["gfx950"]),
"gfx101X-dgpu"  => Some(&["gfx1010", "gfx1011", "gfx1012"]),
"gfx103X-dgpu"  => Some(&["gfx1030", "gfx1031", "gfx1032", "gfx1033",
                          "gfx1034", "gfx1035", "gfx1036"]),
```

I checked these against `normalize_therock_family` — `gfx942`→`gfx94X-dcgpu` (the `starts_with("gfx94")` arm), `gfx950`→`gfx950-dcgpu`, `gfx101*`→`gfx101X-dgpu`, `gfx103*`→`gfx103X-dgpu` — so `known_therock_family_device_chips_round_trip_to_their_family` stays green with all four added. The remaining two (`gfx90X-dgpu`/`gfx90X-dcgpu`) are genuine catch-alls that overlap the exact `gfx900`/`gfx906`/`gfx908`/`gfx90a` families, and `device-all` is a defensible answer there.

The better answer, if you're reconciling with #329 anyway, is #329's shape: derive the extra from the *detected raw arch* (`device-{raw_arch}`) and keep the family table only as the fallback when no raw arch is known. Either way, when `device-all` is chosen, say so in the dry-run output — right now the user sees it in `package_specs` with no indication it means "every GPU ROCm supports."

apps/rocm/src/therock.rs
```diff
@@ -1012,15 +1019,42 @@ fn install_wheel_runtime(
...
 }
 
-fn therock_pip_package_specs(package_versions: &TheRockPipPackageVersions) -> Vec<String> {
+fn therock_pip_package_specs(
+    package_versions: &TheRockPipPackageVersions,
+    rocm_extras: &str,
+) -> Vec<String> {
     vec![
-        format!("rocm[libraries,devel]=={}", package_versions.rocm),
+        format!("rocm[{rocm_extras}]=={}", package_versions.rocm),
```
rominf, line 1027: This shape change breaks a documented acceptance test that CI can't see.

`scripts/therock_sdk_install_test.py` — unchanged by this PR — asserts:

```python
THEROCK_SDK_PACKAGE_SPEC = "rocm[libraries,devel]"   # :29
...
assert_contains(install_output, f"{THEROCK_SDK_PACKAGE_SPEC}==", "sdk install")   # :499
```

After this change the release-channel output is `rocm[libraries,devel,device-gfx1151]==7.14.0`, and `"rocm[libraries,devel]=="` is not a substring of that. The script's default is `--channel release` (`:412`), so every documented invocation in `docs/testing.md:180/186/193/200` and `docs/manual-testing.md:140` fails on the first release run. It's manual-only — not in `ci.yml` — so nothing catches it, and AGENTS.md §8 names `docs/testing.md` checks as part of the gate for touched behavior.

The fix is small: make the constant a prefix, `"rocm[libraries,devel"`, and assert on `assert_contains(install_output, THEROCK_SDK_PACKAGE_SPEC)`. The existing `assert_not_contains(install_output, "rocm[devel]")` negative still does its job.

Same class of staleness in prose at `docs/testing.md:167`, which lists the expected plan as "pinned `rocm[libraries,devel]`, `torch`, `torchvision`, and `torchaudio` versions" — worth a sentence noting the release channel now adds `device-*`.

apps/rocm/src/therock.rs
```diff
@@ -3368,9 +3402,12 @@ fn parse_version(value: &str) -> Option<ParsedVersion> {
 
 fn therock_index_urls(channel: TheRockChannel, family: &str) -> Vec<String> {
```
rominf, line 3403: On the #329 collision you flagged — one detail worth pinning down before you agree a merge order, because it's not just a textual conflict.

#329's replacement for this function still contains the bug:

```rust
// pr/329 apps/rocm/src/therock.rs:3632-3638
TheRockChannel::Release => vec![
    format!("{}/{family}", release_pip_index_base()),
    format!("{}/{family}", release_pip_multi_arch_index_base()),
],
```

Classic first, then the **per-family** multi-arch shape — byte-for-byte `main`'s behavior. #329's new `Next` candidate doesn't cover this: it points at a different host (`whl-next`) and returns `None` outright when `raw_arch` is unknown, so it never reaches `repo.amd.com/rocm/whl-multi-arch`. If #329's `therock.rs` wins the reconciliation, #271 regresses with no test failing — including the new scenario above, for the reason in that comment.

The other direction is the good news: #329's `therock_pip_package_specs` already derives `rocm[libraries,devel,device-{raw_arch}]` from the detected arch (and extends it to `torch[device-*]`/`torchvision[device-*]`, which this PR doesn't), which is the better answer to the `device-all` problem. So the reconciled version wants **this PR's flat candidate** plus **#329's arch-driven extras** — a three-variant generation enum (`Legacy` / `MultiArch` / `Next`) rather than either side winning wholesale.

Whoever merges second should re-verify against #271's actual repro, not just resolve conflicts and trust the suite.

tests/e2e-cucumber/features/runtime_setup.feature
```diff
@@ -84,3 +84,23 @@ Feature: Runtime configuration
...
+  # `@nightly` for the same reason as scenario 8 above: this dry-run still
+  # resolves the real channel index over the network (dry-run only skips the
+  # venv/download, not index resolution), and the no-GPU mock lane's 64-way
+  # concurrency from that extra network work is what pushes
+  # `eai-7960-gen-tps-held-after-scrape-failure` and
+  # `eai-7960-gen-tps-expiry-boundary` past their validity window. Runs on the
+  # nightly lanes instead, where scenarios are serialized.
+  @id:runtime-install-sdk-release-index-shape @nightly
+  Scenario: 4 - Resolving the SDK from the release channel never uses the broken multi-arch URL shape
+    When the user dry-runs installing the SDK for a known family
```
rominf, line 105: Scenario number 4 is already taken — `:30`, "Reinstalling the SDK leaves the installed engine's requirements satisfied."

Nothing fails: the harness's uniqueness assert (`tests/e2e-cucumber/tests/e2e.rs:1179`) covers `@id`, not the numeric prefix. But the numbers are how the file and the HTML report are navigated, and two "Scenario: 4"s in one feature make them indistinguishable there. Next free in this file is 9.

rominf (changes requested):
The core fix is right and I could confirm it end to end. Flattening the multi-arch candidate and putting it first is what #271 needed, and this PR's own GPU lane shows it working: the pre-warm resolves `7.14.0` from `https://repo.amd.com/rocm/whl-multi-arch` where the classic index had it pinned at `7.13.0`. @michaelroy-amd's `package_policy` wording note and @rominf's `@nightly` tagging both landed.

But the same CI log shows the `device-all` fallback going wrong on exactly the hardware that matters most, and the new regression scenario doesn't fail on unpatched `main`.

**`device-all` on Instinct.** Six of the sixteen families return `None` from `known_therock_family_device_chips`, including `gfx94X-dcgpu` and `gfx950-dcgpu`. Job 99541673879 (MI300X) shows the result: `Installing rocm[libraries,devel,device-all]==7.14.0`, then **24 `rocm-sdk-device-*` wheels totalling 4451 MiB** downloaded onto a machine that needs one of them, and the post-install probe reporting `rocm_sdk_target_family: gfx1010` on a gfx943 host. Four of those six buckets are cleanly enumerable from the wheels the index already publishes — details inline, including that the fix keeps the existing round-trip invariant test green.

**The new scenario passes with the bug present.** `!stdout.contains("whl-multi-arch/gfx110X-all")` is satisfied on `main`, because `main` tries the classic index first, it succeeds (stale 7.13.0), the loop returns, and the broken URL is never printed — failed candidates only surface through `bail!`, which is unreachable when the first one works. AGENTS.md §3 wants the opposite.

**Two consumers the change breaks silently.** `scripts/therock_sdk_install_test.py:499` asserts on the literal `rocm[libraries,devel]==`, which is no longer a substring of `rocm[libraries,devel,device-gfx1151]==`; it's manual-only so CI can't catch it, and `docs/testing.md:167` states the old shape as the expected plan. Both are unchanged files that this PR invalidates.

**On "All CI checks are green on `68934fa`".** It isn't — `E2E tests (GPU)` is FAILURE with 4 unexpected failures and `E2E tests (Strix Halo, Ubuntu)` was CANCELLED. I dug into whether they're yours, and the answer is partly. The same four scenario names also fail on #251's lane, so the lane was red before you. But the *mechanism* here is new: your version bump installs a fresh `7.14.0` runtime, and this branch's base (`283811fc`) predates #314's unconditional `ensure_default_engine` in the pre-warm, so the engine venv stays behind on `7.13.0` and every serve dies with `vLLM is not installed in a Linux/WSL ROCm Python environment` — a message that appears zero times on either control lane. Rebasing onto current `main` picks up #314 and should clear that half. Worth correcting the comment either way (AGENTS.md §4).

The #329 collision you flagged is real and worse than the file-level conflict: #329's `legacy_pip_index_candidates` still builds `format!("{}/{family}", release_pip_multi_arch_index_base())`, i.e. this exact bug. Note inline.

Checked and clean: leak scan on the diff and PR text; signature/sign-off gate; `is_multi_arch_pip_index` trailing-slash handling; the nightly path is untouched; `known_therock_family_device_chips_round_trip_to_their_family` is a genuine invariant, not a tautology; the unit tests for the two new helpers are well-targeted.

(One process note: my read here was thorough on the diff but I don't have an Instinct host, so the `device-all` behavior is from your CI log, not from running it.)

michaelroy-amd:
Resolver ownership update: #308 is the canonical owner for current release/nightly aggregate source selection, shared provenance, and runtime-composition behavior. #329 is the ROCm 10 Next-layout extension and will be stacked on #308. This PR's #271 regression behavior will be transferred into #308 with scenario evidence; after that transfer is verified, #272 will be closed as superseded rather than merged. #314/current `main` remains authoritative for the torch owning-runtime settlement; the obsolete alignment work in the older #308 history will not be retained.

For new canonical and Next installs, unknown or incomplete source layouts must fail closed—no silent legacy or alternate-channel fallback. Existing managed runtimes with legacy manifests remain readable and usable for backward compatibility.

Related: #308, #329, #314. Tracking: EAI-8268, EAI-8431, EAI-7956.

michaelroy-amd:
Closing as superseded by merged #308. The complete #271 behavior is now on `main` in merge commit `60abf955`: the release channel resolves the flat canonical aggregate index, selects ROCm 7.14.1 rather than the stale 7.13 per-family stream, and requests exact `device-<gfx>` extras for `rocm`, `torch`, and `torchvision`. The positive `@id:runtime-resolve-canonical-release` scenario asserts the selected source, version/provenance, and exact device payload; it fails against pre-#308 main and passes on the merged implementation. The symlinked install-root behavior is also preserved. No unique accepted #272 behavior remains outside #308, so merging this conflicting duplicate would reintroduce a second resolver model. Related: #308, #329, #314; tracking: EAI-7956, EAI-8268, EAI-8431.

PR #275, opened 2026-08-18
ci(e2e): refresh the shared pre-warmed runtime when a newer one exists (EAI-8057)

tomastola:
## Summary

Nearly every GPU E2E scenario serves against one shared, pre-warmed managed
runtime, so a multi-GiB `rocm install sdk` happens once per runner instead of
once per scenario. That tree lives on the runner's persistent workspace and
survives `git clean`.

**Cause.** Every self-hosted lane guarded the install on directory existence
alone:

```bash
if [ ! -d "$E2E_SHARED_RUNTIMES_DIR/registry" ]; then … fi
```

A persistent directory plus an existence-only guard means that branch runs
exactly once in the lifetime of a runner. After the first run ever, every
subsequent run served against whatever runtime happened to be installed that
day — 16 days old on both MI300X runners when measured. Two consequences: the
lanes validate an increasingly historical SDK, and drift between the shared tree
and what a fresh `install sdk` produces is never exercised at all, so a break in
the install path can only be discovered by a new runner.

**Fix.** Keep the cache, but invalidate it when the channel index has actually
published something newer. `cargo xtask e2e-prewarm` replaces the guard in all
eight self-hosted lanes and resolves to one of four outcomes:

| Index vs. installed | Action |
|---|---|
| nothing installed for the channel | cold `install sdk` |
| index is ahead | install side-by-side, activate, prune |
| `up_to_date` / `ahead_of_index` | reuse |
| freshness unknown (index unreachable) | reuse, warn |

## Why it is shaped this way

**Reuse the CLI's primitives instead of reimplementing version resolution.**
`rocm update` already reports per-runtime freshness against the channel index,
`rocm update --apply --runtime <key> --activate` already installs a newer runtime
side-by-side, and `rocm storage remove-old-installs --keep N` already bounds the
resulting multi-version cache. Comparing versions in workflow shell would have
duplicated logic that exists — and is tested — in the product. It also means the
GPU lanes now exercise `update` and `storage` on real hardware, which nothing
did before.

**Side-by-side, never move the tree.** `install sdk` bakes absolute paths into
the runtime manifest, so a runtime has to be created in its final location. This
is why the update path uses `--apply --activate` rather than installing
elsewhere and swapping.

**In `xtask`, not in the workflows.** The pre-warm block is duplicated across
eight jobs in two shells (bash on the Linux lanes, PowerShell on Strix Windows).
Inline logic would exist eight times in two languages and be untestable; the
Strix Windows lanes lose their PowerShell copy rather than gaining a second
implementation to keep in sync. Same reasoning as the existing `xtask e2e`.

**Anything unclear reuses.** Only a report that positively reads as "a newer
version exists for our channel" triggers work. An unreachable package index
surfaces per-runtime as `status=error`, which reuses and warns — a briefly
offline index must not turn a GPU lane red, nor start a multi-GiB download on
every run. Install and update failures do propagate; a failed prune only warns.

**A scenario pins the output shape — partially, and the gap is worth naming.**
The decision is parsed out of `rocm update` text, so a fixture can drift from the
renderer it imitates. The new `runtime-update-reports-freshness` scenario runs the
real command on hardware and asserts the shape the parser depends on: a
`runtime <key> …` line carrying `channel=` and a recognised `status=`. That covers
structural drift — a renamed prefix or a dropped `channel=` filters every line out
and makes the lane reinstall on every run, which is loud; a renamed `status=`
field fails the scenario outright.

What it does not cover is a rename of the *value* `update_available`. A runner
that is currently up to date reports `up_to_date`, the assertion still passes, and
the decision quietly falls through to "reuse" — the exact defect this PR fixes.
Catching that reliably would need a runner that happens to be stale. The real
cause is that this is an unversioned text contract with more than one parser:
`rocmd` reads the same field by hand in
`update_output_reports_update_available`, also against hand-written fixtures, so
it carries the same exposure today. A machine-readable form of `update`, or a
shared constant for the status values, would close both. That is product-code
work and deliberately out of scope here.

## Test plan

Verified locally:

- `cargo test --workspace --all-targets` — 2215 passed, 0 failed. Includes 12 new
  unit tests covering every freshness verdict and the degraded paths.
- `cargo clippy --workspace --all-targets -- -D warnings` and
  `cargo clippy --locked -p e2e-cucumber --test e2e -- -D warnings` clean;
  `cargo fmt --all --check` clean.
- The parser's fixtures are **captured verbatim from the built binary**, not
  hand-written. That caught a real near-collision: the `update_surfaces` summary
  line is one character away from parsing as a runtime entry, and there is now a
  test pinning that it does not.
- Full orchestration driven against a stub `rocm` on the PATH, so the sequencing
  is verified without a multi-GiB download: `update` → `update --apply --runtime
  <key> --activate` → `storage remove-old-installs --keep 2 --yes`, with
  `ROCM_CLI_{CONFIG,DATA,CACHE}_DIR` correctly scoped to the pre-warm root.

### What the GPU lanes on this PR actually did

Every self-hosted lane reached the new pre-warm and reported the same verdict:

```
pre-warm: reusing the shared release runtime (runtime is up to date with the channel index)
```

Two things follow, one of which contradicts what I expected:

- **The `rocm update` output contract holds on hardware.** The new
  `runtime-update-reports-freshness` scenario passed all three steps on Strix
  Halo/Windows against a real index — the parser's assumption about the report
  shape is confirmed, not just imitated by fixtures.
- **The update path did not fire, and that is the correct verdict.** I predicted
  the 16-day-old MI300X trees would take the update path. They did not, because
  the Release channel currently resolves nothing newer than what those runners
  already hold: `therock_index_urls()` tries the classic per-family
  `whl/{family}` index first, which still succeeds but tops out at ROCm 7.13.0
  (#271, fix open in #272). `up_to_date` is an honest answer to a capped index,
  so reuse is right — but it means **the install, update, activate and prune
  paths are still unexercised on hardware**, and remain covered only by unit
  tests and the stubbed-binary run above.

Worth being plain about the consequence: on the Release channel as it resolves
today, this change is a no-op every run. Its value is that freshness is now
consulted at all — the previous guard would have kept a tree frozen for the life
of a runner no matter what the index published, whereas this self-corrects on the
first run after the index moves. That is also why it has to land before the
nightly-channel follow-up: `therock-nightly` does move daily, and pointing a lane
at a different channel under an existence-only guard would have been a silent
no-op.

### CI status

All required checks are green. Two self-hosted lanes are not, and neither is
caused by this change:

**`E2E tests (Strix Halo, Windows)` — red, pre-existing.** 8 of 41 scenarios
fail here against 8 of 40 on `main`, and the two failure sets are identical; the
+1 is the new scenario, which passes. That lane has failed on every `main` run
since 2026-08-12 (the visible cause is a `llama-server.exe` download failing
after 6 attempts); tracked in #260 and #247.

**`E2E tests (Strix Halo, Ubuntu)` — cancelled at its 35-minute cap.** The two
runners behind that label are not equivalent: `strix-halo-ubuntu` finishes in
13.6–15.9 min, while `strix-halo-ubuntu-2` takes 31.0–31.5 min when it passes and
has hit the cap three times today — the other two on unrelated branches,
including `main`'s own merge queue. This run drew runner-2. The pre-warm was not
the cost: it reused, downloaded nothing, and every serve scenario on that box ran
1.5–20× slower than on runner-1, including a refusal-path scenario that never
serves at all (371s vs under 20s).

One honest caveat on that second point: this PR does add a scenario to a lane
that, on runner-2, already sits within ~3.5 min of its cap, and the new
scenario's shared `a managed runtime is active` precondition is expensive on
that box specifically (0.6s on MI300X; still running at 140s on runner-2 when
the job was killed). So it is a small marginal push on an already-overcommitted
budget. The runner-2 disparity looks worth its own issue rather than a timeout
bump smuggled into this PR — happy to raise one, or to widen the cap here if a
maintainer would rather see that.

## Risk

Low-to-moderate, and concentrated in one place: the lanes can now download.
Previously the pre-warm was a no-op after a runner's first run; now a genuinely
newer index costs one install before the suite. That is the intended behaviour,
but it does mean a lane can be slower than it was, and the conservative bias
exists so this happens only when the index has actually moved. In practice, on
the Release channel today, it has not — see above.

The failure modes are bounded the other way. Every ambiguous outcome reuses, so
the common bad day (index unreachable) behaves exactly like today. The lanes are
`continue-on-error: true`, so even a hard failure in the pre-warm cannot gate a
merge. No product code changes — this is CI orchestration, one new xtask
subcommand, and one new scenario.

Follow-up, tracked separately: with the tree no longer frozen, a scheduled lane
can track the `therock-nightly` channel, which is currently never installed by
CI (EAI-8056). That needs this first, for the reason given above.

- [x] Searched `tests/e2e-cucumber/expectations.toml` for `EAI-8057`; no xfail
  rows reference it. The defect was in CI orchestration, not in any scenario's
  expected outcome, so there is nothing to narrow.

### DIFF
--- a/xtask/src/e2e_prewarm.rs
+++ b/xtask/src/e2e_prewarm.rs
@@ -35,9 +35,9 @@
 //!   cache with a per-channel/format/family retention policy.
 //!
 //! Living in xtask rather than the workflows is deliberate: the pre-warm block is
-//! duplicated across eight jobs in two shells (bash on the Linux lanes, PowerShell
-//! on Strix Windows), so the decision logic would otherwise exist eight times in two
-//! languages and be untestable. Same reasoning as `e2e.rs` — the recipe belongs in
+//! duplicated across multiple jobs in two shells (bash on the Linux lanes,
+//! PowerShell on Strix Windows), so the decision logic would otherwise drift across
+//! copies and be untestable. Same reasoning as `e2e.rs` — the recipe belongs in
 //! one cross-platform place instead of a shell wrapper.
 
 use std::path::{Path, PathBuf};
@@ -81,18 +81,37 @@ pub fn decide(update_report: &str, channel: &str) -> Decision {
     let runtimes: Vec<RuntimeLine> = update_report
         .lines()
         .filter_map(RuntimeLine::parse)
-        .filter(|line| line.channel.as_deref() == Some(channel))
         .collect();
 
-    // Nothing installed for THIS channel. The tree may still hold another
-    // channel's runtime — that is the normal state once a per-channel pre-warm
-    // directory is shared, and the right response is still a cold install here.
-    if runtimes.is_empty() {
+    // A degraded `status=error` line omits `channel=` because resolution failed
+    // before the renderer had a plan. It proves a runtime exists but cannot be
+    // attributed safely, so the conservative choice is reuse, not a fresh
+    // multi-GiB install. A later healthy probe will identify the channel.
+    // In a mixed-channel tree the error may belong to another channel, but the
+    // report has discarded that identity. Reuse remains the safe floor until a
+    // healthy probe can distinguish "missing channel" from "unknown freshness".
+    if !runtimes
+        .iter()
+        .any(|line| line.channel.as_deref() == Some(channel))
+    {
+        if runtimes
+            .iter()
+            .any(|line| line.channel.is_none() && line.status.as_deref() == Some("error"))
+        {
+            return Decision::Reuse {
+                reason: "could not establish runtime freshness; leaving the shared tree untouched"
+                    .to_owned(),
+            };
+        }
+
+        // Nothing installed for THIS channel. The tree may still hold another
+        // channel's runtime, so install this one.
         return Decision::Install;
     }
 
     if let Some(stale) = runtimes
         .iter()
+        .filter(|line| line.channel.as_deref() == Some(channel))
         .find(|line| line.status.as_deref() == Some("update_available"))
     {
         return Decision::Update {
@@ -105,11 +124,13 @@ pub fn decide(update_report: &str, channel: &str) -> Decision {
     // "updating" backwards.
     let reason = if runtimes
         .iter()
+        .filter(|line| line.channel.as_deref() == Some(channel))
         .any(|line| line.status.as_deref() == Some("up_to_date"))
     {
         "runtime is up to date with the channel index"
     } else if runtimes
         .iter()
+        .filter(|line| line.channel.as_deref() == Some(channel))
         .any(|line| line.status.as_deref() == Some("ahead_of_index"))
     {
         "installed runtime is ahead of the channel index"
@@ -387,16 +408,25 @@ source: index\n"
 
     #[test]
     fn index_error_reuses_rather_than_reinstalling() {
-        // The offline/unreachable-index case. Reinstalling here would download
-        // multiple GiB on every run of an isolated runner, and failing would turn
-        // the lane red for a network blip.
+        // The renderer omits `channel=` when resolving the index fails. That is
+        // unknown freshness, not proof that this channel has no runtime, so the
+        // conservative pre-warm decision must reuse rather than download again.
         let text = "update\n  runtime release-wheel-gfx94x-dcgpu-7-13-0 format=wheel \
 status=error message=failed to reach https://repo.amd.com/rocm/whl after 3 tries\n";
-        // The degraded line carries no `channel=`, so it is not attributable to
-        // this channel and the report reads as "nothing installed for release".
-        // That is the one case where the fallback in `run` (registry on disk?)
-        // decides, not this function.
-        assert_eq!(decide(text, "release"), Decision::Install);
+        let Decision::Reuse { reason } = decide(text, "release") else {
+            panic!("an unattributable index error must reuse the existing tree");
+        };
+        assert!(reason.contains("could not establish"), "{reason}");
+    }
+
+    #[test]
+    fn unattributed_error_wins_over_a_known_other_channel() {
+        let text = format!(
+            "{}  runtime release-wheel-gfx94x-dcgpu-7-13-0 format=wheel \
+status=error message=failed to reach the index\n",
+            report("up_to_date", "nightly")
+        );
+        assert!(matches!(decide(&text, "release"), Decision::Reuse { .. }));
     }
 
     #[test]
--- a/xtask/src/e2e_prewarm.rs
+++ b/xtask/src/e2e_prewarm.rs
@@ -1,0 +1,486 @@
+// Copyright © Advanced Micro Devices, Inc., or its affiliates.
+//
+// SPDX-License-Identifier: MIT
+
+//! Prepare the shared managed runtime the GPU E2E lanes serve against.
+//!
+//! Nearly every GPU serve scenario points its `data/runtimes` at ONE shared,
+//! pre-warmed runtime tree (see `E2E_SHARED_RUNTIMES_DIR` and
+//! `E2eWorld::use_shared_runtimes`) so a multi-GiB `install sdk` happens once per
+//! runner instead of once per scenario. The lanes used to guard that install on
+//! directory existence alone:
+//!
+//! ```text
+//! if [ ! -d "$E2E_SHARED_RUNTIMES_DIR/registry" ]; then ... rocm install sdk ...; fi
+//! ```
+//!
+//! which never reinstalls. The tree persists on the runner's PVC, so after the
+//! first run ever the pre-warm was a permanent no-op and every lane kept serving
+//! against whatever runtime happened to be installed that day — 16 days stale on
+//! both MI300X runners when this was measured (EAI-8057). Drift between the
+//! shared tree and what a fresh install produces was therefore untested, and
+//! widened silently.
+//!
+//! This keeps the cache and invalidates it only when the channel index actually
+//! publishes something newer, reusing the primitives the CLI already ships
+//! rather than reimplementing version resolution in workflow shell:
+//!
+//! * `rocm update` reports, per installed runtime, `status=up_to_date |
+//!   update_available | ahead_of_index` by comparing against the channel index.
+//! * `rocm update --apply --runtime <key> --activate` installs the newer runtime
+//!   SIDE BY SIDE and makes it the default. Side-by-side matters: `install sdk`
+//!   bakes ABSOLUTE paths into the runtime manifest, so a runtime must be created
+//!   in its final location and never moved afterwards.
+//! * `rocm storage remove-old-installs --keep N` bounds the resulting multi-version
+//!   cache with a per-channel/format/family retention policy.
+//!
+//! Living in xtask rather than the workflows is deliberate: the pre-warm block is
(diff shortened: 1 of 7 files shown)


xtask/src/e2e_prewarm.rs
```diff
@@ -0,0 +1,486 @@
...
+    // Nothing installed for THIS channel. The tree may still hold another
+    // channel's runtime — that is the normal state once a per-channel pre-warm
+    // directory is shared, and the right response is still a cold install here.
+    if runtimes.is_empty() {
+        return Decision::Install;
+    }
+
+    if let Some(stale) = runtimes
+        .iter()
+        .find(|line| line.status.as_deref() == Some("update_available"))
```
r0x0r, line 115: This routes on the literal string value `update_available`. As you call out in the PR body, the scenario pins the *structure* (a `runtime <key> … channel=… status=…` line) but not this *value* — if the renderer renamed `update_available`, the assertion still passes, this `find` misses, and the pre-warm silently falls through to Reuse, i.e. the exact staleness defect the PR fixes. And `rocmd`'s `update_output_reports_update_available` hand-reads the same field against its own fixtures, so there are now (at least) two independent parsers of an unversioned text contract. Since both parsers already exist, would a single shared `const` for the status values — referenced by the renderer, this function, and rocmd — be cheap enough to land now and close both exposures at once? Understood if you'd rather keep it out of scope; capturing it here so it isn't lost.

r0x0r (commented):
Reviewed — well-engineered and thorough. The existence-only guard genuinely froze the shared runtime after the first install ever, and moving the decision into `xtask e2e-prewarm` (once, cross-platform) instead of eight shell copies in two languages is the right call. Things I verified:

- `keep` defaults to `2` (`default_value_t = 2`), so the eight lanes calling `e2e-prewarm --channel release --prewarm-dir …` without `--keep` prune to 2, not 0 — no risk of pruning the active runtime.
- `RuntimeLine::parse` correctly rejects the `update_surfaces` `runtimes: status=none_configured` line (`strip_prefix("runtime ")` needs the trailing space), so an empty tree still resolves to `Install` rather than a spurious Reuse — and there's a dedicated test pinning exactly that near-collision.
- `decide` ordering is right: `update_available` wins over `up_to_date`/`ahead_of_index`, per-channel filtering keeps a release runtime from satisfying a nightly lane, and every degraded/unknown path is conservative (Reuse, never a red lane or a multi-GiB reinstall on a network blip). The `probe`-failed fallback in `run` preserves the old install-only-if-registry-absent floor.
- Env scoping (`ROCM_CLI_{CONFIG,DATA,CACHE}_DIR` → `prewarm_dir`) matches what the lanes exported by hand, and `resolve_rocm_binary` absolutizes a relative `ROCM_CLI_BINARY` like `e2e.rs`.
- 12 unit tests over verbatim-captured fixtures plus the on-hardware `runtime-update-reports-freshness` scenario.

One substantive thread on the parsing contract below — the gap you already named in the description; filing it as a resolvable thread for tracking rather than as a change request.

xtask/src/e2e_prewarm.rs
tomastola: Agreed the literal is a contract, not a local detail. There are actually four sites spelling it out: two render paths in `therock.rs`, this one, and a pre-existing parser in `rocmd`. A rename on the producing side would compile fine and quietly turn every CI pre-warm back into a no-op reuse — the exact failure this PR removes.

Holding it out of this PR for two reasons: the fix lands in `therock.rs`, which #308 is currently rewriting, and this PR is CI-only. Tracked as EAI-8299 to do right after #308 merges — `xtask`, `apps/rocm` and `apps/rocmd` all already depend on `rocm-core`, so it's a move rather than a restructure.

In the meantime `UPDATE_STATUSES` in `runtime_setup.feature`'s step definitions pins the four values from the consuming side, so a rename fails the `runtime-update-reports-freshness` scenario instead of degrading silently. That's a weaker guard than a shared const — it needs a GPU runner to fire — but it's not nothing. Leaving this open in case you'd rather see it done here.

tomastola:
Force-pushed since your review — rebased onto current `main`, so this is at `81c173c` where you reviewed `6311cc7`. Two changes to the commit itself, flagged so you don't have to re-diff:

**1. Two more lanes converted (8 → 10).** #278 merged after your review and added `e2e-gpu-rad3` and `e2e-gpu-nightly-rad3`, both carrying the old existence-only guard — that PR was mine, so that was my own regression to clean up. Both now call `cargo xtask e2e-prewarm` like the rest. No existence-only guard remains in either workflow.

**2. `Scenario: 4` renumbered to `5`** in `runtime_setup.feature`. #264 landed its own `Scenario: 4` at the same position, so the rebase conflicted there; both scenarios are kept. The `@id:` slug (`runtime-update-reports-freshness`) is unchanged, so the expectation matrix and report grid are unaffected — the numbers are cosmetic.

`xtask/src/e2e_prewarm.rs` is untouched since your review, so every point in your verification list still holds as written.

On the parsing-contract thread — replied inline. Declining it here because the fix lands in `therock.rs`, which #308 is currently rewriting, and this PR is CI-only; tracked as EAI-8299 to do straight after #308 merges. Left the thread open for you rather than resolving it myself.

michaelroy-amd:
Blocking finding before approval:

`rocm update` reports a per-runtime index failure with exit code 0 as `runtime <key> format=<format> status=error ...`, without `channel=` (`apps/rocm/src/therock.rs`). `e2e_prewarm::decide` filters parsed lines by channel, sees no matching runtime, and returns `Decision::Install`. Because the probe itself succeeded, `run` never reaches its registry-existence fallback and executes a fresh multi-GiB `rocm install sdk`—the opposite of the PR's stated invariant that unknown freshness reuses an existing tree.

The regression test named `index_error_reuses_rather_than_reinstalling` currently codifies the bug by asserting `Decision::Install`; its comment says `run` applies the fallback, but that fallback only handles a non-zero/failed probe.

I am preparing a narrow follow-up commit that makes an unattributed `status=error` resolve to conservative reuse, adds regression coverage for the single- and mixed-channel forms, and removes the stale hard-coded workflow count from the docs.

michaelroy-amd:
Fix pushed in `648493ae`.

- An unattributed `status=error` now resolves to conservative reuse instead of `Decision::Install`.
- Added regression coverage for both a lone degraded error line and a mixed-channel tree where the error cannot be attributed safely.
- Removed the stale hard-coded workflow count from the module and CI documentation.

Verification before push:
- `cargo test -p xtask e2e_prewarm`: 13 passed
- `cargo test --workspace --all-targets`: 2243 passed, 8 ignored
- `cargo clippy --workspace --all-targets -- -D warnings`: passed
- `python3 scripts/smoke_local.py`: passed
- pre-push hooks: passed

One unrelated process-state test flaked during the first push attempt (`comfyui::tests::status_reports_stopped_when_saved_comfyui_pid_is_gone`); its immediate isolated rerun passed, and the complete pre-push test gate passed on retry.

michaelroy-amd (approved):
Re-reviewed at `648493ae`. The blocking degraded-index path is fixed: an unattributed `status=error` now conservatively reuses the existing runtime, with single- and mixed-channel regression coverage. Focused tests, full workspace tests, clippy, smoke, signing checks, and required CI checks pass. Approving.

PR #277, opened 2026-08-19
ci: regenerate the generated manifests in git hooks

rominf:
## Summary

`MANIFEST.md`'s dependency table and `THIRD_PARTY_NOTICES.txt` are generated from the dependency graph and gated byte-for-byte in CI, but nothing regenerated them locally. A dependency bump could only be discovered stale a full CI cycle later.

That gap is the single largest source of red CI right now. Across the last 15 failed `ci.yml` runs, the failing steps in the two most-failing jobs were:

| Failing step | Count |
| --- | --- |
| `THIRD_PARTY_NOTICES.txt is current` | 10 |
| `MANIFEST.md dependency table is current` | 9 |
| real `clippy` failures | 2 |

Dependabot already gets this regeneration automatically via `dependabot-manifests.yml`; contributors did not.

- Add two prek hooks running the **write** form of each generator, so a stale file is refreshed and re-staged at commit time instead of failing CI a cycle later.
- They fire only when the dependency graph or a generator changes, so ordinary source edits are unaffected.
- Pin the cargo-about version in one place, and treat a missing **or mismatched** generator as unusable.
- Add `cargo xtask tpn --if-available`, which skips cleanly instead of failing, for use by the hook.

Why the version pinning is part of this change rather than a follow-up: CI's check is byte-for-byte, and other cargo-about versions format the notices differently. A hook that regenerated with whatever version happened to be installed would rewrite the file into a form that fails the very check it exists to prevent — strictly worse than the status quo. `--if-available` therefore treats a wrong version exactly like a missing one.

`--if-available` conflicts with `--check` at the clap level, so CI's staleness gate cannot be silently disabled, and a contract test asserts the pinned constant matches what `ci.yml` and `dependabot-manifests.yml` install, so the three cannot drift.

Risk: low. No product code changes; the only behaviour change outside the hooks is that `cargo xtask tpn` now rejects a non-pinned cargo-about instead of silently generating notices that fail CI.

Non-goals: the local clippy hook still diverges from CI (`--exclude e2e-cucumber`, no `--locked`), which accounts for the 2 real lint failures above. Closing that gap makes pre-push materially slower, so it is worth deciding separately rather than bundling here.

## Test plan

Both failures were reproduced and confirmed fixed, not assumed:

- Removed a generated dependency row from `MANIFEST.md` → `cargo xtask manifest --check` fails (reproducing the CI step) → hook regenerates → check passes → file byte-identical to the committed original.
- Appended a stray line to `THIRD_PARTY_NOTICES.txt` → `cargo xtask tpn --check` fails → hook regenerates → check passes → byte-identical.
- Generator-absent and wrong-version paths exercised with cargo shims: both skip cleanly under `--if-available` and leave the file untouched; both still hard-fail without it.
- `--check --if-available` is rejected by clap.
- Filters: editing a `.rs` file reports `Skipped` for both hooks; editing `Cargo.toml` / `Cargo.lock` fires them.
- `cargo test -p xtask` (94 pass, 3 new), `cargo fmt --all --check`, `cargo clippy --locked -p xtask --all-targets -- -D warnings`, and `prek run --all-files --no-group local-tools` all clean.
- Both hooks ran during this PR's own commit.

Clippy was scoped to `xtask` locally since the Rust changes are confined to that crate; CI runs the full workspace. Not exercised on Windows.

- [x] If this PR fixes a bug, searched `tests/e2e-cucumber/expectations.toml` for the fixed ticket ID and removed/narrowed any now-stale xfail rows. — n/a, no product bug fixed

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -134,3 +134,40 @@ repos:
         language: system
         files: \.ps1$
         groups: [local-tools] # runs in the dedicated `powershell-lint` CI job
+
+  # Generated manifests: MANIFEST.md's dependency table and THIRD_PARTY_NOTICES.txt
+  # are derived from the dependency graph and gated byte-for-byte in CI (`cargo
+  # xtask manifest --check` inside the `clippy` job; the `third-party-notices`
+  # job). Nothing regenerated them locally, so a dependency bump could only be
+  # discovered stale a full CI cycle later — between them these two files caused
+  # the large majority of red CI runs. Dependabot already gets this regeneration
+  # via dependabot-manifests.yml; these hooks give contributors the same.
+  #
+  # Both run the WRITE form, not `--check`: prek fails the commit when a hook
+  # modifies a file, so the refreshed file only needs re-staging. They only fire
+  # when the dependency graph (or the generator) actually changes, so ordinary
+  # source edits are unaffected. pre-commit, not pre-push, so the regenerated
+  # file lands in the same commit as the change that invalidated it.
+  - repo: local
+    hooks:
+      - id: cargo-manifest
+        name: MANIFEST.md dependency table is current
+        entry: cargo xtask manifest
+        language: system
+        files: '^((.*/)?Cargo\.(toml|lock)|xtask/.*|MANIFEST\.md)$'
+        pass_filenames: false
+        stages: [pre-commit]
+        groups: [local-tools] # compiles xtask; CI covers it in the `clippy` job
+      - id: cargo-tpn
+        name: THIRD_PARTY_NOTICES.txt is current
+        # `--if-available` skips cleanly when cargo-about is absent or not the
+        # pinned version, so contributors are never blocked from committing by a
+        # missing optional tool — and a mismatched version cannot rewrite the
+        # notices into a format that fails CI. The `third-party-notices` job
+        # remains the hard gate; clap forbids `--if-available` with `--check`.
+        entry: cargo xtask tpn --if-available
+        language: system
+        files: '^((.*/)?Cargo\.(toml|lock)|xtask/.*|about\.(toml|hbs)|THIRD_PARTY_NOTICES\.txt)$'
+        pass_filenames: false
+        stages: [pre-commit]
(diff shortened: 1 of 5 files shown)


fredespi:
I reviewed this PR at head `1807a6c` (diff against `main`) and did not find anything blocking. This reflects a code-only read of the diff — not CI results or runtime behavior — so it is not itself an approval; it looks safe to approve after final validation (CI + a maintainer pass).

What I checked, and why it holds up:

- **Version pin contract test.** `pinned_version_matches_workflows` asserts `ci.yml` and `dependabot-manifests.yml` both install `cargo-about@0.9.1`. Confirmed both files contain that exact string, so the test passes and the pin can't silently drift from the constant in `tpn.rs`.
- **`tpn::run` control flow.** The `(Generator, if_available)` match is exhaustive and fails safe: a missing or wrong-version generator is a clean skip only under `--if-available`, and a hard error otherwise. `parse_about_version` returns `None` on any unexpected shape, so a garbled `--version` reads as unusable rather than being mistaken for a valid version.
- **`--if-available` / `--check` conflict.** Enforced at the clap level (`conflicts_with = "check"`), so CI's byte-for-byte staleness gate can't be disabled by the new flag.
- **No commit-blocking asymmetry.** The `cargo-manifest` hook has no `--if-available` guard, but `manifest` only shells out to `cargo metadata` (always present in a Rust dev environment) — it never needs an installable external tool the way `tpn` needs `cargo-about`, so it can't block contributors who lack an optional tool.
- **Hook trigger filters vs CI.** The `files:` regexes for both hooks are supersets of CI's `changes` path filters (`Cargo.lock`, `**/Cargo.toml`, `about.toml`, `about.hbs`, `THIRD_PARTY_NOTICES.txt`, `xtask/**`, plus `MANIFEST.md`). They fire at least whenever CI would run the corresponding check — the safe direction — so a stale file can't slip past the hook only to be caught a CI cycle later.

Scope: config + `xtask` only, no product code. The change is well-tested (three new pure unit tests plus the contract test) and the design choices in the PR description match what the code actually does.

fredespi (approved):

PR #280, opened 2026-08-19
test(e2e): xfail the Windows HF-checkpoint serve (EAI-8031)

rominf:
- [x] If this PR fixes a bug, searched `tests/e2e-cucumber/expectations.toml` for the fixed ticket ID and removed/narrowed any now-stale xfail rows. — n/a, this PR *adds* a row for an open bug.

## Summary

The `E2E tests (Strix Halo, Windows)` lane has been red on every run for as long
as the check history goes back, so it is red on every open PR — which is exactly
the state in which a genuine new Windows regression goes unnoticed.

Across the last 17 runs of that lane there is exactly one *host-independent*
unexpected failure: `serve-hf-checkpoint-inference`. It fails on both Windows
runners, every time, and has since the scenario was introduced in #242. Root
cause is tracked in #260: serving `unsloth/Qwen3-0.6B-GGUF:Q4_0` through the
`owner/repo:variant` direct path exits 0, but `/v1/models` never lists the
model, so the endpoint never becomes ready.

This declares that scenario as the known bug it is, and shortens its wait.

**Why scoped to Windows GPU hosts and not wider:** the Strix Halo Ubuntu lane
serves the same checkpoint correctly, and on Windows the short-recipe-name path
(`serve-lemonade-inference`, same checkpoint) still passes. A broader row would
turn a real regression on either into a silently expected failure. Two unit
tests pin that scoping so it cannot drift.

**Why 240s and not the 90s the neighbouring rows use:** the lane leaves
`E2E_SERVE_TIMEOUT_SECS` unset, so today it spends the full 600s default waiting
for a serve that never arrives — about 10 minutes of every run. A *healthy*
lemonade serve of this cache-warm checkpoint takes ~120s on this host (measured
on run 32185662513: the three passing GPU-serve scenarios took 117s, 117s and
120s). 90s would sit below that, which would keep the scenario red even after
#260 is fixed and hide the XPASS that tells us to delete the row. 240s cuts six
minutes off the lane while keeping ~2x headroom.

Risk: low. Test-expectation data plus unit tests; no product code changes.

## Scope boundary — this does not make the lane green on its own

The remaining Windows failures are **not** product bugs and are deliberately not
xfail'd here. One of the two Windows runners cannot complete a TLS connection to
`github.com`:

```
Failed to download llama-server.exe from: .../llama-b9752-bin-win-rocm-7.13-x64.zip
  - Download failed after 6 attempts.
Last error: Download failed: SSL connect error (CURL code: 35)
```

so the lemonade llama.cpp backend install always fails there, taking out
`bench-load-real-serve`, `chat-tool-definitions-accepted`,
`chat-end-to-end-local-model` and `serve-lemonade-inference`. The split is
clean: those four fail 10/10 runs on the affected machine and pass 7/7 on the
healthy one, over a ~23h window.

Those four therefore need a runner fix, not an expectations row. Marking them
xfail would either fail the lane as stale XPASSes whenever it lands on the
healthy runner, or — with `flaky = true` — permanently blind four real GPU
scenarios to regressions while recording an infrastructure fault as a product
bug.

## Test plan

- `cargo test -p e2e-cucumber --lib expectation` — 27 passed, including the two
  new tests. Both fail without the new row (verified by reverting it) and pass
  with it.
- `cargo clippy --locked --workspace --all-targets -- -D warnings` and
  `cargo clippy --locked -p e2e-cucumber --test e2e -- -D warnings` — clean.
- `cargo fmt --check` — clean.
- The lane itself is the real check: expect `serve-hf-checkpoint-inference` to
  reconcile as xfail rather than an unexpected failure, and the lane to pass
  outright when it lands on the healthy runner.

### DIFF
--- a/tests/e2e-cucumber/expectations.toml
+++ b/tests/e2e-cucumber/expectations.toml
@@ -178,3 +178,33 @@ flaky = true
 when = { is_wsl = true, therock_family = "gfx*" }
 bug = "EAI-7998"
 reason = "`examine --json` reports has_amd_gpu:false on a WSL2 host whose own summary names a gfx target."
+
+# --- EAI-8031 (public mirror: ROCm/rocm-cli#260): on a Strix Halo WINDOWS host,
+# serving a canonical Hugging Face checkpoint through the `owner/repo:variant`
+# direct-serve path (`unsloth/Qwen3-0.6B-GGUF:Q4_0`) exits 0 but `/v1/models`
+# never lists the model, so the endpoint never becomes ready. Fails on every run
+# since the scenario was introduced by 37a3b63c (#242, 2026-08-12); it is the
+# lane's only host-independent unexpected failure.
+#
+# os=windows: this is specific to Windows AND to this serve path — the Strix Halo
+# Ubuntu lane passes the scenario, and on Windows the short-recipe-name path
+# (`serve-lemonade-inference`, same checkpoint) passes too. So it is not lemonade
+# managed serve at large, and must not widen to a bare `when = {}`.
+# therock_family/has_amd_gpu mirror the sibling rows and document intent: the
+# scenario is @requires-gpu, so on a no-GPU Windows lane it is SKIPPED, never
+# xfail'd.
+#
+# serve_timeout_secs: the lane leaves `E2E_SERVE_TIMEOUT_SECS` unset, so this
+# waits out the full 600s default on a serve that never arrives — ~10 minutes of
+# every run spent on a known bug. A HEALTHY lemonade serve of this same
+# (cache-warm) checkpoint on this host reaches ready in ~120s: measured on run
+# 32185662513, where the three passing GPU-serve scenarios took 117s, 117s and
+# 120s. 240s therefore keeps ~2x headroom — deliberately NOT the 90s used by the
+# EAI-7423 rows above, which is below a healthy serve here and would mask the
+# fix by keeping the scenario red once the bug is closed. Remove this row when
+# EAI-8031 lands; the resulting XPASS is the signal that it did. ---
+[["serve-hf-checkpoint-inference"]]
+when = { os = "windows", therock_family = "gfx*", has_amd_gpu = true }
+bug = "EAI-8031"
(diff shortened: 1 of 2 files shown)


fredespi:
I reviewed this PR and did not find anything blocking. It looks safe to approve pending final validation (the lane itself, CI, and a maintainer's own pass — things a diff-only read can't see). This is not itself an approval.

**What the PR does:** adds one `serve-hf-checkpoint-inference` row to `expectations.toml` declaring the known EAI-8031 bug as an xfail, scoped to Windows GPU hosts, and shortens its serve-readiness wait to 240s. It also adds two unit tests that pin the scoping and the timeout headroom. No product code changes.

**What I checked:**

- **Scenario id matches.** The row keys on `serve-hf-checkpoint-inference`, which is exactly the `@id` on the scenario in `model_serving.feature:79` (`@requires-gpu @requires-engine:lemonade`), so the row is live rather than a dead no-op.
- **Scoping is correct and can't over-match.** `when = { os = "windows", therock_family = "gfx*", has_amd_gpu = true }` resolves to xfail only on a Windows GPU host: the Strix Ubuntu lane is excluded by `os` (it reports `linux`), and `serve-lemonade-inference` stays expect-pass on Windows because its own row is `os = "linux"`. On a no-GPU Windows lane the scenario is `@requires-gpu`, so `resolve()` returns `Skip` before the xfail check ever runs — consistent with the `has_amd_gpu = true` intent documented in the comment.
- **Timeout is sound.** `serve_timeout_for` returns 240 for this host; it sits between the ~120s healthy-serve measurement and the 600s global default, so it fails fast on the known bug while leaving enough headroom to surface the XPASS signal once the bug is fixed. The two new tests enforce `240 <= secs < 600`.
- **No regression-masking risk.** `deny_unknown_fields` on `Condition`/`XfailEntry` guards against a typo'd key silently becoming an always-xfail, and the row is deliberately not widened to a bare `when = {}`.

I ran `cargo test -p e2e-cucumber --lib expectation` locally: 27 passed, including the two new tests. The change is well-scoped, well-documented, and low-risk.

r0x0r (approved):

PR #282, opened 2026-08-19
ci: stop Dependabot rebasing PRs that do not need it

rominf:
## Summary

`main` no longer requires a branch to be up to date before merging (`strict` is off). The merge queue already builds `main` + PR and runs the full required check set against that merged tree, so an out-of-date branch was never what kept a bad merge out — it only forced churn.

Left on the default `auto` strategy, Dependabot re-pushes **every** open PR each time `main` lands a commit. That is expensive here specifically:

- every path filter in `ci.yml` folds in `.github/workflows/**`, and
- a cargo bump touches `Cargo.lock`, which both the `heavy` and `tpn` filters match,

so each re-push drags the whole suite — self-hosted GPU lanes included — through a rerun that changes no outcome. With 10 open dependency PRs and roughly 2–3 merges a day, that is a large share of what those scarce runners are doing.

Set `rebase-strategy: disabled` on both ecosystems and let the merge queue be the thing that reconciles a PR with `main`.

**Trade-off, stated plainly:** Dependabot also stops auto-resolving *genuine* conflicts. A PR that really does conflict — most often on `Cargo.lock` when two dependency PRs land close together — will now sit in a conflicting state until someone runs `@dependabot recreate`. There is no "rebase only on conflict" setting, so this is the cost of removing the churn. It is rare and visible, unlike what it replaces.

Risk: low and fully reversible — deleting the two lines restores current behaviour.

## Test plan

Configuration-only; there is no local harness for Dependabot behaviour, so this is verified by inspection and observation rather than a test:

- `prek run check-yaml` passes.
- Placement verified against the [`rebase-strategy` schema](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#rebase-strategy) — a top-level key on each `updates` entry, alongside `cooldown` / `groups`.
- The observable effect is negative (absence of re-pushes), so confirmation is that open Dependabot PRs stop being force-pushed when `main` moves. Worth a look a few days after merge.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,6 +19,17 @@ updates:
     # widen it. Cooldown delays only version updates, never security updates.
     cooldown:
       default-days: 7
+    # `main` no longer requires a branch to be up to date before merging: the
+    # merge queue already builds `main` + PR and runs the full required set
+    # against that merged tree, so an out-of-date branch is not a merge blocker.
+    # Left on `auto`, Dependabot re-pushes every open PR each time `main` moves,
+    # and because every path filter in ci.yml folds in `.github/workflows/**`,
+    # each of those re-pushes drags the whole suite — self-hosted GPU lanes
+    # included — through a full rerun that changes no outcome. Disable it and let
+    # the queue be the thing that reconciles with `main`. Cost: Dependabot also
+    # stops auto-resolving genuine conflicts, so a PR that truly conflicts (most
+    # often on `Cargo.lock`) needs `@dependabot recreate` by hand.
+    rebase-strategy: disabled
     groups:
       github-actions:
         patterns:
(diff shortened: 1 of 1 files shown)


fredespi:
I reviewed this change and it looks safe to approve pending final validation (CI and a maintainer's own pass).

What it does: sets `rebase-strategy: disabled` on both the `github-actions` and `cargo` update entries so Dependabot stops force-pushing every open PR each time `main` moves. The key is placed correctly — a top-level key on each `updates` entry, alongside `cooldown` / `open-pull-requests-limit` / `groups` — and `disabled` is a valid value for the option (the only alternative being the default `auto`).

I checked the premise the change rests on and it holds: `main`'s branch protection currently has `strict: false`, so an out-of-date branch genuinely isn't a merge blocker — the merge queue reconciles PRs against `main`. The inline comments accurately state the trade-off (Dependabot no longer auto-resolves genuine conflicts, so a truly conflicting PR needs `@dependabot recreate` by hand), so there's no hidden cost being glossed over.

Config-only, no code paths touched, and fully reversible by deleting the two lines. Nothing blocking from a diff read. Note this reflects what a diff-only review can see, not CI results, and is not itself an approval.

r0x0r (approved):

PR #283, opened 2026-08-19
fix(package): ship THIRD_PARTY_NOTICES.txt in the release bundle

rominf:
## Summary

The release archive ships `bin/rocm`, `bin/rocmd`, `README.md`, `LICENSE.TXT` and the platform installer — but not `THIRD_PARTY_NOTICES.txt`.

Those binaries statically link the crates that file attributes, and most of the licenses involved (MIT, BSD-2/3-Clause, Apache-2.0) require the notice text to accompany the *distributed binary*. Keeping the notices current in the repository — which CI already enforces byte-for-byte — does not discharge that obligation when the artifact users download omits them entirely.

- Add `THIRD_PARTY_NOTICES.txt` to `SHARED_TOP_LEVEL_FILES`, so it lands at the bundle top level next to `LICENSE.TXT` on both platforms.
- Assert its presence in the staging round-trip test.

Found while tracing why generated-manifest staleness caused so many CI failures: the file is gated hard on every PR, yet never actually shipped. The enforcement and the purpose had drifted apart.

Risk: low. One entry in an existing copy list; `copy_required` already fails loudly if a listed file is missing, so a packaging run cannot silently omit it.

Scope note: this fixes what the bundle *contains*. Whether the notices are regenerated at release time is a separate gap — `release.yml` neither regenerates nor verifies them today, relying entirely on the PR-time gate. Worth addressing, but not bundled here.

## Test plan

- `stage_then_checksum_and_sign_round_trips` now asserts the file is staged. Verified it genuinely covers the fix: reverting only the `SHARED_TOP_LEVEL_FILES` change makes it fail with `assertion failed: ...THIRD_PARTY_NOTICES.txt.is_file()`, and it passes with the change in place.
- `cargo test -p xtask` — all package tests pass (11 run).
- `cargo fmt --all --check` and `cargo clippy --locked -p xtask --all-targets -- -D warnings` clean.

### DIFF
--- a/xtask/src/package.rs
+++ b/xtask/src/package.rs
@@ -9,8 +9,8 @@
 //! cross-platform command:
 //!
 //! * a top-level `<dist>/` directory holding `bin/rocm[.exe]`, `bin/rocmd[.exe]`,
-//!   `README.md`, `LICENSE.TXT`, and the platform installer (`install.sh` on
-//!   Unix, `install.ps1` on Windows);
+//!   `README.md`, `LICENSE.TXT`, `THIRD_PARTY_NOTICES.txt`, and the platform
+//!   installer (`install.sh` on Unix, `install.ps1` on Windows);
 //! * an archive of that directory — `<dist>.tar.gz` on Unix, `<dist>.zip` on
 //!   Windows — with the bundle directory as the single top-level entry;
 //! * a `<archive>.sha256` sidecar in `sha256sum`/`Get-FileHash` syntax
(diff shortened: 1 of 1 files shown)


r0x0r (approved):
Correct, minimal, well-tested fix. The statically-linked MIT/BSD/Apache-2.0 crates require their notice text to travel with the distributed binary, and the bundle currently ships only `README.md` and `LICENSE.TXT`. Adding the entry to `SHARED_TOP_LEVEL_FILES` fixes this at the right layer, and `copy_required` already fails loudly if the source is missing, so a packaging run can't silently omit it.

**What's good**

- Root-cause fix in the shared copy list that both platforms consume.
- Regression test genuinely covers the fix: it writes the file into the fake workspace and asserts it lands in the staged bundle. Reverting only the const change leaves the source unstaged and fails the assertion; the test-only revert would make `copy_required` bail on the missing source. Both directions are consistent.
- Module header and constant doc comment both updated, so the docs don't drift from behavior.
- Honest scope note is accurate: `release.yml` calls `cargo xtask package` but neither regenerates nor verifies `THIRD_PARTY_NOTICES.txt`, relying on the PR-time staleness gate in `ci.yml`. Leaving that follow-up out of this PR is the right call.

**Minor, non-blocking**

- The round-trip test asserts the file is *staged* but not that it appears *inside* the archive. Acceptable, since staging is the direct input to `archive()` and archive nesting is covered by `targz_bundle_produces_expected_layout`. A one-line assertion on the archive entry would make it airtight if you want belt-and-suspenders.
- No platform gating needed — the file is legally required on Windows and Linux alike, and the shared list applies to both.

LGTM.

PR #292, opened 2026-08-20
test(e2e): pin the second README-walkthrough batch (EAI-8046..8049)

fredespi:
## Summary

Test-only. Pins the correct behaviour for the second README-walkthrough batch of defects (EAI-8046–8049, tracked upstream as EAI-8050), following the same shape as #216: one behavioural scenario per defect, each asserting the CORRECT behaviour and registered as an `xfail` in `expectations.toml` citing its EAI id. The mock lane stays green now, and each row goes stale (XPASS) the day its bug is fixed. No product behaviour changed.

All four defects were reproduced by hand on an MI300X app-dev pod, then re-confirmed in a plain Linux container against this branch's binary before the scenarios were written. Each scenario was verified to fail on its assertion step (setup steps green) for the stated reason — not a miswiring.

## What each ticket now pins

- **EAI-8046** — `rocm chat` ignores a piped prompt. `chat.feature` scenario 8 pipes a prompt with no `--prompt` and asserts the model actually received it. The load-bearing assertion is the mock's received request (the bug's signature is that no request is made), not stdout. Needed a new `run_rocm_with_stdin` helper — `Command::output` gives the child a null stdin, so no existing step could reach the piped path.
- **EAI-8047** — automation checks listed by `rocm automations list` cannot be enabled from what the listing shows. New `automations.feature` + steps: the scenario derives each check's identifier from the listing (an explicitly-exposed id if present, else the display-name slug as today's fallback) and asserts `automations enable <id>` succeeds. Deriving from the listing — rather than hard-coding the real ids — is the contract.
- **EAI-8048** — `rocm dash --replay <missing>` opens the interactive dashboard instead of refusing. `dash.feature` scenario 10 drives it under a PTY, imposes its own deadline (the current behaviour hangs), and asserts both a prompt non-zero exit and that the alternate-screen interactive view never opened.
- **EAI-8049** — the activation hint printed by `rocm engines shell` names a path that does not exist (the recorded `env_path` already includes the interpreter directory, so the hint doubles that segment). `engine_shell.feature` scenario 2 runs it under a PTY and asserts the hint names a file that exists — phrased as path-exists so the row goes stale on any correct fix.

## Notes

- Harness additions live entirely in the e2e-cucumber test crate: `run_rocm_with_stdin`, a new `automations_steps` module, and PTY-driver support (`wait_for_any_exit`, `screen_logical_lines`, and a race-free latch that detects the alternate-screen enter sequence in the raw byte stream with a bounded post-exit reader drain).
- The four `expectations.toml` rows are all `when = {}` (platform-independent), each with a comment stating the scope and the removal condition.

## Test plan

- [ ] Blocking `e2e` mock job is green: pass-or-xfail, no XPASS. (Locally on a native-Linux container: clippy `-D warnings` clean, e2e-cucumber lib tests pass, mock lane reconciles to 6 xfail / 0 XPASS / 0 unexpected failures.)
- [ ] Each new scenario xfails for its stated reason (verified: each fails on its assertion step with setup steps green).
- No GPU dispatch required — all four run on the fast mock lane.


michaelroy-amd (approved):
Reviewed at `1452ee174a87cf05693284b32a8b16fce0cc094a`. The four scenarios pin the intended user-visible contracts rather than implementation details: piped chat reaches the model, listed automation identifiers are usable, a missing replay file fails without opening the TUI, and the engine-shell activation path exists. Each expectation is scoped to its ticket and removal condition; helper changes have bounded PTY drains and preserve command environment isolation. All 17 required checks pass, including the mock E2E gate and DCO/signoff. No blocking finding. Approving.

rominf:
I read through this PR's diff (scenarios, step definitions, expectations.toml, and the tui_driver additions) and traced each of the four claimed defects against the actual source:

- EAI-8046: confirmed. `Command::Chat` in `apps/rocm/src/main.rs` only reads stdin via the interactive dash path when `interactive_terminal()` is true; with a piped, non-interactive run and no `--prompt`, it falls straight to `render_chat_text`, which never touches stdin or sends a request. That matches the `--prompt` help text's claim ("reads from standard input... when omitted") and the EXAMPLES block, which this scenario correctly proves false today.
- EAI-8047: the derivation logic in the new `automations_steps.rs` (parsing the `automations list` output for an id inline in brackets/parens or on a detail line, falling back to slugifying the display name) looks correct against `render_automations_text`/`watcher_plain_name` in `main.rs`, and won't silently pass once a fix publishes a real id.
- EAI-8048: confirmed `dash::run` in `apps/rocm/src/dash.rs` doesn't validate the `--replay` path before entering the interactive event loop. The new `wait_for_any_exit`/`entered_alternate_screen` latch in `tui_driver.rs` is a sound way to assert "refused before opening the TUI" rather than just "exited non-zero," and the byte-level alternate-screen-enter detection (with a carry buffer for reads split across the escape sequence) is correctly bounded.
- EAI-8049: confirmed by reading `resolve_engine_env` plus `runtime_python_activation_hint`/`runtime_python_env_bin_dir` in `rocm-core`, and the vLLM install's `env_path = command.parent()` in `engines/vllm/src/lib.rs`. The doubled `bin/bin/activate` path is real, and phrasing the assertion as "the hint names a file that exists" (rather than "not doubled") is the right way to keep the scenario meaningful after any valid fix.

I also checked the `expectations.toml` additions against the file's own grammar and existing rows, and the four new `@id:` tags line up 1:1 between the feature files and the xfail table.

One thing worth flagging, though it's not a code issue: GitHub currently reports this PR as having a merge conflict. That's because `dash.feature`'s scenario 10 slot was independently filled by a different, later-merged PR (the "launcher shows live serving instance" scenario) after this branch's base commit — diffed against the actual merge-base with main, this PR's own changes apply cleanly and are a pure addition. It'll need a rebase to reconcile scenario numbering with main, but that's a mechanical rebase, not a defect in the tests or steps themselves.

Nothing else stood out. This looks safe to approve after a rebase and once CI (mock e2e lane, clippy) is green — this is a diff-only read, not a substitute for CI or a maintainer's own pass.

siloteemu:
🔴 Automated review · pr-review-watcher · 1452ee17

## Summary
Test-only PR pinning four README-walkthrough defects as cucumber scenarios that assert the *correct* behaviour, plus four `when = {}` expected-failure rows and the TUI-driver support they need (latched alternate-screen detection, logical-line reads, a non-zero-exit wait). **Approve.** Verified: none of the four scenarios can pass against a binary lacking the behaviour — chat waits for an actual completion request on the mock (`wait_for_chat_request`) rather than for output, automations derives each check's identifier *from the listing* instead of hard-coding the real ids, dash asserts both a prompt non-zero exit *and* that the alternate screen was never entered, and engine-shell asserts the printed hint names an existing file against a fixture planted in the shape vLLM actually records (`command.parent()`); all four run on `ci.yml`'s `e2e` job (ubuntu-latest, mock lane), documented in-repo as a required check, none carrying `@requires-gpu`/`@nightly`/`@lifecycle`/`@merge-queue`, and `@requires-os:linux` matches that runner; no expected-failure row is stale — I read the product source at the base (current main) for all four bugs and each is still present, with `git log HEAD..prw-base` confirming no main-only commit touched the chat stdin arm, `render_automations_text`, the dash replay path, or `runtime_python_activation_hint`/vLLM's `env_path`; no dead rows (every `expectations.toml` key resolves to an existing `@id:` tag) and no dead helpers (all eleven new/changed symbols including `run_rocm_with_stdin`, `slug_of`, `drain_reader` have call sites), every new step phrase has a definition, and a non-`flaky` XPASS exits 1 so each row self-reports the day its bug is fixed. Not run per policy: the e2e suite itself (CI reports 25 passing, 0 failing for this head). Blocking: 0 · Non-blocking: 5.

**None.**

## Non-blocking
- `tests/e2e-cucumber/tests/e2e/dash_steps.rs:250` — the exit deadline is a hard-coded `Duration::from_secs(2)`, while every sibling TUI wait in this suite uses 30s (`SCREEN_TIMEOUT` / `default_timeout()`); harmless today (any failure satisfies the xfail), but once the dash bug is fixed and the row is deleted, a >2s cold binary start on a loaded shared runner turns the required lane red.
- `tests/e2e-cucumber/tests/e2e/chat_steps.rs:141` — `assert_eq!(user_content, "Hello from standard input")` requires the trailing newline to be trimmed; a fix that forwards stdin verbatim would keep the row xfailed rather than letting it go stale, which is the failure mode the file's own comments set out to avoid.
- `tests/e2e-cucumber/tests/e2e/automations_steps.rs:78` — `exposed_identifier` recognises `[id]`, `(id)`, `id:`/`identifier:`; a fix publishing the id in another shape (a trailing column, an em-dash, `slug:`) leaves the row permanently xfailed. Same function: a display name containing a parenthesised single word ("Server recovery (beta)") would be misread as an identifier.
- `tests/e2e-cucumber/tests/e2e/engines_steps.rs:161` — the hint is parsed by stripping a literal `source ` prefix only; a quoted path or a `.`-prefixed hint would fail the existence check for a reason unrelated to the product.
- Branch is 38 commits behind main and last absorbed main at `16252c2`; the green CI was measured against that base. The stale-row risk does not materialise (verified above), but a rebase before merge would re-measure against current main.

fredespi:
Thanks both — all five non-blocking points are addressed in `3490bf6f`, on top of a `main` merge in `832e1064`.

**Rebase / 38 commits behind** (both reviewers): merged `origin/main` rather than rebasing, so the branch stays fast-forwardable and no force-push is needed. The conflict was narrower than it looked but not purely mechanical: `main` adopted the `<feature>-NN` scenario-numbering convention repo-wide in the meantime, and its launcher scenario had claimed the `dash-10` slot. Both scenarios are kept — ours is now `dash-11` — and the other three were renumbered onto the new convention (`chat-09`, `automations-04`, `engine-shell-02`). `main` had also independently added its own `automations.feature` and `automations_steps.rs`; the listing-discoverability scenario folds in alongside those with no duplicate step phrases.

**The four durability notes** — each was a way a *correct* fix could leave the row xfailed instead of going stale, which is precisely what these scenarios exist to prevent, so all four are fixed:

- `chat_steps.rs` — compare the user message trimmed. The step pipes a trailing newline, so forwarding stdin verbatim would otherwise not XPASS.
- `automations_steps.rs` — the identifier is now recognised as a detail line (`id`/`identifier`/`slug`/`key`/`name`), bracketed, parenthesised, dash-separated, or as a trailing column. To avoid the false positive you flagged, a parenthesised/dash/column token must be identifier-shaped (lowercase with a separator, as every real watcher id is), so `Server recovery (beta)` is not read as an id. The slug-of-display-name fallback is unchanged, so today's listing still fails the contract.
- `engines_steps.rs` — accepts `. <path>` and a quoted path as well as `source <path>`.
- `dash_steps.rs` — the bare `Duration::from_secs(2)` is now a documented `REPLAY_REJECTION_TIMEOUT`. I kept it bounded rather than moving to the 30s `default_timeout()`: today's behaviour is a hang, so this bound is paid on every mock-lane run until the bug is fixed. It now has ample headroom over a cold start, which was the flake risk you identified. Happy to align it with the siblings instead if you'd prefer.

**Verification** (Linux container, native-Linux host matching CI): clippy `-D warnings` clean, e2e-cucumber lib 113/113, and the mock lane reconciles **6 xfail, 0 XPASS, 0 stale** — so all four rows still fail for their stated reason and none went stale.

Two failures in that run are pre-existing environment gaps, not regressions, and I verified each by running the identical gate against a pristine `origin/main` checkout in the same container: the 10 `rocm-engine-lemonade` `child_runtime_dir` tests (identical 87 passed / 10 failed on stock `main`; they trip on the container work dir permitting cross-UID rename), and `engine-shell-marks-the-prompt` (also fails on stock `main` there — the container shell prompt ends in `#` rather than `$`). Both are green on CI.

siloteemu:
🔴 Automated review · pr-review-watcher · 3490bf6f

## Summary
Test-only PR adding four cucumber scenarios (plus supporting steps and a PTY-driver capability) that pin correct behaviour for four open CLI defects as documented `xfail` rows. Verdict: **Approve** — all five durability points are addressed, the merge is clean, and all four rows still fail for their stated reason. Verified: the five fixes (trimmed chat comparison, multi-shape identifier recognition, extra hint prefixes, the named replay bound, the merge) all present in `3490bf6`; merge integrity checked four ways — every expectations row resolves to an existing `@id`, no duplicate step phrases or regex/literal ambiguity across the 358 step definitions, every Gherkin step in every feature file resolves, scenario ids and indexes unique and sequential post-renumber (`cargo test --test feature_naming`: 4/4 pass); each scenario re-derived against the product source and confirmed discriminating; no prompt-injection content found in the diff. Blocking: 0 · Non-blocking: 5.

**None.**

## Discrimination check (each row fails only against a binary lacking the behaviour)

- `chat-09` — the `Given` steps start the mock and plant the service record but issue no chat POST, and `rocm chat --provider local` on this path emits only the readiness GET. So `wait_for_chat_request` times out today; it can only be satisfied by a binary that actually forwards stdin. The `.trim()` and the "last user-role message" lookup widen what counts as a correct fix without weakening it. **Discriminating.**
- `automations-04` — `apps/rocm/src/main.rs:16002` prints `  {plain name} ({on|off})` with 4-space detail lines and publishes no id; `watcher_plain_name` maps e.g. `server-recover → "Server recovery"`, so the slug fallback (`server-recovery`) mismatches for all six watchers and `enable` returns rc=1. `exposed_identifier` returns `None` against today's output, and the "guard the guard" non-empty assertion stops a parse failure from vacuously passing. **Discriminating.**
- `dash-11` — today the command hangs, so `wait_for_any_exit` errors out; passing needs both a prompt non-zero exit *and* the latched alternate-screen flag staying false. The raw-byte `\x1b[?1049h` scan with a carry buffer correctly survives an enter+leave in one read. **Discriminating.**
- `engine-shell-02` — `runtime_python_activation_hint` appends `bin/activate` to the recorded `env_path`, and `engines/vllm/src/lib.rs:480` records `command.parent()`, so the hint names `<env>/bin/bin/activate` while the fixture plants `<env>/bin/activate`. **Discriminating.**

## On the 2s → 30s question: keep the dedicated bound, do not align

Keep `REPLAY_REJECTION_TIMEOUT` (`tests/e2e-cucumber/tests/e2e/dash_steps.rs:22`) as its own value. The 30s siblings and this one are not the same kind of wait, so the same number would mean different things:

- Sibling waits are **success** waits — the budget is only spent when something has already gone wrong, so generosity is free in the normal case.
- This is a **failure-detection** wait on a path that hangs by construction. The budget is spent *in full, on every mock-lane run*, for as long as the bug is open. Aligning to 30s triples a standing CI cost and buys no additional safety, because there is nothing slow to wait for — the contract is "refuse before opening the TUI".
- After the fix it reverts to a success wait, and 10s is well over an order of magnitude more than a cold binary start plus PTY spawn and immediate refusal needs on a loaded shared runner — so the row can be deleted without leaving a flake behind. The original bare 2s was the genuinely risky value; raising it to 10s was the right correction, and naming it with the rationale inline is the right way to keep it from being "fixed" later by someone who only sees the inconsistency.

## Non-blocking

- `tests/e2e-cucumber/tests/e2e/engines_steps.rs:91` — the fixture hard-codes the vLLM-shaped manifest, so a fix applied at the *producer* layer (making the install record the env root) leaves the planted manifest unchanged and the row xfailed forever. Defensible (already-installed environments keep the old manifest, so a producer-only fix genuinely doesn't repair them), but that migration requirement isn't what the expectation row's `reason` states — worth a sentence there.
- `tests/e2e-cucumber/tests/e2e/automations_steps.rs:96` — the detail-line key list includes `name:`/`key:` and matches on block order, so a fix printing `name: Server recovery` above an `id:` line would capture the display name. Consider trying `id`/`identifier`/`slug` first, or applying `is_identifier_shaped` to the `name:`/`key:` hits too.
- `tests/e2e-cucumber/tests/e2e/chat_steps.rs:151` — bare `Duration::from_secs(2)` where the sibling dash assertions use `default_timeout()`. The CLI has already exited by the time this `Then` runs, so it is a pure 2s tax on every mock-lane run today; either name it like the dash one or use the shared default.
- `tests/e2e-cucumber/tests/e2e/automations_steps.rs:61` — `is_identifier_shaped` requires a `-`/`_`, so a published id without a separator would go unrecognised in the parenthesised/dash/column shapes. Low risk (every real watcher id has one) and documented, but the bracketed and detail-line shapes are the only unconditional escapes.
- `tests/e2e-cucumber/tests/e2e/chat_steps.rs:151` — uses `wait_for_chat_request` rather than `wait_for_chat_request_where`. The mock module's own docs warn that a local-endpoint readiness probe is itself a chat request; `rocm chat` doesn't issue one on this path today, but if it ever does, this row fails for the wrong reason and never goes stale. The dash managed-chat step already guards exactly this.

siloteemu:
🔴 Automated review · pr-review-watcher · 2c4b6002

## Summary
Test-only PR adding four e2e scenarios (`chat-09`, `automations-04`, `dash-11`, `engine-shell-02`) that pin four README-walkthrough defects, each with an always-on expected-failure row; the two new commits address the prior round's open non-blocking items and nothing else. **Approve.** Verified: merge/renumber integrity is clean (all 16 expectation rows name existing scenarios, the only repeated table key is the documented OR-condition pair, 122 `@id` tags all unique, per-file indexes sequential — `automations-01..04`, `chat-01..09`, `dash-01..11`, `engine-shell-01..02` — 359 step definitions with no ambiguous pair and all 610 Gherkin steps resolving to exactly one definition; `cargo test -p e2e-cucumber --test feature_naming` passes 4/4); staleness is clean — I traced all four bugs in the current base myself and each is still live for its stated reason (`Command::Chat` with no `--prompt` and non-interactive stdin calls `render_chat_text` and never reads stdin; `render_automations_text` prints only `watcher_plain_name` with no id field, and `builtin_watcher` is exact-match, so the slug fallback `server-recovery` ≠ `server-recover`; `dash::run` passes `replay` through with no readability check before the TUI opens; `runtime_python_activation_hint` appends `bin/activate` to a vLLM-recorded `env_path` that is already `command.parent()`); on the revert question every scenario and every assertion is discriminating in both directions — each fails today for the stated reason and would go stale on a correct fix, with no trivially-true assertion, no swallowed failure, no `Then` whose work is done in the `When`, and the one vacuous-pass risk explicitly guarded (`assert!(!checks.is_empty())` in `automations-04`); and all three open items are genuinely fixed — `PIPED_PROMPT_TIMEOUT` and `REPLAY_REJECTION_TIMEOUT` are now named constants with recorded rationale, `exposed_identifier` sweeps the unambiguous `id`/`identifier`/`slug` keys across the whole block before the ambiguous `key`/`name` (which must additionally pass `is_identifier_shaped`), and `wait_for_chat_request_where` matches on the last user message containing the piped prompt so a `"Say ok."` readiness probe cannot satisfy it. Blocking: 0 · Non-blocking: 3.

**None.**

## Non-blocking
- `tests/e2e-cucumber/tests/e2e/dash_steps.rs:29` — the 10s `REPLAY_REJECTION_TIMEOUT` is paid in full on every mock-lane run while the replay-rejection bug stays open; the tradeoff is documented at the constant, just worth tracking if lane runtime becomes a concern.
- `tests/e2e-cucumber/tests/e2e/automations_steps.rs:84-90` — `is_identifier_shaped` requires a `-`/`_`, so a fix that published a separator-less id in an *ambiguous* position (`Server recovery (gpumetrics)`) would leave the row xfailed; the narrowing is deliberate and documented, and the unambiguous `id:`/`[bracketed]` shapes are unaffected.
- `tests/e2e-cucumber/expectations.toml` — the new rows carry internal tracker references, matching the file's own documented grammar and every pre-existing row on the base; flagging only because this is a public repository and the convention is repo-wide rather than introduced here.

siloteemu (approved):
Reviewed at `2c4b6002`: no blocking issues, third consecutive clean pass.

Verified this round: merge and renumber integrity after absorbing the base's repo-wide scenario renumbering — every expectation row names a scenario that exists, all scenario tags unique, per-file indexes sequential, no ambiguous or duplicate step definitions, and every step resolves. Staleness checked directly: all four pinned defects were traced in the current base and each is still live for its stated reason, so none of the rows has quietly become an unexpected pass.

Each scenario discriminates in both directions — it fails today for the reason stated, and would go stale on a correct fix rather than silently continuing to pass. Three minor notes in the review comment; none needs action before merge.

tests/e2e-cucumber/expectations.toml
```diff
@@ -208,3 +208,57 @@ when = { os = "windows", therock_family = "gfx*", has_amd_gpu = true }
...
+# bug is fixed. None is `flaky`: each reproduced deterministically.
+
+# --- EAI-8046: `rocm chat` ignores a piped prompt ---
+# With no `--prompt` and a non-terminal stdin, the command renders an assistant
+# status summary and exits 0 without ever reading stdin or contacting the model,
+# though `--prompt`'s own help says it reads standard input when omitted and the
+# command's EXAMPLES block shows `echo … | rocm chat`. Container-measured: the
+# `--prompt` control produced a POST on the served model, the piped run produced
+# only the readiness GET. Remove this row once a piped prompt reaches the model.
+[["chat-cli-stdin-prompt"]]
```
rominf, line 227: **Blocking — this row is already fixed on `main`, and leaving it here turns the required lane red.**

Commit `51036072` ("fix(chat): read a one-shot prompt from piped stdin (#339)") is on `main`. It added `read_piped_prompt()` wired into `Command::Chat`, and it shipped its own `chat-09` regression scenario under the **identical** tag `@id:chat-cli-stdin-prompt` (now `features/chat.feature:103` on `main`) as a permanent expect-pass — with a stronger assertion than this one: byte-exact whitespace preservation, versus `.contains()` against a `.trim()`ed value here.

Two consequences:

1. This row isn't marked `flaky`, so once the scenario passes, `tests/e2e-cucumber/tests/e2e.rs:1354-1366` classifies it as `stale_xpass` and calls `std::process::exit(1)`. That takes the required `E2E tests` check red.
2. Keeping both scenarios makes `tests/feature_naming.rs:166-179` panic on `duplicate @id:chat-cli-stdin-prompt` — the map is suite-wide, so a same-file duplicate trips it just the same.

Suggested fix: drop this PR's `chat-09` scenario, its step definitions, and this expectations row. #339 already covers the behaviour, and covers it better.

tests/e2e-cucumber/features/chat.feature
```diff
@@ -92,3 +92,11 @@ Feature: Chat and endpoint detection
     And the model is registered with the CLI
     When the user sends a one-shot chat prompt through the CLI
     Then the CLI prints the assistant's reply
+
+  @id:chat-cli-stdin-prompt
+  Scenario: chat-09 - The chat CLI answers a prompt received through standard input
```
rominf, line 97: This is the scenario half of the stale `chat-cli-stdin-prompt` row (see my comment on `expectations.toml`). `main` now carries its own `chat-09` under the same `@id:chat-cli-stdin-prompt` tag at `chat.feature:103`, so this one needs to go along with its steps — otherwise `feature_naming.rs` panics on the duplicate id before the suite even runs.

tests/e2e-cucumber/expectations.toml
```diff
@@ -208,3 +208,57 @@ when = { os = "windows", therock_family = "gfx*", has_amd_gpu = true }
...
+bug = "EAI-8047"
+reason = "`automations list` publishes no usable identifier, so the checks it lists cannot be enabled from what it shows."
+
+# --- EAI-8048: `dash --replay` does not validate its recording ---
+# Pointed at a file that does not exist, the dashboard enters its alternate
+# screen and waits for input instead of refusing. Container-measured under a
+# pseudo-terminal: no output, killed at a 12s timeout (rc=124). The scenario's
+# step imposes its own deadline, so this xfails promptly rather than hanging.
+# Remove this row once an unreadable recording is rejected before the TUI opens.
+[["dash-replay-rejects-an-unreadable-session"]]
```
rominf, line 249: **Blocking — same drift as the chat row: fixed upstream since this PR's last green run.**

Commit `0595b926` ("fix(dash): validate --replay path before entering the TUI (#331)") is on `main`. It added its own `dash-11 - Replaying a missing recording fails before entering the dashboard` (`@id:dash-replay-missing-file-fails-fast`, now `features/dash.feature:127`) along with `wait_for_refusal` / `in_alternate_screen` / `drain_final_screen` in `tui_driver.rs`.

This PR's `dash-11` asserts the same contract — prompt non-zero exit, alternate screen never entered — so against fixed `main` it passes, which makes it a stale XPASS and takes the required lane red exactly like the chat row.

Keeping both also leaves two scenarios numbered `dash-11`, which fails **two** `feature_naming` tests, not one: `scenario_indexes_are_unique_across_the_suite` and `scenario_names_are_indexed_sequentially_per_feature` (with #331's scenario present, the twelfth must be `dash-12`).

tests/e2e-cucumber/features/dash.feature
```diff
@@ -123,3 +123,16 @@ Feature: Interactive dashboard
...
+  # Driven under a pseudo-terminal like the rest of this feature: the replay path
+  # opens the same interactive event loop, and a piped run is refused before it
+  # gets there (it fails on the terminal, which proves nothing about the file).
+  #
+  # The step imposes its own deadline and treats expiry as a FAILURE rather than
+  # waiting for the scenario timeout. Today the dashboard takes an unreadable
+  # session as far as the alternate screen and then sits waiting for input, so a
+  # plain "assert a non-zero exit" step would hang for as long as CI allowed.
+  @id:dash-replay-rejects-an-unreadable-session @requires-os:linux
+  Scenario: dash-11 - Replaying a session that cannot be read is refused promptly
```
rominf, line 136: The scenario half of the stale `dash-replay-rejects-an-unreadable-session` row. #331's `dash-11` on `main` covers this contract now, so this scenario and its steps come out with the expectations row.

The reasoning captured in the comment block above it — why this must be driven under a pty, and why the step owns its own deadline instead of leaning on the scenario timeout — is genuinely good and is worth preserving somewhere, even if the scenario itself doesn't survive the rebase.

tests/e2e-cucumber/features/automations.feature
```diff
@@ -25,3 +25,16 @@ Feature: Automation watchers
...
+
+  # Scenarios 1-3 enable a watcher by an id the test already knows. This one pins
+  # the complementary contract: the listing is the only place a user learns which
+  # background checks exist, so whatever it shows has to be enough to act on.
+  # Deriving each check's identifier from the listing the way a reader would IS
+  # the contract under test — the identifiers are deliberately not written into
+  # the scenario or the steps, because hard-coding them would pass against a
+  # listing that exposes nothing. No fixtures, no GPU, no network.
+  @id:automations-listed-checks-can-be-enabled
+  Scenario: automations-04 - Every background check that is listed can be turned on
```
rominf, line 37: **Blocking, but this is the easy one — it just needs renumbering.**

`main` independently added `automations-04 - Enabling during an in-flight daemon spawn starts no second daemon` (`@id:automations-autostart-defers-during-spawn-window`). The `@id` tags, the step phrases and the Rust symbols do **not** collide with yours — only the index does.

That's not cosmetic though: `feature_naming.rs` hard-fails on both `scenario_indexes_are_unique_across_the_suite` and `scenario_names_are_indexed_sequentially_per_feature`. So this becomes `automations-05` after the rebase.

Unlike the chat and dash rows, the defect this one pins is still live on `main` — `render_automations_text` still prints only `watcher_plain_name` with no identifier — so the scenario itself stays.

tests/e2e-cucumber/expectations.toml
```diff
@@ -208,3 +208,57 @@ when = { os = "windows", therock_family = "gfx*", has_amd_gpu = true }
 bug = "EAI-8031"
 reason = "Managed lemonade serve of a Hugging Face `owner/repo:variant` checkpoint exits 0 but the endpoint never becomes ready on Windows."
 serve_timeout_secs = 240
+
+# ── Second README-walkthrough batch (EAI-8050) ─────────────────────────────
```
rominf, line 212: **Blocking (mechanics), anchored here because this block is one of the conflict sites.**

`gh` reports `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`, and `git merge-tree` against current `main` produces conflicts in **eight** files: `expectations.toml`, `automations.feature`, `chat.feature`, `dash.feature`, `automations_steps.rs`, `chat_steps.rs`, `dash_steps.rs`, and `tui_driver.rs`.

Worth flagging that this is not a mechanical rebase. Resolving it means deciding — twice — which of two competing implementations of the same scenario survives. This file conflicts in particular because `main` deleted the block that this batch appends after.

Once the two stale scenarios are dropped, the header comment above needs rewriting too: it describes a four-defect batch, and after the rebase it's two.

tests/e2e-cucumber/tests/e2e/tui_driver.rs
```diff
@@ -259,13 +268,54 @@ impl TuiSession {
         self.finished
     }
 
+    /// Whether the child ever switched the terminal to its alternate screen —
+    /// i.e. opened the full-screen interactive view. Latched, so it stays true
+    /// after the child switches back on exit. Lets a "must be refused before the
+    /// TUI opens" contract be asserted, not just "exited non-zero".
+    pub fn entered_alternate_screen(&self) -> bool {
```
rominf, line 275: **Not blocking — a salvage request. Please don't throw this away.**

This latch is strictly stronger than what `main` ended up with. `main`'s `in_alternate_screen()` snapshots `parser.screen().alternate_screen()` after the child has exited, without draining first. A binary that entered the alternate screen and then left it reads `false` under that approach — which is precisely the race this latched `AtomicBool`, set from the raw byte stream with a carry buffer across read boundaries and paired with a bounded `drain_reader()` on exit, is built to close.

Even though this PR's `dash-11` gets dropped in favour of #331's, this mechanism is the better primitive and is worth porting into `main`'s `dash-11` as a follow-up.

rominf (changes requested):
Requesting changes — but I want to lead with the important part: **there is nothing wrong with this diff.** The test design here is careful, and I went through all four scenarios specifically hunting for ways they could pass without proving anything. They came back clean.

What changed is the ground underneath. `main` moved after this PR's last green CI run (CI completed 11 Sep; the upstream fixes landed 14 and 15 Sep), and **two of the four defects this PR pins as xfails have since been fixed upstream**. The earlier reviews that checked for staleness were *correct when they were made* — this is drift, not a reviewer error and not an author error. Most of the work still lands.

### What drifted

**`chat-cli-stdin-prompt` is now fixed on `main`.** #339 landed `read_piped_prompt()` wired into `Command::Chat`, and shipped its own `chat-09` regression scenario under the *identical* `@id:chat-cli-stdin-prompt` tag as a permanent expect-pass — with a stronger assertion than this one (byte-exact whitespace preservation vs. `.contains()` on a trimmed value). Two consequences, both of which turn the required `E2E tests` lane red rather than just going unnoticed. Details inline.

**`dash-replay-rejects-an-unreadable-session` is now fixed on `main`.** #331 added its own `dash-11` (`@id:dash-replay-missing-file-fails-fast`) asserting the same contract — non-zero exit, alternate screen never entered — plus the `wait_for_refusal` / `in_alternate_screen` / `drain_final_screen` helpers in `tui_driver.rs`. Same stale-XPASS consequence, plus a second scenario index collision.

**`automations-04` collides by index only.** `main` independently added its own `automations-04`. The `@id` tags, step phrases and Rust symbols do *not* collide — but `feature_naming.rs` hard-fails on both the uniqueness and the sequential-numbering assertions, so the index isn't cosmetic. This one just needs renumbering to `automations-05`.

**The rebase is not mechanical.** `gh` reports `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`, and `git merge-tree` produces conflicts in eight files. Resolving it means deciding, twice, which of two competing implementations of the same scenario survives.

### Net effect

After a correct rebase this PR shrinks from four pinned defects to two — the automations-listing scenario (renumbered) and `engine-shell-02` — plus the harness support those two still need. That's a large enough scope change that the description and test plan, which describe a four-defect batch, should be rewritten to match.

### Two things worth knowing

**Both approvals are stale.** michaelroy-amd approved at `1452ee17` (4 commits behind) and siloteemu at `2c4b6002` (1 behind). For what it's worth, the only commit since `2c4b6002` is the merge `0f6be360`, and its `tui_driver.rs` conflict resolution — converging on `main`'s `wait_for_exit_code` while keeping this branch's `drain_reader()` — is sound. That is emphatically not where the problem is.

**Please don't throw away the alt-screen latch.** This branch's `entered_alternate_screen` is strictly stronger than what `main` ended up with, and I'd like to see it ported even though the scenario that uses it gets dropped. Left a comment on it inline.

### What I checked that held up

The vacuous-pass audit was real, not a formality, so a couple of specifics: in `automations_steps.rs` the guard-the-guard `assert!(!checks.is_empty(), …)` fires before the enable loop, and the real assertion sits *outside* the loop so it always executes; `engines_steps.rs:161-197` terminates all four failure modes in `panic!`/`assert!`; and all four scenarios carry only `@requires-os:linux`, so none of them is silently skipped-but-counted-green on any lane.

I also independently re-confirmed against current `main` that the **two non-stale rows are still live**: `render_automations_text` still prints only `watcher_plain_name` with no identifier, and `runtime_python_activation_hint` still appends `bin/activate` to a vLLM `env_path` that was recorded as `command.parent()` — so the `<env>/bin/bin/activate` defect is real and this PR's characterisation of it is accurate.

### One process observation

The advisory `xfail-hint` job (`ci.yml:173`) only reads the *current PR's own* title, body and commits. It structurally cannot warn when somebody else's merged fix stales a row here — which is exactly how this landed silently. Probably worth a follow-up issue rather than anything in this PR.

siloteemu:
🔴 Automated review · pr-review-watcher · 0f6be360

## Our earlier approval on this PR is stale — do not read it as current

An approval from this account sits on an earlier revision of this pull request. It should not be taken as a current judgement, and the reason is a change in the base branch rather than anything the author did.

The point another reviewer raised about the piped-stdin scenario checks out on all three of its factual premises, verified directly rather than taken on trust:

- the commit that added `read_piped_prompt()` is present in the base branch's history;
- the base branch's `tests/e2e-cucumber/features/chat.feature` already carries `@id:chat-cli-stdin-prompt` at line 103;
- this branch still adds the same `@id:chat-cli-stdin-prompt` in `features/chat.feature` and a matching row in `expectations.toml`.

So after a merge the identifier would exist twice. That is a genuine defect, not a preference, and it is the kind the repository's own naming guard exists to catch.

No new gate is filed here: another reviewer already holds a change request covering this, and the branch does not currently merge cleanly. This comment exists so that the approval standing above it is not mistaken for a current one. Nothing else in this pull request was re-reviewed at this revision.

_No check-run counts are quoted here because this note rests on repository content rather than on check state._

PR #295, opened 2026-08-21
fix(dash): launcher shows live serving instances; amd-smi detection off critical path (eai-8190)

r0x0r:
## Summary

Two related fixes so the dashboard's model/instance count reflects reality promptly:

1. **Launcher front door always showed "Idle"**, even with a model actively serving. `run_launcher` built an empty `AppState`; it now reads the managed-service registry (the same authority `rocm services` reads) once per hub-loop pass and seeds the front door's instances from it. Also normalizes "serving" everywhere in the dash UI to `is_serving()` (`Ready` + `Running`), since a served model reports `Ready`, not `Running` — the running-instance counts previously undercounted.
2. **~15-20s dashboard startup lag**: `amd-smi` detection (`amd-smi version` + first `system_info()`) ran inline before the telemetry run loop's first tick, blocking managed-service discovery and the first snapshot broadcast behind it. Detection is now spawned in the background and adopted via a `oneshot` channel the moment it lands, so serving instances surface within one discovery tick instead of waiting on GPU detection.

## Commits

- `fix(dash): seed launcher front door with live serving instances`
- `fix(dash-daemon): run amd-smi detection off the run loop's critical path`

## Testing

- `cargo test -p rocm-dash-daemon -p rocm-dash-tui -p rocm --lib --tests` — all passing, including two new unit regression tests (`launcher_state_seeds_serving_from_registry_instances`, `serving_detail_counts_ready_instances_as_running`) and one new daemon-level regression test (`slow_gpu_detection_does_not_delay_service_discovery`, which asserts on snapshot **ordering** rather than wall-clock timing to avoid flakiness under subscriber starvation).
- `cargo clippy -p rocm -p rocm-dash-tui -p rocm-dash-daemon --all-targets -- -D warnings` — clean.
- Local pre-commit/pre-push hooks (signing, DCO sign-off, fmt, clippy, cargo test) all passed.

**Cucumber e2e coverage gap:** no existing scenario in `tests/e2e-cucumber/features/dash.feature` covers the bare-`rocm` launcher front door or GPU-detection timing, and none were added here — the fixes are covered at the unit/integration level (above) instead. Flagging per repo convention rather than silently relying on existing coverage; happy to add a scenario if maintainers want dedicated e2e coverage for the launcher hub screen.

## GPU hardware testing

Not run against real GPU hardware in this environment (no `/dev/kfd` available); the amd-smi-detection fix is validated via the fake-binary regression test described above, which exercises the real detection/adoption code path.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3569,6 +3569,7 @@ dependencies = [
  "keyring-core",
  "libc",
  "rocm-core",
+ "rocm-dash-core",
  "rocm-dash-daemon",
  "rocm-dash-tui",
  "rocm-deps",
--- a/apps/rocm/Cargo.toml
+++ b/apps/rocm/Cargo.toml
@@ -22,7 +22,9 @@ rocm-deps = { path = "../../crates/rocm-deps" }
 # rocm-dash unified dashboard launch. The
 # telemetry daemon + ratatui-0.30 TUI are launched from the `dash` verb; tokio
 # drives the async daemon/TUI from the otherwise-sync `rocm` binary.
-# (rocm-dash-core is pulled in transitively by the two below — not a direct dep.)
+# rocm-dash-core is a direct dep so the launcher front door can build serving
+# `Instance`s from the managed-service registry.
+rocm-dash-core = { path = "../../crates/rocm-dash-core" }
 rocm-dash-daemon = { path = "../../crates/rocm-dash-daemon" }
 rocm-dash-tui = { path = "../../crates/rocm-dash-tui" }
 tokio = { workspace = true }
--- a/apps/rocm/src/dash.rs
+++ b/apps/rocm/src/dash.rs
@@ -358,7 +358,12 @@ pub fn run_launcher(chat_mock: bool) -> Result<()> {
     let config = RocmCliConfig::load(&paths).unwrap_or_default();
     let theme = config.dashboard.tui.theme;
     loop {
-        match rocm_dash_tui::ui::launcher::run_launcher(&theme)? {
+        // Re-read the managed-service registry on each pass so the front door
+        // reflects models started (or stopped) during a prior flow. This is a
+        // cheap status-only file read — no telemetry daemon and no network
+        // readiness probes, so the front door stays instant.
+        let serving = launcher_serving_instances(&paths);
+        match rocm_dash_tui::ui::launcher::run_launcher(&theme, serving)? {
             None => return Ok(()),
             Some(choice) => match launcher_route(choice) {
                 LauncherRoute::Focused(focus) => run_focused(focus)?,
@@ -367,6 +372,28 @@ pub fn run_launcher(chat_mock: bool) -> Result<()> {
             },
         }
     }
+}
+
+/// Serving instances for the launcher front door, read from the managed-service
+/// registry (the same authority `rocm services` reads).
+///
+/// Deliberately cheap: a status-only registry read with no network readiness
+/// probes and no telemetry daemon, so the front door renders instantly.
+fn launcher_serving_instances(paths: &AppPaths) -> Vec<rocm_dash_core::metrics::Instance> {
+    use rocm_dash_daemon::registry::{discover_managed_services, load_service_records};
+    let records = load_service_records(&paths.services_dir());
+    discover_managed_services(&records)
+        .svcs
+        .into_iter()
+        .map(|svc| rocm_dash_core::metrics::Instance {
+            container_id: svc.container_id,
+            container_name: svc.container_name,
+            model_name: svc.model_name,
+            status: svc.status,
+            port: svc.port,
+            ..Default::default()
+        })
+        .collect()
 }
 
 /// Entry point for a focused launcher flow (Set up / Serve / Diagnose).
--- a/apps/rocm/src/dash.rs
+++ b/apps/rocm/src/dash.rs
@@ -66,6 +66,9 @@ pub fn runner_options(
         // amd-smi ships inside the managed runtime wheel's bin dir, not on PATH;
         // resolve the path so the GPU collector can find it.
         amd_smi_binary: Some(rocm_core::resolve_amd_smi_binary()),
+        // Production always runs the real `/dev/kfd` pre-flight; only daemon
+        // integration tests with a fake binary skip it.
+        amd_smi_skip_kfd_preflight: false,
     }
 }
 
@@ -942,5 +945,78 @@ mod tests {
             first_summary.preferred_engine.as_ref(),
             first.preferred_engines.first()
         );
+    }
+
+    /// The launcher front-door seam that regressed: a `ready` managed vLLM
+    /// service recorded on disk must map into a live `Instance` (id, model,
+    /// port, status) so bare `rocm` shows it instead of "Idle". Reads the same
+    /// registry `rocm services` reads — status-only, no daemon, no network.
+    #[test]
+    fn launcher_serving_instances_maps_ready_registry_record() {
+        let root = std::env::temp_dir().join(format!(
+            "rocm-cli-launcher-serving-test-{}-{}",
+            std::process::id(),
+            rocm_core::unix_time_millis()
+        ));
+        let p = AppPaths {
+            config_dir: root.join("cfg"),
+            data_dir: root.join("data"),
+            cache_dir: root.join("cache"),
+        };
+        let services_dir = p.services_dir();
+        std::fs::create_dir_all(&services_dir).unwrap();
+        std::fs::write(
+            services_dir.join("svc.json"),
+            r#"{"service_id":"vllm-launcher","engine":"vllm",
+                "model_ref":"meta-llama/Llama-3.1-8B","canonical_model_id":"m",
+                "host":"127.0.0.1","port":11435,
+                "endpoint_url":"http://127.0.0.1:11435/v1","mode":"managed",
+                "status":"ready","created_at_unix_ms":1}"#,
+        )
+        .unwrap();
+
+        let instances = launcher_serving_instances(&p);
+
+        assert_eq!(instances.len(), 1, "the ready managed service must surface");
+        let inst = &instances[0];
+        assert_eq!(inst.container_id, "vllm-launcher");
+        assert_eq!(inst.model_name, "meta-llama/Llama-3.1-8B");
+        assert_eq!(inst.port, Some(11435));
+        assert_eq!(inst.status, rocm_dash_core::metrics::InstanceStatus::Ready);
+
+        let _ = std::fs::remove_dir_all(&root);
+    }
+
+    /// A non-scrapeable record (unbound `port:0`) must NOT surface as a live
+    /// instance on the front door — the registry is the authority and `:0` is
+    /// never a real serving endpoint.
+    #[test]
+    fn launcher_serving_instances_skips_unbound_record() {
+        let root = std::env::temp_dir().join(format!(
+            "rocm-cli-launcher-serving-skip-test-{}-{}",
+            std::process::id(),
+            rocm_core::unix_time_millis()
+        ));
+        let p = AppPaths {
+            config_dir: root.join("cfg"),
+            data_dir: root.join("data"),
+            cache_dir: root.join("cache"),
+        };
+        let services_dir = p.services_dir();
+        std::fs::create_dir_all(&services_dir).unwrap();
+        std::fs::write(
+            services_dir.join("svc.json"),
+            r#"{"service_id":"vllm-unbound","engine":"vllm","model_ref":"m",
+                "canonical_model_id":"m","host":"127.0.0.1","port":0,
+                "mode":"managed","status":"ready","created_at_unix_ms":1}"#,
+        )
+        .unwrap();
+
+        assert!(
+            launcher_serving_instances(&p).is_empty(),
+            "an unbound (:0) record must not surface as a live instance"
+        );
+
+        let _ = std::fs::remove_dir_all(&root);
     }
 }
--- a/crates/rocm-dash-collectors/src/amd_smi.rs
+++ b/crates/rocm-dash-collectors/src/amd_smi.rs
@@ -43,7 +43,32 @@ impl AmdSmiCollector {
     /// directory rather than on `PATH`, so callers resolve the path or command
     /// name (via `rocm_core::resolve_amd_smi_binary`) and pass it here.
     pub async fn detect_with_binary(binary: impl Into<OsString>) -> Option<Self> {
-        if !kfd_accessible() {
+        Self::detect_with_binary_inner(binary, false).await
+    }
+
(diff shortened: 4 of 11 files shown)


volen-silo:
## Review: changes requested

Solid diagnosis on both bugs, and the `is_serving()` normalization checks out — but two of the three fixes ship without working coverage, and the amd-smi change introduces a user-visible false alarm on healthy hardware.

---

### Blocking

**1. False "amd-smi unavailable" alarm on every dashboard startup with a working GPU**

`crates/rocm-dash-daemon/src/runner.rs:281`

```rust
warnings.push("amd-smi unavailable (no /dev/kfd or binary missing)".into());
```

Before this change, `gpu == None` meant exactly one thing: GPU permanently unavailable, decided before the loop started. After it, `gpu == None` also means "detection is still in flight" for the whole window the PR describes as ~15s. `gpu_tick` defaults to 1s, so on a healthy GPU host this warning is now pushed into roughly 15 consecutive snapshots, renders as the `⚠ N` header badge (`crates/rocm-dash-tui/src/ui/mod.rs:346-356`) and the Hardware-tab warning panel, and then self-clears. On such a host this string previously never appeared at all.

That is a regression in exactly the property the PR is trying to improve — the dashboard reflecting reality promptly.

Smallest fix is to gate on the flag the loop already tracks:

```rust
} else if gpu_init_done {
    warnings.push("amd-smi unavailable (no /dev/kfd or binary missing)".into());
}
```

Worth noting the codebase already has vocabulary for "known but not yet settled" — `InstanceStatus::Starting { phase }` and `ObservationFreshness` — if you'd rather surface "detection pending" explicitly than suppress.

**2. The new daemon regression test provides no coverage on any host without `/dev/kfd`**

`crates/rocm-dash-daemon/tests/end_to_end.rs:346` (`slow_gpu_detection_does_not_delay_service_discovery`)

`AmdSmiCollector::detect_with_binary` short-circuits on `!kfd_accessible()` *before* invoking the supplied binary (`crates/rocm-dash-collectors/src/amd_smi.rs:45-48`). So the `sleep 4` fake is never executed, detection resolves instantly to `None`, `gpu_system_info` stays `None` forever, and the assertion `surfaced_before_gpu == Some(true)` holds — **identically with the fix reverted**, since pre-fix the inline detection also returns immediately on such a host.

Verified empirically rather than by inspection:

```
test slow_gpu_detection_does_not_delay_service_discovery ... ok
test result: ok. 1 passed ... finished in 0.03s
```

0.03s, against a fake that sleeps 4 seconds. The fake never ran. On the whole CI matrix except the GPU lane, this test is a no-op that reads as coverage.

The test's own doc comment anticipates this but draws the wrong conclusion — "exactly the case where this bug cannot occur" conflates *the bug can't happen here* with *the fix is exercised here*. A test that passes against reverted code isn't a regression test.

Options: make the KFD pre-flight injectable/overridable for tests so the fake is actually reached; or inject the detection future directly at the runner seam; or, if neither is palatable, mark the test as requiring `/dev/kfd` and name the gated lane that runs it, so it doesn't read as green coverage where it can't fail.

While you're in there — this test sits in precisely the right window to have caught finding 1, but only inspects `gpu_system_info`. Adding `assert!(!snap.warnings.iter().any(|w| w.contains("amd-smi unavailable")))` would have failed today.

**3. Launcher fix has no behavioral coverage — scenario requirement not discharged**

Two parts:

- `AGENTS.md` §3 requires a Gherkin scenario when a change alters what a CLI user can observe, and explicitly says "a unit test asserting the internal helper does NOT discharge this; it proves the function, not the behavior." The one new test (`launcher_state_seeds_serving_from_registry_instances`) tests a private helper that didn't exist before this commit, fed hand-built `Instance`s. That's the exact pattern the rule names.

- I checked whether this is structurally untestable, and it isn't. `TuiSession::spawn(world, args)` already drives the real crossterm TUI under a PTY (`tests/e2e-cucumber/tests/e2e/tui_driver.rs:125`), and bare `rocm` routes to `dash::run_launcher` whenever `interactive_terminal()` is true (`apps/rocm/src/main.rs:1210-1218`) — which a PTY slave satisfies, as the existing `dash`/`chat` scenarios rely on. Spawning with an empty args slice reaches exactly the changed code. So this is an unwritten scenario, not a harness limitation.

The PR body's up-front flagging is genuinely appreciated and much better than silence, but §3's escapes are: cite an existing scenario id, name a gated lane, or classify as purely internal. Deferring the decision to maintainers isn't one of them. (The amd-smi timing fix is a fairer argument for "hard to cover at PTY level" — but then §3 wants the gated lane named.)

- Related: `launcher_serving_instances` (`apps/rocm/src/dash.rs:381`) is the actual fix seam — it reads real service-record JSON and maps it to `Instance` — and it has no test at all. The mapping is the part that could plausibly be wrong; the added test bypasses it entirely.

---

### Non-blocking

- **Silent hole on the oneshot's closed path.** `gpu_init_rx.try_recv()` under `if !gpu_init_done && let Ok(...)` — if the detection task ever drops the sender without sending, `try_recv` returns `Err(Closed)` forever, `gpu_init_done` never flips, and there is no diagnostic at all. No live panic path exists in `AmdSmiCollector` today (it's `Option`/`Result`-only), so this is latent rather than a live bug, but a `warn!` on `Err(Closed)` would cost one line and close it. This matters more once finding 1 is fixed, since the per-tick warning is currently the only signal that GPU is off.

- **The `None`-means-two-things ambiguity is broader than the badge.** The agent/chat GPU tool (`crates/rocm-dash-tui/src/agent.rs:243-267`) answers `gpu_count: 0` during the detection window with no indication that detection is pending — an assistant asked "how many GPUs" in the first seconds gets a confidently wrong answer. Other consumers degrade gracefully (`efficiency.rs`, `vram.rs` render `—`), so the badge is the only active misrender, but a `Pending` state rather than suppression would close the set.

- **Redundant nesting** at `runner.rs:252-262`: `if detected.is_some() { if let Some(i) = &info { ... } }` — `info` is `Some` exactly when `detected` is `Some` (the spawned task constructs it that way), so the outer check is unreachable-when-false. A single `if let Some(i) = &info` reads the same and can't drift.

- **`..Default::default()`** in `launcher_serving_instances`: correct today (the launcher's render path reads none of the defaulted fields), but it means any future launcher line that reads e.g. VRAM silently renders 0 rather than "unknown". Worth a brief comment pinning why the defaults are safe here.

- **Platform coverage**: the new daemon test is `#[cfg(unix)]`, leaving Windows — a supported platform per `AGENTS.md` §6 — uncovered for this path.

---

### Tradeoffs

- **Two independent fixes in one PR.** `AGENTS.md` §11 asks for one logical change. The commits are cleanly split with disjoint file sets and independent tests, so bisect/revert of either is trivial and the cost here is reviewability rather than maintainability — flagging it as a judgment call for maintainers, not asking for a split.

- **Front-door freshness is entry-time only.** `run_launcher` takes the registry snapshot once, then `event::read()` blocks on a `let state` that never changes. A model started in another terminal while the front door is displayed won't appear until the user navigates away and back. The doc comment scopes this honestly ("during a prior flow"), and for a front door it seems like the right call — noting it so the choice is visible.

---

### Verified clean (worth stating, since these were the load-bearing claims)

- **"Normalizes serving everywhere" is accurate, not overclaimed.** Audited every remaining `InstanceStatus::Running`/`Ready` comparison workspace-wide: `pane.rs:407` really was the last narrow `== Running` count site, and post-PR that grep is empty. Every other site is either fixture data or a `Ready | Running` arm already treating them identically. The new semantics also match what `rocm services` itself does (`ready|running → running`, `apps/rocm/src/main.rs:14700-14707`), so the counts will genuinely agree.

- **`serving_detail_counts_ready_instances_as_running` is a real behavioral test** — traced it against pre-fix code and both assertions fail there. This is the one new test that does what a regression test should.

- **The "deliberately cheap" claim on `launcher_serving_instances` holds.** Traced `load_service_records` (`read_dir`/`read` only) and `discover_managed_services` (pure string/status logic) end to end — no HTTP readiness probes, no subprocess, no daemon. The front door stays instant, and the doc comment is accurate.

- Both commits signed with DCO sign-off, no AI footers, substantive commit messages, leak scan clean over the diff and PR text, CI green.

r0x0r:
Thanks for the thorough review — pushed 422abff addressing all three blocking items.

**1. Premature `amd-smi unavailable` warning.** The `else` that produced the warning is now gated on `gpu_init_done`: while detection is still in flight `gpu == None` is transient, so the loop stays silent instead of flashing the badge on healthy hardware. If the detection task ever ends without sending a result (a latent path today — no fallible await), the adoption block settles `gpu_init_done = true` and emits the honest "detection task ended without a result" warning once, so the badge resolves rather than never clearing.

**2. The regression test was a no-op on GPU-less CI.** You were right: `kfd_accessible()` short-circuited before the fake `amd-smi` ever ran, so the test passed even with the fix reverted. I added an `amd_smi_skip_kfd_preflight` seam on `RunnerOptions` (plumbed to a `#[doc(hidden)]` `AmdSmiCollector::detect_with_binary_skipping_kfd_preflight`) that **only** the daemon test flips — production always runs the mandatory `/dev/kfd` pre-flight (the D-state guard is intact). The fake now genuinely runs the slow detection through the real path, so the test fails if detection moves back onto the critical path. It also now asserts the surfaced snapshot carries **no** `amd-smi unavailable` warning, which would have caught finding #1.

**3. Behavioural coverage for the front-door fix.** Added two direct unit tests for `launcher_serving_instances` (a `ready` registry record maps to a live `Instance` with the right id/model/port/status; an unbound `:0` record is dropped), plus a new PTY scenario `@id:launcher-shows-live-serving-instance` that drives bare `rocm` (the launcher front door) with a live managed service staged on disk and asserts the strip shows `Serving <model>` rather than `Idle — nothing serving`.

**Non-blocking, called out as follow-ups (not in this PR):**
- The launcher/daemon regression test is `#[cfg(unix)]` (the fake is a shell script); Windows/ConPTY coverage for this path stays a follow-up, consistent with the other `@requires-os:linux` dash scenarios.
- Pending/detection-in-flight state for the agent GPU tool (`gpu_count: 0` during the detection window) is unchanged here.

Local verification: `rocm-dash-daemon` / `rocm-dash-collectors` / `rocm-dash-tui` suites and the new `apps/rocm` `dash::` tests pass; `e2e-cucumber` compiles. The macOS clippy const-fn noise in `rocm-core` is pre-existing and unrelated; the Linux CI clippy gate is authoritative.

volen-silo:
## Follow-up review: all three blocking findings resolved — LGTM

Re-verified against the current head `422abff3` (no force-push; the delta since my review is exactly that one commit).

### 1. Premature `amd-smi unavailable` alarm — **resolved**

`runner.rs:307` now gates the warning on `gpu_init_done`, and the in-flight branch is silent. The honest case is not swallowed: once detection settles with no GPU, `gpu_init_done` is true and the warning fires on that same tick and every tick after.

Not taken on faith — I flipped the gate back to unconditional locally and the new daemon test failed with exactly the premature-warning assertion:

```
assertion `left == right` failed: managed service surfaced with a premature
`amd-smi unavailable` warning while detection was still in flight
  left: Some(false)  right: Some(true)
```

So the assertion you added genuinely guards this, as you said it would.

### 2. Daemon regression test was a no-op on GPU-less CI — **resolved, and I want to be precise about the evidence**

The test still finishes in **0.04s** against a fake that sleeps 4s. That is *not* a red flag, and I should correct my original reasoning: 0.04s is the correct post-fix runtime, because the test breaks out and aborts the runner the moment the instance surfaces — it never waits for detection. My earlier "0.03s, the fake never ran" inference happened to reach the right conclusion for the wrong reason; timing alone can't distinguish the two cases here.

What settles it is reverting the fix. I put detection back on the critical path (awaited the oneshot before the loop, keeping the new seam) and ran it:

```
thread 'slow_gpu_detection_does_not_delay_service_discovery' panicked at
crates/rocm-dash-daemon/tests/end_to_end.rs:434:
  managed service only surfaced after GPU detection completed
  (the ~15-20s dashboard startup lag regressed)
test result: FAILED ... finished in 13.16s
```

13.16s means the fake really executed (three invocations x 4s) through the real detection path, and the test really fails when the behaviour regresses. Before your change that same experiment would have passed. This is now a regression test.

On the seam itself: I traced every `RunnerOptions` construction site and every caller of the detection functions. `true` is reachable from exactly one place — the `#[cfg(unix)]` daemon test — the default is `false`, the production constructor in `dash.rs` hardcodes `false`, and `detect_with_binary`'s behaviour is unchanged for existing callers. The KFD D-state guard is intact in production. Non-blocking suggestion only: a `#[cfg(feature = "test-util")]` gate (enabled via the daemon's dev-dependency on the collectors crate) would make "this function does not exist in a production build" compile-time-enforced rather than doc-comment-enforced. `#[doc(hidden)] pub` on a `publish = false` workspace crate is a fine tradeoff as shipped.

### 3. Launcher behavioural coverage / scenario requirement — **resolved**

`@id:launcher-shows-live-serving-instance` has all five steps defined, the `Given` stages a `ready` registry record into the isolated data dir the spawned `rocm` actually reads, and bare `rocm` under the PTY reaches `run_launcher`. Pre-fix `run_launcher` passed no instances at all, so the strip rendered the literal `Idle — nothing serving` — your `Then` assertions are a genuine red/green pair across the fix.

Ran it rather than just reading it:

```
Scenario: 10 - The launcher front door shows a live serving model rather than idle
 ✔ Given a running managed model is available locally
 ✔ When the user opens the launcher
 ✔ Then the launcher shows the model serving
 ✔ When the user quits the launcher
 ✔ Then the launcher exits successfully
1 scenario (1 passed), 5 steps (5 passed)
```

It resolves to expected-pass with no expectations entry, and the diff touches `**/*.rs` and `**/*.feature`, so it runs in the blocking Linux lane — real coverage, not a gated lane. It also passed on the GPU lane in CI. The two unit tests on `launcher_serving_instances` cover the mapping seam I flagged as untested.

### Non-blocking follow-ups

- The `..Default::default()` comment suggestion wasn't picked up — still just a suggestion, not blocking.
- Deferring the agent GPU-tool pending state and Windows coverage is reasonable. One small note on the former: now that the badge correctly stays quiet during the detection window, the agent's confident `gpu_count: 0` is the only remaining artefact of the ambiguity. Still non-blocking.

### Verification run here

- `cargo test --workspace --all-targets` — exit 0 (first pass tripped the known `providers::tests::local_provider_*` flake; clean on re-run)
- `cargo xtask e2e -- -n "launcher front door"` — 1/1 passed

### One thing to do before merge

The branch is 15 commits behind `main` and the GPU e2e lane is red with 4 unexpected failures — none of them yours. They're staleness: the `no 'Folder:' line in examine output` failure is precisely what #316/#317 on `main` fix, and the serve-on-GPU ones are covered by other merged commits. Your new launcher scenario passes on that lane. A rebase onto `main` should take it green; worth doing so the lane isn't red at merge time.

Everything blocking is resolved. LGTM.

volen-silo (approved):

PR #298, opened 2026-08-22
fix(comfyui): pin managed torch stack during install (EAI-8051)

fredespi:
## Summary

`rocm comfyui install` installs ComfyUI's dependencies into the machine's managed ROCm runtime. It already filters `torch`/`torchvision`/`torchaudio` out of ComfyUI's *direct* requirement specs, but nothing scopes package resolution — so a **transitive** ComfyUI dependency can still resolve `torch` from PyPI, pull a CUDA build plus `nvidia-*` wheels, and displace the runtime's ROCm torch. The runtime the whole machine serves models with then becomes a CUDA build with no AMD GPU support: installing an optional app breaks the base runtime (and vLLM serving with it).

This pins the managed runtime's existing torch stack to its exact installed versions so the resolver can never swap in a CUDA build.

## Changes

- Before running `uv pip install` for ComfyUI's deps, probe the managed runtime's currently-installed `torch`/`torchvision`/`torchaudio` versions via `importlib.metadata` (no native import needed).
- Write those as a uv `--constraint` file and pass it to the install. The resolver treats the stack as already satisfied and never replaces it; because ROCm torch has no `nvidia-*` dependencies, none are pulled in.
- A genuine version conflict (a dep needing an incompatible torch) now fails the install loudly instead of silently corrupting the runtime.
- Added unit tests for the constraint-line generation and the `--constraint` argument wiring.

## Test plan

- [x] `cargo test -p rocm` (Linux) — 477 tests pass, including 3 new ComfyUI tests.
- [x] `cargo clippy` / `cargo fmt --check` clean.
- [ ] End-to-end proof is the `@requires-gpu @nightly` scenario `comfyui-install-preserves-the-rocm-runtime` from #291 (EAI-8071), which runs on a GPU host after that contract lands.

### DIFF
--- a/apps/rocm/src/comfyui.rs
+++ b/apps/rocm/src/comfyui.rs
@@ -1438,6 +1438,16 @@ const TORCH_STACK_VERSION_PROBE: &str = "import json,sys\n\
 /// each package currently present, so the ComfyUI dependency resolution is pinned
 /// to them and cannot pull a CUDA build. An absent package yields no line — we
 /// only constrain what the runtime actually ships.
+///
+/// Constraining only the installed packages is safe here because the full
+/// torch/torchvision/torchaudio trio is an invariant of a managed runtime:
+/// `therock` installs all three atomically and refuses to install unless it finds
+/// a mutually-compatible set (see `select_compatible_stack` — it bails with "no
+/// mutually compatible ... torch, torchvision, and torchaudio versions"), and
+/// ComfyUI install only ever targets such a runtime (`select_runtime` requires a
+/// managed wheel install). So there is no "torch present, torchvision/torchaudio
+/// absent" runtime in which an unpinned CUDA `torchvision` could resolve against
+/// the release part of the pinned ROCm torch and install a mixed CUDA/ROCm stack.
 fn torch_stack_constraints(
     python: &Path,
     runtime_env: Option<&ComfyUiRuntimeEnvironment>,
--- a/apps/rocm/src/comfyui.rs
+++ b/apps/rocm/src/comfyui.rs
@@ -347,10 +347,18 @@ pub(crate) fn install(
         let _ = io::stdout().flush();
         let uv = ensure_uv_binary(paths)
             .context("failed to acquire uv binary for ComfyUI dependency install")?;
+        // Pin the managed runtime's ROCm torch stack to its exact installed
+        // versions so a transitive ComfyUI dependency cannot resolve `torch` (and
+        // its `nvidia-*` CUDA deps) from PyPI and clobber it (EAI-8051). Filtering
+        // torch out of the direct specs is not enough — the resolver reaches it
+        // transitively. A constraint fails loudly on a genuine version conflict,
+        // which leaves the runtime intact rather than silently corrupting it.
+        let constraints = torch_stack_constraints(&runtime.python, Some(&runtime_env), &mut log)?;
+        let constraints_path = write_torch_constraints(&app_root, &constraints, &mut log)?;
         run_uv_logged_command(
             paths,
             &uv,
-            uv_install_args(&runtime.python, &packages),
+            uv_install_args(&runtime.python, &packages, constraints_path.as_deref()),
             Some(&runtime_env),
             &mut log,
             "install ComfyUI dependencies",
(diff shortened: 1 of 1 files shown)


apps/rocm/src/comfyui.rs
```diff
@@ -1393,13 +1401,124 @@ fn requirement_package_name(spec: &str) -> Option<String> {
...
+        .with_context(|| format!("failed to parse torch stack version probe output: {stdout}"))?;
+    Ok(parsed.into_iter().collect())
+}
+
+/// Build `name==version` constraint lines for the torch-stack packages that were
+/// found, in a stable order. Pure and side-effect free for unit testing.
+fn torch_constraint_lines(versions: &[(String, String)]) -> Vec<String> {
+    TORCH_STACK_PACKAGES
+        .iter()
+        .filter_map(|package| {
```
r0x0r, line 1502: The constraint set is deliberately limited to torch-stack packages that are *currently installed* (absent ones yield no line). That fully protects the common case where the managed runtime ships all three. But if a runtime ever has `torch` present while `torchvision`/`torchaudio` are not, a transitive ComfyUI dep could still resolve those two from PyPI as CUDA builds — a CUDA `torchvision` that only requires `torch==<ver>` (no local tag) is satisfied by the pinned ROCm torch, so it installs alongside without conflict and you get a mixed CUDA/ROCm stack. Is the full torch/torchvision/torchaudio stack an invariant of the managed ROCm runtime? If so, worth asserting/annotating that here so the guarantee is explicit; if not, consider always emitting constraints for all three (pinning absent ones to the runtime torch's compatible versions, or refusing to add them) rather than only what's found.

r0x0r (commented):
Reviewed — the approach is sound. Pinning the runtime's torch stack to its exact installed versions (local `+git…` identifiers and all) via a uv `--constraint` file is the right lever: a transitive dep that wants a compatible torch is satisfied by what's already there (no reinstall), and a genuinely incompatible one fails loudly instead of silently swapping in a CUDA build. I checked the `TORCH_STACK_VERSION_PROBE` string — the Rust `\`-line-continuations strip the source indentation and the `\x20` escapes re-supply it, so the emitted Python has valid, consistent block indentation. The pure helpers are unit-tested, and the end-to-end CUDA-rejection proof is the `@requires-gpu @nightly` `comfyui-install-preserves-the-rocm-runtime` scenario in #291 — a gated lane, named per AGENTS §3.

One substantive thread on the scope of the guarantee below.

apps/rocm/src/comfyui.rs
r0x0r: Thanks -- the follow-up (f287bb76) documents exactly the invariant I was asking about, and I verified the two references hold: `therock` resolves rocm+torch+torchvision+torchaudio as one mutually-compatible set and bails otherwise ("no mutually compatible ... torch, torchvision, and torchaudio versions were found", therock.rs), and ComfyUI's `select_runtime` refuses any non-`wheel`/non-ready runtime (comfyui.rs), so the "torch present, torchvision/torchaudio absent" case cannot occur. The guarantee is now explicit. (Tiny aside, non-blocking: the note names `select_compatible_stack`, but the actual resolver is `select_matching_pip_package_versions`.) Resolving.

r0x0r (approved):
Approving. The transitive-torch defect (EAI-8051) is fixed at the right layer: the managed runtime's exact torch/torchvision/torchaudio versions are pinned via a uv `--constraint` file, so the resolver treats the ROCm stack as satisfied and can never swap in a CUDA build, while a genuine conflict fails loudly instead of silently corrupting the runtime. The `\x20`-escaped Python probe reconstructs to valid indentation, and the pure helpers are unit-tested. My one thread -- whether "constrain only what's installed" leaves a partial-stack gap -- is resolved: the follow-up commits document (and I verified) that `therock` installs the full trio atomically as a mutually-compatible set (`select_matching_pip_package_versions` bails otherwise) and ComfyUI's `select_runtime` only ever targets a managed `wheel` runtime, so the partial-stack case cannot arise. End-to-end CUDA-rejection proof is the `@requires-gpu @nightly` scenario in #291, named per AGENTS. 24/24 CI green.

PR #308, opened 2026-08-24
fix(install): resolve canonical ROCm streams (EAI-8268)

michaelroy-amd:
## Summary
- make EAI-8268 / #308 the canonical owner for current release and nightly aggregate TheRock sources
- fail closed on unknown or incomplete source layouts; never silently fall back to legacy or another channel
- select exactly the published `device-<gfx target>` payload and preserve legacy-manifest readability
- key managed wheel runtimes by source/package composition and repair stale same-version runtimes side by side
- make update/prewarm select and activate the exact planned replacement runtime
- absorb #272's #271 regression as positive source/version/provenance/device-extra coverage

## Scope and dependencies
- #329 is the ROCm 10 Next-layout extension and remains stacked on this PR until this base lands.
- #272 is superseded after its observable #271 behavior and coverage are verified here.
- #314/current `main` remains authoritative for torch owning-runtime settlement. The obsolete torch/vLLM alignment experiments from the previous history were removed.

## Behavior
New canonical installs use exactly one aggregate source per channel and fail closed when its layout or exact device payload is unavailable. Existing managed runtimes with legacy manifests remain readable and receive a one-time side-by-side composition repair when reproducible.

## Verification
- [x] `cargo test --workspace --all-targets` — 2,467 passed across 30 suites; 8 ignored
- [x] `python3 scripts/smoke_local.py`
- [x] `cargo xtask manifest --check`
- [x] `cargo xtask tpn --check`
- [x] `cargo xtask verify-commits --base origin/main` — all 10 commits signed and signed off
- [x] `cargo xtask e2e -- -n 'Previewing a release SDK install resolves the canonical aggregate'` — 1 scenario, 4 steps passed
- [x] live release wheel dry-run — ROCm 7.14.1, canonical aggregate source, exact `device-gfx1103`
- [x] live nightly wheel dry-run — ROCm 10.1.0a20260822, canonical aggregate source, exact `device-gfx1103`
- [x] live release/nightly tarball dry-runs — canonical channel-specific catalogs, no fallback

Local `python` is unavailable, so the documented smoke script was run with the equivalent `python3` interpreter. Local Hawkeye 6.5.1 cannot parse the repository's current config; commits used the CI-pinned, checksum-verified Hawkeye 7.0.0 binary, and the required CI license-header job remains authoritative.

### DIFF
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -30053,7 +30053,7 @@ ID_LIKE="suse opensuse"
             "gfx120X-all",
             "7.14.0",
         );
-        let manifests = vec![stale.clone(), wrong_family, repaired.clone()];
+        let manifests = vec![stale, wrong_family, repaired.clone()];
 
         let selected = select_installed_update_runtime(&manifests, &repaired.runtime_key)
             .expect("the side-by-side repair must be selected by its exact key");
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -15823,11 +15823,6 @@ fn apply_runtime_update(
     let manifests = therock::load_runtime_manifests(paths)?;
     let source = select_runtime_update_source(&manifests, config, runtime_selector)?;
     let plan = therock::runtime_update_plan(paths, source, &manifests)?;
-    // The plan resolved the index for the SOURCE runtime's family; letting the
-    // install re-detect one would compose a different runtime than the plan just
-    // predicted whenever this host's GPU disagrees with the runtime being
-    // updated, and the key lookup below would then find nothing.
-    let family_override = Some(source.family.as_str());
     let mut output = String::new();
     let _ = writeln!(output, "runtime update");
     let _ = writeln!(output, "  source_runtime_key: {}", source.runtime_key);
@@ -15855,13 +15850,12 @@ fn apply_runtime_update(
 
     if dry_run {
         let _ = writeln!(output, "  mode: dry-run");
-        let install_plan = therock::install_sdk(
+        let install_plan = therock::install_sdk_for_update(
             paths,
             &source.channel,
             &source.format,
-            None,
-            None,
-            family_override,
+            &source.family,
+            plan.device_target.as_deref(),
             true,
         )?;
         let _ = writeln!(output, "  install_plan:");
@@ -15871,13 +15865,12 @@ fn apply_runtime_update(
         return Ok(output);
     }
 
-    let install_output = therock::install_sdk(
+    let install_output = therock::install_sdk_for_update(
         paths,
         &source.channel,
         &source.format,
-        None,
-        None,
-        family_override,
+        &source.family,
+        plan.device_target.as_deref(),
         false,
     )?;
     let manifests_after = therock::load_runtime_manifests(paths)?;
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -9194,6 +9194,7 @@ fn adopt_runtime_from_probe(
         // Adoption does not install torch, so the build is derived from the SDK
         // version instead.
         sdk_torch: None,
+        wheel_composition: None,
         read_only: true,
         imported_from: Some(install_root),
         installed_at_unix_ms: rocm_core::unix_time_millis(),
@@ -15821,7 +15822,12 @@ fn apply_runtime_update(
 ) -> Result<String> {
     let manifests = therock::load_runtime_manifests(paths)?;
     let source = select_runtime_update_source(&manifests, config, runtime_selector)?;
-    let plan = therock::runtime_update_plan(paths, source)?;
+    let plan = therock::runtime_update_plan(paths, source, &manifests)?;
+    // The plan resolved the index for the SOURCE runtime's family; letting the
+    // install re-detect one would compose a different runtime than the plan just
+    // predicted whenever this host's GPU disagrees with the runtime being
+    // updated, and the key lookup below would then find nothing.
+    let family_override = Some(source.family.as_str());
     let mut output = String::new();
     let _ = writeln!(output, "runtime update");
     let _ = writeln!(output, "  source_runtime_key: {}", source.runtime_key);
@@ -15840,6 +15846,7 @@ fn apply_runtime_update(
         therock::runtime_version_display(&plan.latest_version)
     );
     let _ = writeln!(output, "  status: {}", plan.status);
+    let _ = writeln!(output, "  target_runtime_key: {}", plan.target_runtime_key);
     let _ = writeln!(output, "  activate_after_install: {activate}");
     if !plan.update_available {
         let _ = writeln!(output, "  result: no newer runtime found");
@@ -15854,7 +15861,7 @@ fn apply_runtime_update(
             &source.format,
             None,
             None,
-            None,
+            family_override,
             true,
         )?;
         let _ = writeln!(output, "  install_plan:");
@@ -15870,12 +15877,21 @@ fn apply_runtime_update(
         &source.format,
         None,
         None,
-        None,
+        family_override,
         false,
     )?;
     let manifests_after = therock::load_runtime_manifests(paths)?;
-    let installed = select_installed_update_runtime(&manifests_after, source, &plan.latest_version)
-        .context("updated runtime install completed but the new runtime manifest was not found")?;
+    // By exact key, never by version: a same-version repair installs a sibling
+    // that shares the source's channel, format, family AND version, so a
+    // version match would just as happily return the stale runtime this update
+    // was meant to replace, and then activate it.
+    let installed = select_installed_update_runtime(&manifests_after, &plan.target_runtime_key)
+        .with_context(|| {
+            format!(
+                "runtime install completed but no manifest was written for the planned runtime key `{}`",
+                plan.target_runtime_key
+            )
+        })?;
     let _ = writeln!(output, "  installed_runtime_key: {}", installed.runtime_key);
     let _ = writeln!(
         output,
@@ -15940,15 +15956,11 @@ fn select_runtime_update_source<'a>(
 
 fn select_installed_update_runtime<'a>(
     manifests: &'a [therock::InstalledRuntimeManifest],
-    source: &therock::InstalledRuntimeManifest,
-    latest_version: &str,
+    target_runtime_key: &str,
 ) -> Option<&'a therock::InstalledRuntimeManifest> {
-    manifests.iter().find(|manifest| {
-        manifest.channel == source.channel
-            && manifest.format == source.format
-            && manifest.family == source.family
-            && manifest.version == latest_version
-    })
+    manifests
+        .iter()
+        .find(|manifest| manifest.runtime_key == target_runtime_key)
 }
 
 pub(crate) fn render_automations_text(paths: &AppPaths, config: &RocmCliConfig) -> Result<String> {
@@ -30018,32 +30030,40 @@ ID_LIKE="suse opensuse"
     }
 
     #[test]
-    fn installed_update_runtime_matches_latest_version_and_family() {
-        let mut source = test_runtime_manifest_for_update(
-            "old-gfx120",
+    fn installed_update_runtime_is_selected_by_exact_target_key() {
+        // Everything a version match would have keyed on is identical here:
+        // same channel, format, family and version. Only the composition-keyed
+        // runtime key tells the freshly installed repair apart from the stale
+        // runtime it was installed to replace.
+        let stale = test_runtime_manifest_for_update(
+            "release-wheel-multi-arch-7-14-0",
             "therock-release:gfx120X-all",
             "gfx120X-all",
-            "7.13.0a20260416",
+            "7.14.0",
         );
-        source.channel = "release".to_owned();
         let wrong_family = test_runtime_manifest_for_update(
-            "new-gfx110",
+            "release-wheel-multi-arch-7-14-0-ffffffffffffffff",
             "therock-release:gfx110X-all",
             "gfx110X-all",
-            "7.14.0a20260531",
+            "7.14.0",
         );
-        let target = test_runtime_manifest_for_update(
-            "new-gfx120",
+        let repaired = test_runtime_manifest_for_update(
+            "release-wheel-multi-arch-7-14-0-0123456789abcdef",
             "therock-release:gfx120X-all",
             "gfx120X-all",
-            "7.14.0a20260531",
+            "7.14.0",
         );
-        let manifests = vec![wrong_family, target.clone()];
+        let manifests = vec![stale.clone(), wrong_family, repaired.clone()];
 
-        let selected = select_installed_update_runtime(&manifests, &source, "7.14.0a20260531")
-            .expect("matching updated runtime should be selected");
+        let selected = select_installed_update_runtime(&manifests, &repaired.runtime_key)
+            .expect("the side-by-side repair must be selected by its exact key");
+        assert_eq!(selected.runtime_key, repaired.runtime_key);
 
-        assert_eq!(selected.runtime_key, target.runtime_key);
+        assert!(
+            select_installed_update_runtime(&manifests, "release-wheel-multi-arch-7-15-0")
+                .is_none(),
+            "an install that wrote no manifest for the planned key must not resolve to a sibling"
+        );
     }
 
     fn write_test_pip_runtime(
@@ -30117,6 +30137,7 @@ ID_LIKE="suse opensuse"
                 ..therock::RocmSdkPythonProbe::default()
             }),
             sdk_torch: None,
+            wheel_composition: None,
             read_only: false,
             imported_from: None,
             installed_at_unix_ms,
@@ -30156,6 +30177,7 @@ ID_LIKE="suse opensuse"
             pip_cache_dir: None,
             rocm_sdk: None,
             sdk_torch: None,
+            wheel_composition: None,
             read_only: false,
             imported_from: None,
             installed_at_unix_ms: 1,
--- a/apps/rocm/src/therock.rs
+++ b/apps/rocm/src/therock.rs
@@ -295,12 +295,6 @@ impl AggregateDeviceTarget {
         }
     }
 
-    /// The `rocm` extras this target implies. Always three extras, so a plan
-    /// never reads as though the GPU backend were optional.
-    fn rocm_extras(&self) -> String {
-        format!("libraries,devel,device-{}", self.as_str())
-    }
-
     fn reason(&self) -> Option<&str> {
         match self {
             Self::Exact(_) => None,
@@ -1493,17 +1487,23 @@ fn install_wheel_runtime(
 
 fn therock_pip_package_specs(
     package_versions: &TheRockPipPackageVersions,
-    rocm_extras: &str,
+    device_target: &str,
 ) -> Vec<String> {
+    let device_extra = format!("device-{device_target}");
     vec![
-        format!("rocm[{rocm_extras}]=={}", package_versions.rocm),
-        format!("torch=={}", package_versions.torch),
-        format!("torchvision=={}", package_versions.torchvision),
+        format!(
+            "rocm[libraries,devel,{device_extra}]=={}",
+            package_versions.rocm
+        ),
+        format!("torch[{device_extra}]=={}", package_versions.torch),
+        format!(
+            "torchvision[{device_extra}]=={}",
+            package_versions.torchvision
+        ),
         format!("torchaudio=={}", package_versions.torchaudio),
     ]
 }
-
-/// The exact install intent for `resolution` under `rocm_extras`.
+/// The exact install intent for `resolution` and its source-validated device target.
 ///
 /// Kept beside [`therock_pip_package_specs`] so the specs that identify a
 /// runtime are, by construction, the specs that get installed.
@@ -1515,7 +1515,7 @@ fn wheel_runtime_composition(
         source_layout_generation: THEROCK_SOURCE_LAYOUT_GENERATION.to_owned(),
         package_specs: therock_pip_package_specs(
             &resolution.package_versions,
-            &device_target.rocm_extras(),
+            device_target.as_str(),
         ),
         rocm_sdk_target: matches!(device_target, AggregateDeviceTarget::Exact(_))
             .then(|| device_target.as_str().to_owned()),
@@ -5125,7 +5125,7 @@ mod tests {
         );
 
         assert_eq!(target, AggregateDeviceTarget::Exact("gfx1201".to_owned()));
-        assert_eq!(target.rocm_extras(), "libraries,devel,device-gfx1201");
+        assert_eq!(target.as_str(), "gfx1201");
     }
 
     #[test]
@@ -5164,10 +5164,7 @@ mod tests {
         let target =
             AggregateDeviceTarget::resolve(None, "gfx110X-all", &published_device_targets());
 
-        assert_eq!(
-            target.rocm_extras(),
-            "libraries,devel,device-<undetermined>"
-        );
+        assert_eq!(target.as_str(), "<undetermined>");
         assert!(
             target
                 .reason()
@@ -5434,15 +5431,14 @@ mod tests {
             torchaudio: "2.10.0+rocm7.13.0a20260513".to_owned(),
             compatibility_key: "7.13.0a20260513".to_owned(),
         };
-        let package_specs =
-            therock_pip_package_specs(&package_versions, "libraries,devel,device-gfx942");
+        let package_specs = therock_pip_package_specs(&package_versions, "gfx942");
 
         assert_eq!(
             package_specs,
             vec![
                 "rocm[libraries,devel,device-gfx942]==7.13.0a20260513".to_owned(),
-                "torch==2.10.0+rocm7.13.0a20260513".to_owned(),
-                "torchvision==0.25.0+rocm7.13.0a20260513".to_owned(),
+                "torch[device-gfx942]==2.10.0+rocm7.13.0a20260513".to_owned(),
+                "torchvision[device-gfx942]==0.25.0+rocm7.13.0a20260513".to_owned(),
                 "torchaudio==2.10.0+rocm7.13.0a20260513".to_owned(),
             ]
         );
--- a/apps/rocm/src/therock.rs
+++ b/apps/rocm/src/therock.rs
@@ -27,11 +27,11 @@ use std::path::{Path, PathBuf};
 use std::process::{Command, Output, Stdio};
 use std::time::Duration;
 
-const THEROCK_NIGHTLY_PIP_INDEX_BASE: &str = "https://rocm.nightlies.amd.com/v2";
-const THEROCK_RELEASE_PIP_INDEX_BASE: &str = "https://repo.amd.com/rocm/whl";
-const THEROCK_RELEASE_PIP_MULTI_ARCH_INDEX_BASE: &str = "https://repo.amd.com/rocm/whl-multi-arch";
+const THEROCK_NIGHTLY_PIP_INDEX_BASE: &str = "https://rocm.nightlies.amd.com/whl-multi-arch";
+const THEROCK_RELEASE_PIP_INDEX_BASE: &str = "https://repo.amd.com/rocm/whl-multi-arch";
 const THEROCK_RELEASE_TARBALL_BASE: &str = "https://repo.amd.com/rocm/tarball/";
 const THEROCK_NIGHTLY_TARBALL_BASE: &str = "https://rocm.nightlies.amd.com/tarball/";
+const THEROCK_SOURCE_LAYOUT_GENERATION: &str = "multi-arch-v2";
 const DEFAULT_MANAGED_PYTHON_VERSION: &str = "3.12";
 const STARTUP_UPDATE_CHECK_INTERVAL_MS: u128 = 12 * 60 * 60 * 1_000;
 const STARTUP_UPDATE_CHECK_TIMEOUT_SECS: u64 = 2;
@@ -51,6 +51,13 @@ const THEROCK_MAX_PLAUSIBLE_TARBALL_BYTES: u64 = 256 * 1024 * 1024 * 1024;
 enum TheRockChannel {
     Release,
     Nightly,
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+struct CanonicalSource {
+    wheel_index: &'static str,
+    tarball_catalog: &'static str,
+    layout_generation: &'static str,
 }
 
 #[derive(Debug, Clone, Eq, PartialEq)]
@@ -112,12 +119,53 @@ impl TheRockChannel {
         }
     }
 
-    const fn tarball_base_url(self) -> &'static str {
-        match self {
-            Self::Release => THEROCK_RELEASE_TARBALL_BASE,
-            Self::Nightly => THEROCK_NIGHTLY_TARBALL_BASE,
-        }
+    const fn canonical_source(self) -> CanonicalSource {
+        canonical_source(self)
     }
+}
+
+const fn canonical_source(channel: TheRockChannel) -> CanonicalSource {
+    match channel {
+        TheRockChannel::Release => CanonicalSource {
+            wheel_index: THEROCK_RELEASE_PIP_INDEX_BASE,
+            tarball_catalog: THEROCK_RELEASE_TARBALL_BASE,
+            layout_generation: THEROCK_SOURCE_LAYOUT_GENERATION,
+        },
+        TheRockChannel::Nightly => CanonicalSource {
+            wheel_index: THEROCK_NIGHTLY_PIP_INDEX_BASE,
+            tarball_catalog: THEROCK_NIGHTLY_TARBALL_BASE,
+            layout_generation: THEROCK_SOURCE_LAYOUT_GENERATION,
+        },
+    }
+}
+
+fn render_canonical_provenance(
+    output: &mut String,
+    channel: TheRockChannel,
+    source_url: &str,
+    layout_generation: &str,
+    version: &str,
+) {
+    let build_date = runtime_version_build_date(version)
+        .unwrap_or_else(|| "<not published by canonical source>".to_owned());
+    let _ = writeln!(output, "  channel: {}", channel.as_str());
+    let _ = writeln!(output, "  canonical_source: {source_url}");
+    let _ = writeln!(output, "  selected_rocm_version: {version}");
+    let _ = writeln!(output, "  build_date: {build_date}");
+    let _ = writeln!(output, "  source_layout_generation: {layout_generation}");
+}
+
+fn validate_aggregate_index_layout(html: &str) -> Result<()> {
+    let required_packages = ["rocm/", "torch/", "torchvision/", "torchaudio/"];
+    if required_packages
+        .iter()
+        .all(|package| html.contains(package))
+    {
+        return Ok(());
+    }
+    bail!(
+        "unknown canonical TheRock aggregate index layout: expected package links for rocm, torch, torchvision, and torchaudio"
+    )
 }
 
 #[derive(Debug, Clone)]
@@ -849,7 +897,14 @@ fn install_wheel_runtime(
         output,
         "  summary: rocm-cli will install the ROCm SDK and matching PyTorch packages for this Python and operating system"
     );
-    let _ = writeln!(output, "  channel: {}", channel.as_str());
+    let source = channel.canonical_source();
+    render_canonical_provenance(
+        &mut output,
+        channel,
+        source.wheel_index,
+        source.layout_generation,
+        &resolution.latest_version,
+    );
     let _ = writeln!(output, "  format: wheel");
     if let Some(selector) = version_selector {
         let _ = writeln!(output, "  requested: {}", selector.describe());
@@ -1062,7 +1117,14 @@ fn install_tarball_runtime(
 
     let mut output = String::new();
     let _ = writeln!(output, "sdk install");
-    let _ = writeln!(output, "  channel: {}", channel.as_str());
+    let source = channel.canonical_source();
+    render_canonical_provenance(
+        &mut output,
+        channel,
+        source.tarball_catalog,
+        source.layout_generation,
+        &artifact.version,
+    );
     let _ = writeln!(output, "  format: tarball");
     let _ = writeln!(output, "  family: {}", artifact.family);
     let _ = writeln!(output, "  family_source: {}", artifact.family_source);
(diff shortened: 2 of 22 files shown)


apps/rocm/src/therock.rs
```diff
@@ -1143,33 +1205,44 @@ fn resolve_pip_runtime_with_timeout(
...
+        channel,
+        &family_resolution,
+        source.wheel_index,
+        wheel_compatibility,
+        version_selector,
+        download_timeout_secs,
     )
+    .with_context(|| {
+        format!(
+            "failed to resolve TheRock {} wheel runtime from canonical source {}\n\n{}",
```
fredespi, line 1711: The wheel index is now family-independent, but the failure guidance still tells the user their family is the problem.

`resolve_pip_runtime_from_index` is now called with `source.wheel_index` and no `/<family>` segment, so `family_resolution` no longer influences which wheels are found — it is only carried into the result for display. When resolution fails, this context still appends `family_resolution_hint(...)`, which for an auto-detected family says:

```
Re-run with an explicit package family: `rocm install sdk --family <FAMILY>`.
Recognized families: ...
```

On the wheel path that advice can no longer change anything: every `--family` value now yields the identical index URL and the identical resolution, so a user who follows it gets the same error a second time.

Two related spots while you are here:

- the dry-run still prints `family:` / `family_source:` immediately under `canonical_source:` (line 904), which reads as though the family scoped the selection;
- `runtime_key(channel, "wheel", &resolution.family, ...)` (line 877) still keys the managed venv by family, so `--family A` and `--family B` on one host now build two separate multi-gigabyte environments with identical content.

Suggestion: for `format == "wheel"`, drop the family half of the hint and keep the `--channel` / `--format` half, which is still actionable — then decide deliberately whether the family should stay in the wheel runtime key.

apps/rocm/src/therock.rs
```diff
@@ -112,14 +119,55 @@ impl TheRockChannel {
...
+    source_url: &str,
+    layout_generation: &str,
+    version: &str,
+) {
+    let build_date = runtime_version_build_date(version)
+        .unwrap_or_else(|| "<not published by canonical source>".to_owned());
+    let _ = writeln!(output, "  channel: {}", channel.as_str());
+    let _ = writeln!(output, "  canonical_source: {source_url}");
+    let _ = writeln!(output, "  selected_rocm_version: {version}");
+    let _ = writeln!(output, "  build_date: {build_date}");
```
fredespi, line 155: `build_date: <not published by canonical source>` will print on every release-channel install.

`runtime_version_build_date` only returns a date when the version string embeds eight date digits. Nightly versions do (`10.1.0a20260822`); stable release versions never do (`7.14.0`). So every `--channel release` preview reports `<not published by canonical source>`, which reads as a provenance gap at the source when nothing is actually missing — a stable version simply does not carry a build date.

It also contradicts the convention already set in this file: `runtime_version_display("7.14.0")` returns plain `7.14.0` and just omits the `(build ...)` suffix when there is no embedded date (see `runtime_version_display_mentions_embedded_build_date`).

Suggestion: omit the `build_date` line when there is no embedded date, or use neutral wording such as `not encoded in this version`.

apps/rocm/src/therock.rs
```diff
@@ -27,11 +27,11 @@ use std::path::{Path, PathBuf};
 use std::process::{Command, Output, Stdio};
 use std::time::Duration;
 
-const THEROCK_NIGHTLY_PIP_INDEX_BASE: &str = "https://rocm.nightlies.amd.com/v2";
-const THEROCK_RELEASE_PIP_INDEX_BASE: &str = "https://repo.amd.com/rocm/whl";
-const THEROCK_RELEASE_PIP_MULTI_ARCH_INDEX_BASE: &str = "https://repo.amd.com/rocm/whl-multi-arch";
+const THEROCK_NIGHTLY_PIP_INDEX_BASE: &str = "https://rocm.nightlies.amd.com/whl-multi-arch";
+const THEROCK_RELEASE_PIP_INDEX_BASE: &str = "https://repo.amd.com/rocm/whl-multi-arch";
```
fredespi, line 32: Not about this line specifically — anchored here because `MANIFEST.md` is not part of the diff.

`MANIFEST.md` lines 649-650 still document the endpoints this PR removes:

```
Release channel wheels are served from `https://repo.amd.com/rocm/whl/<gpu-family>/`.
Nightly channel wheels are served from `https://rocm.nightlies.amd.com/v2/<gpu-family>/`.
```

Both are now wrong in host and in shape — the family segment is gone. That section is the repository's statement of what the CLI downloads at runtime, so it is worth keeping exact. The tarball paragraph directly below it is still correct and needs no change.

tests/e2e-cucumber/features/runtime_setup.feature
```diff
@@ -42,3 +42,8 @@ Feature: Runtime configuration
     When the user tries to adopt the existing install
     Then the adoption is refused
     And the error explains which install types can be adopted
+
+  @id:runtime-resolve-canonical-nightly
```
fredespi: This scenario carries no capability tag, so it runs in the blocking mock `E2E tests` job — and it needs live network access to pass. `install sdk --dry-run` still performs the full resolution: the new root layout fetch plus four simple-index fetches.

Every other scenario on that lane is served locally or by the in-process mock server. This one makes an external service's availability and layout a merge gate for the whole repository, so an outage or a layout change lands as a red required check on unrelated PRs rather than on the change that needs to know.

It does pass today — I confirmed the full nightly stack resolves for cp312 on Linux and Windows from the aggregate index. The concern is the standing coupling, not the current result.

Options: tag it for a lane where a live-stream dependency is expected, or split it so the blocking lane asserts the rendered provenance against a locally served index and only the tagged lane touches the real stream.

tests/e2e-cucumber/features/runtime_setup.feature
```diff
@@ -42,3 +42,8 @@ Feature: Runtime configuration
...
+  Scenario: 4 - Previewing a nightly SDK install reports canonical provenance
```
fredespi: Duplicate scenario number: `Scenario: 4 - Reinstalling the SDK leaves the installed engine's requirements satisfied` is already number 4 in this file (line 30). The numbers are used for name filtering (`cargo xtask e2e -- -n ...`) and in the report grid, so this one probably wants to be 5.

fredespi (changes requested):
I reviewed this against the current head (69ee614) and checked the new canonical streams directly rather than trusting the constants: both aggregate indexes serve the expected `rocm/`, `torch/`, `torchvision/` and `torchaudio/` package pages, and a resolver run against each one from a clean environment produces a complete, mutually consistent stack — release and nightly, Linux and Windows, without any family segment and without needing an `amd-torch-device-*` package. Device coverage on the aggregate stream matches what the old family indexes offered. So the central move is sound, and deleting the dead `whl/<family>` fallback is a real improvement.

Five things I would fix before merge. None of them break resolution; three are user-visible and two are test hygiene.

The one that matters most: the wheel index is now family-independent, but the surrounding code still treats the family as the thing that selects wheels. On a resolution failure the CLI still tells the user to re-run with an explicit `--family`, which can no longer change the outcome, and the managed runtime is still keyed by family, so two family values now produce two identical multi-gigabyte installs. Details inline.

Also worth noting for the record: the new scenario is untagged, so it puts a live dependency on the nightly index into the blocking merge gate for the whole repository.

A few things I suspected and then ruled out, so they do not need your time: the metadata signature policy is inert by default (the pinned key is empty), so the new root fetch does not require a new `.sig` sidecar for anyone who has not opted in; the aggregate stream is not missing older families; and the multi-arch payload is not larger than the family-specific one it replaces.

apps/rocm/src/therock.rs
michaelroy-amd: Fixed in 05edff0c. Wheel failures now use aggregate-specific guidance without `--family`; dry-run labels the value as `target_family`; and managed wheel keys are now `channel-wheel-multi-arch-version`, preventing duplicate aggregate environments across family values.

apps/rocm/src/therock.rs
michaelroy-amd: Fixed in 05edff0c. Stable versions now report `build_date: not encoded in stable version`; dated nightly versions continue to report the parsed date.

apps/rocm/src/therock.rs
michaelroy-amd: Fixed in 05edff0c. `MANIFEST.md` now documents the canonical release and nightly `whl-multi-arch` endpoints without family segments.

tests/e2e-cucumber/features/runtime_setup.feature
michaelroy-amd: Fixed in 05edff0c. The live canonical-stream scenario is now tagged `@nightly`, so it no longer adds an external network dependency to the blocking mock lane.

tests/e2e-cucumber/features/runtime_setup.feature
michaelroy-amd: Fixed in 05edff0c. The canonical provenance scenario is now numbered 5.

tests/e2e-cucumber/features/runtime_setup.feature
```diff
@@ -57,3 +57,8 @@ Feature: Runtime configuration
     When the user tries to adopt the existing install
     Then the adoption is refused
     And the error explains which install types can be adopted
+
+  @id:runtime-resolve-canonical-nightly @nightly
+  Scenario: 5 - Previewing a nightly SDK install reports canonical provenance
```
fredespi: This is again a duplicate `Scenario: 5`. Commit `05edff0c` correctly changed this scenario from 4 to 5, but the subsequent merge brought in `The update check reports the active runtime's freshness` as scenario 5 at line 45. Scenario numbers are used in name filters and report output, so please renumber this new canonical-provenance scenario to 6.

fredespi (changes requested):
I re-reviewed the two follow-up commits against the full current PR. The five issues from my previous review are addressed: wheel guidance and runtime identity now match the aggregate stream, stable provenance wording is neutral, the manifest is current, and the live scenario is isolated to the nightly lane. The merge from main introduced one small numbering regression that still needs correction before merge; details inline.

tests/e2e-cucumber/features/runtime_setup.feature
michaelroy-amd: Fixed in 0eefc897. The canonical-provenance scenario is now numbered 6. The same commit also activates the refreshed canonical shared runtime in isolated GPU scenarios, addressing the multi-runtime CI failure exposed by the new aggregate key.

fredespi:
Re-review complete at `0eefc897`. The duplicate scenario number is corrected: canonical provenance is now scenario 6. I also reviewed the added shared-runtime activation path and found no blocking issue in this follow-up. The PR looks safe to approve after final CI and runtime validation; this comment is not a formal approval.

xtask/src/e2e_prewarm.rs
```diff
@@ -232,6 +232,8 @@ pub fn run(channel: &str, keep: usize, prewarm_dir: &Path) -> Result<()> {
         }
     }
 
+    ensure_default_engine(&rocm, prewarm_dir)?;
```
fredespi, line 531: This call is unreachable when the runtime is reused: `Decision::Reuse` returns at line 231. That is the important repair case here. The previous run already updated the persistent shared tree to `release-wheel-multi-arch-7-14-0` without vLLM; on the next run `rocm update` reports that runtime as `up_to_date`, `decide` returns `Reuse`, and pre-warm exits before installing the engine, so the same serve scenarios still fail. Please arrange for the default-engine check/install to run for a reused current runtime too (while keeping the intended degraded-index behavior), and cover the reuse path.

fredespi (changes requested):
I reviewed only the new pre-warm commit against the previously cleared head. Scenario numbering remains correct, and the shared-runtime activation path is unchanged. One control-flow issue prevents this commit from repairing the persisted runtime state it targets; details inline.

fredespi:
Re-review complete at `74418ccf`. The prior reuse-path issue is fixed: `Decision::Reuse` now runs `ensure_default_engine` before returning, so an existing current shared runtime is repaired too. Scenario numbering remains unique through scenario 6, and the shared-runtime activation refactor preserves the behavior cleared earlier. Focused `xtask` engine-inventory and `e2e-cucumber` canonical-runtime-key tests pass. This head looks safe to approve after final self-hosted CI and runtime validation; this comment is not a formal approval.

apps/rocm/src/therock.rs
```diff
@@ -1014,7 +1068,10 @@ fn install_wheel_runtime(
 
 fn therock_pip_package_specs(package_versions: &TheRockPipPackageVersions) -> Vec<String> {
     vec![
-        format!("rocm[libraries,devel]=={}", package_versions.rocm),
+        format!(
+            "rocm[libraries,devel,device-all]=={}",
```
fredespi: Adding `device-all` changes the required runtime contents, but it does not change the runtime identity or freshness decision. The persistent MI300X tree already contains `release-wheel-multi-arch-7-14-0` without device payloads. Its manifest still says version `7.14.0`, and `runtime_update_plan` compares only that version with the index, so the next pre-warm reports `up_to_date`, takes `Decision::Reuse`, and only ensures vLLM; it never reruns this SDK install. The same cached environment therefore remains without the payload needed to fix the observed `Failed to infer device type` failures. Please add a repair/migration for same-version runtimes, or include the package/layout composition in runtime identity and freshness, with coverage for an existing old-composition cache.

fredespi (changes requested):
I reviewed only the new aggregate-device-payload commit against the head cleared in the last round. The reuse-path engine repair, unique scenario numbering, and shared-runtime activation remain intact. One cache-identity issue means this commit does not repair the persistent runtime that exposed the missing device payload; details inline. This is now an accumulating design problem rather than another isolated test slip: the persistent runtime cache is version-keyed, while successive fixes change the required package composition without changing that identity.

Separately, `device-all` installs every device wheel available for the current OS, and individual published device wheels can be hundreds of megabytes to over 1.5 GB. If the generic or target-specific device extra can satisfy the detected family, it would avoid a very large managed environment; please confirm that all payloads are truly required.

apps/rocm/src/therock.rs
```diff
@@ -822,16 +870,16 @@ fn install_wheel_runtime(
         &wheel_compatibility,
         version_selector,
     )?;
+    let device_target = aggregate_device_target(&resolution.family);
+    let rocm_extras = device_target.as_deref().map_or_else(
```
fredespi: This improves fresh installs by selecting `device-gfx942` instead of every device payload, but it still changes only the install arguments. `wheel_runtime_key` remains `channel + version`, `InstalledRuntimeManifest` records no package/layout composition, and `runtime_update_plan` still compares only `manifest.version` with `latest_version`. Therefore the persistent `release-wheel-multi-arch-7-14-0` that lacks any device payload still reports `up_to_date`; pre-warm takes `Reuse`, only ensures the engine, and never executes this target-specific SDK install. No migration or invalidation path was added. Please either include the required composition in runtime identity/freshness, or explicitly detect and repair same-version runtimes, with a regression test starting from the old 7.14.0 cache shape.

fredespi (changes requested):
I re-reviewed the cache design at the full current head, not only the latest patch. The reuse-path engine repair, scenario numbering, and shared-runtime activation remain intact. However, the round-6 cache-identity issue is **not resolved**: this commit narrows the payload chosen for a new install, but it still does not invalidate or repair an existing same-version runtime. Details inline.

engines/vllm/src/lib.rs
```diff
@@ -1387,6 +1387,9 @@ print(json.dumps({"present": spec is not None, "version": version}))
 
 fn apply_therock_env(command: &mut ProcessCommand, runtime: &VllmRuntime) -> Result<()> {
     command.env("VLLM_TARGET_DEVICE", "rocm");
+    if let Some(target_family) = therock_device_target(runtime) {
```
fredespi: This is a launch-time workaround that only runs when `sdk_library_paths` already contains an `_rocm_sdk_device_*` directory. The stale persistent 7.14.0 runtime lacks that device package, so `therock_device_target` returns `None` and this adds nothing. More importantly, the actual cache path is unchanged: `runtime_update_plan` compares only the version, reports `up_to_date`, `e2e-prewarm` takes `Decision::Reuse`, runs only `ensure_default_engine`, and returns. No SDK reinstall, migration, invalidation, or re-key occurs, so this commit cannot repair the old cache. Please make composition participate in runtime identity/freshness, or explicitly detect and repair the same-version runtime before reuse.

fredespi (dismissed):
The round-6/7 cache-identity issue is **STILL OPEN** at this head. I traced the complete existing-cache path again. The new launch environment handling does not migrate, invalidate, repair, or re-key a same-version runtime. The previously cleared engine-prewarm repair, unique scenario numbering, and shared-runtime activation remain intact.

apps/rocm/src/therock.rs
michaelroy-amd: Fixed in 9f63cfb5. device-all is removed. Aggregate installs require one target-specific device extra (including gfx943 -> device-gfx942); absent or mismatched exact detection now fails clearly instead of installing every published payload.

engines/vllm/src/lib.rs
michaelroy-amd: Addressed at the cache boundary in 9f63cfb5. The launch variable remains necessary after a target package exists, but it is no longer treated as the migration: an old same-version manifest now becomes repair_available and installs a composition-keyed replacement before reuse.

apps/rocm/src/therock.rs
michaelroy-amd: Fixed in 9f63cfb5. Managed wheel manifests now record source-layout generation plus exact pinned package specs. Freshness reports repair_available when either composition or its composition-keyed runtime identity is stale. Applying that repair creates and activates a side-by-side runtime selected by its exact target key. Regression coverage starts from the old 7.14.0 manifest shape with no composition.

xtask/src/e2e_prewarm.rs
michaelroy-amd: Fixed in 9f63cfb5. The decision match now joins at one shared ensure_default_engine call, so Reuse cannot return before engine validation. Repair is a distinct decision path and both reuse and repair routing are covered.

juhovainio:
## Conflict check against #329

Verified via a real trial merge (`git merge --no-commit --no-ff`, not just diffing) that this PR conflicts with #329 in 4 files: `apps/rocm/src/therock.rs` (14 hunks), `docs/ci-hardware-testing.md`, `tests/e2e-cucumber/src/capability.rs`, and `tests/e2e-cucumber/tests/e2e/runtime_steps.rs`. (`comfyui.rs`, `main.rs`, `storage.rs` are also touched by both but auto-merge cleanly.)

The two PRs are not solving the same problem, but they rewrite the same seam of code:

- **This PR** hardens resolution against the *existing* release/nightly layout: pick one canonical aggregate index/tarball URL per channel (drops the old multi-candidate-URL fallback list), validates the index has the expected `rocm/torch/torchvision/torchaudio` layout, switches package specs from grouped-family extras (`rocm[libraries,devel]`) to a single detected device-target extra (`rocm[libraries,devel,device-gfx942]`), and adds a composition-aware runtime key + repair mechanism for stale caches.
- **#329** adds a brand-new ROCm 10 "next" layout at different URLs (`stable.repo.amd.com`), dispatched by generation, explicitly designed so legacy resolution stays byte-for-byte unchanged.

Whichever of these merges second will need a real rebase, not a mechanical conflict resolution: this PR already changes legacy's package-spec shape, URL candidate list, and runtime-key scheme, which #329's "legacy stays untouched" premise assumes doesn't happen. The `CanonicalSource`/`canonical_source()` model here and `TheRockIndexGeneration`/`env_override_base()` model in #329 will need to be reconciled into one design, not textually merged.

Flagging early so whoever rebases second isn't surprised by the depth of the conflict.

tomastola (commented):
Not a full review of the canonical-stream work — this is findings from debugging #314's red GPU lane that turned out to be about this PR, plus one concrete rebase instruction. Two are actionable here; two are evidence that the design in this PR is right.

## 1. This PR's hardware runs contaminate the shared pre-warm cache for every other PR

This is the one worth acting on. It is measured, not inferred, and it cost about a day of red lanes on #314.

The self-hosted GPU runners keep the pre-warm runtime tree in retained shared storage across jobs. This branch can compose a runtime `main` cannot, and it leaves it behind. Registry entries found in the shared cache:

```
release-wheel-gfx94x-dcgpu-7-13-0  version 7.13.0  index_url https://repo.amd.com/rocm/whl/gfx94X-dcgpu
release-wheel-gfx94x-dcgpu-7-14-0  version 7.14.0  index_url https://repo.amd.com/rocm/whl-multi-arch
```

That second `index_url` cannot be produced by `main`: `therock_index_urls` builds `whl/{family}` and `whl-multi-arch/{family}`, the latter 403s, so the newest version `main` can reach for this family is 7.13.0. The flat `whl-multi-arch` path is reachable only from this branch, which redefines `THEROCK_RELEASE_PIP_INDEX_BASE` to it.

The reason it then sticks is a design decision in `xtask/src/e2e_prewarm.rs::decide` — and it is doing exactly what it says:

```rust
// `ahead_of_index` means the installed runtime is NEWER than anything the
// index offers (a hand-placed or pinned build). Reuse it rather than
// "updating" backwards.
```

7.14.0 is ahead of `main`'s reachable index, so every subsequent lane — running `main`'s code, which cannot compose a multi-arch runtime — resolves `Decision::Reuse` onto it and inherits a runtime it has no way to complete. On one runner the active-runtime pointer had been left on 7.14.0 and on the other on 7.13.0, so the failure looked intermittent and PR-dependent when it was neither.

Note the fence here is **one-way**. `source_layout_generation` is hashed into the composition key (`therock.rs:3911`), so *this* branch correctly refuses to reuse a `main`-composed runtime. Nothing protects the other direction: `main` has no notion of layout generation and reuses anything newer. So the identity work in this PR, which is right, does not remove the hazard — it removes it only for this PR's own runs.

Worth fixing before the full hardware matrix runs again. Options, roughly in order of how much I like them:

- Give the pre-warm cache an identity that includes the layout generation (install root or `runtime_key`), so the two generations coexist instead of one shadowing the other. This also fits the round-6/7 cache-identity thread already open on this PR.
- Or fence `decide`: treat `ahead_of_index` as reusable only when the installed runtime's `index_url` is one this build can itself reach, and install alongside otherwise.
- Or isolate the pre-warm cache per branch. Simplest, but gives up the warm-tree coverage that the reuse path exists to provide.

I cleaned the 7.14.0 runtime out of the shared cache to unblock #314 (details in ROCm/rocm-cli#314). `main` cannot re-create it, so that is not a treadmill — but it will come straight back the next time this PR's matrix runs.

## 2. `settle_engine_install` here still has the bug #314 just fixed — drop this copy on rebase

At `c0c4409`, `settle_engine_install(paths, engine, runtime_key, response)` derives the target torch build and the device-probe library paths from a **caller-supplied selector**, but installs into the **environment's own** interpreter. Three of the four call sites (`main.rs:3802`, `4011`, `4329`) pass a `runtime_id`, and `runtime_id` is shared by every side-by-side install of one channel and family:

```
runtime_key: release-wheel-gfx94x-dcgpu-7-13-0   version 7.13.0
runtime_key: release-wheel-gfx94x-dcgpu-7-14-0   version 7.14.0
runtime_id : therock-release:gfx94X-dcgpu        <- both
```

So with two runtimes installed side by side the selector can name a different runtime than the one the engine env lives in, and the alignment installs one tree's SDK torch build into the other tree's python. Side-by-side is not an edge case here — the pre-warm's `Decision::Update` path installs alongside on purpose.

#314 fixes this by resolving the owning runtime from `response.python_executable` (longest containing `install_root`, compared both verbatim and canonicalized, since the CLI stores `install_root` canonicalized while an engine adapter echoes back whatever path it was handed). Selector stays as the fallback for interpreters outside every install root.

Concretely, when rebasing this PR onto the resulting `main`: **drop this PR's torch-alignment commits rather than merging them**, and keep #314's `settle_engine_install`. The two implementations collide across `main.rs`, `therock.rs` and `engines/vllm/src/lib.rs`, and this one lacks the ownership resolution.

## 3. Supporting evidence that the composition work in this PR is the right fix

Measured on hardware while debugging the above, comparing the two published `torch 2.11.0` wheels:

| wheel | `.hip_fatbin` | `libtorch_hip.so` | offload targets | device check |
|---|---|---|---|---|
| `+rocm7.14.0` (multi-arch) | `NOBITS`, 0 bytes | 136 MB | none | `hipErrorInvalidImage` |
| `+rocm7.13.0` (family index) | `PROGBITS` | 188 MB | gfx942 | usable, 1 device |

The multi-arch wheel is not corrupt and 7.14 is not a bad build — on-disk size matches the published wheel byte for byte, and `+rocm7.13.0` under `whl-multi-arch` is stripped identically. Multi-arch wheels ship without device code by design; kernels live in `.kpack` archives carried by the aggregate device package. Generalised: **resolving a version published only in the multi-arch channel into a family-specific tree always yields a kernel-less torch**, and the tree that had it was carrying the family-specific `_rocm_sdk_libraries_gfx94X_dcgpu` (no `.kpack`) rather than the aggregate.

That is precisely the failure `WheelRuntimeComposition` / `required_aggregate_device_target` / the `rocm[libraries,devel,device-{target}]` extras exist to prevent, so this is a concrete argument for the approach, not against it.

It also settles an earlier reading of the red MI300X lane as "reported target is gfx943 and no compatible published device package is available". That is not what was happening — the target resolved fine; the failure was a kernel-less torch from a cross-composed tree. `therock.rs:4699` already covers `required_aggregate_device_target(Some("gfx943"), "gfx94X-dcgpu")`.

## 4. Known gap neither PR closes

`engines/vllm/src/lib.rs` still builds the env id from `runtime_id`:

```rust
env_id = format!("external-vllm-{}", stable_id_component(runtime_id))
```

which drops the version, so both side-by-side runtimes map to one env id and the registry keeps whichever registered first. The **active** runtime is therefore not necessarily the one an engine uses. #314 makes the used runtime work; it does not make the right runtime get used, and this PR does not change it either. Not a blocker for either — flagging it so it does not get assumed fixed. Happy to take it as a separate change.

apps/rocm/src/therock.rs
```diff
@@ -822,17 +979,15 @@ fn install_wheel_runtime(
...
     ));
-    let runtime_key = runtime_key(
-        channel,
-        "wheel",
-        &resolution.family,
-        Some(&resolution.latest_version),
-    );
-    let install_root = resolved_install_root(paths, "wheel", &runtime_key, prefix);
+    let runtime_key = wheel_runtime_key(channel, &resolution.latest_version, &wheel_composition);
+    let install_root = prefix.unwrap_or_else(|| managed_runtime_root(paths, "wheel", &runtime_key));
```
rominf: This drops the symlink resolution that #317 added, and I think unintentionally — the tarball path at `:1263` still calls `resolved_install_root(paths, "tarball", &runtime_key, prefix)`, so only the wheel side changed.

`main:843` was:

```rust
let install_root = resolved_install_root(paths, "wheel", &runtime_key, prefix);
```

which is this same expression plus `rocm_core::resolve_path_through_symlinks(&requested)`. `resolved_install_root` is still defined on this branch at `:4028` and its doc comment still explains exactly why the wheel path needs it:

> `install sdk` writes this path into three places that all outlive the command: the registry manifest, the sidecar beside the runtime, and — via `uv` — the `#!` line of every console script in the venv.

`git merge-base --is-ancestor ff78075b pr/308` confirms #317 is in this branch's base, so this reverts it rather than predating it. The consequence is the original bug: reaching `data/runtimes` through a link bakes the link path into the manifest and every console-script shebang, so removing the link leaves the runtime claiming to live somewhere that isn't there while the files sit next door.

That's not hypothetical here — the e2e harness creates that link for the shared pre-warm tree, and scenario 8 (`runtime-install-records-the-real-folder`) exists to pin this behavior.

Restoring `resolved_install_root(paths, "wheel", &runtime_key, prefix)` should be all it needs.

apps/rocm/src/therock.rs
```diff
@@ -822,17 +979,15 @@ fn install_wheel_runtime(
         &wheel_compatibility,
         version_selector,
     )?;
+    let device_target =
+        required_aggregate_device_target(detect_host_gfx_target().as_deref(), &resolution.family)?;
```
rominf: This gate runs unconditionally, before the `if dry_run` early return at `:1055`, and `required_aggregate_device_target` errors on `None`:

```rust
let detected = detected.with_context(|| {
    format!("unable to detect an exact GPU target for canonical aggregate family `{family}`; refusing to install every device payload")
})?;
```

So `rocm install sdk --dry-run` now fails on any host without a detectable AMD GPU. Two scenarios in this PR's feature file depend on it succeeding:

- **Scenario 8** (pre-existing, `runtime-install-records-the-real-folder`) runs `install sdk --family gfx110X-all --dry-run` through `run_rocm_ok` (asserts rc==0). Its step comment says why: *"`--family` because a host with no AMD GPU has no target to detect, and the preview resolves the install folder before it needs one."* That premise is what this gate removes.
- **Scenario 6** (new, `runtime-resolve-canonical-nightly`) has the same shape with `--family gfx120X-all`, also via `run_rocm_ok`, and carries no `@requires-gpu` — so it's scheduled on hosts that can't satisfy it. On the gfx94X GPU lane it fails the other way, since `canonical_aggregate_device_target` returns `None` when the detected chip doesn't normalize to the requested family.

Neither ran: both are `@nightly` and the failing run had `E2E_INCLUDE_NIGHTLY:` empty.

The `--family` case seems like the more interesting one. A user passing `--family` explicitly has stated their target, but `resolve_family` returns it from the override branch and then this gate re-derives the device target from host detection anyway, so the flag doesn't get them past it. On a preview there's nothing to protect — no payload is downloaded — so skipping the requirement on the `dry_run` path and rendering something like `device_target: <undetermined>` would keep both scenarios honest. For the real install path, falling back to the family's device set when `--family` is supplied would be the equivalent of what #272 does (see the last comment).

apps/rocm/src/main.rs
```diff
@@ -7520,21 +8182,40 @@ fn engine_dependency_check(
...
         }
         Err(error) => EngineDependencyCheck::NotVerified(error.to_string()),
     }
 }
 
+/// Separate a deliberate divergence from a genuine violation.
+///
+/// Only a divergence about the realigned package is expected; anything else in
+/// the same run is still a violation and must keep saying so.
+fn classify_dependency_details(
```
rominf, line 8819: I think this is what's red on the GPU lane, and it looks like a normal case rather than an edge one.

`details.iter().all(|d| d.contains(&marker))` only tolerates the divergence when *every* violation names the realigned package. But realigning torch drags its stack along, so a second violation is the expected outcome, not an anomaly. From run 33096888884:

```
violation: The package `vllm` requires `torch==2.11.0+gitd0c8b1f`, but `2.11.0+rocm7.14.0` is installed
violation: The package `vllm` requires `triton==3.6.0`, but `3.7.1+git0263a6a6.rocm7.14.0` is installed
dependency_check: violated
device_check: no_devices
 1: rocm::unusable_runtime_error
```

The torch line is the deliberate realignment this code exists to tolerate; the triton line is its consequence. Because `all()` sees the triton detail, the whole run is classified `Violated`, the user is advised to run `rocm engines install vllm --reinstall` (which would undo the alignment), and `install_left_runtime_unusable` then exits 1.

Partitioning instead of requiring `all()` would fit the doc comment's intent better — expected-divergence details for the realigned package and the packages that pin against it, genuine violations for everything else — and the `--reinstall` action line should be suppressed whenever any detail is an expected divergence, since following it makes things worse.

(The `no_devices` half of this is the selector/`env_path` mismatch @tomastola described at `runtime_id: release-wheel-multi-arch-7-14-0-…` vs `env_path: …/release-wheel-gfx94x-dcgpu-7-13-0/bin`, which I gather #314 fixes — flagging only the classification half here.)

apps/rocm/src/main.rs
```diff
@@ -14511,8 +15209,8 @@ fn apply_runtime_update(
         false,
     )?;
     let manifests_after = therock::load_runtime_manifests(paths)?;
```
rominf, line 15876: `apply_runtime_update` predicts a runtime key with one family, then installs with another.

`plan.target_runtime_key` comes from `runtime_update_plan` → `resolve_latest_for_manifest` (`therock.rs:734`), which passes `Some(manifest.family.as_str())` as the family override. But the apply path just above calls `install_sdk(paths, &source.channel, &source.format, None, None, None, false)` — `family_override = None`. So the install re-derives the family from `resolve_family`'s detection chain, and the family feeds the `device-<target>` extra, which feeds the composition, which feeds the key.

What makes that fatal now rather than merely surprising is the lookup narrowing on the next line. `main` matched on channel/format/family/version:

```rust
manifest.channel == source.channel && manifest.format == source.format
    && manifest.family == source.family && manifest.version == latest_version
```

This PR replaces it with an exact `manifest.runtime_key == target_runtime_key`. Any divergence between predicted and actual now aborts with *"runtime install completed but the target runtime manifest was not found"* — after a multi-gigabyte install has already succeeded and been written to disk.

The dry-run branch a few lines up has the same `None`, so the preview describes a different install than the apply performs.

Passing `Some(&source.family)` on both branches would keep the planned and applied keys in agreement. Worth doing regardless of whether a divergence is reachable today — `xtask e2e-prewarm`'s new `Decision::Repair` drives exactly this path, and the failure mode is expensive.

apps/rocm/src/therock.rs
```diff
@@ -1012,15 +1180,60 @@ fn install_wheel_runtime(
...
+        )
+    })?;
+    canonical_aggregate_device_target(detected, family).with_context(|| {
+        format!(
+            "detected GPU target `{detected}` does not match resolved family `{family}`; refusing to install every device payload"
+        )
+    })
+}
+
+fn canonical_aggregate_device_target(detected: &str, family: &str) -> Option<String> {
```
rominf: Only `gfx94*` is canonicalized to a published package name; everything else passes the raw detected chip straight through as `device-<target>` (`:1203`).

That's fine for the single-chip families, but `normalize_therock_family` maps four chips into `gfx110X-all` (`gfx1100`/`1101`/`1102`/`1103`) and two into `gfx120X-all` (`gfx1200`/`1201`). So a gfx1103 host resolves family `gfx110X-all` and then requests `rocm[libraries,devel,device-gfx1103]`. Whether that resolves depends on the index publishing a device payload per detected chip rather than per family bucket — and if it doesn't, it's an install-time resolution failure, not a warning.

I couldn't verify the index contents from here (the endpoint isn't reachable from my environment), so I can't tell you whether `device-gfx1103` exists today. But #272 came to a different conclusion about this: it added `known_therock_family_device_chips` in `rocm-core` mapping `gfx110X-all -> [gfx1100, gfx1101, gfx1102, gfx1103]` and requests *all* of a family's chips, with a `device-all` fallback for unmapped families.

Either the two PRs disagree about the index layout, or one of them is carrying a workaround the other doesn't need — worth settling explicitly rather than by merge order. The safest local version would be to verify the requested `device-<target>` extra exists in the index before pinning it, so a wrong guess produces a clear message instead of a resolver error.

apps/rocm/src/therock.rs
```diff
@@ -561,46 +666,76 @@ pub(crate) fn render_update_report(paths: &AppPaths) -> Result<String> {
...
-                "    activate: add `--activate` to make the newly installed runtime the default after install"
+                "    activate: add `--activate` to make the installed runtime the default after install"
             );
         }
     }
 
     Ok(output)
 }
 
+fn runtime_freshness(
```
rominf, line 904: Runtimes installed before this PR look permanently repairable, and the daemon re-notifies about them every six hours.

A manifest written by `main` has `wheel_composition: None`, so at equal versions the `Ordering::Equal` guard's `manifest.wheel_composition.as_ref() != required_composition` is always true and freshness is `RepairAvailable`. The repair then installs a *new* composition-keyed runtime alongside rather than amending the old manifest, so the old one keeps reporting `RepairAvailable` on every subsequent check.

That state reaches the daemon: `rocmd/src/lib.rs:1229-1231` folds `repair_available` into `update_output_reports_update_available`, and `:3861` calls `record_update_available_notification` whenever it's true, which emits *"A ROCm runtime repair is available because its package composition changed"* — every `THEROCK_UPDATE_INTERVAL_MS` (6h), forever, for a repair the user already performed.

`xtask/src/e2e_prewarm.rs::decide` was given an explicit `up_to_date`-wins guard (`:124`) which suggests this state is known-reachable; the daemon has no equivalent.

Suppressing `RepairAvailable` when a sibling runtime of the same channel/format/family already carries the required composition would cover both. Backfilling legacy manifests would work too, but the sibling check also handles the case where the user repaired and kept the old runtime deliberately.

apps/rocm/src/therock.rs
```diff
@@ -1012,15 +1180,60 @@ fn install_wheel_runtime(
...
     vec![
-        format!("rocm[libraries,devel]=={}", package_versions.rocm),
+        format!("rocm[{rocm_extras}]=={}", package_versions.rocm),
         format!("torch=={}", package_versions.torch),
         format!("torchvision=={}", package_versions.torchvision),
         format!("torchaudio=={}", package_versions.torchaudio),
     ]
 }
 
+fn wheel_runtime_composition(
```
rominf, line 1510: Not about this function specifically — anchoring here because it's where the three PRs collide. Raising it because it isn't visible from inside any one of them.

**#329** builds a two-generation index model: `TheRockIndexGeneration{Legacy, Next}`, `whl-next` plus a second release host, `raw_arch` plumbing, and per-generation cache-key namespacing. It keeps `therock_index_urls()` and adds candidate ordering on top.

**#308** goes the other way: one hardcoded `CanonicalSource` per channel with a constant `layout_generation`, and removes `therock_index_urls()` entirely.

These aren't mergeable in sequence — whichever lands second has to be substantially rewritten against the other's model, and a clean textual rebase would produce something neither author designed.

**#272** overlaps differently: it makes the *identical* `therock_pip_package_specs(package_versions, rocm_extras)` signature change this PR makes, with a competing extras policy (see the `device-<target>` comment above). Same root cause, two incompatible fixes; these will conflict semantically even if git merges them cleanly.

Probably worth a maintainer picking the index model first and having the other two rebase onto it, rather than resolving it at merge time. Happy to be told this is already settled somewhere I haven't seen.

rominf (dismissed):
I looked at this mainly from the resolution-correctness angle, since @fredespi and @tomastola have already covered the pre-warm cache identity and the `settle_engine_install` selector issues in depth. Version handling is the part I'd most expected to find problems in and it's genuinely solid: `parse_version` strips `+local`, `ParsedVersion`/`VersionStage` derive `Ord` with `Alpha < Rc < Stable`, and `compare_version_strings` only falls back to byte comparison when both sides fail to parse. "Latest" is a descending sort with a per-candidate torch/torchvision/torchaudio completeness check, and Release additionally filters to `is_stable_runtime_version`. That's semver, not lexicographic, and it's the right shape.

What I found is mostly around the new mandatory device-target gate and one thing I think is an accident.

The accident first: the wheel install root stopped going through `resolved_install_root`, so it no longer resolves symlinks. #317 (`ff78075b`) is in this branch's history — I confirmed with `git merge-base --is-ancestor` — and the tarball path still calls it, so this is one-sided rather than a deliberate removal. It matters because it re-opens the exact bug #317 fixed, and the e2e harness creates the very symlink that triggers it.

The device-target gate is the bigger design question. `required_aggregate_device_target` hard-errors when host detection returns `None`, and it runs before the `dry_run` early return, so a preview on a machine with no AMD GPU now fails — including via `--family`, since the gate re-checks detection rather than trusting the flag. Two scenarios in this PR's own feature file depend on that no-GPU preview working, one of them pre-existing and one added here. Both are `@nightly` and the run had `E2E_INCLUDE_NIGHTLY` empty, so CI didn't catch it.

On the GPU lane failure: I traced run 33096888884 and I don't think it's a flake — `classify_dependency_details` requires *every* detail to name the realigned package, and realigning torch predictably drags `triton` out of pin too, so the deliberate divergence gets reported as a violation and the run exits 1. Details inline.

Also worth raising: this bundles canonical stream resolution with a ~1300-line torch-alignment/device-probe subsystem. AGENTS.md §11 asks for one logical change per PR, and here it's load-bearing — the torch-alignment defect above is what's keeping the stream-resolution work red.

There's an ordering problem with #329 and #272 that isn't visible from any single PR; I've noted it in the last comment.

juhovainio:
@michaelroy-amd Rebased this branch onto latest `main` (all 21 commits preserved individually, no squashing).

While verifying the rebased result (`cargo test --workspace --all-targets`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo xtask e2e`), clippy surfaced two silent merge artifacts from conflict resolution that weren't caught by `cargo check`/`cargo test` alone. Both are fixed, folded into the commits where the conflicts originated (not left as separate cleanup commits):

- `xtask/src/e2e_prewarm.rs`: conflict resolution had reverted a `main`-side ordering fix back to an older pattern where `runtime_changed` was computed inline inside the decision's own match arm instead of via the free function read after the engine check. Restored the correct ordering; fixed up into `fix(runtime): migrate stale wheel compositions`, the commit where the regression was introduced.
- `apps/rocm/src/main.rs`: stray blank line between a doc comment and `classify_runtime_device_probe` left over from resolving the final commit's conflicts. Fixed up into that same commit.

Full verification after the fixups: tests and clippy clean, E2E suite 71/74 passing (2 expected xfail, 1 pre-existing environment-only failure unrelated to this branch, confirmed it reproduces identically on unmodified `main`).

michaelroy-amd:
Resolver ownership update: this PR/EAI-8268 is the canonical owner for current release/nightly aggregate source selection, shared provenance, and runtime-composition behavior. #329/EAI-8431 will be rebased as the ROCm 10 Next-layout extension on top of this model. #272's #271 regression behavior will be transferred and verified here with scenario evidence, after which #272 will close as superseded. #314/current `main` remains authoritative for the torch owning-runtime settlement; obsolete alignment work in this branch's older history will be dropped during the rebase.

For new canonical and Next installs, unknown or incomplete source layouts must fail closed—no silent legacy or alternate-channel fallback. Existing managed runtimes with legacy manifests remain readable and usable for backward compatibility.

Related: #272, #329, #314. Tracking: EAI-8268, EAI-8431, EAI-7956.

michaelroy-amd:
Rebased and narrowed onto current `main` at `7fa847cc`. This version preserves #314 as the authoritative torch-owning-runtime settlement, absorbs #272's observable #271 regression with positive source/version/provenance/device-extra coverage, and leaves #329 as the dependent ROCm 10 Next-layout extension. All local workspace, clippy, smoke, manifest, notices, signature, focused Cucumber, and live canonical dry-run gates are green; CI is now running against the rewritten head. Please re-review this head because the prior review context targeted the superseded history.

michaelroy-amd:
CI diagnosis at `f603779c`: the prior MI300X run installed the correct `device-gfx942` package, but the post-install `rocm_sdk` probe selected its unrelated default `gfx1010`; vLLM settlement then tested the wrong library target and failed with `hipErrorInvalidImage` before Cucumber started. The fix passes the already source-validated exact device target only to the newly installed runtime probe. Adopted legacy-runtime probes remain unchanged, and #314 torch settlement remains authoritative. Focused TheRock tests (101 passed) and clippy pass locally; fresh CI is running.

juhovainio (approved):

PR #316, opened 2026-08-27
ci(e2e): repair a pre-warm tree that records a folder elsewhere

### LINKED ISSUE #315: E2E harness: a scenario that installs while using the shared runtimes symlink poisons the shared pre-warm tree
## Summary

The e2e harness lets a scenario reuse an expensive pre-warmed runtime by symlinking the scenario's `data/runtimes` at the shared tree. That is safe for scenarios that only read it. A scenario that *installs* while that symlink is in place writes its own ephemeral, absolute paths into the shared runtime manifest. When the scenario's temp directory is cleaned up, the shared runtime is left recording a path that no longer exists, and every later command that resolves it fails.

## Mechanism

1. Each scenario runs against a throwaway data directory, `/tmp/rocm-e2e-XXXXXX/data`.
2. `use_shared_runtimes()` (`tests/e2e-cucumber/tests/e2e.rs:284`) symlinks `<scenario-data>/runtimes` to the directory named by `E2E_SHARED_RUNTIMES_DIR`.
3. The scenario runs `rocm install sdk` with `ROCM_CLI_DATA_DIR` set to the scenario's own data directory.
4. `install sdk` bakes absolute paths — `install_root`, `python_executable` — derived from that data directory into the runtime manifest.
5. Because `data/runtimes` is a symlink, the manifest is written into the **shared** registry, recording `/tmp/rocm-e2e-XXXXXX/...` paths.
6. The scenario ends and its temp directory is removed.
7. The shared runtime manifest now points into a deleted directory.

## Observed on a self-hosted GPU runner

Checked directly against a runner's shared pre-warm tree after it had completed nightly runs. Two runtimes are registered; one is poisoned and one is intact:

```
=== release-wheel-gfx94x-dcgpu-7-13-0.json
"install_root":      "/tmp/rocm-e2e-7MidR2/data/runtimes/wheel/release-wheel-gfx94x-dcgpu-7-13-0"
"python_executable": "/tmp/rocm-e2e-7MidR2/data/runtimes/wheel/.../bin/python"

=== release-wheel-multi-arch-7-14-0.json
"install_root":      "<prewarm>/data/runtimes/wheel/release-wheel-multi-arch-7-14-0"
"python_executable": "<prewarm>/data/runtimes/wheel/.../bin/python"
```

The first manifest's recorded interpreter is gone, while the venv it should be describing is present:

```
manifest path exists: NO
actual venv exists:   YES
```

The `/tmp/rocm-e2e-7MidR2` directory no longer exists; other, newer `/tmp/rocm-e2e-*` directories from later runs do. So the poisoned entry outlived the scenario that wrote it, which is the defining symptom.

Subsequent commands against that runtime fail with:

```
runtime `<key>` did not resolve to a managed TheRock Python environment for automatic vLLM install
```

Re-running `install sdk` against the shared tree directly repairs it, because that rewrites the manifest with correct absolute paths.

## Impact

A run can leave the host's shared pre-warmed runtime unusable for later runs. The failure surfaces long after the run that caused it, on an unrelated scenario, and the message names neither the scenario nor the symlink — so it reads as random corruption of the pre-warm tree.

Symptoms seen from this state include an `active.json` referring to a deleted install root under `/tmp`, and console scripts whose shebang points at that same dead path (a missing interpreter reports `No such file or directory` against a script that plainly exists, which is its own detour).

Note in the evidence above that only one of the two registered runtimes is poisoned. A tree in this state is not obviously broken — it can pass whatever happens to use the intact entry, which makes the fault look intermittent.

## Relationship to the existing guard

`e2e-selfhosted.yml` already documents this hazard from a different angle:

> the shared dir IS the pre-warm's own `data/runtimes`, and we NEVER move it. `install sdk` bakes ABSOLUTE paths (`install_root`, `python_executable`) into the runtime manifest; a post-install `mv` would leave those pointing at a deleted location and every serve would fail instantly

The workflow avoids *moving* the directory. The symlink reaches the same end state without moving anything: the install itself supplies the wrong prefix.

## How to check a host

```bash
grep -o '"install_root":[^,]*' <prewarm>/data/runtimes/registry/*.json
```

A `/tmp/rocm-e2e-*` prefix means that entry has been poisoned.

## Possible directions

Not prescribing one, since the right answer depends on how much the harness wants to protect:

- Have installing scenarios use their own runtimes directory rather than the shared symlink, accepting the cold-install cost for the few that write.
- Make the shared tree read-only for scenarios, so an install into it fails loudly instead of silently rewriting it.
- Have `install sdk` resolve the real path of its data directory before baking absolute paths, so a symlinked runtimes dir records the shared location rather than the scenario's.
- Have the pre-warm decision validate that the recorded `python_executable` exists, so a poisoned tree is repaired rather than reused.

The last one only limits the blast radius; the first three address the cause.

## Scope

Pre-existing, and independent of any change currently in flight — found while verifying an unrelated fix on a GPU host.

volen-silo:
Fixes #315 (the cause is addressed separately; see *Scope* below).

## Summary

- Before deciding whether the shared pre-warm tree is fresh, drop any managed runtime in it whose recorded `install_root` is outside that tree, so the pre-warm reinstalls instead of serving a dead one.
- Read-only runtimes are exempt — `runtimes adopt` records an external folder on purpose.
- Nine tests over a new pure `assess`, mirroring the existing `decide` contract. Fixtures captured verbatim from the built binary, per the convention the module's `EMPTY` docblock already sets.

## Root cause

A scenario reaches the shared tree through a symlink at its own `data/runtimes` (`E2eWorld::use_shared_runtimes`). `install sdk` run that way bakes the *link's* path — a per-scenario temp dir — into the manifest that lands in the **shared** registry, and into the venv's console-script shebangs. The temp dir dies with the scenario; the shared runtime is left naming a folder that no longer exists. The tree lives on a runner PVC, so the damage outlives the run and surfaces later on an unrelated scenario.

`rocm update` cannot see this. It compares versions against the channel index, not the runtime against the disk.

## Why the folder is deleted rather than reinstalled over

This is the repair, not a precaution. Measured on uv 0.9.30 with a venv created through a symlink:

| Action | Shebang afterwards |
|---|---|
| re-run `uv venv` | still the dead path (warns "already exists") |
| reinstall an already-satisfied package | still the dead path (`Audited 1 package`) |
| `--reinstall` | rewritten, script runs |

A poisoned venv keeps a working `bin/python` — it is a symlink to the base interpreter, which is still there — so `ensure_uv_venv` reuses it, and the install audits rather than reinstalls. Re-running the install over a poisoned tree reports success and repairs nothing. Only removing the folder first works.

This also contradicts the issue's note that "re-running `install sdk` against the shared tree directly repairs it": that holds only when the SDK version actually changed, which is exactly the case where the pre-warm would not be reusing.

## Why the signal is not `status=unusable`

The obvious approach — repair whatever `runtimes list` calls unusable — is unsafe, and the captured fixtures show why. A healthy managed runtime reports:

```
mode=managed status=unusable (pip runtime manifest is missing rocm_sdk probe data)
```

This code deletes what it selects. Hanging a multi-GiB delete on a status with many causes — and one a future tightening of `validate_runtime_manifest_for_activation` could start emitting for healthy runtimes — is not a good trade. An out-of-tree install root, for a runtime this tree is supposed to own, has exactly one cause.

## Non-obvious decisions

- **Repair runs before `probe`**, so `decide` reads a registry with nothing dead in it and needs no new arm.
- **Failure to repair is fatal**, unlike the rest of this module. Everywhere else a conservative fallback keeps the lane green; here the alternative is serving a runtime already known to be broken, which fails later, elsewhere, and names none of this. An unreadable `runtimes list` still removes nothing.
- **`assess` is pure**, so every branch is testable without a GPU, a network, or an install.
- `runtimes uninstall` drops the registry entry but reports `folder_removed: no` for a dead root, which is why the explicit folder removal exists. It is guarded to stay inside the pre-warm tree.

## Verification

`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace --all-targets`, `python scripts/smoke_local.py` — all clean.

Exercised against real trees, not only fixtures:

- **Poisoned** (files in the right place, manifest naming a dead `/tmp/rocm-e2e-*` path): warning emitted, registry entry dropped, CLI reported `folder_removed: no`, folder removed here, `decide` then chose `Install`.
- **Healthy managed runtime plus a read-only runtime adopted from outside**: nothing removed, payload intact, tree correctly reused.

Not run locally: a real GPU lane against a live PVC.

## Scope and risk

CI tooling only — no product code, so no runtime behaviour changes for users. Low risk, and it is the part that helps runners that are poisoned *today*, independent of the cause fix.

This limits the blast radius; it does not stop recurrence. The cause — `install sdk` recording the path it was handed rather than the folder the files land in — is a product change with a different risk profile and follows separately.

### DIFF
--- a/xtask/src/e2e_prewarm.rs
+++ b/xtask/src/e2e_prewarm.rs
@@ -143,6 +143,132 @@ pub fn decide(update_report: &str, channel: &str) -> Decision {
     }
 }
 
+/// A managed runtime in the shared tree that records an install root somewhere
+/// else, so serving against it cannot work.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct PoisonedRuntime {
+    pub runtime_key: String,
+    pub format: String,
+    pub install_root: String,
+}
+
+/// Managed runtimes in the pre-warm tree whose recorded install root is not in
+/// that tree.
+///
+/// Pure so every branch is unit-testable without a GPU, a network, or an install
+/// — same contract as [`decide`].
+///
+/// A scenario reaches the shared tree through a symlink at its own
+/// `data/runtimes` (`E2eWorld::use_shared_runtimes`). `install sdk` run that way
+/// writes the *link's* path — a per-scenario temp dir — into the manifest that
+/// lands in the SHARED registry, and into the venv's console-script shebangs. Once
+/// the scenario's temp dir is gone the shared runtime records a path that no
+/// longer exists, and every later run that resolves it fails.
+///
+/// The signal is deliberately "install root is outside the tree" and NOT
+/// `status=unusable`. Unusable has many causes — a missing `rocm_sdk` probe block
+/// alone reports it — and this function's caller DELETES what it returns. Keying
+/// a multi-GiB delete off a status that a future tightening of
+/// `validate_runtime_manifest_for_activation` could start emitting for healthy
+/// runtimes is not a trade worth making. An out-of-tree install root, for a
+/// runtime this tree is supposed to own, has exactly one cause.
+///
+/// `mode=read-only` is exempt: `runtimes import` / `runtimes adopt` record an
+/// external folder on purpose, and that is what read-only means.
+#[must_use]
+pub fn assess(runtimes_list_report: &str, runtimes_dir: &Path) -> Vec<PoisonedRuntime> {
+    // The recorded root of a poisoned runtime does not exist, so it cannot be
+    // canonicalized. Compare textually instead, against both the path as given and
+    // its resolved form, so a caller that passes a path through a symlinked parent
+    // still matches the roots the CLI wrote.
+    let mut roots = vec![runtimes_dir.to_path_buf()];
+    if let Ok(resolved) = runtimes_dir.canonicalize()
+        && resolved != *runtimes_dir
+    {
+        roots.push(resolved);
+    }
+
+    let lines: Vec<&str> = runtimes_list_report.lines().collect();
+    let mut poisoned = Vec::new();
+    for (index, line) in lines.iter().enumerate() {
+        let Some(entry) = RuntimeEntry::parse(line) else {
+            continue;
+        };
+        if entry.read_only {
+            continue;
+        }
+        // `install_root:` is written immediately under its entry by
+        // `render_runtimes_text`. Without it there is nothing to judge, and
+        // guessing a path we are about to delete is not acceptable.
+        let Some(install_root) = lines
+            .get(index + 1)
+            .and_then(|next| next.trim().strip_prefix("install_root: "))
+            .map(str::trim)
+            .filter(|value| !value.is_empty())
+        else {
+            continue;
+        };
+        if roots
+            .iter()
+            .any(|root| Path::new(install_root).starts_with(root))
+        {
+            continue;
+        }
+        poisoned.push(PoisonedRuntime {
+            runtime_key: entry.runtime_key,
+            format: entry.format,
+            install_root: install_root.to_owned(),
+        });
+    }
+    poisoned
+}
+
+/// One `  {marker} <key> runtime_id=… format=… mode=… status=…` line from
+/// `rocm runtimes list`.
+///
+/// `status=` is last and its value carries spaces and parentheses
+/// (`unusable (install root is missing: /x)`), so it is not read here — every
+/// field this needs appears before it and is space-free.
+struct RuntimeEntry {
+    runtime_key: String,
+    format: String,
+    read_only: bool,
+}
+
+impl RuntimeEntry {
+    fn parse(line: &str) -> Option<Self> {
+        let mut fields = line.split_whitespace();
+        let first = fields.next()?;
+        // The active/rollback markers are separate tokens; anything else is the key.
+        let runtime_key = if matches!(first, "*" | "-") {
+            fields.next()?
+        } else {
+            first
+        };
+        let mut format = None;
+        let mut mode = None;
+        let mut is_entry = false;
+        for field in fields {
+            match field.split_once('=') {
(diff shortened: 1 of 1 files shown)


r0x0r (approved):
## Approve

Reviewed the full change and verified it against the code it drives. This is a tight, well-scoped fix.

### What I checked
- **Parser matches the real renderer.** `assess` / `RuntimeEntry::parse` line up exactly with `render_runtimes_text` in `apps/rocm/src/main.rs`: the `  {marker} {key} runtime_id=… format=… mode=… status=…` entry line followed by an indented `install_root:` line. Marker tokens (`*` / `-`), header lines (`active_runtime_id:`, `registry:`, `installed: none`), and the read-only exemption are all handled correctly, and reading `install_root` from `index + 1` is right.
- **Folder removal is genuinely necessary.** Confirmed in `uninstall_runtime`: it only deletes `manifest.install_root` when it `exists()`. For a poisoned runtime that path is the dead per-scenario temp dir, so the registry entry is cleared but the physically-planted tree at `<runtimes>/<format>/<key>` is left behind — exactly what the explicit, `starts_with`-guarded removal cleans up.
- **Conservative floors are intact.** An unreadable `runtimes list` deletes nothing and keeps the lane green; the healthy-but-`unusable` and read-only-adopted entries in the fixture are correctly left alone. Keying off out-of-tree install root rather than `status=unusable` is the right call and is well justified.
- **Local verification:** `cargo test -p xtask e2e_prewarm` (22 pass), `cargo clippy -p xtask --all-targets -- -D warnings` clean, `cargo fmt -p xtask -- --check` clean.

The reasoning in the docstrings and PR description is thorough, the tests cover each branch, and the risk is limited to CI tooling with no product behavior change. LGTM.

PR #317, opened 2026-08-27
fix(install): record the folder the files land in, not the link

### LINKED ISSUE #315: E2E harness: a scenario that installs while using the shared runtimes symlink poisons the shared pre-warm tree

volen-silo:
Fixes #315. Stacked on #316 — that one repairs runners poisoned today; this one stops it recurring. Either can merge first.

## Summary

- `install sdk` now resolves the install root to its real location before anything is written there, so a runtime reached through a linked `data/runtimes` records the folder the files land in. Covers `--prefix` and the tarball format too.
- Second commit, independent: when a launch fails because a console script's `#!` interpreter is gone, say so instead of reporting ENOENT against a file that is plainly there.

## Root cause

The install root was built by joining onto the data dir and recorded as-is. Reaching `data/runtimes` through a symlink makes that a route rather than a place — correct while the link exists, wrong the moment it goes.

The path outlives the command. It goes into the registry manifest, the sidecar, and — through uv — the `#!` line of every console script in the environment. So the runtime keeps reporting itself installed at a folder that is not there while the files sit untouched next door, and every entry point fails to start.

The E2E harness creates exactly this shape when a scenario opts into the shared pre-warmed runtime, which is how a runner's shared tree got poisoned. Nothing about the fault is test-only: any data dir reached through a link has it.

## Reproduced first (§3)

Measured directly with uv 0.9.30, creating a venv through a symlink:

| Artifact | Records |
|---|---|
| `pyvenv.cfg` | only `home` — clean |
| `bin/python` | symlink to the **base** interpreter — stays alive |
| `bin/activate` | the **link** path |
| console-script shebang | the **link** path |

Remove the link and the reported symptom appears exactly: the script exists, `bin/python --version` still answers, and running the script fails `cannot execute: required file not found`.

## Why only the root is resolved

`python_executable` is derived from the root and must keep the venv's own `bin/python`, which is itself a symlink to the base interpreter — resolving it independently would record the system Python and break venv semantics. The sidecar and the `rocm_sdk` probe fields are likewise derived from the root and need no separate handling; the probe reports paths relative to the interpreter it runs under.

The adopt path already draws the line in exactly this place: `install_root.canonicalize()` sits next to `absolute_existing_file_path_preserving_symlink` for the interpreter. This gives the install path the same treatment. Resolve the root, preserve the interpreter.

## Non-obvious decisions

- **Not at `AppPaths::discover`.** The link is at `data/runtimes`, one level *below* `data_dir`, so canonicalizing the data dir resolves nothing here.
- **Not at manifest-write time.** By then uv has already written the shebangs. The manifest would be honest and the runtime still broken — worse, because `runtimes list` would report `ready` for an environment where nothing starts.
- **Rejected: rewriting `install_root` on load.** It looks tempting and would turn a true `status=unusable (install root is missing: …)` into a false `ready`.
- The helper walks to the deepest existing ancestor and re-attaches the tail, since the leaf does not exist yet. Windows verbatim `\\?\` prefixes are stripped with the existing `strip_verbatim_prefix`; junctions resolve through the same call, so no separate branch.

## Verification

`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo clippy -p e2e-cucumber --test e2e -- -D warnings`, `cargo test --workspace --all-targets`, `python scripts/smoke_local.py` — all clean.

Regression tests fail before the fix and pass after — verified by reverting the one-line change and re-running: `install_root_resolves_a_symlinked_runtimes_folder`, `install_root_resolves_a_symlinked_prefix`, and `install_root_resolves_the_tarball_format_too` all FAIL, while `install_root_is_unchanged_when_nothing_is_linked` stays green, so the tests are not merely restating the new behaviour.

## Scenario, and the lane that runs it (§3)

`@id:runtime-install-records-the-real-folder` in `runtime_setup.feature`. Previewing the install is enough to pin this and needs no GPU and no download: the folder is resolved before the preview prints it.

**It is tagged `@nightly`, and not for its own cost.** It runs in about eight seconds, nearly all of it resolving the channel index. The problem is that the no-GPU mock lane runs 64 scenarios concurrently, and that much extra network work pushes two existing dash scenarios past the validity window they assert on:

- `eai-7960-gen-tps-held-after-scrape-failure`
- `eai-7960-gen-tps-expiry-boundary`

Measured: both fail 3/3 with this scenario present on the mock lane, and both pass with the very same scenario once the suite is serialized. Those two are timing-fragile under load — a pre-existing property this scenario merely exposes, and worth its own fix. Until then this runs on the five self-hosted nightly lanes, where `E2E_INCLUDE_NIGHTLY` is set and scenarios are serialized because a GPU is present. Confirmed there: all four steps pass.

Also confirmed the per-PR mock lane is clean with the scenario skipped (0 unexpected failures), and that the product change alone — without the scenario — leaves the mock lane at 0 unexpected failures.

## Risk

Low-to-moderate; one behaviour change worth naming. `finalize_successful_sdk_install` derives the managed root from `install_root`, so a user whose data dir sits behind a link and who does not set `ROCM_CLI_DATA_DIR` will see the resolved path in `rocm examine` on later runs. Same files, different string. Not a regression, but it is user-visible.

`--prefix` resolution is likewise deliberate — same failure mode, and `runtimes adopt` already canonicalizes its root. If a maintainer would rather leave `--prefix` alone, the change narrows cleanly to the generated-path branch.

## Unrelated flakes seen while verifying

Both reproduce on clean `main`, neither is touched here:

- `providers::tests::local_provider_*` — fails roughly 1 run in 3 under `--workspace --all-targets`, passes in isolation; a different test in the module each time, so shared state in `providers::tests`.
- `rocm-core::fix::tests::the_path_fix_finds_the_install_the_rest_of_the_cli_found` — two tests in that module mutate the process-global `ROCM_PATH` with no lock.


r0x0r (approved):
Reviewed the full diff. This is a well-targeted, well-tested fix and the reasoning in the description holds up against the code.

**What's right**

- **Root cause and fix location.** Resolving the install root before it is written — rather than at `AppPaths::discover` (the link is one level below `data_dir`) or at manifest-write time (uv has already stamped the shebangs by then) — is the correct seam. Resolving the root while deliberately preserving the venv's own `bin/python` interpreter matches the existing adopt path and keeps venv semantics intact.
- **`resolve_path_through_symlinks`.** Walking up to the deepest existing ancestor, canonicalizing, and re-attaching the not-yet-existing tail is the right shape for a path that is about to be *recorded* rather than used. The bail-outs are sound: `file_name()` is `None` for both root and a trailing `..`, so a `..` in the tail returns the caller's path untouched rather than guessing, and a dangling link falls back to the absolutized input. Stripping the Windows verbatim prefix so stored paths still `starts_with`-match ordinary ones is a good catch.
- **Test coverage is genuinely load-bearing, not restating.** The regression trio (`_symlinked_runtimes_folder`, `_symlinked_prefix`, `_the_tarball_format_too`) plus the negative `_is_unchanged_when_nothing_is_linked` guard against gratuitous rewriting, and the `resolve_path_through_symlinks` unit tests cover missing leaf, linked ancestor, dangling link, idempotency, `..`, and the verbatim prefix. Canonicalizing scratch roots up front so assertions don't turn on a linked temp dir is the correct precaution.
- **Second commit (stale-interpreter hint).** Turning the kernel's ENOENT-against-the-script into an explanation that names the dead `#!` interpreter is a real usability win, and it's appropriately conservative — `None` for a live interpreter, a real binary, a genuinely absent file, or an `env`-style shebang — so it only speaks up when it adds something. `describe_skipped_managed_runtimes` is best-effort and only decorates an error already being returned, which is the right contract.

**Notes, non-blocking**

- The user-visible behaviour change (a resolved path appearing in `rocm examine` for data dirs behind a link) is called out honestly in the description and is not a regression.
- `--prefix` resolution is consistent with `runtimes adopt` already canonicalizing its root, so applying the same treatment is defensible.

The `@nightly` tagging rationale — the scenario's extra concurrent index resolution pushing the two timing-fragile `eai-7960-gen-tps-*` dash scenarios past their validity window on the 64-wide mock lane — is a reasonable interim call, with the underlying fragility correctly flagged as its own follow-up.

Approving.

Issue #294, opened 2026-08-20
ci(e2e): deduplicate the self-hosted e2e lanes into a composite action


rominf:
### Problem

The self-hosted e2e lane definitions in `.github/workflows/e2e-selfhosted.yml` and `.github/workflows/nightly.yml` are largely copy-paste. Measured across the 10 lanes in those two files, **62% of step-body lines (502 of 806) sit inside a block that appears verbatim in at least one other lane.**

| Repeated block | Copies | Lines |
|---|---|---|
| `GPU preflight (bounded wait…)` bash | 5 | 130 |
| GPU preflight — advisory + PowerShell variants | 2 + 2 | 128 |
| `Upload E2E report` (differs only in `name:`) | 10 | 60 |
| `Reclaim GPU from stray E2E processes` | 8 | 66 |
| WSL bootstrap / toolchain / checkout | various | 106 |

The five copies of the GPU preflight script differ in **exactly one integer** (`MIN_FREE_GIB`, 16 vs 8); the other ~26 lines are byte-identical. The main `Run E2E tests` script only escapes exact matching because of path substitutions — by sequence similarity it is 97.1% identical between `e2e-gpu` and `e2e-gpu-rad3`, and 97.6% between their nightly equivalents.

The cost is drift, and it is already visible. Examples found while reviewing #278:

- `e2e-gpu-rad3` omits `E2E_MERGE_QUEUE`, which all four sibling lanes set — so it silently runs the cheap PR canary during merge-queue.
- `e2e-gpu-nightly` is the only GPU lane with no preflight step at all.
- The per-PR `e2e-gpu-strix-windows` lane has no `E2E_SERVE_TIMEOUT_SECS` while its nightly twin sets `"300"`.
- `e2e-gpu-nightly-rad3` drops the pre-warm diagnostics its own comment says it mirrors.

Each of these is a one-line divergence in a block that is otherwise identical five ways over. There is no mechanism that would catch any of them.

### Proposed approach

Extract the repeated blocks into a **composite action** (e.g. `.github/actions/e2e-gpu-lane-linux/action.yml`) parameterised on the axes that genuinely vary: `min-free-gib`, preflight strict-vs-advisory, shared uv cache dir, scenario name filter, artifact name.

[...]

PR #319, opened 2026-08-27
test(ci): derive the hardware-docs guard from the workflows

volen-silo:
Part of #294 — acceptance criterion 4 ("adding a new lane without updating `docs/ci-hardware-testing.md` fails CI"). The drift fixes and the composite-action extraction follow separately.

## Summary

- `hardware_testing_docs_cover_all_self_hosted_platforms` now derives its expectations from `e2e-selfhosted.yml` / `nightly.yml` / `ci.yml` instead of comparing the doc against literals in the test source.
- Two surfaces gain a guard they never had: the **runner-labels column**, which was parsed and then discarded (`.map(|row| (row[0], row[2]))`), and the **dispatch-choices list**, which had none.
- No documentation changes — see below.

## Why the old guard could not work

Both sides were hand-maintained. Adding a lane changed neither the doc nor the `vec!`, so the test stayed green with the table silently short a row. It only fired when someone edited one side without the other — which fails the person *fixing* the docs, not the person who made them wrong. #278 shows the shape: the author had to update the workflow, the doc, the README and the test literals in lockstep, and the test went red mid-way through doing so.

## New assertions

Derived per self-hosted job, in workflow order: table rows, runner labels vs actual `runs-on`, artifact list (via `uploaded_e2e_artifacts` across all three workflows), `workflow_dispatch.inputs.platform.options` vs the documented choices, and both "the self-hosted jobs (…)" prose enumerations. The `Platform` column stays free prose — it isn't derivable from YAML — and is asserted non-empty only.

## Correcting the issue on two points

**The docs are not currently stale.** #294 says adding a lane in #278 "left the doc table, the artifact list, two sentences and the dispatch-options list stale, with the test green throughout". #278 (`01ba5257`) in fact updated `docs/ci-hardware-testing.md` (22 lines), `tests/e2e-cucumber/README.md` (6) and the test literals (13) in the same commit. All six derived assertions pass against the doc exactly as it is on `main`, which is why this PR touches no documentation. The defect is structural — hardcoded expectations cannot catch *future* drift — not a divergence sitting in the tree today.

**"Three of those lanes" (`ci-hardware-testing.md:188`) is not stale either.** Exactly three lanes carry `strix-halo`. Left alone.

## Non-obvious decisions

- No YAML crate. The file deliberately uses line-scanning helpers and documents why; this matches that style.
- `runs_on_values` generalized to `flattened_values(text, key)` so the same continuation-joining logic can read `options:`. `runs_on_values` is now a wrapper and its existing extractor guard passes untouched.
- `uploaded_e2e_artifacts` lived inside `#[cfg(test)] mod tests` in `e2e_report.rs` and had to be hoisted to `#[cfg(test)] pub(crate)` to be shared. Body unchanged; both existing consumers still pass.
- Job-id scanning is scoped to `top_level_block(text, "jobs")`, so `push:` / `group:` / `contents:` at the same indent cannot be mistaken for jobs. Covered by a dedicated extractor guard alongside the file's other helper tests.

## Verification

`cargo fmt --all -- --check` = 0 · `cargo clippy --workspace --all-targets -- -D warnings` = 0 · `cargo test --workspace --all-targets` = 0 · `python scripts/smoke_local.py` = 0.

A guard whose failure mode is "passes while wrong" proves nothing by passing, so it was mutation-tested. Each of these fails, and the tree restores byte-identical:

| Mutation | Result |
|---|---|
| Delete a lane row from the doc table | FAIL — job-column assertion, naming the missing lane |
| Change a `Runner labels` cell | FAIL — labels assertion, showing documented vs actual |
| Blank a `Platform` cell | FAIL — non-empty assertion |
| Remove one dispatch choice from the doc | FAIL — dispatch-choices assertion |
| **Add a sixth self-hosted lane, docs untouched** | FAIL, and it keeps failing through each surface in turn — table row, then artifact list, then dispatch options, then the first prose enumeration, then the second — going green only after all six follow |

Load-bearing tests confirmed still passing: `ci_yml_schedules_no_self_hosted_job`, `workflows_use_distinct_concurrency_groups`, `self_hosted_workflow_owns_the_gpu_lanes`, all four `*_prebuilt_e2e_lanes_*`, and both consumers of `uploaded_e2e_artifacts`.

## Scope

CI/test-only; no CLI-observable behaviour changes, so no Gherkin scenario is required (AGENTS.md §3 exempts CI plumbing and tests).

One CI-visibility note worth stating: `xtask/src/affected.rs` forces a full-workspace test run for any `.github/workflows/` change, so this test runs on PRs that touch the workflows. A **docs-only** PR selects nothing (`is_ignorable` drops `docs/`), so the doc half is enforced on push/merge_group and by `windows-build-and-test` rather than on the PR itself.

### DIFF
--- a/.github/workflows/e2e-selfhosted.yml
+++ b/.github/workflows/e2e-selfhosted.yml
@@ -509,6 +509,10 @@ jobs:
       # Raise the wait budget rather than let that read as a product failure; a
       # genuine hang still fails, just later.
       E2E_TUI_TIMEOUT_SECS: "90"
+      # Same cold-start budget as every other lane and as this lane's own nightly
+      # twin: a first serve on shared hardware loads the model before it answers,
+      # and the default is short enough to read that as a product failure.
+      E2E_SERVE_TIMEOUT_SECS: "300"
       # Match the Linux Strix dispatch path: opt into the platform-adaptive
       # large-model scenario only when the manual input is enabled.
       E2E_INCLUDE_NIGHTLY: "${{ inputs.include_nightly && '1' || '' }}"
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -363,6 +363,46 @@ jobs:
           rm -rf /tmp/rocm-e2e-* 2>/dev/null || true
           echo "reclaimed"
 
+      # GPU preflight: fail fast (~90s) instead of hanging to the job cap when the
+      # GPU is missing, the driver is wedged, or VRAM is still saturated by a
+      # leftover serve. A BOUNDED POLL, not a one-shot check: transient contention
+      # (e.g. the reclaim step's kills still draining VRAM) self-heals within
+      # seconds, so we retry up to a ceiling and succeed the moment the GPU is both
+      # responsive AND has enough free VRAM. Only a genuinely absent/wedged/held
+      # GPU reaches the ceiling and fails — with a per-reason message.
+      - name: GPU preflight (bounded wait for an available GPU)
+        run: |
+          # A serve here needs most of the card; require a generous free-VRAM
+          # floor so a leftover serve (which the reclaim step should have killed)
+          # is caught, while normal baseline (~300 MB used) passes immediately.
+          MIN_FREE_GIB="${GPU_PREFLIGHT_MIN_FREE_GIB:-16}"
+          CEILING_SECS="${GPU_PREFLIGHT_CEILING_SECS:-90}"
+          min_free=$(( MIN_FREE_GIB * 1024 * 1024 * 1024 ))
+          deadline=$(( SECONDS + CEILING_SECS ))
+          reason="rocm-smi never returned within its timeout (driver wedged or GPU absent)"
+          while [ "$SECONDS" -lt "$deadline" ]; do
+            # rocm-smi itself can hang on a wedged driver — bound it with timeout.
+            out=$(timeout 15 rocm-smi --showmeminfo vram 2>/dev/null) || { sleep 5; continue; }
+            # Parse the byte value AFTER the colon; the line prefix "GPU[0]" would
+            # otherwise make a naive first-number match pick up the "0".
+            total=$(printf '%s\n' "$out" | grep -i 'VRAM Total Memory' | sed 's/.*: *//' | grep -oE '[0-9]+' | tail -1)
+            used=$(printf '%s\n' "$out" | grep -i 'VRAM Total Used Memory' | sed 's/.*: *//' | grep -oE '[0-9]+' | tail -1)
+            if [ -z "$total" ] || [ -z "$used" ]; then
+              reason="rocm-smi returned no VRAM figures (no AMD GPU detected)"
+              sleep 5; continue
+            fi
+            free=$(( total - used ))
+            if [ "$free" -ge "$min_free" ]; then
+              echo "GPU ready: $(( free / 1024 / 1024 / 1024 )) GiB free (>= ${MIN_FREE_GIB} GiB)."
+              exit 0
+            fi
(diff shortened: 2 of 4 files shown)


rominf:
I reviewed this PR (diff-only, plus building and running the affected xtask tests locally against the PR's actual worktree content) and didn't find anything blocking.

What I checked in particular, since this is a CI guard whose whole point is not to pass silently when it shouldn't:

- Ran `hardware_testing_docs_cover_all_self_hosted_platforms` and the rest of `workflow_contract`/`e2e_report` against the PR head — all 21 workflow-contract tests and the full `xtask` suite (110 tests) pass as-is, and `cargo fmt --check` / `cargo clippy --all-targets -D warnings` are clean.
- Reproduced a few of the mutation cases described in the PR body myself (adding a 6th self-hosted lane without touching the docs, adding a dispatch option, and drifting a runner-labels cell) — each one fails the new guard with a clear, specific assertion message naming the mismatch, so the derived assertions do catch the drift they claim to.
- Traced the new helpers (`flattened_values`, `flattened_list_items`, `self_hosted_e2e_jobs`, the `nested_block` reuse for `workflow_dispatch.inputs.platform.options`) against the actual `e2e-selfhosted.yml` content — the job-id scan is correctly scoped to the `jobs:` block, and the label/list comparisons work for both the flow (`[a, b]`) and block (`- a`) YAML spellings.
- The PR's claim that the docs aren't currently stale and that this touches no documentation checks out — `git diff` confirms only `xtask/src/e2e_report.rs` and `xtask/src/workflow_contract.rs` changed.

This is a diff-and-local-test read, not a substitute for CI or a maintainer's own pass — but from that read, this looks safe to merge once CI is green.

r0x0r (commented):
This is a solid structural fix — deriving the expectations from the workflows is the right call, and the mutation table (especially "add a sixth lane, docs untouched" failing through each surface in turn) is convincing evidence the guard actually bites now rather than passing-while-wrong.

One forward-looking note on `flattened_list_items`: for the block form it splits the `" - "`-joined string that `flattened_values` produced back on `" - "`. That round-trip is lossy — it's safe for runner labels and the current dispatch `options:` values (none contain `" - "`), but now that `flattened_values` is generalized over arbitrary keys, a future derived value that legitimately contained `" - "` would mis-split silently. It's a reasonable assumption to make given the inputs; worth a one-line comment stating that the helper assumes label-like tokens with no embedded `" - "`, so the constraint is explicit for whoever reuses it next.

Everything else — scoping the job-id scan to `top_level_block(text, "jobs")`, hoisting `uploaded_e2e_artifacts` to shared with the body unchanged, and keeping `runs_on_values` as a thin wrapper so its existing guard still applies — reads well.

r0x0r (commented):
## Re-review (rebased to `dfe11b037`)

Re-read the full diff against the current head. This head is a rebase of what I reviewed on 2026-08-28 — the derivation logic is byte-for-byte the same, so my earlier read stands: deriving the doc expectations from the workflows is the right structural fix, and the mutation table (especially "add a sixth lane, docs untouched" failing through each surface in turn) is convincing that the guard now bites instead of passing-while-wrong.

**One earlier point is still open** (not a blocker, just unaddressed by the rebase): `flattened_list_items` parses the block form by splitting the `" - "`-joined string that `flattened_values` produced back on `" - "`. That round-trip is lossy — safe for runner labels and the current dispatch `options:` values, but now that `flattened_values` is generalized over arbitrary keys, a future derived value that legitimately contained `" - "` would mis-split silently. A one-line comment on the helper stating it assumes label-like tokens with no embedded `" - "` would make the constraint explicit for the next reuse. Purely a robustness note.

No new findings on this pass.

**CI note (§12):** the red **E2E tests (GPU)** / **E2E tests (Strix Halo, Ubuntu)** on this PR are not a real signal — they come from run `33371799973`, whose overall conclusion is **cancelled** (superseded), with the Strix job itself cancelled. A re-run is needed to get a clean status; there's nothing here for a test-only, workflow-deriving change to have broken.

r0x0r (approved):

PR #320, opened 2026-08-27
ci(e2e): close two self-hosted lane divergences, and guard them

volen-silo:
Part of #294 — acceptance criterion "the drift instances are resolved or deliberately documented".

**Stacked on #319** (base is `ci/derive-hardware-docs-guard`), because both new tests derive the lane list using the `self_hosted_e2e_jobs` helper that PR adds. Retarget to `main` once #319 lands.

## Summary

- `e2e-gpu-nightly` gains the GPU preflight step it never had — byte-identical to its per-PR twin's, verified programmatically.
- Per-PR `e2e-gpu-strix-windows` gains `E2E_SERVE_TIMEOUT_SECS: "300"`.
- Two contract tests so neither can re-drift, and so a lane added tomorrow is covered without anyone remembering.

## What was wrong

**`e2e-gpu-nightly` had no preflight.** Its step list was checkout → Reclaim → toolchain → Run → Upload. It was the only one of the ten self-hosted lanes across both files without one. The consequence is not cosmetic: on a wedged driver, or a GPU still held by a leftover serve, the lane hangs to its 90-minute cap instead of failing in ~90s with a named reason. Its own per-PR twin (`e2e-selfhosted.yml:200`, `MIN_FREE_GIB` 16) and its sibling `e2e-gpu-nightly-rad3` (`nightly.yml:446`) both fail fast.

**Per-PR `e2e-gpu-strix-windows` had no `E2E_SERVE_TIMEOUT_SECS`.** The other four per-PR lanes set `"300"`, as does its own nightly twin. A first serve on shared hardware loads the model before it answers, and the default is short enough that the load reads as a product failure.

Both are one-line divergences inside blocks that are otherwise identical five ways over, and nothing existed that would have caught either.

## Why tests, not just fixes

The fixes are trivial; the guards are the point. Both derive the lane list from the workflows rather than restating it, so:

- a lane added tomorrow is covered the day it lands, without anyone remembering to extend a list;
- a lane that genuinely should *not* wait for a GPU needs an exemption written into the test **with its reason** — which makes it a decision someone makes, rather than one a copy-paste makes for them.

This is the same shape as #319's docs guard: derive from the workflow, don't restate it in the test.

## Verification

`cargo fmt --all -- --check` = 0 · `cargo clippy --workspace --all-targets -- -D warnings` = 0 · `cargo test --workspace --all-targets` = 0 · `python scripts/smoke_local.py` = 0.

Both guards mutation-tested — reverting each fix makes its test fail, naming the lane, and the tree restores byte-identical:

| Mutation | Result |
|---|---|
| Remove the preflight from `e2e-gpu-nightly` | FAIL — ``nightly.yml job `e2e-gpu-nightly` runs on self-hosted GPU hardware but has no GPU preflight step`` |
| Remove `E2E_SERVE_TIMEOUT_SECS` from `e2e-gpu-strix-windows` | FAIL — ``e2e-selfhosted.yml job `e2e-gpu-strix-windows` must give a cold serve the same budget as every other self-hosted lane`` |

Also verified programmatically that the inserted preflight is byte-identical to `e2e-gpu`'s (including `MIN_FREE_GIB` 16, the Instinct floor — not the 8 used by the Strix lanes), and that all ten self-hosted lanes now carry one.

## Scope

CI-only; no CLI-observable behaviour changes, so no Gherkin scenario is required (AGENTS.md §3 exempts CI plumbing).

The other two drift instances listed in #294 do not reproduce — `e2e-gpu-rad3` does set `E2E_MERGE_QUEUE` (`e2e-selfhosted.yml:848`) and `e2e-gpu-nightly-rad3` drops only comment prose. Evidence is in [this issue comment](https://github.com/ROCm/rocm-cli/issues/294#issuecomment-5436472995).

### DIFF
--- a/.github/workflows/e2e-selfhosted.yml
+++ b/.github/workflows/e2e-selfhosted.yml
@@ -509,6 +509,10 @@ jobs:
       # Raise the wait budget rather than let that read as a product failure; a
       # genuine hang still fails, just later.
       E2E_TUI_TIMEOUT_SECS: "90"
+      # Same cold-start budget as every other lane and as this lane's own nightly
+      # twin: a first serve on shared hardware loads the model before it answers,
+      # and the default is short enough to read that as a product failure.
+      E2E_SERVE_TIMEOUT_SECS: "300"
       # Match the Linux Strix dispatch path: opt into the platform-adaptive
       # large-model scenario only when the manual input is enabled.
       E2E_INCLUDE_NIGHTLY: "${{ inputs.include_nightly && '1' || '' }}"
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -363,6 +363,46 @@ jobs:
           rm -rf /tmp/rocm-e2e-* 2>/dev/null || true
           echo "reclaimed"
 
+      # GPU preflight: fail fast (~90s) instead of hanging to the job cap when the
+      # GPU is missing, the driver is wedged, or VRAM is still saturated by a
+      # leftover serve. A BOUNDED POLL, not a one-shot check: transient contention
+      # (e.g. the reclaim step's kills still draining VRAM) self-heals within
+      # seconds, so we retry up to a ceiling and succeed the moment the GPU is both
+      # responsive AND has enough free VRAM. Only a genuinely absent/wedged/held
+      # GPU reaches the ceiling and fails — with a per-reason message.
+      - name: GPU preflight (bounded wait for an available GPU)
+        run: |
+          # A serve here needs most of the card; require a generous free-VRAM
+          # floor so a leftover serve (which the reclaim step should have killed)
+          # is caught, while normal baseline (~300 MB used) passes immediately.
+          MIN_FREE_GIB="${GPU_PREFLIGHT_MIN_FREE_GIB:-16}"
+          CEILING_SECS="${GPU_PREFLIGHT_CEILING_SECS:-90}"
+          min_free=$(( MIN_FREE_GIB * 1024 * 1024 * 1024 ))
+          deadline=$(( SECONDS + CEILING_SECS ))
+          reason="rocm-smi never returned within its timeout (driver wedged or GPU absent)"
+          while [ "$SECONDS" -lt "$deadline" ]; do
+            # rocm-smi itself can hang on a wedged driver — bound it with timeout.
+            out=$(timeout 15 rocm-smi --showmeminfo vram 2>/dev/null) || { sleep 5; continue; }
+            # Parse the byte value AFTER the colon; the line prefix "GPU[0]" would
+            # otherwise make a naive first-number match pick up the "0".
+            total=$(printf '%s\n' "$out" | grep -i 'VRAM Total Memory' | sed 's/.*: *//' | grep -oE '[0-9]+' | tail -1)
+            used=$(printf '%s\n' "$out" | grep -i 'VRAM Total Used Memory' | sed 's/.*: *//' | grep -oE '[0-9]+' | tail -1)
+            if [ -z "$total" ] || [ -z "$used" ]; then
+              reason="rocm-smi returned no VRAM figures (no AMD GPU detected)"
+              sleep 5; continue
+            fi
+            free=$(( total - used ))
+            if [ "$free" -ge "$min_free" ]; then
+              echo "GPU ready: $(( free / 1024 / 1024 / 1024 )) GiB free (>= ${MIN_FREE_GIB} GiB)."
+              exit 0
+            fi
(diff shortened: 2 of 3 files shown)


rominf:
I reviewed this PR's diff (both workflow changes and the new `xtask` contract tests) and didn't find anything to flag.

What I checked:
- The inserted `e2e-gpu-nightly` preflight step in `nightly.yml` is byte-identical to the `e2e-gpu" preflight in `e2e-selfhosted.yml` (diffed the two blocks directly), including the 16 GiB floor.
- `e2e-gpu-strix-windows` now sets `E2E_SERVE_TIMEOUT_SECS: "300"`, matching its nightly twin and the other four per-PR lanes.
- I built the `xtask` crate and ran the two new contract tests — both pass against the current workflows. I then reproduced the PR's own mutation testing: removing the `e2e-gpu-nightly` preflight step fails `every_self_hosted_lane_waits_for_an_available_gpu` with a message naming the lane, and removing `E2E_SERVE_TIMEOUT_SECS" from `e2e-gpu-strix-windows` fails `every_self_hosted_lane_allows_for_a_cold_serve" the same way. Both guards actually catch the regression they're meant to catch, not just superficially.
- Confirmed all ten self-hosted jobs across both workflow files now carry a GPU preflight step and the same `E2E_SERVE_TIMEOUT_SECS`, so the fix's scope matches what the PR claims.

This looks safe to approve after final validation (CI results, and the maintainer's own pass) — this is a diff-only review and doesn't substitute for those.

r0x0r (commented):
Nice catch on both divergences, and I like that the guards derive the lane list from the workflows instead of restating it — the bounded-poll preflight with a per-reason `::error::` message is a real improvement over hanging to the 90-minute cap.

One substantive concern on `every_self_hosted_lane_allows_for_a_cold_serve`: it asserts `E2E_SERVE_TIMEOUT_SECS == "300"` exactly, for every self-hosted lane.

- Exact equality makes 300 both a floor and a ceiling. A genuinely heavier lane (say a large nightly model that needs a longer cold-load budget) can't raise the value without tripping the guard, forcing a workaround or an exemption. If the intent is "no lane gets less than the shared budget," `>= 300` expresses that without capping legitimately larger budgets.
- It also assumes every self-hosted lane runs a serve scenario. A future self-hosted lane that doesn't serve (a build/TUI-only lane on the same hardware) would be forced to carry an irrelevant env var. The preflight guard's rationale explicitly anticipates "a lane that genuinely should not wait for a GPU" and asks for a documented exemption; the serve-timeout guard has no equivalent escape hatch. Worth deciding whether the same exemption shape applies here.

Neither blocks the two fixes, which are clearly correct — it's about how tightly the guard constrains lanes added later.

r0x0r (commented):
## Re-review (rebased to `071a8c81c`)

Re-read the full diff. This head is a rebase of the 2026-08-28 version — the two workflow fixes and both guard tests are unchanged — so my earlier read stands: the divergences are real, the bounded-poll preflight with a per-reason `::error::` message is a genuine improvement over hanging to the 90-minute cap, and deriving the lane list from the workflows (rather than restating it) is the right shape. Still stacked on #319.

**Earlier point still open** on `every_self_hosted_lane_allows_for_a_cold_serve` (not addressed by the rebase): it asserts `E2E_SERVE_TIMEOUT_SECS == "300"` exactly, for every self-hosted lane.

- Exact equality makes 300 both a floor and a ceiling — a genuinely heavier lane (e.g. a large nightly model needing a longer cold-load budget) can't raise the value without tripping the guard. If the intent is "no lane gets *less* than the shared budget," `>= 300` expresses that without capping legitimately larger budgets.
- It also assumes every self-hosted lane runs a serve scenario; a future self-hosted lane that doesn't serve would be forced to carry an irrelevant var. The preflight guard's comment explicitly anticipates "a lane that genuinely should not wait for a GPU" and asks for a documented exemption — the serve-timeout guard has no equivalent escape hatch. Worth deciding whether the same exemption shape applies here.

Neither blocks the two fixes, which are correct — it's about how tightly the guard constrains lanes added later. No new findings this pass.

**CI note (§12):** the red **E2E tests (GPU)** here is the recurring shared-GPU-contention flake, not this change. The run failed the identical three serve/inference scenarios — `bench-load-real-serve`, `chat-end-to-end-local-model`, `serve-vllm-inference` — that failed together on the unrelated #326 (a TUI-only change) the same day, and #326's re-run of that lane went green. A CI-only change that adds a preflight and an env var can't affect vLLM serve/chat/bench; a re-run should clear it. (Mildly on-topic: this PR's preflight is aimed at exactly this class of contention, though these three are the serve scenarios failing under load rather than the wedged-driver hang the preflight guards.)

r0x0r (approved):

PR #327, opened 2026-08-31
fix(bench): populate the engine column and keep bench load latency windowed

r0x0r:
## Summary

`bench load` writes a CSV whose header includes `engine` and `tpot_ms`, but the
`engine` field was always blank in the data row, and `tpot_ms` was frequently
blank too (while `ttft_ms` was populated). Since the CSV feeds the dashboard
Bench panel, `engine` was missing there as well.

- **`engine` — the behaviour fix.** The load generator's only view of the
  backend is its Prometheus `/metrics` endpoint, and the scraper only
  understands vLLM's `vllm:` series. A scrape carrying any recognised vLLM field
  now labels the row, taking the string from the engine registry itself
  (`EngineKind::Vllm.label()`) rather than respelling it, so the CSV value cannot
  drift from the registry. A non-vLLM / 404 endpoint leaves the column blank
  rather than guessing.
- **`tpot_ms` — semantics kept, now documented and covered.** A blank `tpot_ms`
  is not an arithmetic bug. Latency stays strictly windowed —
  `Δsum/Δcount × 1000` over the cell's before/after scrapes — and a missing
  scrape, a flat counter, or a counter reset each leave the column blank. It is
  deliberately *not* backfilled from the endpoint's lifetime `sum/count`
  average: that average covers earlier ramp cells and other clients, so writing
  it into this cell's immutable row would describe a different population than
  the row names, with nothing to flag it. A blank column therefore honestly
  means "not measured for this cell". The helpers are renamed
  (`prom_deltas`/`latency_delta_ms` → `prom_latency`/`latency_ms`, dropping the
  unused placeholder return values). The docstring now records where this
  deliberately *differs* from the telemetry daemon's `avg_ms_from_histogram`,
  which does fall back because it is a repeated live poll with a rolling
  baseline that persists nothing.
- **One CSV cell changes value, not only its name.** The guard on the deltas is
  now phrased as an acceptance (`Δcount > 0 && Δsum >= 0`) rather than a
  rejection (`if Δcount <= 0.0 || Δsum < 0.0 { return None }`). Every comparison
  against NaN is false, so the old form let a NaN delta fall through and return
  `Some(NaN)`, which `opt_f64` wrote into `results.csv` as a literal `NaN` cell;
  the new form returns `None`, so that cell is blank. Blank is the honest
  answer, and a `NaN` token in a numeric column is a downstream hazard — but it
  is a change to what lands in the file, not a pure rename. (An earlier revision
  of this PR body claimed the computed values were unchanged. That was wrong.)
- **Non-finite deltas are rejected too.** `±Inf` reaches the arithmetic by the
  same route as NaN — they are legal exposition tokens that `parse::<f64>()`
  accepts verbatim — but unlike NaN they survive the ordered comparisons: an
  infinite `Δsum` satisfies `>= 0.0` and reached `opt_f64`, whose `{:.6}` wrote
  a literal `inf` into a numeric column, while an infinite `Δcount` collapsed a
  real `Δsum` to `0.000000`. Both deltas must now be finite, which makes the
  "blank when unmeasurable" contract exact rather than NaN-only.

## Reproduce

```
rocm bench load --endpoint http://127.0.0.1:11435/v1 --requests 2 --concurrency 1 --out /tmp/b.csv
head -2 /tmp/b.csv
```

Before: `engine` empty.
After: `engine` populated (`vllm`). `tpot_ms` carries the windowed value when the
endpoint's TPOT counter advanced across the two scrapes, and stays blank when it
did not.

## Migration: a shared `results.csv` written across this change

**`engine` is a rollup key**, and `rocm bench load` appends to one shared
`<data_dir>/bench/results.csv` by default. Trials of a cell are rows differing
only by `run`, so a file spanning this change holds two groups for the same
cell: the pre-change trials (blank `engine`) and the post-change ones (`vllm`).
N trials become two smaller groups — a Pass^N over N trials becomes two N=1
groups, which `at_n_mark` renders as a muted dash. Nothing on screen explains
it: the Bench rollup table renders `cell/model/tp·dtype/concurrency` and never
`engine`, so the two groups look identical, and the table caps at 8 data rows,
so doubling the group count can push cells off it entirely.

This is not only an upgrade artefact — the same split follows any two runs of a
cell where `/metrics` was reachable for one and not the other (a remote
endpoint, metrics disabled, a restart mid-series).

**Rotate or move `results.csv` if you compare trials across the change.** The
run now says so itself rather than leaving it to a source comment: while holding
the append lock, a row that carries an `engine` scans the target file for
earlier rows of the same cell without one, and the CLI warns once per run,
naming the file, the cell, and the fix. The row is still appended — this is a
warning, not a refusal:

```
warning: /home/u/.rocm/bench/results.csv already has rows for bench-c1 with a blank engine
column; `engine` is a rollup key, so those trials group separately from this run's
engine=vllm rows
hint: rotate or move that file if you are comparing trials across the change
```

Rendering `engine` in the Bench rollup table would also make a split group
explicable rather than mysterious. That is a TUI change interacting with the
8-row height cap, so it is deliberately not in this PR.

## The NaN guard: corrected rationale

Commit `163224ad`'s message, and the code comments it added, justified the guard
by claiming Prometheus reports an unobserved histogram sum as NaN. **That is
wrong.** A histogram's `_sum`/`_count` are counters that `prometheus_client` —
which is what vLLM uses for `vllm:time_to_first_token_seconds` and
`vllm:time_per_output_token_seconds` — initialises to `0.0` and `0`; NaN appears
in text exposition for *summary quantiles* with no observations and for gauges
explicitly set to NaN, neither of which is among the four series `vllm_prom`
reads.

The guard stays, on accurate ground: `vllm_prom::extract` ends in
`parse::<f64>()`, which accepts `NaN` and `±Inf` tokens alike, so a malformed or
non-conforming exporter can still put one into the arithmetic, and this is the
last place to stop it before an immutable CSV row. **The guard is therefore defensive, not a
response to an observed exposition.** That is also why it is covered by a unit
test rather than a Gherkin scenario (AGENTS.md §3): the path is unreachable from
any conforming endpoint, so there is no user-observable behaviour to script
against — bench-04's mock counters always advance and never reach it, and a mock
that emitted `NaN` would be asserting on a fixture nobody can produce in the
field. The blank-vs-`NaN` cell change above is the user-visible part, and it is
stated here rather than left in a commit message.

The code comments carrying the wrong claim are corrected in this PR. `163224ad`'s
message cannot be corrected without rewriting a pushed branch, so it stands as
superseded by the paragraph above.

## Tests

The user-observable behaviour — the columns the CLI writes into the CSV, and the
warning it prints — is covered by two Gherkin scenarios in
`tests/e2e-cucumber/features/bench.feature`, both ungated, so they run on every
lane:

- `@id:bench-load-records-engine-and-tpot` (`bench-04`) benchmarks a mock
  endpoint whose vLLM-flavoured `/metrics` histograms advance on every scrape,
  reads the `--out` CSV back, and asserts the emitted row carries `engine=vllm`
  and a `tpot_ms` of 20 ms. The mock's TPOT histogram adds 0.4 s of `_sum` per
  20 `_count` on every scrape, so `Δsum/Δcount` is exactly 0.020 s however many
  scrapes the window spans; asserting the value rather than merely its sign is
  what makes the scenario pin the windowed arithmetic, since a lifetime
  `sum/count` backfill or a unit slip would also be "positive".
- `@id:bench-load-warns-on-engine-split` (`bench-05`, new) runs the benchmark
  once, blanks the `engine` column of the row it wrote — exactly what a
  pre-change build, or a run with no reachable `/metrics`, leaves behind — then
  benchmarks again and asserts the run warns, names the cell and the file, and
  still exits 0. The fixture is produced by the CLI rather than pasted so it
  cannot drift from `CSV_HEADER` (a mismatched header is refused outright).

The pre-existing GPU scenario is renumbered `bench-06`, otherwise unchanged. The
sentence in `tests/e2e-cucumber/expectations.toml` that justifies the
`bench-load-real-serve` xfail is updated to match: it still said "the three
MockServer-backed bench scenarios ... pin the request path and the failure
reporting", which undercounted them and omitted the CSV columns.

Unit tests cover `latency_ms` (windowed value; blank on a flat counter, on a
missing before-scrape, on a counter reset, on a NaN delta, on an infinite
delta, and with no data at all), `detect_engine` (recognised sample → `vllm`; absent or empty sample →
blank), the split detection (flagged for the same cell; quiet when the earlier
rows already carry the engine, when they belong to another cell, when the new
row has no engine, and on a fresh file; raised once per sweep, not once per
cell; and a guard that the scanned column indexes still match `CSV_HEADER`), and
two bench-row tests over a mock endpoint: one reproducing the real symptom (TPOT
counter flat while TTFT advances), asserting `engine` is populated and `tpot_ms`
stays blank, and one where the counter advances, asserting `tpot_ms` is
populated. The existing `t7` non-vLLM test is extended to assert `engine` stays
blank.

## Verification

- `cargo test --workspace --all-targets` — passed
- `cargo xtask e2e` — bench feature green (bench-01 … bench-06)
- Every added test was falsified: each was confirmed red against a deliberate
  break of the code it covers, and green again after restoring it.

### DIFF
--- a/crates/rocm-dash-collectors/src/bench_load.rs
+++ b/crates/rocm-dash-collectors/src/bench_load.rs
@@ -22,6 +22,8 @@ use serde_json::Value;
 use tokio::sync::Semaphore;
 use tokio::task::JoinSet;
 
+use crate::engine_registry::EngineKind;
+
 /// Timeout for /metrics scrapes (short; must never stall the cell sweep).
 const METRICS_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(1);
 
@@ -390,12 +392,25 @@ async fn run_cell_with_clients(
             (Some(running), Some(waiting))
         });
 
-    // TTFT/TPOT deltas come from the before/after histogram scrapes (unchanged).
-    let (_, _, ttft_ms, tpot_ms) = prom_deltas(prom_before.as_ref(), prom_after.as_ref());
+    // TTFT/TPOT latency from the before/after histogram scrapes: the windowed
+    // mean over just the requests this cell issued. A field stays blank when the
+    // window is not measurable (a scrape missing, the counter flat, or a reset)
+    // rather than borrowing the endpoint's lifetime average, which would fold in
+    // traffic from earlier ramp cells or other clients.
+    let (ttft_ms, tpot_ms) = prom_latency(prom_before.as_ref(), prom_after.as_ref());
+
+    // The `/metrics` scraper only understands vLLM's `vllm:` series, so a
+    // recognisable sample is proof the endpoint is vLLM; a non-vLLM endpoint
+    // leaves the column blank rather than guessing. `engine` is a new column:
+    // rows appended before this change have it blank, and since `engine` is a
+    // rollup key, a shared `results.csv` spanning the upgrade splits a cell's
+    // trials into blank and `vllm` groups — rotate the file to regroup them.
+    let engine = detect_engine(prom_before.as_ref(), prom_after.as_ref());
 
     let row = BenchmarkRow {
         cell: format!("bench-c{concurrency}"),
         run: 1,
+        engine,
         model: Some(spec.model.clone()),
         concurrency: Some(concurrency),
         input_len: Some(spec.input_len),
@@ -423,32 +438,85 @@ async fn run_cell_with_clients(
     })
 }
 
-/// Compute latency deltas from two Prometheus samples.
+/// Best-effort engine label for the CSV `engine` column from a scrape pair.
 ///
-/// Returns `((), (), ttft_ms, tpot_ms)` — the first two elements are `None`
-/// placeholder for the deprecated peak fields (peaks now come from the
-/// mid-cell poller; this function only computes histogram deltas).
-/// If either sample is `None`, both latency fields are `None`.
-fn prom_deltas(
+/// The load generator's only view of the backend is its Prometheus `/metrics`
+/// endpoint, and [`crate::vllm_prom::parse`] only recognises vLLM's `vllm:`
+/// series. A sample carrying any recognised field is therefore proof the
+/// endpoint is vLLM. Returns `None` when neither scrape produced a recognisable
+/// sample (a non-vLLM endpoint, a 404, or a malformed body), leaving the column
+/// blank rather than guessing.
+fn detect_engine(
     before: Option<&InstanceSample>,
     after: Option<&InstanceSample>,
-) -> (Option<u32>, Option<u32>, Option<f64>, Option<f64>) {
-    let (Some(b), Some(a)) = (before, after) else {
-        return (None, None, None, None);
-    };
-
-    let ttft_ms = latency_delta_ms(b.ttft_sum_s, b.ttft_count, a.ttft_sum_s, a.ttft_count);
-    let tpot_ms = latency_delta_ms(b.tpot_sum_s, b.tpot_count, a.tpot_sum_s, a.tpot_count);
-
-    (None, None, ttft_ms, tpot_ms)
+) -> Option<String> {
+    let recognised = [after, before].into_iter().flatten().any(sample_is_vllm);
+    // Take the label from the registry (`engine_registry.rs`) rather than
+    // respelling it here: `engine` is a rollup key, so a spelling that drifted
+    // from the registry would silently split rollup groups.
+    recognised.then(|| EngineKind::Vllm.label().to_string())
 }
 
-/// Compute `Δsum / Δcount * 1000` (ms).
+/// Whether a parsed sample carries any vLLM-specific field.
 ///
-/// Returns `None` if either input is `None`, if the count delta is not
-/// positive (avoids division by zero and nonsense from stale counters),
-/// or if the sum delta is negative (counter reset).
-fn latency_delta_ms(
+/// `gen_tps` is deliberately excluded: it is the rate-reporting seam other
+/// engines (e.g. Lemonade) populate, not a vLLM signal, and the vLLM parser
+/// never sets it.
+const fn sample_is_vllm(s: &InstanceSample) -> bool {
+    s.kv_cache_usage_pct.is_some()
+        || s.running_reqs.is_some()
+        || s.waiting_reqs.is_some()
+        || s.gen_tokens_total.is_some()
+        || s.ttft_sum_s.is_some()
+        || s.ttft_count.is_some()
+        || s.tpot_sum_s.is_some()
+        || s.tpot_count.is_some()
+}
+
+/// Compute TTFT/TPOT latency (ms) from two Prometheus samples.
+///
+/// Returns `(ttft_ms, tpot_ms)`, each the windowed mean over just the requests
+/// this cell issued (see [`latency_ms`]). Either field is `None` when its
+/// window is not measurable — including a flat TPOT counter, which for vLLM
+/// means no inter-token gap was recorded in the window, i.e. the value is
+/// genuinely unmeasured for this cell rather than zero.
+fn prom_latency(
+    before: Option<&InstanceSample>,
+    after: Option<&InstanceSample>,
+) -> (Option<f64>, Option<f64>) {
+    let ttft_ms = latency_ms(
+        before.and_then(|s| s.ttft_sum_s),
+        before.and_then(|s| s.ttft_count),
+        after.and_then(|s| s.ttft_sum_s),
+        after.and_then(|s| s.ttft_count),
+    );
+    let tpot_ms = latency_ms(
+        before.and_then(|s| s.tpot_sum_s),
+        before.and_then(|s| s.tpot_count),
+        after.and_then(|s| s.tpot_sum_s),
+        after.and_then(|s| s.tpot_count),
+    );
+    (ttft_ms, tpot_ms)
+}
+
+/// Windowed latency (ms) across the cell's two histogram scrapes.
+///
+/// Computes `Δsum/Δcount × 1000` — the mean latency of just the requests this
+/// cell issued between the before- and after-scrape. Returns `None` when the
+/// window cannot be measured: either scrape missing, the observation count did
+/// not advance (`Δcount ≤ 0`, no request recorded this metric in the window),
+/// or the counter reset (`Δsum < 0`, a server restart).
+///
+/// The value is deliberately *not* backfilled from the after-scrape's lifetime
+/// `sum/count` average. That average covers every request the server process
+/// ever handled — earlier ramp cells and other clients included — so writing it
+/// into this cell's immutable CSV row would describe a different population than
+/// the row names, with nothing to flag it. A blank column therefore honestly
+/// means "not measured for this cell". (This is narrower than the telemetry
+/// daemon's `avg_ms_from_histogram`, which *does* fall back: it is a repeated
+/// live poll with a rolling baseline that self-corrects on the next tick and
+/// persists nothing, so the trade-offs differ.)
+fn latency_ms(
     sum_before: Option<f64>,
     count_before: Option<f64>,
     sum_after: Option<f64>,
@@ -456,10 +524,8 @@ fn latency_delta_ms(
 ) -> Option<f64> {
     let delta_sum = sum_after? - sum_before?;
     let delta_count = count_after? - count_before?;
-    if delta_count <= 0.0 || delta_sum < 0.0 {
-        return None;
-    }
-    Some(delta_sum / delta_count * 1000.0)
+    // Guard division by zero (flat counter) and counter resets.
+    (delta_count > 0.0 && delta_sum >= 0.0).then_some(delta_sum / delta_count * 1000.0)
 }
 
 /// Concurrency levels tried by [`run_auto_ramp`] in order.
@@ -1067,6 +1133,10 @@ mod tests {
         );
         assert_eq!(row.ttft_ms, None, "ttft_ms should be None for 404 /metrics");
         assert_eq!(row.tpot_ms, None, "tpot_ms should be None for 404 /metrics");
+        assert_eq!(
+            row.engine, None,
+            "engine should be blank for a non-vLLM (404 /metrics) endpoint"
+        );
         assert!(
             row.gen_tps.unwrap_or(0.0) > 0.0,
             "gen_tps must still be positive"
@@ -1505,5 +1575,224 @@ mod tests {
             reason.contains("usage"),
             "the reason should name the missing field, got: {reason}"
         );
+    }
+
+    // ---------- engine + tpot column population ----------
+
+    fn sample_with_tpot(
+        ttft_sum: Option<f64>,
+        ttft_count: Option<f64>,
+        tpot_sum: Option<f64>,
+        tpot_count: Option<f64>,
+    ) -> InstanceSample {
+        InstanceSample {
+            running_reqs: Some(1),
+            ttft_sum_s: ttft_sum,
+            ttft_count,
+            tpot_sum_s: tpot_sum,
+            tpot_count,
+            ..Default::default()
+        }
+    }
+
+    #[test]
+    fn latency_ms_prefers_the_window_when_the_counter_advanced() {
+        // Δsum=1.0s over Δcount=2 → 500 ms.
+        assert_eq!(
+            latency_ms(Some(1.0), Some(10.0), Some(2.0), Some(12.0)),
+            Some(500.0)
+        );
+    }
+
+    #[test]
+    fn latency_ms_is_none_when_the_counter_did_not_advance() {
+        // A flat window (Δcount = 0) is unmeasured for this cell, not the
+        // endpoint's lifetime average. For vLLM's TPOT this is the load-bearing
+        // case: a flat counter means zero inter-token gaps were recorded here,
+        // so a blank column is the honest answer.
+        assert_eq!(
+            latency_ms(Some(2.0), Some(100.0), Some(2.0), Some(100.0)),
+            None
+        );
+    }
+
+    #[test]
+    fn latency_ms_is_none_when_the_before_scrape_is_missing() {
+        // With no baseline there was never a window over this cell; the row must
+        // not borrow the after-scrape's lifetime average.
+        assert_eq!(latency_ms(None, None, Some(3.0), Some(30.0)), None);
+    }
+
+    #[test]
+    fn latency_ms_is_none_on_a_counter_reset() {
+        // sum dropped (server restart) → the negative window is discarded and
+        // the row stays blank rather than reporting an average from a different
+        // process lifetime.
+        assert_eq!(
+            latency_ms(Some(10.0), Some(100.0), Some(0.2), Some(2.0)),
+            None
+        );
+    }
+
+    #[test]
+    fn latency_ms_is_none_without_any_data() {
+        assert_eq!(latency_ms(None, None, None, None), None);
+        // count == 0 is not divisible → None, never a divide-by-zero number.
+        assert_eq!(latency_ms(None, None, Some(0.0), Some(0.0)), None);
+    }
+
+    #[test]
+    fn prom_latency_computes_ttft_and_tpot_independently() {
+        // ttft advanced (windowed → 500 ms); tpot flat across the window, so it
+        // is genuinely unmeasured for this cell and stays blank rather than
+        // reporting the endpoint's lifetime average.
+        let before = sample_with_tpot(Some(1.0), Some(10.0), Some(2.0), Some(100.0));
+        let after = sample_with_tpot(Some(2.0), Some(12.0), Some(2.0), Some(100.0));
+        let (ttft_ms, tpot_ms) = prom_latency(Some(&before), Some(&after));
+        assert_eq!(ttft_ms, Some(500.0));
+        assert_eq!(tpot_ms, None);
+    }
+
+    #[test]
+    fn prom_latency_reports_tpot_when_the_counter_advances() {
+        // The window advances for both histograms → both populated. ttft:
+        // (2.0-1.0)/(12-10)*1000 = 500 ms; tpot: (2.4-2.0)/(120-100)*1000 = 20 ms.
+        let before = sample_with_tpot(Some(1.0), Some(10.0), Some(2.0), Some(100.0));
+        let after = sample_with_tpot(Some(2.0), Some(12.0), Some(2.4), Some(120.0));
+        let (ttft_ms, tpot_ms) = prom_latency(Some(&before), Some(&after));
+        assert_eq!(ttft_ms, Some(500.0));
+        let tpot = tpot_ms.expect("tpot_ms should be Some when the counter advances");
+        assert!((tpot - 20.0).abs() < 1e-9, "expected tpot≈20 got {tpot}");
+    }
+
+    #[test]
+    fn detect_engine_labels_a_recognised_sample_vllm() {
+        let sample = sample_with_tpot(Some(1.0), Some(10.0), Some(2.0), Some(100.0));
+        assert_eq!(detect_engine(Some(&sample), None).as_deref(), Some("vllm"));
+        assert_eq!(detect_engine(None, Some(&sample)).as_deref(), Some("vllm"));
+    }
+
+    #[test]
+    fn detect_engine_leaves_an_unrecognised_endpoint_blank() {
+        // No scrape at all (404 / non-vLLM), and an all-`None` (malformed body)
+        // sample must both leave the column blank rather than guess "vllm".
+        assert_eq!(detect_engine(None, None), None);
+        let empty = InstanceSample::default();
+        assert_eq!(detect_engine(Some(&empty), Some(&empty)), None);
+    }
+
+    fn prom_body_full(
+        running: u32,
+        waiting: u32,
+        ttft_sum: f64,
+        ttft_count: f64,
+        tpot_sum: f64,
+        tpot_count: f64,
+    ) -> String {
+        format!(
+            "vllm:num_requests_running {running}\n\
+             vllm:num_requests_waiting {waiting}\n\
+             vllm:time_to_first_token_seconds_sum {ttft_sum}\n\
+             vllm:time_to_first_token_seconds_count {ttft_count}\n\
+             vllm:time_per_output_token_seconds_sum {tpot_sum}\n\
+             vllm:time_per_output_token_seconds_count {tpot_count}\n"
+        )
+    }
+
+    // A vLLM endpoint whose TPOT counter did not advance between the before and
+    // after scrapes (while TTFT did): the `engine` column is populated from the
+    // recognised scrape, `ttft_ms` carries its windowed value, and `tpot_ms`
+    // stays blank — a flat TPOT window is unmeasured for this cell, not a value
+    // borrowed from the endpoint's lifetime average.
+    #[tokio::test]
+    async fn bench_row_labels_engine_and_leaves_flat_tpot_blank() {
+        let server = MockServer::start().await;
+
+        Mock::given(method("POST"))
+            .and(path("/v1/chat/completions"))
+            .respond_with(stub_response(100, 50))
+            .mount(&server)
+            .await;
+
+        // Before scrape: ttft cumulative 1.0s/10, tpot cumulative 2.0s/100.
+        Mock::given(method("GET"))
+            .and(path("/metrics"))
+            .respond_with(
+                ResponseTemplate::new(200)
+                    .set_body_string(prom_body_full(5, 2, 1.0, 10.0, 2.0, 100.0)),
+            )
+            .up_to_n_times(1)
+            .mount(&server)
+            .await;
+        // After/poller: ttft advanced (sum 2.0/count 12) but tpot counter is
+        // unchanged (still 2.0/100) — the exact real-vLLM symptom.
+        Mock::given(method("GET"))
+            .and(path("/metrics"))
+            .respond_with(
+                ResponseTemplate::new(200)
+                    .set_body_string(prom_body_full(8, 1, 2.0, 12.0, 2.0, 100.0)),
+            )
+            .mount(&server)
+            .await;
+
+        let mut spec = make_spec(&server.uri());
+        spec.requests = 2;
+        let row = run_cell(&spec, 1).await.unwrap().row;
+
+        assert_eq!(
+            row.engine.as_deref(),
+            Some("vllm"),
+            "engine must be labelled from the recognised vLLM scrape"
+        );
+        // ttft: windowed (2.0-1.0)/(12-10)*1000 = 500 ms.
+        let ttft = row.ttft_ms.expect("ttft_ms should be Some");
+        assert!((ttft - 500.0).abs() < 0.01, "expected ttft≈500 got {ttft}");
+        // tpot: flat window → genuinely unmeasured for this cell → blank.
+        assert_eq!(
+            row.tpot_ms, None,
+            "a flat TPOT window must stay blank, not borrow the lifetime average"
+        );
+    }
+
+    // When the TPOT counter advances across the cell's scrapes, the windowed
+    // value is populated — the ordinary served case.
+    #[tokio::test]
+    async fn bench_row_populates_tpot_when_the_counter_advances() {
+        let server = MockServer::start().await;
+
+        Mock::given(method("POST"))
+            .and(path("/v1/chat/completions"))
+            .respond_with(stub_response(100, 50))
+            .mount(&server)
+            .await;
+
+        // Before: tpot cumulative 2.0s/100. After: 2.4s/120 → Δ 0.4s/20 = 20 ms.
+        Mock::given(method("GET"))
+            .and(path("/metrics"))
+            .respond_with(
+                ResponseTemplate::new(200)
+                    .set_body_string(prom_body_full(5, 2, 1.0, 10.0, 2.0, 100.0)),
+            )
+            .up_to_n_times(1)
+            .mount(&server)
+            .await;
+        Mock::given(method("GET"))
+            .and(path("/metrics"))
+            .respond_with(
+                ResponseTemplate::new(200)
+                    .set_body_string(prom_body_full(8, 1, 2.0, 12.0, 2.4, 120.0)),
+            )
+            .mount(&server)
+            .await;
+
+        let mut spec = make_spec(&server.uri());
+        spec.requests = 2;
+        let row = run_cell(&spec, 1).await.unwrap().row;
+
+        assert_eq!(row.engine.as_deref(), Some("vllm"));
+        let tpot = row
+            .tpot_ms
+            .expect("tpot_ms must be populated when the counter advanced");
+        assert!((tpot - 20.0).abs() < 0.01, "expected tpot≈20 got {tpot}");
     }
 }
--- a/crates/rocm-dash-collectors/src/bench_load.rs
+++ b/crates/rocm-dash-collectors/src/bench_load.rs
@@ -80,6 +80,33 @@ pub struct CellReport {
     pub failed: u32,
     /// Reason from the first failure observed, for the operator-facing warning.
     pub first_error: Option<String>,
+    /// Set on the first cell of a sweep whose append landed in a file that
+    /// already held blank-`engine` rows for that cell; `None` on every other
+    /// report, so the CLI warns once per run rather than once per cell. Always
+    /// `None` from [`run_cell`], which appends nothing.
+    pub engine_split: Option<EngineSplitNotice>,
+}
+
+/// Evidence that the results file being appended to straddles the point at
+/// which the `engine` column started being populated.
+///
+/// `engine` is a rollup key (`RollupKey` in `rocm_dash_core::bench_rollup`), so
+/// a cell's rows written before the column was populated (blank `engine`) do
+/// not group with the rows written after it (`vllm`): one Pass^N over N trials
+/// becomes two smaller groups, and nothing on the dashboard says why — the
+/// Bench rollup table renders `cell/model/tp·dtype/concurrency` and never
+/// `engine`. The same split follows any run whose `/metrics` was unreachable
+/// (a remote endpoint, metrics disabled, a restart mid-series), so this is not
+/// a one-off upgrade notice; it fires whenever the two groups actually exist.
+///
+/// Carried out to the CLI rather than printed here so the collector stays free
+/// of process-wide output.
+#[derive(Debug, Clone)]
+pub struct EngineSplitNotice {
+    /// The cell whose earlier rows in the file carry a blank `engine`.
+    pub cell: String,
+    /// The results file holding both groups.
+    pub path: String,
 }
 
 /// Why a single benchmark request produced no usable measurement.
@@ -401,10 +428,10 @@ async fn run_cell_with_clients(
 
     // The `/metrics` scraper only understands vLLM's `vllm:` series, so a
     // recognisable sample is proof the endpoint is vLLM; a non-vLLM endpoint
-    // leaves the column blank rather than guessing. `engine` is a new column:
-    // rows appended before this change have it blank, and since `engine` is a
-    // rollup key, a shared `results.csv` spanning the upgrade splits a cell's
-    // trials into blank and `vllm` groups — rotate the file to regroup them.
+    // leaves the column blank rather than guessing. `engine` is a new column and
+    // a rollup key, so a shared `results.csv` spanning the upgrade splits a
+    // cell's trials into blank and `vllm` groups; the append path detects that
+    // and the CLI warns about it (see [`EngineSplitNotice`]).
     let engine = detect_engine(prom_before.as_ref(), prom_after.as_ref());
 
     let row = BenchmarkRow {
@@ -435,6 +462,8 @@ async fn run_cell_with_clients(
         succeeded: n_success,
         failed: n_failed,
         first_error,
+        // Filled in by the sweep, which is what knows the target file.
+        engine_split: None,
     })
 }
 
@@ -506,7 +535,7 @@ fn prom_latency(
 /// window cannot be measured: either scrape missing, the observation count did
 /// not advance (`Δcount ≤ 0`, no request recorded this metric in the window),
 /// the counter reset (`Δsum < 0`, a server restart), or either delta is `NaN`
-/// (Prometheus reports an unobserved histogram sum that way).
+/// (defensive: see the guard below).
 ///
 /// The value is deliberately *not* backfilled from the after-scrape's lifetime
 /// `sum/count` average. That average covers every request the server process
@@ -527,7 +556,18 @@ fn latency_ms(
     let delta_count = count_after? - count_before?;
     // Phrased as acceptance, not rejection: every comparison against NaN is
     // false, so this rejects a NaN delta where `if delta_count <= 0.0 || ...`
-    // would fall through and emit Some(NaN).
+    // would fall through and emit Some(NaN) — which `opt_f64` then writes into
+    // the CSV as a literal `NaN` cell.
+    //
+    // A conforming vLLM cannot produce that NaN: `_sum`/`_count` are counters
+    // that `prometheus_client` initialises to `0.0`/`0` and never renders as
+    // NaN (NaN appears in text exposition for summary quantiles with no
+    // observations and for gauges explicitly set to NaN, neither of which is
+    // among the four series `vllm_prom` reads). The guard is therefore
+    // defensive, not a response to an observed exposition: `extract` finishes
+    // with `parse::<f64>()`, which accepts a `NaN` token from a malformed or
+    // non-conforming exporter, and this is the last place to stop it before it
+    // reaches an immutable CSV row.
     (delta_count > 0.0 && delta_sum >= 0.0).then_some(delta_sum / delta_count * 1000.0)
 }
 
@@ -537,13 +577,30 @@ pub const RAMP_SEQUENCE: &[u32] = &[1, 2, 4, 8, 16, 32, 64, 128];
 /// Minimum fractional `gen_tps` improvement to keep ramping.
 pub const PLATEAU_GAIN: f64 = 0.05;
 
+/// Zero-based position of the `cell` column in [`CSV_HEADER`].
+const CELL_COLUMN: usize = 0;
+
+/// Zero-based position of the `engine` column in [`CSV_HEADER`].
+///
+/// Both indexes are only read after the file's header has been validated as
+/// byte-identical to [`CSV_HEADER`], so they cannot address a foreign layout.
+const ENGINE_COLUMN: usize = 4;
+
 /// Open (or create) `csv_path` once, take an exclusive advisory lock, validate
 /// or write the header, then append one newline-terminated row.
 ///
 /// The lock serializes cooperating `rocm bench load` processes so concurrent
 /// first writers cannot both emit the header. `O_APPEND` keeps each row write
 /// at the end of the file.
-fn append_one_row(row: &BenchmarkRow, csv_path: &Path) -> Result<(), BenchLoadError> {
+///
+/// Returns an [`EngineSplitNotice`] when the row carries an `engine` and the
+/// file already holds rows for the same cell without one — the two sets are
+/// separate rollup groups, which the caller reports. Detected under the same
+/// lock as the append so a concurrent writer cannot slip a row in between.
+fn append_one_row(
+    row: &BenchmarkRow,
+    csv_path: &Path,
+) -> Result<Option<EngineSplitNotice>, BenchLoadError> {
     use std::io::{BufRead, Seek};
 
     let mut file = OpenOptions::new()
@@ -553,6 +610,7 @@ fn append_one_row(row: &BenchmarkRow, csv_path: &Path) -> Result<(), BenchLoadEr
         .open(csv_path)?;
     file.lock()?;
 
+    let mut notice = None;
     if file.metadata()?.len() == 0 {
         file.write_all(CSV_HEADER.as_bytes())?;
     } else {
@@ -564,11 +622,59 @@ fn append_one_row(row: &BenchmarkRow, csv_path: &Path) -> Result<(), BenchLoadEr
                 path: csv_path.display().to_string(),
             });
         }
+        if row.engine.is_some() {
+            file.seek(std::io::SeekFrom::Start(0))?;
+            if has_blank_engine_row(&file, &row.cell) {
+                notice = Some(EngineSplitNotice {
+                    cell: row.cell.clone(),
+                    path: csv_path.display().to_string(),
+                });
+            }
+        }
     }
 
+    // `O_APPEND` puts this at the end of the file regardless of where the
+    // scan above left the read cursor.
     let line = serialize_row_to_line(row)?;
     file.write_all(&line)?;
-    Ok(())
+    Ok(notice)
+}
+
+/// Whether `file` already holds a row for `cell` whose `engine` column is blank.
+///
+/// Best-effort by design: an unreadable record is skipped rather than failing
+/// the append, because this only decides whether to print a warning and a
+/// half-written or hand-edited line is not worth refusing a benchmark over.
+/// The reader is `flexible` for the same reason. Stops at the first hit, so the
+/// common case (a file the current build wrote) costs one pass at most.
+fn has_blank_engine_row(file: &std::fs::File, cell: &str) -> bool {
+    csv::ReaderBuilder::new()
+        .has_headers(true)
+        .flexible(true)
+        .from_reader(std::io::BufReader::new(file))
+        .records()
+        .flatten()
+        .any(|record| {
+            record.get(CELL_COLUMN) == Some(cell)
+                && record
+                    .get(ENGINE_COLUMN)
+                    .is_none_or(|engine| engine.trim().is_empty())
+        })
+}
+
+/// Attach `notice` to `report` unless this sweep already raised one.
+///
+/// The notice describes the *file*, so repeating it for every concurrency level
+/// of a ramp would be noise; `already` carries the once-per-sweep state.
+fn attach_engine_split(
+    report: &mut CellReport,
+    notice: Option<EngineSplitNotice>,
+    already: &mut bool,
+) {
+    if !*already && notice.is_some() {
+        report.engine_split = notice;
+        *already = true;
+    }
 }
 
 /// Run a concurrency sweep and append one aggregate row per cell to `csv_path`.
@@ -586,11 +692,13 @@ pub async fn run_and_append_csv(
     let clients = BenchClients::new()?;
 
     let mut reports = Vec::with_capacity(concurrency_levels.len());
+    let mut split_reported = false;
     for &conc in concurrency_levels {
-        let report = run_cell_with_clients(spec, conc, &clients).await?;
+        let mut report = run_cell_with_clients(spec, conc, &clients).await?;
         // A fully-failed cell is still appended: the dashboard tailer expects one
         // row per cell, and the caller reports the failure separately.
-        append_one_row(&report.row, csv_path)?;
+        let notice = append_one_row(&report.row, csv_path)?;
+        attach_engine_split(&mut report, notice, &mut split_reported);
         reports.push(report);
     }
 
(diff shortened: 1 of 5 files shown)


crates/rocm-dash-collectors/src/bench_load.rs
```diff
@@ -423,43 +432,96 @@ async fn run_cell_with_clients(
...
+        (sum_before, count_before, sum_after, count_after)
+    {
+        let delta_sum = sa - sb;
+        let delta_count = ca - cb;
+        // Windowed average, guarding division by zero and counter resets.
+        if delta_count > 0.0 && delta_sum >= 0.0 {
+            return Some(delta_sum / delta_count * 1000.0);
+        }
+    }
+    // Cumulative-average fallback from the after-scrape (first/only scrape, no
```
rominf: **This fallback puts data from outside the benchmarked cell into that cell's row.**

`bench load` measures the requests *this cell* issued. When the window is unusable, `sum/count*1000` is the server process's lifetime average — every request it has served, from every client, since it started.

The ramp makes this concrete. `run_auto_ramp` (`:643-664`) walks `RAMP_SEQUENCE` = 1,2,4,8,… against one `BenchClients` and one server, appending a row per cell. So a `bench-c8` row that falls back reports a TPOT dominated by the earlier concurrency-1 and -2 traffic — which is precisely the variable the ramp exists to isolate. Nothing in the row marks it, so a consumer reading `results.csv` can't tell a windowed value from a lifetime one.

And the PR's own target case is the worst one. vLLM records `time_per_output_token_seconds` only for inter-token gaps, so a flat TPOT counter across the window means **zero TPOT observations occurred in this cell**. That's the one situation where the lifetime average is guaranteed not to describe this run — and it's exactly when the fallback fires. A blank there was correct: it meant "not measurable", which is true.

`latency_ms_falls_back_on_a_counter_reset` (`:1621-1628`) is the sharpest version — it asserts that after a server restart, we report 100.0 ms from a 2-observation histogram belonging to a *different process lifetime* than the before-scrape. That test blesses the behaviour rather than catching it.

I'd keep window-only semantics for the CSV, where blank honestly means unmeasured. If a fallback is genuinely wanted, narrow it to the one defensible case — `prom_before` entirely absent, so there was never a window — and never on a flat or reset counter. And whichever way it goes, the row should carry the provenance so a consumer can tell them apart.

crates/rocm-dash-collectors/src/bench_load.rs
```diff
@@ -423,43 +432,96 @@ async fn run_cell_with_clients(
...
-/// Returns `None` if either input is `None`, if the count delta is not
-/// positive (avoids division by zero and nonsense from stale counters),
-/// or if the sum delta is negative (counter reset).
-fn latency_delta_ms(
+/// Prefers `Δsum/Δcount × 1000` across the two scrapes — the latency of just
+/// the requests this cell issued. When that window is unavailable (a scrape is
+/// missing, the counter did not advance, or it reset) but the after-scrape
+/// carries a non-empty cumulative histogram, falls back to its lifetime average
+/// `sum/count × 1000`. This mirrors the telemetry daemon's
+/// `avg_ms_from_histogram` so both consumers of the same histograms agree.
```
rominf: **This doesn't mirror the daemon, and the parity claim is what justifies the fallback.**

`avg_ms_from_histogram` (`crates/rocm-dash-daemon/src/runner.rs:978`) gates its window on `dt > 0.0 && dt <= MAX_RATE_INTERVAL_S`, with `MAX_RATE_INTERVAL_S = 60.0` and a comment at `:950-955` saying exactly why: "A gap longer than this makes the windowed delta stale (outage or forward wall-clock jump / NTP / VM resume), so the histogram falls back to the cumulative average for that sample."

`latency_ms` takes no interval at all. A default bench cell — 128 requests at 1024 output tokens — runs well past 60 s, so under the daemon's own rule that window would be rejected as stale, not accepted.

The contracts differ in kind too. The daemon is a repeated-poll live gauge with a rolling per-instance baseline (`prev_ttft`/`prev_tpot`), where a one-cycle cumulative fallback self-corrects on the next poll and nothing is persisted. A two-scrape benchmark writes one immutable row to a file people compare across runs; there is no next poll to correct it.

Since `adb1d7d9` was specifically a doc-accuracy commit, this line is worth getting right: say what's actually shared (the `sum/count` formula) and what isn't (the interval guard, the rolling baseline, the persistence), and drop "so both consumers agree" — they don't.

crates/rocm-dash-collectors/src/bench_load.rs
```diff
@@ -423,43 +432,96 @@ async fn run_cell_with_clients(
...
+fn detect_engine(
     before: Option<&InstanceSample>,
     after: Option<&InstanceSample>,
-) -> (Option<u32>, Option<u32>, Option<f64>, Option<f64>) {
-    let (Some(b), Some(a)) = (before, after) else {
-        return (None, None, None, None);
-    };
+) -> Option<String> {
+    let recognised = [after, before].into_iter().flatten().any(sample_is_vllm);
+    recognised.then(|| "vllm".to_string())
```
rominf: Minor: this hardcodes the label that `engine_registry.rs` already owns.

`EngineKind::Vllm.label()` returns `"vllm"` and its doc calls it the "Stable lowercase label (matches `rocm engines` / config engine names)" — the registry is documented as "one place that maps a discovered serving engine to its per-engine sample parser". A second in-crate spelling can drift from it, and since `engine` is a `RollupKey` field, drift would silently split rollup groups.

`EngineKind::Vllm.label().to_string()` is the one-line version.

Adjacent, not asking for it here: `try_scrape_prom` (`:162`) also hardcodes `vllm_prom::parse` rather than dispatching through `EngineKind::parse_sample`. Routing through the registry is what would eventually let a Lemonade endpoint be labelled instead of leaving the column blank.

rominf (dismissed):
I read this at `adb1d7d9`. The `engine` column half is straightforward and correct — `sample_is_vllm` only checks fields the vLLM parser sets, `gen_tps` is rightly excluded since that's Lemonade's seam, and I couldn't find a mislabelling path. Column order, the 18-field record, and the append guard are byte-identical to `main`, and the only reader (`bench_tail.rs:41-50`) deserializes by header name, so pre-upgrade files keep working. Units check out too: `_sum` in seconds over `_count` observations gives ms per output token.

My concern is the other half. The cumulative fallback doesn't just fill a blank — it writes a number measured over a different population than the row describes, with nothing in the CSV to distinguish the two. Details inline.

Two things about scope worth raising here:

**`ttft_ms` changed and the PR doesn't say so.** On `main`, `prom_deltas` returned `(None, None)` unless *both* scrapes existed, and `latency_delta_ms` returned `None` on a flat or reset counter. At this head, a missing before-scrape, a flat counter, or a reset all produce a lifetime average for `ttft_ms` as well. The title, the body ("populate engine and tpot_ms"), and the docstring you corrected in `adb1d7d9` all frame this as tpot-only — so a column that used to be strictly windowed quietly loosened without anyone being told. Either scope the fallback to `tpot_ms`, or say plainly in the body and the docstring that `ttft_ms` moves too.

**Rollup groups split across the upgrade.** `--out` defaults to the shared appended `<data_dir>/bench/results.csv`, so a user's file ends up with pre-upgrade rows (`engine` empty → `None`) and post-upgrade rows (`Some("vllm")`) for the same cell/model/concurrency. `RollupKey` includes `engine` (`crates/rocm-dash-core/src/bench_rollup.rs:39`), so those trials now land in two Pass^N groups instead of one and the Bench panel's rollup changes. That's inherent to populating the column and I don't think it needs a code change, but users should be told to rotate `results.csv`.

**Process notes.** This changes the contents of a file the CLI creates, which AGENTS.md §3 puts squarely in "user-observable" — and it's explicit that a unit test asserting the internal helper doesn't discharge it. The new wiremock `#[tokio::test]` is in-crate, and none of the four `bench.feature` scenarios assert CSV column content. Either add a scenario asserting the emitted row carries `engine`/`tpot_ms` against the metrics mock, or name the `@id:` you're relying on. Separately, §8's local gate includes `python scripts/smoke_local.py`, which the Verification section doesn't mention either way — §8 asks for the gap to be stated if it couldn't run.

Checks are green apart from skipped lanes.

One thing I checked and am not flagging: a NaN `_sum` would propagate through `opt_f64`'s `{f:.6}` into the CSV, but `main` has the identical exposure, so it's pre-existing rather than introduced here.

r0x0r:
Thanks @rominf — this was the right call. Reworked to **drop the cumulative fallback entirely** and keep the CSV window-only, per your preference. Rebased onto `main` and force-pushed as a single commit (`419052a`).

**Blocking — fallback wrote a different population into the row.** Removed it. `latency_ms` is now strictly `Δsum/Δcount × 1000`; a missing scrape, a flat counter, or a reset returns `None`. Blank honestly means "not measured for this cell". The `latency_ms_falls_back_on_a_counter_reset` test that blessed the old behaviour is gone; the reset/flat/missing cases now assert `None`. Your point about the ramp is exactly why: a `bench-c8` row must not carry TPOT dominated by the earlier concurrency-1/-2 traffic, and for vLLM a flat TPOT window means zero inter-token gaps were recorded here — so blank is the correct answer.

**Doc parity claim (`:501`).** Removed — it was wrong. The docstring no longer claims to mirror the daemon or that "both consumers agree". It now says the daemon's `avg_ms_from_histogram` *does* fall back because it's a repeated live poll with a rolling baseline that self-corrects and persists nothing, whereas this writes one immutable row, so the trade-offs differ.

**`ttft_ms` scope (raised in the summary).** No longer moves: with the fallback gone, both `ttft_ms` and `tpot_ms` stay exactly as windowed as on `main`. Nothing loosened silently.

**Minor (`:448`) — hardcoded label.** Now `EngineKind::Vllm.label().to_string()`, with a comment noting that `engine` is a rollup key so a drifted spelling would split groups. (Left `try_scrape_prom`'s `vllm_prom::parse` alone as you said — not this PR.)

**Rollup groups split across the upgrade.** Not a code change, but users are now told: a comment at the `engine` assignment and the commit body both say a shared `results.csv` spanning the upgrade mixes blank and `vllm` rows for the same cell and should be rotated.

**§3 — user-observable coverage.** Added `@id:bench-load-records-engine-and-tpot` (bench-04), an ungated MockServer-backed scenario that runs on every lane. It benchmarks a metrics-backed endpoint whose histogram counters advance each scrape, then reads the emitted `--out` file and asserts the row carries `engine=vllm` and a positive `tpot_ms` — proving the columns reach the file, not just the helper. Unit tests cover windowed / flat / reset / missing-scrape / advancing-counter.

**§8.** `cargo fmt`, `cargo clippy --all-targets -D warnings`, `cargo test`, and `python3 scripts/smoke_local.py` all green locally. PTAL.

crates/rocm-dash-collectors/src/bench_load.rs
r0x0r: Agreed, and the fallback is gone as of `419052a` (current PR head).

`latency_ms` is now strictly windowed — `Δsum/Δcount × 1000` — returning `None` when either scrape is missing, when `Δcount ≤ 0` (flat counter), or when `Δsum < 0` (reset). Blank once again means "not measured for this cell", so a ramp row can no longer report a TPOT dominated by the earlier concurrency-1/-2 traffic, and the vLLM case you singled out — a flat TPOT counter meaning zero inter-token gaps were recorded in this cell — stays blank instead of borrowing the process lifetime average.

`latency_ms_falls_back_on_a_counter_reset`, the test that blessed the old behaviour, is deleted. The reset case is now `latency_ms_is_none_on_a_counter_reset` and asserts `None`; the flat, missing-before, and no-data cases assert `None` as well.

On provenance: with the fallback removed, every populated value is windowed by construction, so there is no longer a lifetime-vs-windowed distinction for a consumer to disambiguate and I did not add a marker column. Happy to add one if you still want it for a future fallback.

Verified on Linux: `cargo test --workspace --all-targets` and `cargo clippy --workspace --all-targets -- -D warnings` both clean. The end-to-end coverage is the ungated scenario `@id:bench-load-records-engine-and-tpot` (bench-04), which I confirmed fails before this change (row emitted with a blank `engine`) and passes after.

crates/rocm-dash-collectors/src/bench_load.rs
r0x0r: You are right, and the claim is removed in `419052a`.

The docstring no longer asserts that this mirrors the daemon, and "so both consumers agree" is gone. Because the fallback itself was dropped, there is also no interval guard left to match — with the value strictly windowed there is no staleness decision for `latency_ms` to make, so the `MAX_RATE_INTERVAL_S` question no longer arises here.

What remains is a note recording the difference rather than a parity claim:

> The value is deliberately *not* backfilled from the after-scrape's lifetime `sum/count` average. That average covers every request the server process ever handled — earlier ramp cells and other clients included — so writing it into this cell's immutable CSV row would describe a different population than the row names, with nothing to flag it. A blank column therefore honestly means "not measured for this cell". (This is narrower than the telemetry daemon's `avg_ms_from_histogram`, which *does* fall back: it is a repeated live poll with a rolling baseline that self-corrects on the next tick and persists nothing, so the trade-offs differ.)

crates/rocm-dash-collectors/src/bench_load.rs
r0x0r: Done in `419052a` — the label now comes from the registry:

```rust
recognised.then(|| EngineKind::Vllm.label().to_string())
```

with a comment at the call site recording why it matters: `engine` is a `RollupKey` field, so a spelling that drifted from the registry would silently split rollup groups.

Left `try_scrape_prom`'s hardcoded `vllm_prom::parse` alone, per your note that it is not being asked for here. Agreed that routing it through `EngineKind::parse_sample` is what would eventually let a Lemonade endpoint be labelled rather than leaving the column blank.

siloteemu:
🔴 Automated review · pr-review-watcher · 419052a9

## Summary
Populates the previously always-blank `engine` CSV column in `rocm bench load` by detecting vLLM from the `/metrics` scrape, renames/redocuments the TTFT/TPOT latency helpers, and adds unit plus e2e (bench-04) coverage for both columns. Verdict: **Approve**. Verified: I answered the revert question per assertion — the `engine` assertions (`bench_load.rs:1743`, `1792`, `bench_steps.rs:187`) all fail on revert, so the headline fix is non-vacuously tested, and `bench_load.rs:1137`'s `engine == None` is not vacuous either since it catches over-labelling; but `latency_ms`/`prom_latency` are **byte-for-byte the same arithmetic** as the removed `latency_delta_ms`/`prom_deltas` (same guard, same `?` short-circuit), so every new `tpot_ms`/`ttft_ms` assertion passes identically against the base branch — that is new coverage of unchanged behaviour, not proof of a fix, and the commit message ("keep latency windowed") is honest about it. Units confirmed correct: `vllm:time_per_output_token_seconds_*` is a per-token histogram in seconds, `Δsum/Δcount×1000` yields ms/token, and the unit tests pin the exact 20.0 ms and 500.0 ms values; blank (not zero, not the lifetime average) is written when the window is unmeasurable. I confirmed by reading source that `engine` is a real `RollupKey` field (`bench_rollup.rs:39`), that `CSV_HEADER` already contained `engine` so no file-format break, that `try_scrape_prom` only ever uses `vllm_prom::parse`, that lemonade's parser sets only `gen_tps` (correctly excluded from `sample_is_vllm`), and that the e2e mock's counters genuinely advance. No prompt-injection content anywhere in the diff. Relied on CI for the full matrix/e2e execution; ran one `cargo check -p rocm-dash-collectors` (clean). Blocking: 0 · Non-blocking: 5.

**None.**

## Non-blocking
- `tests/e2e-cucumber/tests/e2e/bench_steps.rs:193` — `tpot.parse::<f64>().is_ok_and(|v| v > 0.0)` is the one loose assertion: the mock's `Δsum/Δcount` is a deterministic **exactly 0.020 s → 20 ms** on every scrape pair, so `> 0.0` would happily accept `0.02` if the `×1000` were ever dropped. Assert `(v - 20.0).abs() < 0.01` to make the e2e discriminate on units, as the unit tests already do.
- `tests/e2e-cucumber/tests/e2e/bench_steps.rs:177,183` — the CSV is parsed with `header.split(',')` / `data.split(',')`, but the writer is `csv::Writer`, which quotes embedded commas; no populated field contains one today (`launcher` is `"rocm bench load (local smoke)"`), so it works, but a future model/launcher string with a comma would silently shift the column index and assert the wrong value instead of failing. Use the `csv` crate reader.
- `tests/e2e-cucumber/expectations.toml:156-157` — this PR made the comment stale: it still says "the **three** MockServer-backed bench scenarios run unxfail'd on every lane" when bench-04 makes four. The parallel comment in `bench.feature` was updated; this one was missed.
- `crates/rocm-dash-collectors/src/bench_load.rs:449` — `detect_engine` can only ever emit `"vllm"`; `EngineKind::Lemonade` is unreachable from this path (bench only probes `/metrics`, lemonade serves JSON at `/api/v1/stats`), so the `engine` column stays blank for lemonade endpoints — the default engine on Strix Halo. Intentional per the doc comment, but worth a follow-up if the column is meant to be fleet-wide.
- `crates/rocm-dash-collectors/src/bench_load.rs:519-527` — the guard rewrite quietly changes NaN handling: the old `if delta_count <= 0.0 || delta_sum < 0.0` fell through on NaN and could write `Some(NaN)` (rendered as `"NaN"` in a numeric CSV column), while `(delta_count > 0.0 && delta_sum >= 0.0).then_some(...)` returns `None`. Strictly an improvement, but undocumented in a diff otherwise presented as a pure rename.

r0x0r:
Retitled, and added a regression test for a NaN edge case found while reconciling the description with the code.

**Retitle.** Was _"populate engine and tpot_ms columns"_ → now _"populate the engine column and keep bench load latency windowed"_. `tpot_ms` was never blank: it was already computed and written before this branch, via `prom_deltas`/`latency_delta_ms` on the same four histogram fields. The old title described the cumulative-average fallback that was dropped during review, not what ships. The only newly-populated column is `engine`.

**NaN guard (`163224ad`).** Reconciling the two implementations turned up one genuine behaviour change that was neither documented nor tested. The guard was rewritten from a rejection to an acceptance:

```rust
// before
if delta_count <= 0.0 || delta_sum < 0.0 { return None; }
Some(delta_sum / delta_count * 1000.0)
// after
(delta_count > 0.0 && delta_sum >= 0.0).then_some(delta_sum / delta_count * 1000.0)
```

Those are exact complements for ordinary floats, but not for `NaN`: every comparison against `NaN` is false, so the old form fell through and wrote `Some(NaN)` into an immutable CSV row, while the new one yields a blank. Reachable from a real scrape — Prometheus reports an unobserved histogram sum as `NaN`, and `vllm_prom.rs:155` parses it verbatim with `parse::<f64>().ok()`; there is no `is_nan`/`is_finite` guard anywhere on that path.

The new test is discriminating: with the guard reverted to the old form it fails with `left: Some(NaN), right: None`. Verified on Linux — `cargo test --workspace --all-targets` and `cargo clippy --workspace --all-targets -- -D warnings` both clean.

siloteemu:
🔴 Automated review · pr-review-watcher · 163224ad

## Summary
`rocm bench load` now labels the CSV `engine` column from the vLLM `/metrics` scrape and keeps TTFT/TPOT strictly windowed (`Δsum/Δcount`, blank when unmeasurable), plus a NaN guard and tests. **Approve** — no blocking issues. Verified: the retitle is accurate against the base (`engine` was absent from the `BenchmarkRow` literal and fell to `..Default::default()` → always blank; `tpot_ms` was already computed by `latency_delta_ms` and written, so the old "two newly populated columns" title was the wrong claim); the NaN reachability argument holds at the path level — `try_scrape_prom` → `vllm_prom::parse` → `extract` ends in `value_str.parse::<f64>().ok()`, and there is no `is_finite`/`is_nan` check anywhere between the scrape and `latency_ms` (the finite guards in `rocm-dash-core/src/observation.rs` are on the telemetry-daemon path, not this one); the new NaN test genuinely discriminates — I compiled both guard forms side by side and the old `if delta_count <= 0.0 || delta_sum < 0.0` returns `Some(NaN)` for all four of the test's cases while the new form returns `None`, and `"NaN".parse::<f64>()` does succeed; the earlier loose end-to-end assertion was **not** addressed. Blocking: 0 · Non-blocking: 5.

**None.**

## Non-blocking
- `tests/e2e-cucumber/tests/e2e/bench_steps.rs:191-194` — the earlier item is unaddressed: `tpot.parse::<f64>().is_ok_and(|v| v > 0.0)` still accepts any positive value, while `mock_server.rs:399-407` makes `Δsum/Δcount` a deterministic 20.0 ms by construction. Tighten to `(v - 20.0).abs() < 0.5`. As written this assertion would also pass against the base arithmetic, so `col("engine") == "vllm"` (line 186) is the only part of the scenario that discriminates.
- `crates/rocm-dash-collectors/src/bench_load.rs:531` — the guard rejects NaN but not ±Inf: `"+Inf"` parses to `f64::INFINITY`, `delta_sum >= 0.0` accepts it, and `opt_f64`'s `{f:.6}` writes `inf` into the numeric column — the same defect class this PR is closing. `.then(|| …).filter(|v| v.is_finite())` (or `is_finite` on the deltas) closes it and makes the doc claim exact.
- `crates/rocm-dash-collectors/src/bench_load.rs:508` and the test commit body — "Prometheus reports an unobserved histogram sum as NaN" is not right: a histogram `_sum` with zero observations is `0`; `NaN` is what summary *quantiles* emit. The guard is still worth having, but justify it as "NaN is a legal exposition value and the parser accepts it verbatim with no finite check" rather than a mechanism that does not produce it.
- `crates/rocm-dash-collectors/src/bench_load.rs:404-411` — the rollup-split footgun (a `results.csv` spanning the upgrade groups a cell's trials into blank and `vllm`, silently lowering N) is documented only in a source comment; nothing warns at load time and no user-facing doc mentions rotating the file.
- `tests/e2e-cucumber/tests/e2e/bench_steps.rs:115-141` duplicates `run_bench` (88-113) solely to add `--out`; and `col()` at line 183 splits on `,` without CSV quoting — harmless with today's values, fragile if a column ever contains a comma.

Also checked and clean: leak scan over the diff found no internal references; sign-off present on both commits; `sample_is_vllm` cannot mislabel a Lemonade endpoint (`try_scrape_prom` hard-wires the vLLM parser, and Lemonade exposes JSON, not Prometheus text); no dead code left by the earlier removal; no docs reference the bench CSV columns. Note on the test sweep: every other added unit test fails on revert only because it names symbols that do not exist at base, so behaviourally only the NaN test and the three `engine` assertions distinguish new logic from old — the rest are characterization tests for the rewritten function, which is a legitimate use, just not regression protection.

*Posting note: this review was completed at this same head in an earlier cycle but was not published due to an error on our side. The head has not moved since, so the findings stand as written.*

crates/rocm-dash-collectors/src/bench_load.rs
```diff
@@ -423,43 +438,97 @@ async fn run_cell_with_clients(
...
     sum_after: Option<f64>,
     count_after: Option<f64>,
 ) -> Option<f64> {
     let delta_sum = sum_after? - sum_before?;
     let delta_count = count_after? - count_before?;
-    if delta_count <= 0.0 || delta_sum < 0.0 {
-        return None;
-    }
-    Some(delta_sum / delta_count * 1000.0)
+    // Phrased as acceptance, not rejection: every comparison against NaN is
```
rominf, line 557: The PR body says the helpers are renamed and simplified but "the computed values are unchanged". That isn't quite true, and the difference is worth stating because it's user-visible.

The old guard was `if delta_count <= 0.0 || delta_sum < 0.0 { return None; }`. Every comparison against NaN is false, so a NaN delta fell straight through and returned `Some(NaN)` — which `opt_f64` then writes into `results.csv` as a literal `NaN` cell. This acceptance-phrased form returns `None` instead, so that cell is now blank.

The new behaviour is clearly the better one and I'm not asking you to change it. But it is a change to what lands in the CSV, so it belongs in the body rather than only in the tip commit's message — which currently argues the opposite of the body it sits under. Per AGENTS.md §3 that also wants a scenario rather than only the unit test, since bench-04's mock counters always advance and never reach this path; if you'd rather argue NaN is unreachable in practice and the guard is purely defensive, saying that in the PR text works too.

crates/rocm-dash-collectors/src/bench_load.rs
```diff
@@ -423,43 +438,97 @@ async fn run_cell_with_clients(
...
 
-/// Compute `Δsum / Δcount * 1000` (ms).
+/// Windowed latency (ms) across the cell's two histogram scrapes.
+///
+/// Computes `Δsum/Δcount × 1000` — the mean latency of just the requests this
+/// cell issued between the before- and after-scrape. Returns `None` when the
+/// window cannot be measured: either scrape missing, the observation count did
+/// not advance (`Δcount ≤ 0`, no request recorded this metric in the window),
+/// the counter reset (`Δsum < 0`, a server restart), or either delta is `NaN`
+/// (Prometheus reports an unobserved histogram sum that way).
```
rominf: This rationale isn't right, and the same claim appears in the test comment below and in commit `163224ad`'s message.

A Prometheus histogram's `_sum` and `_count` are counters initialised to zero. The Go and Python clients — and `prometheus_client` is what vLLM uses for `vllm:time_to_first_token_seconds` and `vllm:time_per_output_token_seconds` — emit `0.0` and `0` for a histogram with no observations, never NaN. NaN shows up in text exposition for *summary quantiles* with no observations, and for gauges explicitly set to NaN. Neither is one of the four series `vllm_prom.rs` reads.

Keep the guard — it's still worth having, because `extract` does `parse::<f64>()` and will happily accept a `NaN` token from a malformed or non-conforming exporter. It's the justification that needs fixing, and it matters because this premise is exactly what's being used to argue the guard must never be simplified back to the rejection form. An accurate version of that argument is stronger than a wrong one.

crates/rocm-dash-collectors/src/bench_load.rs
```diff
@@ -390,12 +392,25 @@ async fn run_cell_with_clients(
...
+    // rather than borrowing the endpoint's lifetime average, which would fold in
+    // traffic from earlier ramp cells or other clients.
+    let (ttft_ms, tpot_ms) = prom_latency(prom_before.as_ref(), prom_after.as_ref());
+
+    // The `/metrics` scraper only understands vLLM's `vllm:` series, so a
+    // recognisable sample is proof the endpoint is vLLM; a non-vLLM endpoint
+    // leaves the column blank rather than guessing. `engine` is a new column:
+    // rows appended before this change have it blank, and since `engine` is a
+    // rollup key, a shared `results.csv` spanning the upgrade splits a cell's
+    // trials into blank and `vllm` groups — rotate the file to regroup them.
```
rominf: This comment is the only place the migration is disclosed, and I think it needs to be louder than a source comment.

`engine` is a `RollupKey` field (`bench_rollup.rs:39`) and the default output is one shared append-only file (`$HOME/.rocm/bench/results.csv`). Trials of a cell are rows differing only by `run`, so after this lands every pre-upgrade trial (blank `engine`) and every post-upgrade trial (`vllm`) of the same cell group separately — halving N and turning a Pass^N over N trials into two N=1 groups, which `at_n_mark` then renders as a muted dash. The same split happens any time `/metrics` is reachable on one run of a cell and not another: a remote endpoint, metrics disabled, a restart mid-series.

What makes it hard to diagnose is that the Bench rollup table renders `cell/model/tp·dtype/concurrency` but never `engine` — I checked, it appears in `bench.rs` only in a test fixture. So the two groups are visually identical with no on-screen explanation, and `compute_rollup_height` caps the table at 8 data rows, so doubling the group count can push cells off-screen entirely.

"Rotate the file to regroup them" is the right advice; it just needs to reach someone who isn't reading this file. A line in the PR body plus a one-time warning when appending a `vllm` row to a file that already has blank-`engine` rows for the same cell would cover it. Rendering `engine` in the rollup table would also make a split group explicable rather than mysterious.

tests/e2e-cucumber/features/bench.feature
```diff
@@ -6,9 +6,9 @@ Feature: Benchmarking a served endpoint
   # one of the two 404'd — and every request failure was swallowed, so the run
   # still exited 0 with a row of blanks.
   #
-  # bench-01 to bench-03 run on every lane (MockServer-backed, no GPU needed)
-  # and are what pin the request path and the failure reporting. bench-04 is the
-  # hardware proof against a really served model.
+  # bench-01 to bench-04 run on every lane (MockServer-backed, no GPU needed)
```
rominf: Anchoring here because the file I mean isn't in this diff: the same sentence exists in `tests/e2e-cucumber/expectations.toml:156-158`, and it didn't get the update this one correctly did.

It still reads "the three MockServer-backed bench scenarios run unxfail'd on every lane ... and they are what pin the request path and the failure reporting." With bench-04 added, the count is wrong and the enumeration is short by the CSV columns. That sentence is the justification for the `bench-load-real-serve` xfail not weakening coverage, so it's the one place you'd want it accurate.

Nothing breaks functionally — `expectations.toml` is keyed by `@id:` and `bench-load-real-serve` is untouched, so the renumber to bench-05 doesn't invalidate the row. Just mirror the wording you already used here.

rominf (dismissed):
All three points from my last review are properly fixed, and I checked each against the code rather than the replies. The fallback is gone and the value is strictly windowed; the daemon-parity claim is replaced by a note whose three distinctions I verified against `avg_ms_from_histogram` (it really does fall back, really does keep a rolling baseline, really does persist nothing); and the label comes from the registry. Thanks for taking the narrow reading on the fallback rather than papering over it with a provenance column.

Four things below, one of which I think matters more than its size suggests.

The rollup split is the real one. `engine` is a `RollupKey` field and the default output is a single append-only file, so every pre-upgrade trial of a cell and every post-upgrade trial now land in different groups — a Pass^N over N trials quietly becomes two N=1 groups. And the Bench tab never renders `engine`, so on screen those groups are indistinguishable with nothing to explain them. Right now that's disclosed only in a source comment.

The other three are accuracy rather than behaviour, but two of them are load-bearing for a future reader deciding whether the guard can be simplified back.

I did not find problems in the windowing arithmetic itself — the acceptance-phrased guard is strictly stronger than the old rejection form, the divide-by-zero is covered, and a mid-cell restart lands on the reset path. The new scenario also looks deterministic rather than flaky: the mock's `/metrics` tick advances per GET and the poller is awaited before the after-scrape.

r0x0r:
Thanks — all four are addressed in `26ff4e26`.

**1. The rollup split.** Both halves are in: the PR body now has a "Migration: a shared `results.csv` written across this change" section stating the rollup-key consequence (N trials → two groups, `at_n_mark`'s muted dash, no `engine` column on screen, the 8-row cap), and the run warns for itself. While holding the append lock, a row that carries an `engine` scans the target file for earlier rows of the same cell without one; the sweep raises it at most once and the CLI prints a warning naming the file, the cell, and the fix. The row is still appended — warning, not refusal. Detection is best-effort (a malformed line is skipped, not fatal) and only runs for `engine`-carrying rows. Covered by unit tests (flagged for the same cell; quiet when the earlier rows already carry the engine, when they are another cell, when the new row has no engine, on a fresh file; raised once per multi-cell sweep; plus a guard that the scanned column indexes still match `CSV_HEADER`) and by a new ungated scenario `@id:bench-load-warns-on-engine-split` (bench-05; the GPU scenario renumbers to bench-06). The scenario builds its fixture by running the benchmark once and blanking the `engine` column of the row it wrote, rather than pasting one — a hand-written header would rot into a header-mismatch failure that proves nothing, and blanking is also exactly what a run with no reachable `/metrics` leaves behind.

You are right that rendering `engine` in the rollup table would make a split group explicable rather than mysterious, and I'd like that — but it is a TUI change that interacts with `compute_rollup_height`'s 8-row cap (an extra column is cheap, the extra *groups* are what overflow), so I've left it out of this PR and said so in the body. Happy to open it as a follow-up if you agree that's the right split.

**2. PR body vs. NaN.** Corrected. The body now has this as its own bullet: the old rejection-phrased guard let a NaN delta return `Some(NaN)`, which `opt_f64` wrote as a literal `NaN` cell, and the acceptance form makes that cell blank — a change to what lands in the file, not a rename. The sentence claiming the computed values were unchanged is gone.

On scenario-vs-unit-test, I took your second option and argued it in the body: the guard is **purely defensive**, so there is no user-observable behaviour to script. That follows directly from point 3 — if no conforming exporter can produce the NaN, then any scenario would be asserting against a fixture that cannot occur in the field, and it would pin the mock rather than the product. The user-visible part of this change is the blank-vs-`NaN` cell, and that is now stated in the body rather than only in a commit message.

**3. The NaN rationale was wrong — my error.** You are right and I checked it rather than taking it on faith. A histogram's `_sum`/`_count` are counters that `prometheus_client` initialises to `0.0`/`0`; NaN in text exposition comes from summary quantiles with no observations and from gauges explicitly set to NaN, and none of the four series `vllm_prom` reads is either. I kept the guard on the accurate ground you gave: `extract` finishes with `parse::<f64>()`, which accepts a `NaN` token from a malformed or non-conforming exporter, and this is the last place to stop one before an immutable CSV row. Both code sites are fixed — the docstring/inline comment at `latency_ms` and the comment in `latency_ms_is_none_when_a_counter_is_nan`.

The third site is `163224ad`'s commit message, which says "Prometheus reports an unobserved histogram sum as NaN and vllm_prom parses it verbatim". That commit is pushed and I'm not force-pushing this branch to reword it, so **it stands with a superseded rationale**: the accurate one is the "The NaN guard: corrected rationale" section of the PR body and the comments in `26ff4e26`. Anyone reading the guard from the code will find the correct reasoning; only the commit log carries the old claim.

**4. `expectations.toml`.** Fixed to mirror the feature file: five MockServer-backed bench scenarios (bench-01 to bench-05), pinning the request path, the failure reporting, the CSV columns, and the rollup-split warning.

**Verification.** `cargo test --workspace --all-targets` and `cargo clippy --workspace --all-targets -- -D warnings` clean. `cargo xtask e2e`: the bench feature is green end to end (bench-01 … bench-06); the run's only unexpected failures were `serve-vllm-inference` and `serve-hf-checkpoint-inference`, which are the GPU-serving scenarios that fail on my GPU-less box and pass on the hardware lanes. Every test added here was falsified first — each confirmed red against a deliberate break of the code it covers (suppressing the notice, dropping the once-per-sweep guard, dropping the cell match, pointing the scan at the wrong column, and suppressing the CLI warning for bench-05) and green again after restoring it.

r0x0r:
Thanks — all five non-blocking items triaged against the current head. Two were already addressed by `26ff4e26`/`1d341d82`, which landed after the revision you reviewed (`163224ad`); three are fixed in `8a18a68f` and `72577a0b`.

**1. Loose `tpot_ms` assertion in `bench-04` — fixed (`72577a0b`).** Correct, and the arithmetic checks out: the mock adds 0.4 s of `_sum` per 20 `_count` per scrape, so `Δsum/Δcount` is exactly 0.020 s regardless of how many scrapes the window spans. Now asserts `(v - 20.0).abs() < 0.5`. Falsified: with the constant set to 40.0 the scenario fails (`bench-load-records-engine-and-tpot` — unexpected failure); at 20.0 it passes.

**2. `±Inf` slips past the guard — fixed (`8a18a68f`).** Confirmed: `"+Inf"` parses via `parse::<f64>()`, an infinite `Δsum` satisfies `>= 0.0`, and `opt_f64`'s `{:.6}` renders it `inf`. Guarded both deltas rather than only the result, because the reverse case (infinite `Δcount`) would otherwise collapse a real `Δsum` to `0.000000` instead of blanking it. New test `latency_ms_is_none_when_a_counter_is_infinite`; falsified by dropping the two `is_finite` conjuncts — fails with `left: Some(inf), right: None`, while the NaN test still passes, so the two cover distinct defects.

**3. NaN rationale — already corrected at `1d341d82`.** The comment at `bench_load.rs` and the unit test now say what you asked for: `prometheus_client` initialises histogram `_sum`/`_count` to zero and never emits NaN; NaN appears for summary quantiles and explicitly-set gauges, neither of which `vllm_prom` reads. The guard is justified as "`extract` ends in `parse::<f64>()`, which accepts the token verbatim". `163224ad`'s commit message cannot be corrected without rewriting a pushed branch; the PR body records it as superseded.

**4. Rollup split documented only in a comment — already addressed at `26ff4e26`.** A one-time warning now fires under the append lock when an `engine`-carrying row is appended to a file already holding blank-`engine` rows for the same cell; it names the file, the cell, and says to rotate. Covered by scenario `@id:bench-load-warns-on-engine-split` (`bench-05`), and by the PR body's migration section. There is no user-facing doc that describes the bench CSV to update — as you noted, nothing in `docs/` references these columns.

**5a. Duplicated `run_bench` — fixed (`72577a0b`).** `run_bench` now takes an optional output path and builds the argv once; `run_bench_recording` is a two-line wrapper.

**5b. `col()` splits on `,` without CSV quoting — not changing this one.** Agreed it is fragile in principle, but the production reader is not: `has_blank_engine_row` uses `csv::Reader`, so the indexed `ENGINE_COLUMN` lookup is quoting-aware. The exposure is confined to two test helpers whose inputs are fixed by the scenarios (`bench-c1`, `TestModel/E2E-1B`, numerics). Making them quoting-aware means adding a `csv` dependency to the E2E crate for test-only plumbing, which is a separate change from this one (AGENTS.md §11). Happy to take it as a follow-up if you would rather it not wait.

Head is now `72577a0b`. `cargo test --workspace --all-targets` and `cargo clippy --workspace --all-targets -- -D warnings` both clean; `cargo xtask e2e` green across bench-01 … bench-06 (the one unexpected failure on my box is `serve-08`, a lemonade GGUF serve scenario failing on a missing `libdrm_amdgpu.so.1` — unrelated to this change and not in the diff's blast radius).

siloteemu:
🔴 Automated review · pr-review-watcher · 72577a0b

## Summary
Populates the bench CSV's `engine` column from the vLLM `/metrics` scrape, makes TTFT/TPOT strictly windowed (Δsum/Δcount) with a NaN/±Inf guard, and warns once per run when a results file already holds blank-`engine` rows for the same cell — plus two new MockServer-backed e2e scenarios. No blocking findings. Verified: `cargo test -p rocm-dash-collectors --lib bench_load` → 46 passed, 0 failed; confirmed `engine` really is a `RollupKey` field (`crates/rocm-dash-core/src/bench_rollup.rs:39`) so the split-warning premise holds; confirmed `CSV_HEADER` order matches `serialize_row_to_line`'s explicit 18-field write and `CELL_COLUMN`/`ENGINE_COLUMN` (0/4); confirmed the e2e mock's TPOT really yields exactly 20 ms (`tests/e2e-cucumber/src/mock_server.rs:406-407`, `tpot_sum_s/tpot_count = 0.020 s` for any tick, so the windowed delta is constant), matching `EXPECTED_TPOT_MS`; confirmed the bench-04→bench-06 renumber was *required* by the `scenario_names_are_indexed_sequentially_per_feature` guard in `tests/e2e-cucumber/tests/feature_naming.rs:131`, not cosmetic. On focus (a): the new-feature tests (`detect_engine_*`, `append_flags_*`, `append_is_quiet_*`, `engine_split_is_reported_once_per_sweep`, both `bench_row_*`, `scan_column_indexes_match_the_header`) all exercise code that does not exist at `prw-base`, so a revert breaks them; the NaN and ±Inf tests are the only ones that fail on *behaviour* rather than on the rename (the old `if delta_count <= 0.0 || delta_sum < 0.0` guard returns `Some(NaN)`/`Some(inf)` for those inputs) — the other `latency_ms_*` cases assert behaviour identical to the pre-PR `latency_delta_ms`. Leak scan over the diff: clean. No prompt-injection content found. CI at this head is inconclusive (most checks unfinished), so nothing here rests on it. Blocking: 0 · Non-blocking: 5.

**None.**

## Non-blocking
- `crates/rocm-dash-collectors/src/bench_load.rs:431` — the comment calls `engine` "a new column", but `CSV_HEADER` already contained `engine` at `prw-base` (only its *population* is new); a reader who believes the header changed would conclude old files fail the byte-identical header check and that this whole split-detection path is unreachable, so "newly populated column" is worth the two-word fix.
- `crates/rocm-dash-collectors/src/bench_load.rs:656` — "the common case … costs one pass at most" is backwards: a file with *no* blank-`engine` rows for the cell (the steady state) costs a **full** scan, once per cell, so an 8-cell ramp is up to 8 full passes over a growing file; reword to say the early exit only helps when a hit exists.
- `crates/rocm-dash-collectors/src/bench_load.rs:1113` — the new `assert!(…is_none())` in `concurrent_first_appends_write_one_header` is vacuous: every row is `..Default::default()` so `engine` is `None` and the scan is gated out by `if row.engine.is_some()`, meaning the comment "a notice here would mean the scan fires on the wrong rows" is false (the scan never runs); either drop the assertion or give one writer `engine: Some("vllm")`.
- `tests/e2e-cucumber/tests/e2e/bench_steps.rs:191` and `:220` — `blank_engine_column` and the `col()` helper parse the CSV with naive `.split(',')`/`join(",")` while production writes and reads it with the `csv` crate; safe for today's comma-free fixtures (`launcher` is `rocm bench load (local smoke)`, model is `TestModel/E2E-1B`) but it silently misaligns the row for the next author who uses a comma-bearing model name.
- `apps/rocm/src/dash.rs:601` — the warning hardcodes `engine=vllm` rather than reading the row's actual engine; correct today because `detect_engine` can only return the vLLM label, but it goes stale silently the moment a second engine is detected here.

siloteemu:
🔴 Automated review · pr-review-watcher · 46b580d3

## Summary

Populates the `engine` column in `bench load`'s results CSV, keeps `tpot_ms`/`ttft_ms` strictly windowed, tightens the latency guard to reject non-finite deltas, and — new since the blocking review — detects and warns when a results file straddles the point at which `engine` started being populated. Verified: ran `cargo test -p rocm-dash-collectors --lib bench_load` (46 passed, 0 failed); confirmed against `git show prw-base:...` that `CSV_HEADER` puts `cell` at index 0 and `engine` at index 4 (matching `CELL_COLUMN`/`ENGINE_COLUMN`), that `latency_delta_ms` at the base was already windowed with no fallback, and that `engine` was present-but-never-populated at the base; confirmed each new unit test fails against the pre-PR guard, and that the new e2e scenario asserts on stderr the base never emits. Blocking: 0 · Non-blocking: 5.

## Status of the existing change request

The operative change request is the 2026-09-10 one at `163224ad`; its own body confirms the three earlier points were fixed, and raises four new ones. The author has pushed five commits since (`26ff4e2`, `1d341d8`, `8a18a68`, `72577a0`, `46b580d`). Point by point:

**1. "The rollup split is the real one … it needs to be louder than a source comment" — a line in the PR body plus a one-time warning when appending a `vllm` row to a file that already has blank-`engine` rows for the same cell; rendering `engine` in the rollup table "would also" help.**

- The warning: **appears addressed.** `append_one_row` now scans under the same advisory lock it already holds and returns an `EngineSplitNotice`; `attach_engine_split` gates it to once per sweep; `apps/rocm/src/dash.rs:604-612` prints it to stderr naming the file and the cell, plus a rotate hint. The append still succeeds — warn, not refuse. Covered by three unit tests and by a new ungated e2e scenario (`@id:bench-load-warns-on-engine-split`). I confirmed the once-per-sweep state is a fresh local per call to `run_and_append_csv`/`run_auto_ramp`, and that both entry points feed the same `for report in &reports` loop the warning sits in.
- The PR-body line: **cannot tell.** This automation does not contact GitHub, so I cannot read the PR description at this head.
- Rendering `engine` in the rollup table: **still outstanding**, but the reviewer framed it as an additional "would also", not part of the ask. Both underlying facts they cited still hold and are unchanged by this PR: the rollup header in the Bench tab is `cell, model, cfg, N, Pass^N, Pass@N, meanPTPS, meanGTPS` with no `engine`, and `compute_rollup_height` still caps at 8 data rows. Those files are outside this diff, and the new `EngineSplitNotice` doc comment names both as the reason the warning exists.

**2. "`the computed values are unchanged` isn't quite true … it belongs in the body rather than only in the tip commit's message. Per AGENTS.md §3 that also wants a scenario … or argue in the PR text that NaN is unreachable and the guard is purely defensive."**

- The substance of the "defensive / unreachable in practice" argument: **appears addressed in source** — the inline comment at the guard now states plainly that a conforming vLLM cannot produce that NaN and that the guard is "defensive, not a response to an observed exposition", with the reachable route (`extract` ending in `parse::<f64>()` accepting a `NaN` token from a non-conforming exporter) spelled out. The test comment carries the same corrected framing.
- Whether that landed in the **PR body**: **cannot tell** — same reason as above.
- No e2e scenario for the NaN path was added. That is consistent with the second branch the reviewer explicitly offered ("if you'd rather argue NaN is unreachable in practice and the guard is purely defensive, saying that in the PR text works too"), so it is only outstanding if the reviewer wanted the scenario branch specifically.
- Beyond the ask, the author widened the guard to `is_finite()` on both deltas. I checked this is not redundant: an infinite `Δsum` satisfies `>= 0.0` and an infinite `Δcount` satisfies `> 0.0`, so both conjuncts are load-bearing, and the new `latency_ms_is_none_when_a_counter_is_infinite` fails against the base guard.

**3. "This rationale isn't right … `_sum`/`_count` are counters initialised to zero; NaN shows up for summary quantiles and explicitly-NaN gauges. Keep the guard, fix the justification."**

**Appears addressed.** The docstring no longer claims Prometheus reports an unobserved histogram sum as NaN; the replacement reproduces the reviewer's correction (counters initialised to `0.0`/`0` by `prometheus_client`; NaN belongs to summary quantiles and explicit gauges; neither is among the four series read). The test comment below it is updated too. One residue: commit `163224a`'s message still carries the refuted claim — it cannot be corrected in place without a history rewrite, and `26ff4e2`'s message says so explicitly and supersedes it. Reasonable, but a reader of `git log` alone will still meet the wrong version.

**4. "`expectations.toml:156-158` still says three MockServer-backed bench scenarios; mirror the wording you already used in the feature file."**

**Appears addressed**, and the counts are now right. I counted the feature file myself: `bench-01`..`bench-05` are MockServer-backed and ungated, `bench-06` is the `@requires-gpu` hardware proof. `expectations.toml` now reads "the five MockServer-backed bench scenarios (bench-01 to bench-05)" and enumerates the request path, failure reporting, CSV columns and rollup-split warning — matching `bench.feature`'s wording and the actual scenario set. The `bench-load-real-serve` row is keyed by `@id:` and is untouched by the bench-05→bench-06 renumber, as the reviewer noted.

Net: of the four, three appear substantively addressed in code; the two PR-body asks cannot be judged from here, and one optional suggestion (rendering `engine` in the rollup table) was not taken.

## Nothing new is wrong at the current head

I found no blocking defect in the code pushed after the 09-10 review. Specifically checked and clean: the `seek(0)` + rescan inside `append_one_row` does not disturb the append position (`OpenOptions` sets `append(true)`, so `O_APPEND` puts the write at EOF regardless of the read cursor) and happens under the exclusive lock already held, so a concurrent writer cannot slip a row between the scan and the append; `CellReport`'s new public field has exactly one construction site and no pattern-matches elsewhere in the workspace; both `append_one_row` callers consume the notice; `Option::is_none_or` is well under the pinned MSRV of 1.96; the leak scan over the diff is clean (the one `EAI-` ticket id in `expectations.toml` is explicitly permitted by this repo and is not flagged). No prompt-injection content was found in the diff, comments, commit messages or the reviewer-comment file.

Per the standing focus on tests: every new unit test fails if the matching production change is reverted (I traced the NaN/Inf cases by hand against the base guard `delta_count <= 0.0 || delta_sum < 0.0`, and the three engine-split tests against the base `append_one_row`, which returned `Result<(), _>`). The new e2e scenario likewise fails on the base, which emits no such warning. Two exceptions are noted below as non-blocking.

## Measured check state

26 success, 1 skipped, 0 failure, 0 pending at this head. No per-lane detail is available to me; I cannot attribute any of those outcomes to a named job.

**None.**

## Non-blocking

- `apps/rocm/src/dash.rs:606` — the warning text hardcodes `engine=vllm`, but the row's engine now comes from the registry and `EngineSplitNotice` does not carry it; this is the same drift class the reviewer flagged earlier at the label site, just in a user-facing string.
- `crates/rocm-dash-collectors/src/bench_load.rs:658` — `has_blank_engine_row` does a full linear pass on every labelled append, not once per sweep, so a long-lived shared `results.csv` costs O(cells × file size) per run and degrades as it grows; the doc's "one pass at most" is true per call but does not surface that.
- `tests/e2e-cucumber/tests/e2e/bench_steps.rs:188` — `blank_engine_column` splits on `,` rather than using the `csv` crate, so it would corrupt a row if any field ever contained a comma; safe for today's fixture data only.
- `tests/e2e-cucumber/tests/e2e/bench_steps.rs:279` — the tightened `tpot_ms ≈ 20.0` assertion in bench-04 would still pass with this PR's production changes reverted: the base already computed `Δsum/Δcount × 1000` windowed. It is a genuine coverage addition, but only the `engine == vllm` half of that scenario pins new behaviour, and the scenario comment reads as if both do.
- `crates/rocm-dash-collectors/src/bench_load.rs:1108` — the `assert!(…is_none())` added to the concurrent-append test never reaches the detection branch, since every row it writes has `engine == None`; it guards a false positive rather than exercising the scan.

crates/rocm-dash-collectors/src/bench_load.rs
jussielo-amd: Fixed — the PR body now has a dedicated section ("One CSV cell changes value, not only its name") saying exactly this: the acceptance-phrased guard (introduced in `419052a`) turns a NaN delta from `Some(NaN)` written as a literal `NaN` cell into `None` written as blank, and retracts the earlier body text that claimed values were unchanged. `8a18a68` extends the same guard to reject ±Inf deltas for the same reason. Left `419052a`'s message as-is since rewriting a pushed commit isn't worth it — the body is the corrected record.

crates/rocm-dash-collectors/src/bench_load.rs
jussielo-amd: Fixed in `26ff4e2` — reworded both the doc comment and the test comment to the accurate reasoning (histogram `_sum`/`_count` are counters initialised to 0, not NaN; NaN only shows up for unobserved summary quantiles or explicit-NaN gauges, neither of which applies here) and kept the guard on the correct justification (`parse::<f64>()` accepts a NaN token from a non-conforming exporter). `163224ad`'s commit message still has the old claim — can't fix that without rewriting a pushed commit — so the PR body notes it's superseded.

crates/rocm-dash-collectors/src/bench_load.rs
jussielo-amd: Fixed in `26ff4e2` — added `EngineSplitNotice` detection in `bench_load.rs` and a runtime warning (in `apps/rocm/src/dash.rs`) printed when a `vllm`-labeled row is appended alongside earlier blank-`engine` rows for the same cell, naming the file and the cell. Also added a "Migration" section to the PR body covering the mechanics. Didn't add `engine` to the rollup table itself — felt like a bigger UI change and out of scope here; can open a follow-up if you'd like that tracked.

tests/e2e-cucumber/features/bench.feature
jussielo-amd: Fixed in `26ff4e2` — `expectations.toml`'s xfail justification now mirrors `bench.feature`'s wording (five MockServer-backed scenarios, bench-01 to bench-05, covering request path/failure reporting/CSV columns/rollup-split warning).

jussielo-amd (approved):
Reviewed with an automated multi-phase review pass (fan-out review, cross-file/blast-radius synthesis, design review, verification against source, plus a clean local test/clippy/e2e run). No blocking issues.

**Summary:** populates the previously-blank `engine` column in `results.csv` (sourced from the `EngineKind` registry, not a re-typed literal), switches `tpot_ms` to a windowed delta between consecutive Prometheus scrapes instead of a lifetime average, guards that computation against non-finite (NaN/Inf) histogram deltas from a malformed exporter, and warns when a `results.csv` mixes rows with/without `engine` (since it participates in rollup grouping). Clean, well-scoped commit history — one commit even publicly corrects a wrong rationale from an earlier commit's message.

**Non-blocking:** `dash.rs` (~line 606) hardcodes `"engine=vllm"` in the split-warning text, but `EngineSplitNotice` doesn't actually carry an engine name. Harmless today since `detect_engine` can only return `Some("vllm")` or `None`, but it's a forward-coupling risk if a second engine is added later — worth a low-priority follow-up when that happens.

**Disclosed tradeoffs (accepted, not gaps):** TUI rollup table doesn't surface `engine` yet (deferred due to the 8-row height cap); `has_blank_engine_row` is intentionally best-effort and skips malformed CSV rows rather than failing the append — both documented in-line with clear reasoning.

**Verified:** clippy clean; 103 unit tests passed (2 pre-existing hardware-gated ignores, unrelated); e2e bench scenarios (mock-backed, including the two new ones) all passed. The GPU-hardware scenario failed only due to no GPU in this sandbox — pre-existing, unrelated to this PR.

Safe to merge as-is.

PR #331, opened 2026-09-01
fix(dash): validate --replay path before entering the TUI

r0x0r:
## Summary

`rocm dash --replay <path>` did no upfront validation of the replay path. A missing or unreadable recording was passed straight into the dashboard, so the failure only surfaced as a disconnect inside the TUI *after* the alt-screen had already taken over the terminal (EAI-8366).

This validates the user-supplied replay path in `dash::run` before building the runtime or entering the alt-screen. A missing file, a directory, or an unreadable regular file now fails fast with a clear error and a non-zero exit. `--demo` generates its own session and is exempt.

**Before:** under a real terminal, `rocm dash --replay /nonexistent/file` enters the alt-screen and hangs there.
**After:** it prints `replay file not found: /nonexistent/file` and exits non-zero without taking over the terminal.

## Changes

- `apps/rocm/src/dash.rs`: add `validate_replay_path()` and call it first in `run()`.
  - Failure kinds are distinguished off `io::Error::kind()`, so a permission-denied `stat` reports `cannot read replay file: …` rather than being mislabelled "not found".
  - Only a *directory* is rejected on shape. Requiring a regular file would break inputs the replay reader (`std::fs::read_to_string`) handles today — FIFOs, `/dev/stdin`, process substitution (`--replay <(zcat rec.ndjson.gz)`).
  - The readability probe (`File::open`) runs only for regular files, so it never blocks on or consumes a one-shot stream.
- Unit tests for the missing / directory / readable-file cases, plus — on unix — the unreadable (mode `0o000`, skipped as root) and non-regular-input cases. Five tests, one per documented rejection/acceptance branch.
  - The non-regular case uses a **writer-less FIFO** (`libc::mkfifo`), validated on a worker thread that must report back inside a 10s timeout. It pins both halves of the contract: a FIFO is not a shape error, *and* validation does not pre-open it. An earlier `/dev/null` version could not do the latter — a character device opens instantly, so it passed with or without the `is_file()` narrowing.
- E2E scenario `@id:dash-replay-missing-file-fails-fast` (`dash-11`) + steps, driven through a **real pseudo-terminal** via `TuiSession`, matching the rest of `dash.feature`. It asserts the process exits non-zero *without* ever entering the alt-screen, and that the "replay file not found" message reaches the screen.
- `tests/e2e-cucumber/tests/e2e/tui_driver.rs`: `in_alternate_screen()`, `wait_for_refusal()` and `drain_final_screen()` helpers backing the scenario.

## Testing

Verified on Linux (the fail-fast property is unobservable through a pipe — a piped `dash` cannot enter the alt-screen either way — so the scenario has to run under a PTY):

- `cargo test --workspace --all-targets` — green; the five `dash::tests::validate_replay_path_*` tests pass.
- `cargo xtask e2e -- -n "Replaying a missing recording"` — 1 scenario, 3 steps, passed.
- **Regression proof (scenario):** with the `validate_replay_path` call neutralised (pre-fix behaviour) the same scenario fails: `timed out after 30s waiting for 'dash --replay <missing>' to be refused — it did not exit`. So it genuinely pins the bug rather than passing either way.
- **Regression proof (`is_file()` narrowing):** with the `if meta.is_file()` guard removed so `File::open` runs unconditionally, `validate_replay_path_accepts_a_fifo_without_opening_it` fails after 10s (`validate_replay_path never returned for a writer-less FIFO (timed out waiting on channel)`) while the other four tests still pass; restored, all five pass.

**Gated lane:** the scenario is tagged `@requires-os:linux`, like every other scenario in `dash.feature` (portable-pty compiles on Windows via ConPTY but that path is not yet a blocking contract). It is exercised by the Linux E2E lanes, not by the Windows ones.

The author's local machine is macOS, an unsupported platform for this repo, so all gates above were run on a Linux host rather than locally; CI on Linux is authoritative.

### DIFF
--- a/apps/rocm/src/dash.rs
+++ b/apps/rocm/src/dash.rs
@@ -294,8 +294,54 @@ fn prune_stale_demo_sessions(dir: &Path, keep: &Path) {
     }
 }
 
+/// Validate a user-supplied `--replay` path before the dashboard takes over the
+/// terminal. Fails fast with a clear error (and, via the returned `Err`, a
+/// non-zero exit) when the file is missing, is a directory, or is not readable —
+/// rather than entering the alt-screen and only then surfacing a disconnect
+/// inside the TUI. Pure filesystem check → unit-testable without a terminal.
+///
+/// Only a *directory* is rejected on shape, not "not a regular file": the replay
+/// reader is `std::fs::read_to_string`, which happily reads FIFOs, `/dev/stdin`,
+/// and process-substitution paths (`--replay <(zcat rec.ndjson.gz)`), all of
+/// which `metadata().is_file()` would reject. A missing-vs-unreadable distinction
+/// is made by matching the `metadata` error kind, so a permission-denied parent
+/// is not mislabelled "not found".
+fn validate_replay_path(path: &Path) -> Result<()> {
+    let meta = std::fs::metadata(path).map_err(|err| match err.kind() {
+        std::io::ErrorKind::NotFound => {
+            anyhow::anyhow!("replay file not found: {}", path.display())
+        }
+        _ => anyhow::anyhow!("cannot read replay file: {}: {err}", path.display()),
+    })?;
+    if meta.is_dir() {
+        anyhow::bail!(
+            "replay path is a directory, not a recording: {}",
+            path.display()
+        );
+    }
+    // Confirm a *regular* file actually opens for reading — a successful `stat`
+    // only needs a traversable parent, not a readable file — so a mode-0
+    // recording fails here instead of after the alt-screen has taken over. Only
+    // regular files are pre-opened: opening a FIFO / `/dev/stdin` / process
+    // substitution would block on, or consume, the one-shot stream the replay
+    // reader (`read_to_string`) is about to read, so those are left to the reader.
+    if meta.is_file() {
+        std::fs::File::open(path)
+            .with_context(|| format!("replay file is not readable: {}", path.display()))?;
+    }
+    Ok(())
+}
+
 /// Entry point for `rocm dash`. Builds a tokio runtime and runs the dashboard.
 pub fn run(replay: Option<PathBuf>, demo: bool, chat_mock: bool) -> Result<()> {
+    // Validate a user-supplied `--replay` path up front — before building the
+    // runtime or entering the alt-screen — so a missing/unreadable file fails
+    // fast with a clear error and non-zero exit instead of silently taking over
+    // the terminal and surfacing a disconnect inside the TUI. `--demo` writes
+    // its own session below, so it is exempt.
+    if !demo && let Some(path) = replay.as_deref() {
+        validate_replay_path(path)?;
+    }
     let paths = AppPaths::discover()?;
     let config = RocmCliConfig::load(&paths)?;
     // `--demo` writes a synthetic session and replays it, so the dashboard shows
@@ -1018,5 +1064,111 @@ mod tests {
         );
 
         let _ = std::fs::remove_dir_all(&root);
+    }
+
+    /// EAI-8366 regression: `rocm dash --replay <missing>` must fail the upfront
+    /// validation with a clear "not found" error (→ non-zero exit) instead of
+    /// entering the alt-screen and surfacing a disconnect inside the TUI.
+    #[test]
+    fn validate_replay_path_rejects_missing_file() {
+        let missing = std::env::temp_dir().join(format!(
+            "rocm-cli-replay-missing-{}-{}.ndjson",
+            std::process::id(),
+            rocm_core::unix_time_millis()
+        ));
+        // Guard against a pre-existing file from a prior crashed run.
+        let _ = std::fs::remove_file(&missing);
+
+        let err = validate_replay_path(&missing).expect_err("missing file must be rejected");
+        let msg = format!("{err:#}");
+        assert!(
+            msg.contains("replay file not found"),
+            "error must name the missing replay file: {msg}"
+        );
+    }
+
+    /// A directory passed to `--replay` is present but is not a replayable
+    /// recording; validation must reject it up front rather than in the TUI.
+    #[test]
+    fn validate_replay_path_rejects_directory() {
+        let dir = std::env::temp_dir().join(format!(
+            "rocm-cli-replay-dir-{}-{}",
+            std::process::id(),
+            rocm_core::unix_time_millis()
+        ));
+        std::fs::create_dir_all(&dir).unwrap();
+
+        let err = validate_replay_path(&dir).expect_err("a directory must be rejected");
+        let msg = format!("{err:#}");
+        assert!(
+            msg.contains("is a directory"),
+            "error must explain the path is a directory: {msg}"
+        );
+
+        let _ = std::fs::remove_dir_all(&dir);
+    }
+
+    /// EAI-8366 regression for the readability branch the PR advertises but did
+    /// not cover: a present, regular recording that cannot be opened (mode 0o000)
+    /// must be rejected up front — not after the alt-screen has taken over. Unix
+    /// only (Windows has no equivalent open-time read bit) and skipped as root,
+    /// where the mode is not enforced.
+    #[cfg(unix)]
+    #[test]
+    fn validate_replay_path_rejects_unreadable_file() {
+        use std::os::unix::fs::PermissionsExt;
+
+        if rocm_core::openmpi::running_as_root() {
+            // Mode 0o000 is not enforced for uid 0, so the open would succeed and
+            // the branch under test cannot be exercised.
+            return;
+        }
+
+        let file = std::env::temp_dir().join(format!(
+            "rocm-cli-replay-unreadable-{}-{}.ndjson",
+            std::process::id(),
+            rocm_core::unix_time_millis()
+        ));
+        std::fs::write(&file, "{}\n").unwrap();
+        std::fs::set_permissions(&file, std::fs::Permissions::from_mode(0o000)).unwrap();
+
+        let err = validate_replay_path(&file).expect_err("an unreadable file must be rejected");
+        let msg = format!("{err:#}");
+        assert!(
+            msg.contains("not readable") || msg.contains("Permission denied"),
+            "error must explain the file is not readable: {msg}"
+        );
+
+        // Restore permissions so cleanup can remove the file.
+        let _ = std::fs::set_permissions(&file, std::fs::Permissions::from_mode(0o600));
+        let _ = std::fs::remove_file(&file);
+    }
+
+    /// EAI-8366 regression: a readable *non-regular* input (here `/dev/null`, a
+    /// character device — a stand-in for the FIFO / `/dev/stdin` / process-
+    /// substitution paths the replay reader handles) must NOT be rejected on
+    /// shape. Only a directory is a shape error; requiring a regular file would
+    /// break streaming replays.
+    #[cfg(unix)]
+    #[test]
+    fn validate_replay_path_accepts_non_regular_readable_input() {
+        validate_replay_path(Path::new("/dev/null"))
+            .expect("a readable non-regular input must pass validation");
+    }
+
+    /// A present, readable recording passes validation so the normal replay
+    /// flow proceeds into the dashboard unchanged.
+    #[test]
+    fn validate_replay_path_accepts_readable_file() {
+        let file = std::env::temp_dir().join(format!(
+            "rocm-cli-replay-ok-{}-{}.ndjson",
+            std::process::id(),
+            rocm_core::unix_time_millis()
+        ));
+        std::fs::write(&file, "{}\n").unwrap();
+
+        validate_replay_path(&file).expect("a readable file must pass validation");
+
+        let _ = std::fs::remove_file(&file);
     }
 }
--- a/apps/rocm/src/dash.rs
+++ b/apps/rocm/src/dash.rs
@@ -338,7 +338,11 @@ pub fn run(replay: Option<PathBuf>, demo: bool, chat_mock: bool) -> Result<()> {
     // runtime or entering the alt-screen — so a missing/unreadable file fails
     // fast with a clear error and non-zero exit instead of silently taking over
     // the terminal and surfacing a disconnect inside the TUI. `--demo` writes
-    // its own session below, so it is exempt.
+    // its own session below, so it is exempt: `--demo` overwrites `replay` with
+    // that generated path, so a caller-supplied one is never read in demo mode
+    // and must not be rejected. `conflicts_with` rules the pair out on the CLI,
+    // but this is a `pub fn` whose two flags arrive independently, so the guard
+    // is a real precondition at this boundary rather than dead code.
     if !demo && let Some(path) = replay.as_deref() {
         validate_replay_path(path)?;
     }
@@ -1158,6 +1162,11 @@ mod tests {
     ///
     /// A readable character device such as `/dev/null` cannot stand in here: it
     /// opens instantly, so it passes with or without the narrowing.
+    ///
+    /// Unix only: `mkfifo` has no Windows equivalent, and Windows named pipes
+    /// live in the `\\.\pipe\` namespace rather than on the filesystem, so the
+    /// blocking-open behaviour this pins cannot be staged there. The narrowing
+    /// under test is platform-independent, so this covers it for every target.
     #[cfg(unix)]
     #[allow(unsafe_code)] // libc FFI: mkfifo has no std equivalent
     #[test]
--- a/apps/rocm/src/dash.rs
+++ b/apps/rocm/src/dash.rs
@@ -1144,16 +1144,56 @@ mod tests {
         let _ = std::fs::remove_file(&file);
     }
 
-    /// EAI-8366 regression: a readable *non-regular* input (here `/dev/null`, a
-    /// character device — a stand-in for the FIFO / `/dev/stdin` / process-
-    /// substitution paths the replay reader handles) must NOT be rejected on
-    /// shape. Only a directory is a shape error; requiring a regular file would
-    /// break streaming replays.
+    /// EAI-8366 regression covering *both* halves of the non-regular-file
+    /// contract, on a real one-shot stream rather than a stand-in:
+    ///
+    /// 1. a FIFO must not be rejected on shape — only a directory is a shape
+    ///    error, and requiring a regular file would break `--replay <(...)`;
+    /// 2. validation must not *pre-open* it. `open(2)` on a writer-less FIFO
+    ///    blocks until a writer arrives, so dropping the `is_file()` narrowing
+    ///    around the readability probe hangs `rocm dash` before the TUI starts
+    ///    (and, with a writer, would consume the stream the replay reader is
+    ///    about to read). The validation therefore runs on a worker thread and
+    ///    must report back well inside the timeout below.
+    ///
+    /// A readable character device such as `/dev/null` cannot stand in here: it
+    /// opens instantly, so it passes with or without the narrowing.
     #[cfg(unix)]
+    #[allow(unsafe_code)] // libc FFI: mkfifo has no std equivalent
     #[test]
-    fn validate_replay_path_accepts_non_regular_readable_input() {
-        validate_replay_path(Path::new("/dev/null"))
-            .expect("a readable non-regular input must pass validation");
+    fn validate_replay_path_accepts_a_fifo_without_opening_it() {
+        use std::os::unix::ffi::OsStrExt;
+
+        let fifo = std::env::temp_dir().join(format!(
+            "rocm-cli-replay-fifo-{}-{}",
+            std::process::id(),
+            rocm_core::unix_time_millis()
+        ));
+        let c_path = std::ffi::CString::new(fifo.as_os_str().as_bytes()).unwrap();
+        // SAFETY: `c_path` is a valid NUL-terminated C string that outlives the
+        // call, and `mkfifo` only reads it.
+        let rc = unsafe { libc::mkfifo(c_path.as_ptr(), 0o600) };
+        assert_eq!(rc, 0, "mkfifo failed: {}", std::io::Error::last_os_error());
+
+        // No writer is ever opened for this FIFO, so any `open` of it blocks
+        // indefinitely. The worker thread is deliberately not joined: if the
+        // validation regresses it stays parked in `open` until the test process
+        // exits, and the timeout below is what fails the test.
+        let (tx, rx) = std::sync::mpsc::channel();
+        let probe = fifo.clone();
+        std::thread::spawn(move || {
+            let _ = tx.send(validate_replay_path(&probe).map_err(|err| format!("{err:#}")));
+        });
+        let outcome = rx.recv_timeout(std::time::Duration::from_secs(10));
+
+        let _ = std::fs::remove_file(&fifo);
+        let validation = outcome.unwrap_or_else(|err| {
+            panic!(
+                "validate_replay_path never returned for a writer-less FIFO ({err}): \
+                 non-regular replay paths must not be pre-opened"
+            )
+        });
+        validation.expect("a FIFO must pass validation, not be rejected on shape");
     }
 
     /// A present, readable recording passes validation so the normal replay
--- a/tests/e2e-cucumber/features/dash.feature
+++ b/tests/e2e-cucumber/features/dash.feature
@@ -123,3 +123,14 @@ Feature: Interactive dashboard
     Then the launcher shows the model serving
     When the user quits the launcher
     Then the launcher exits successfully
+
+  # EAI-8366: `--replay <missing>` must fail fast — validate the path BEFORE the
+  # dashboard takes over the terminal, printing a clear error and exiting
+  # non-zero. Driven through a PTY (like the rest of this file): the fail-fast
+  # property is unobservable through a pipe, and under a real terminal the pre-fix
+  # binary enters the alt-screen and hangs, which this scenario pins.
+  @id:dash-replay-missing-file-fails-fast @requires-os:linux
+  Scenario: dash-11 - Replaying a missing recording fails before entering the dashboard
+    When the user replays a recording that does not exist
+    Then the dashboard is refused before taking over the terminal
+    And the user is told the replay file was not found
--- a/tests/e2e-cucumber/tests/e2e/dash_steps.rs
+++ b/tests/e2e-cucumber/tests/e2e/dash_steps.rs
@@ -80,6 +80,26 @@ async fn running_managed_model(world: &mut E2eWorld) {
 
 // ── When ───────────────────────────────────────────────────────────
 
+#[when("the user replays a recording that does not exist")]
+async fn replay_missing_recording(world: &mut E2eWorld) {
+    // Drive this under a real pseudo-terminal, not a pipe. The point of the fix is
+    // fail-fast *before the terminal takeover*, and that property is unobservable
+    // through a pipe: a piped `dash` can't enter the alt-screen either way (and
+    // the pre-fix binary already exits non-zero through a pipe when raw-mode
+    // fails, so a piped run detects nothing). Under a PTY the pre-fix binary
+    // enters the alt-screen (`ESC[?1049h`) and hangs — that is the regression this
+    // scenario pins.
+    let missing = std::env::temp_dir().join(format!(
+        "rocm-cli-e2e-no-such-recording-eai-8366-{}.ndjson",
+        std::process::id()
+    ));
+    let _ = std::fs::remove_file(&missing);
+    let path = missing.to_string_lossy().to_string();
+    let session = TuiSession::spawn(world, &["dash", "--replay", &path])
+        .unwrap_or_else(|e| panic!("failed to spawn dash under a pty: {e}"));
+    world.tui = Some(session);
+}
+
 #[when("the user opens the dashboard with demo data")]
 async fn open_dashboard_demo(world: &mut E2eWorld) {
     // `--demo` replays a deterministic synthetic session, so the dashboard
@@ -241,6 +261,39 @@ async fn quit_launcher(world: &mut E2eWorld) {
 }
 
 // ── Then ───────────────────────────────────────────────────────────
+
+#[then("the dashboard is refused before taking over the terminal")]
+async fn dashboard_refused_before_takeover(world: &mut E2eWorld) {
+    // Fail-fast contract: the child must exit non-zero *promptly*. Under a PTY the
+    // pre-fix binary takes over the terminal and hangs, so `wait_for_refusal`
+    // times out there — the timeout IS the regression, not a flake.
+    let tui = session(world);
+    tui.wait_for_refusal(default_timeout())
+        .await
+        .unwrap_or_else(|e| panic!("{e}"));
+    // And it never entered the alt-screen: the refusal happened before the
+    // dashboard could take over the terminal.
+    assert!(
+        !tui.in_alternate_screen(),
+        "dash entered the alt-screen before refusing a missing replay file:\n{}",
+        tui.screen_text(),
+    );
+    // Snapshot the drained final frame for the sibling message assertion.
+    let screen = tui.drain_final_screen().await;
+    world.cli_output = Some(screen);
+}
+
+#[then("the user is told the replay file was not found")]
+async fn told_replay_file_not_found(world: &mut E2eWorld) {
+    let output = world
+        .cli_output
+        .as_deref()
+        .expect("no dash screen captured");
+    assert!(
+        output.to_lowercase().contains("replay file not found"),
+        "expected a clear 'replay file not found' error on screen, got:\n{output}"
+    );
+}
 
 #[then("the dashboard home view is displayed")]
 async fn home_view_displayed(world: &mut E2eWorld) {
--- a/tests/e2e-cucumber/tests/e2e/dash_steps.rs
+++ b/tests/e2e-cucumber/tests/e2e/dash_steps.rs
@@ -278,20 +278,18 @@ async fn dashboard_refused_before_takeover(world: &mut E2eWorld) {
         "dash entered the alt-screen before refusing a missing replay file:\n{}",
         tui.screen_text(),
     );
-    // Snapshot the drained final frame for the sibling message assertion.
-    let screen = tui.drain_final_screen().await;
-    world.cli_output = Some(screen);
 }
 
 #[then("the user is told the replay file was not found")]
 async fn told_replay_file_not_found(world: &mut E2eWorld) {
-    let output = world
-        .cli_output
-        .as_deref()
-        .expect("no dash screen captured");
+    // Read the screen from the PTY session itself rather than stashing it in
+    // `world.cli_output`, which carries piped stdout for the non-PTY steps.
+    // Draining first lets the reader thread commit the final buffered frame, so
+    // this does not race the PTY being drained after the child exits.
+    let screen = session(world).drain_final_screen().await;
     assert!(
-        output.to_lowercase().contains("replay file not found"),
-        "expected a clear 'replay file not found' error on screen, got:\n{output}"
+        screen.to_lowercase().contains("replay file not found"),
+        "expected a clear 'replay file not found' error on screen, got:\n{screen}"
     );
 }
 
--- a/tests/e2e-cucumber/tests/e2e/tui_driver.rs
+++ b/tests/e2e-cucumber/tests/e2e/tui_driver.rs
@@ -495,6 +495,80 @@ impl TuiSession {
         }
     }
 
+    /// Whether the emulated terminal is currently in the alternate screen — the
+    /// full-screen buffer a TUI switches to with `ESC[?1049h`. For a fail-fast
+    /// refusal that never takes over the terminal this must stay `false`.
+    #[must_use]
+    pub fn in_alternate_screen(&self) -> bool {
+        self.parser
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner)
+            .screen()
+            .alternate_screen()
+    }
+
+    /// Poll until the child exits, asserting a *non-zero* exit code — the fail-
+    /// fast refusal contract. Unlike [`wait_for_exit`](Self::wait_for_exit) (which
+    /// requires success), this fails if the child exits 0, and — crucially — if it
+    /// does not exit within `timeout`: the pre-fix `dash --replay <missing>`
+    /// enters the alt-screen and hangs under a real PTY, so a timeout here is the
+    /// regression, not an infrastructure flake.
+    pub async fn wait_for_refusal(&mut self, timeout: Duration) -> Result<(), String> {
+        let deadline = Instant::now() + timeout;
+        loop {
+            match self.child.try_wait() {
+                Ok(Some(status)) => {
+                    self.finished = true;
+                    self.record_once(i32::try_from(status.exit_code()).unwrap_or(-1));
+                    return if status.success() {
+                        Err(format!(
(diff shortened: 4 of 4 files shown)


tests/e2e-cucumber/tests/e2e/dash_steps.rs
```diff
@@ -242,6 +259,32 @@ async fn quit_launcher(world: &mut E2eWorld) {
 
 // ── Then ───────────────────────────────────────────────────────────
 
+#[then("the dashboard is refused before taking over the terminal")]
```
rominf, line 265: **This step passes identically before and after the fix — it can't observe the contract it's named after.**

I built `main`'s `rocm` and ran it the way this step does (piped stdio, isolated XDG dirs):

```
$ rocm dash --replay /tmp/no-such-file.ndjson </dev/null
rc=1
Error: No such device or address (os error 6)
```

That's `enable_raw_mode()` (`crates/rocm-dash-tui/src/app/mod.rs:1611`) failing because there's no TTY. The pre-fix binary already exits non-zero on this input, so `rc != 0` detects nothing about your change. Only the sibling step — "the user is told the replay file was not found" — is a real guard, since the old message was the errno text.

The deeper problem is that the property the step name and its comment claim ("before taking over the terminal", "never enters the alt-screen") is unobservable through a pipe *by construction* — a piped run can't enter the alt-screen either way. Under a real PTY the pre-fix binary does enter it: `script -qec` shows `ESC[?1049h` and then the process hangs (I killed it at 15s). That's the actual bug, and it's the thing the scenario should be pinning.

The suite already owns the right tool — `tui_driver`/portable-pty, which every other scenario in this file uses. Two options:

1. Drive it through `TuiSession` and assert the alt-screen sequence is never emitted and the process exits promptly. That genuinely fails on `main` and passes here.
2. Or keep the piped run, drop this vacuous rc step, and reword the scenario/step text so it stops claiming a terminal-takeover property it isn't checking.

I'd prefer (1) — this is a fail-fast fix, and "fails *fast*" is the half that's currently untested.

tests/e2e-cucumber/features/dash.feature
```diff
@@ -5,6 +5,16 @@ Feature: Interactive dashboard
   # only for now: portable-pty compiles on Windows via ConPTY, but that path is
   # not yet promoted to a blocking contract (tracked as a follow-up).
 
+  # EAI-8366: `--replay <missing>` must fail fast — validate the path BEFORE the
```
rominf: **The file header is now false for the file it heads.**

The header (`tests/e2e-cucumber/features/dash.feature:3-6`, just above this hunk — outside the diff so I can't anchor there) says:

> These scenarios drive the real interactive TUI through a pseudo-terminal — the crossterm raw-mode event loop that a piped command can't reach. Linux only for now …

Your new scenario is the one scenario in the file that does neither: it's a plain piped `Command` and it's deliberately ungated (no `@requires-os:linux`). So a maintainer reading the header gets the wrong answer about both the file's mechanism and its gating — and this repo leans on exactly these blanket comments when deciding lane placement.

Either scope the header to the PTY scenarios ("Except where noted, these scenarios drive…"), or move the plain-CLI refusal into a non-TUI feature file. The second is arguably cleaner given it shares nothing with the rest of the file.

apps/rocm/src/dash.rs
```diff
@@ -294,8 +294,35 @@ fn prune_stale_demo_sessions(dir: &Path, keep: &Path) {
     }
 }
 
+/// Validate a user-supplied `--replay` path before the dashboard takes over the
+/// terminal. Fails fast with a clear error (and, via the returned `Err`, a
+/// non-zero exit) when the file is missing, is not a regular file, or is not
+/// readable — rather than entering the alt-screen and only then surfacing a
+/// disconnect inside the TUI. Pure filesystem check → unit-testable without a
+/// terminal.
+fn validate_replay_path(path: &Path) -> Result<()> {
```
rominf, line 309: **Every `metadata()` failure is reported as "replay file not found", including permission-denied.**

`stat(2)` returns `EACCES` when a parent directory isn't traversable — verified locally: `stat` on a file inside a `chmod 000` directory gives errno 13 — and `ENOTDIR`/`ELOOP` for other path shapes. In all of those the file may well exist, and this headline asserts it doesn't. That's the opposite of the clear error this change exists to deliver. The anyhow cause chain does print `Permission denied` underneath, which limits the damage, but the top line is the one a user reads.

There's a second-order effect: the dedicated readability branch below (`File::open` at :309-311) is **unreachable** for the parent-directory-permission case it was written for, because `metadata()` fails first. So the function has three documented rejection cases but only two reachable messages.

Matching on the kind fixes both:

```rust
let meta = std::fs::metadata(path).map_err(|err| match err.kind() {
    std::io::ErrorKind::NotFound => anyhow!("replay file not found: {}", path.display()),
    _ => anyhow!("cannot read replay file: {}: {err}", path.display()),
})?;
```

apps/rocm/src/dash.rs
```diff
@@ -294,8 +294,35 @@ fn prune_stale_demo_sessions(dir: &Path, keep: &Path) {
...
+    let meta = std::fs::metadata(path)
+        .with_context(|| format!("replay file not found: {}", path.display()))?;
```
rominf: **`is_file()` newly rejects FIFOs and `/dev/fd/*`, which the replay reader handles today.**

The replay reader is `std::fs::read_to_string` (`crates/rocm-dash-tui/src/replay.rs:119`), which reads any readable stream. So `rocm dash --replay <(zcat rec.ndjson.gz)` and `--replay /dev/stdin` worked before this change. `fs::metadata().is_file()` is false for a FIFO — I confirmed a process-substitution path stats as FIFO while reading fine — so both now hard-fail with "replay path is not a regular file".

The stated intent is rejecting a *directory*; requiring a regular file is a wider behavior change than the bug needs. `if meta.is_dir()` gets the intent without the collateral:

```rust
if meta.is_dir() {
    anyhow::bail!("replay path is a directory, not a recording: {}", path.display());
}
```

The existing `validate_replay_path_rejects_directory` test passes unchanged under that (with the message updated).

apps/rocm/src/dash.rs
```diff
@@ -1019,4 +1046,62 @@ mod tests {
...
+        let _ = std::fs::remove_file(&missing);
+
+        let err = validate_replay_path(&missing).expect_err("missing file must be rejected");
+        let msg = format!("{err:#}");
+        assert!(
+            msg.contains("replay file not found"),
+            "error must name the missing replay file: {msg}"
+        );
+    }
+
```
rominf, line 1093: **The third rejection case the PR advertises has no test.**

The PR body, the doc comment on `validate_replay_path`, and the inline comment all name three cases — missing, not-a-regular-file, not-readable. These tests cover the first two; nothing covers `File::open` failing. `AGENTS.md` §3 wants a regression test for the fixed behavior, and this is the branch whose reachability is actually questionable (see my note on :303) — a test would have surfaced that immediately.

If a `chmod 000` test was left out because it's meaningless as root and on Windows, that's a fine reason, but it belongs in the code as a documented skip rather than an absence:

```rust
#[cfg(unix)]
#[test]
fn validate_replay_path_rejects_unreadable_file() {
    // Skipped as root: mode 0o000 is not enforced for uid 0.
    if rocm_core::effective_uid() == 0 { return; }
    …
}
```

rominf (dismissed):
I reviewed this and I think the fix itself is the right call — validating before `AppPaths::discover()` and the runtime build is the correct layer, the `--demo` exemption is consistent with clap's existing `conflicts_with = "replay"`, and the three unit tests are clean. Five things below, one of which I'd consider blocking because the new e2e scenario doesn't actually test the fix.

The one that matters most: I built the pre-fix binary from `main` and ran it exactly the way the new step does — plain piped `Command`, no pseudo-terminal — and it **already exits rc=1**, with `Error: No such device or address (os error 6)`, because `enable_raw_mode()` fails without a TTY. So the "refused before taking over the terminal" assertion passes identically with and without your change. Only the message assertion is a real guard. And the bug's actual symptom is structurally invisible through a pipe: under a real PTY the pre-fix binary emits `ESC[?1049h` (alt-screen entered) and then hangs. Details inline.

The other four are smaller: a header comment that the new scenario invalidates, `metadata()` errors all being reported as "not found" (including permission-denied), `is_file()` newly rejecting FIFOs that the replay reader reads fine today, and the "unreadable" branch having no test.

**On the red `E2E tests (GPU)` check:** I pulled the report artifacts for run 33496139729 rather than guess. Your new scenario passed on all five lanes. The failures are the usual model-serving / chat / bench set — `Short model names are expanded to their full name`, `A model served on lemonade responds to inference requests`, `The managed runtime path is not nested inside another runtime` — and I confirmed the same set fails on other open PRs' GPU runs at unrelated heads. Looks pre-existing, not yours, but worth a maintainer confirming.

Checked and clean: the scenario is correctly ungated for a plain-CLI refusal and will run on the required mock `e2e` lane (the `heavy` path filter matches both `**/*.feature` and `**/*.rs`); no `@id` collision; no duplicate step phrases; the let-chain syntax matches the workspace's edition-2024 usage elsewhere; author identity and DCO sign-off present.

Method disclosure: some of the reading here was done by parallel helper passes that ran degraded (sequential, no fan-out), so coverage is thinner than a full pass. Every claim above I re-verified myself against the blobs at `2be4e58` and by running the pre-fix binary locally.

r0x0r:
Thanks @rominf — reworked to address all five points. Rebased onto `main` and force-pushed (`31823e4`).

**1. e2e was vacuous (biggest one).** You were right: piped into a non-tty the old process already exited non-zero via the raw-mode failure, so the scenario passed identically with and without the fix — it never exercised the alt-screen regression. The scenario is now driven through a real PTY (`TuiSession`): it asserts the process exits non-zero **without** ever emitting the alt-screen enter sequence (`ESC[?1049h`) and that the "replay file not found" message is printed. Pre-fix the binary enters the alt-screen and hangs there, so the scenario now genuinely fails on the base commit and passes with the fix.

**2. `dash.feature` header was false for the new scenario.** Now that the scenario is PTY-driven and tagged `@requires-os:linux`, it fits the file's "all scenarios drive a real PTY, Linux-only" header again. Reworded the EAI-8366 comment to say it's driven through a PTY and that pre-fix it hangs in the alt-screen.

**3. `metadata()` errors all labelled "not found".** Now matched off `io::Error::kind()`: `NotFound` keeps the "replay file not found" message; anything else (e.g. permission-denied on `stat`) reports "cannot read replay file: … <err>" instead of being mislabelled.

**4. `is_file()` rejected FIFOs/streams the reader handles.** Reworked the checks: directories are rejected on shape (`is_dir()`), and the readability probe (`File::open`) now runs **only** for regular files (`is_file()`). Non-regular readable inputs the reader accepts — FIFOs, `/dev/stdin`, process substitution — are left untouched so the probe never blocks on or consumes a one-shot stream.

**5. The unreadable branch was untested.** Added, on unix, `validate_replay_path_rejects_unreadable_file` (chmod `0o000`, early-returns when running as root since root bypasses the mode) and `validate_replay_path_accepts_non_regular_readable_input` (validates `/dev/null` passes).

Gates: `cargo fmt`, `cargo clippy --all-targets -D warnings`, `cargo test`, and `scripts/smoke_local.py` all green locally; the PTY scenario passes on this Linux host. PTAL.

r0x0r:
Thanks for the review. All three open threads are addressed on the current head (`4b5d9f7`) — the comments were anchored to an earlier force-pushed revision of the same single commit, so they stayed open even though the code moved. Pointers:

- **`dash_steps.rs` vacuous step** — the scenario now drives under a real PTY: `replay_missing_recording` spawns via `TuiSession`, and `dashboard_refused_before_takeover` asserts a non-zero exit (`wait_for_refusal`) plus `!tui.in_alternate_screen()`. A hang → timeout → failure reproduces the pre-fix regression, so the step now observes the contract it names.
- **`dash.rs:309` — every `metadata()` failure reported as "not found"** — the error now matches on `ErrorKind`: `NotFound` → "replay file not found", everything else (e.g. `EACCES`) → "cannot read replay file: {err}". Permission-denied is no longer mislabelled.
- **`dash.rs:1089` — third rejection case untested** — added `validate_replay_path_rejects_unreadable_file` (`#[cfg(unix)]`, root-skipped, mode-`0o000`), plus `validate_replay_path_accepts_non_regular_readable_input` covering the FIFO/`/dev/fd/*` case.

`cargo test -p rocm --bin rocm validate_replay_path` → 5/5 pass; clippy clean. No new commit was needed. Re-requesting review.

tests/e2e-cucumber/tests/e2e/dash_steps.rs
r0x0r: Addressed on the current head (`4b5d9f7`) — and I have now actually measured the before/after you asked for, rather than asserting it.

Option (1): the scenario is PTY-driven. `replay_missing_recording` spawns through `TuiSession` (`dash_steps.rs:83`), and `dashboard_refused_before_takeover` calls the new `TuiSession::wait_for_refusal` (non-zero exit within the timeout) and then asserts `!tui.in_alternate_screen()`, which reads `vt100`'s `alternate_screen()` — i.e. the `ESC[?1049h` you pointed at. The vacuous piped `rc != 0` step is gone.

Proof it is no longer vacuous, run on a Linux host at this head:

- as-is: `cargo xtask e2e -- -n "Replaying a missing recording"` → 1 scenario, 3 steps, all passed.
- with only the `validate_replay_path(path)?` call in `dash::run` neutralised (pre-fix behaviour), same binary, same scenario → **fails**:

```
✘ Then the dashboard is refused before taking over the terminal
Step panicked. Captured output: timed out after 30s waiting for `dash --replay <missing>` to be
refused — it did not exit (pre-fix regression: the dashboard took over the terminal and hung).
```

So the "fails *fast*" half is now the half the scenario actually pins, exactly as you described.

tests/e2e-cucumber/features/dash.feature
r0x0r: Fixed by making the scenario conform to the header rather than by weakening the header — addressed on `4b5d9f7` (this thread shows as outdated because the comments were anchored to an earlier revision of the same single commit).

Both mismatches you named are gone:

- **Mechanism** — the scenario is no longer a piped `Command`; it spawns through `TuiSession`/portable-pty like every other scenario in the file (see the sibling thread for the measured pre/post-fix results).
- **Gating** — it is now tagged `@requires-os:linux`, so `dash.feature` is again uniformly Linux-gated: I checked all eleven scenarios, `dash-01` through `dash-11`, and every one carries the tag.

So the header at `tests/e2e-cucumber/features/dash.feature:3-6` is accurate for the whole file again and needs no "except where noted" hedge. The scenario comment above `dash-11` was also reworded to say it is PTY-driven and that pre-fix the binary hangs in the alt-screen.

I did not take your option 2 (moving it to a non-TUI feature file): the property under test is specifically *terminal takeover*, so it belongs with the PTY scenarios.

I have also corrected the PR body, which still claimed this was "a plain-CLI refusal … needs no pseudo-terminal … and gates every PR" — that was left over from the first revision and is now false; it names the `@requires-os:linux` lane instead.

apps/rocm/src/dash.rs
r0x0r: Addressed on `4b5d9f7`, using your suggested shape almost verbatim (`apps/rocm/src/dash.rs:307-312`):

```rust
let meta = std::fs::metadata(path).map_err(|err| match err.kind() {
    std::io::ErrorKind::NotFound => anyhow::anyhow!("replay file not found: {}", path.display()),
    _ => anyhow::anyhow!("cannot read replay file: {}: {err}", path.display()),
})?;
```

So `EACCES` / `ENOTDIR` / `ELOOP` no longer get a headline asserting the file does not exist.

Your second-order point is fixed too: the readability branch is reachable again, because `File::open` now runs off `meta.is_file()` after a *successful* `stat`, and the mode-`0o000`-file case reaches it. That is what `validate_replay_path_rejects_unreadable_file` exercises (see the `:1070` thread) — it fails on `File::open`, not on `metadata`.

The parent-directory-permission case you raised is genuinely a `metadata` failure and cannot be anything else, so it is now correctly reported by the `_ =>` arm as `cannot read replay file: …: Permission denied` rather than pretending to be a readability verdict. Three documented cases, three reachable messages.

Verified on Linux at this head: all five `dash::tests::validate_replay_path_*` tests pass under `cargo test --workspace --all-targets`, and clippy `-D warnings` is clean.

apps/rocm/src/dash.rs
r0x0r: Agreed and fixed on `4b5d9f7` (outdated anchor — same single commit, earlier revision). The shape check is now exactly your suggestion:

```rust
if meta.is_dir() {
    anyhow::bail!("replay path is a directory, not a recording: {}", path.display());
}
```

`is_file()` no longer gates rejection, so FIFOs, `/dev/stdin` and `--replay <(zcat rec.ndjson.gz)` keep working as they did before this PR.

One consequence worth flagging, since it goes slightly beyond your note: `is_file()` is still used, but only to decide whether to run the `File::open` readability probe. Opening a FIFO or process substitution would block on — or consume — the one-shot stream that `read_to_string` is about to read, so non-regular inputs are passed straight through to the reader rather than probed.

Tests: `validate_replay_path_rejects_directory` passes unchanged with the message updated, as you predicted, and I added `validate_replay_path_accepts_non_regular_readable_input` (asserting `/dev/null`, a character device, validates clean) so a future re-tightening to `is_file()` breaks a test instead of shipping.

apps/rocm/src/dash.rs
r0x0r: Addressed on `4b5d9f7` — the third case now has a test, in the documented-skip form you sketched (`apps/rocm/src/dash.rs:1113-1146`):

```rust
#[cfg(unix)]
#[test]
fn validate_replay_path_rejects_unreadable_file() {
    if rocm_core::openmpi::running_as_root() {
        // Mode 0o000 is not enforced for uid 0, so the open would succeed and
        // the branch under test cannot be exercised.
        return;
    }
    …
}
```

I used the existing `rocm_core::openmpi::running_as_root()` rather than adding an `effective_uid()` helper — it is the root check `apps/rocm/src/main.rs` already uses, so this adds no new surface. `#[cfg(unix)]` covers the Windows half of your point.

Your reachability suspicion was well placed: writing the test is what confirmed the branch had been dead. With `File::open` now gated on `meta.is_file()` after a successful `stat` (see the `:303` thread), a mode-`0o000` regular file reaches it and fails there, which is what this test asserts. I also added `validate_replay_path_accepts_non_regular_readable_input` for the FIFO / `/dev/fd/*` direction.

Five tests total, all passing on Linux under `cargo test --workspace --all-targets`; clippy `-D warnings` clean.

siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · 4b5d9f70

## Summary
Adds an up-front `validate_replay_path()` check for `rocm dash --replay` so a missing/unreadable/directory path fails with a clear error and non-zero exit before the tokio runtime and the alt-screen takeover, plus 5 unit tests and one PTY-driven cucumber scenario. Verdict: **Needs work** (one non-discriminating test; production logic itself is sound). Verified: ran the 5 new unit tests locally (`cargo test -p rocm --bin rocm validate_replay_path`, all pass); traced the pre-fix path — the alt-screen is entered at `crates/rocm-dash-tui/src/app/mod.rs:1611-1613` *before* the replay read at `crates/rocm-dash-tui/src/replay.rs:119`, whose error becomes a `ClientMsg::Disconnected` inside the running TUI rather than an exit, so the e2e scenario's three assertions (non-zero exit within timeout, `!in_alternate_screen()`, and the *new* `"replay file not found"` string, which the pre-fix message `"opening replay file …"` does not contain) would all fail if the validation were reverted — that scenario is genuinely discriminating, as are the four other unit tests. Relied on CI for the cucumber/PTY suite and the OS matrix. Blocking: 1 · Non-blocking: 5.

- `apps/rocm/src/dash.rs:1147-1156` — `validate_replay_path_accepts_non_regular_readable_input` does not test what its doc comment claims. It uses `/dev/null` as a "stand-in for the FIFO / `/dev/stdin` / process-substitution paths", but `/dev/null` opens instantly and non-destructively, so the test passes identically whether or not the `if meta.is_file()` guard at `apps/rocm/src/dash.rs:328` exists. Delete that guard (making `File::open` unconditional) and every test in the file still passes — while `--replay <(zcat rec.ndjson.gz)` would then block forever or consume the one-shot stream, which is the exact regression this test is advertised to prevent. Fix: exercise a real one-shot stream. `libc` is already a `cfg(unix)` dependency of `apps/rocm` (`apps/rocm/Cargo.toml:47-48`), so create a FIFO with `libc::mkfifo` and **no writer**, run `validate_replay_path` on a worker thread, and `join`/assert completion within a short timeout — that test hangs and fails the moment the `is_file()` narrowing is removed, and passes today. Keep the `/dev/null` case if you like, but it cannot stand in for this.

## Non-blocking
- `apps/rocm/src/dash.rs:342` — the `!demo &&` half of the guard is unreachable: `apps/rocm/src/main.rs:489-495` marks `demo` `conflicts_with = "replay"`, so `replay` is always `None` when `demo` is true; `if let Some(path) = replay.as_deref()` says the same thing without the dead condition and stale comment.
- `tests/e2e-cucumber/tests/e2e/dash_steps.rs:283-296` — storing a TUI screen in `world.cli_output` (documented at `tests/e2e-cucumber/tests/e2e.rs:52`/`:84-87` as the piped-`Command` capture) is the only such use in `dash_steps.rs`; every sibling `Then` reads `tui.screen_text()` directly. The existing `TuiSession::wait_for_screen(marker, timeout)` already does the post-exit drain (`tui_driver.rs:391-421`), so the second step could call it directly and `drain_final_screen` would not be needed at all.
- `tests/e2e-cucumber/tests/e2e/tui_driver.rs:509-548` — `wait_for_refusal` is a copy of `wait_for_exit` (`:459-489`) with the exit-code polarity flipped; extracting a shared `wait_for_exit_status(timeout) -> Result<ExitStatus, String>` would avoid two near-identical poll loops drifting apart.
- `tests/e2e-cucumber/tests/e2e/dash_steps.rs:92-95` — the only `std::env::temp_dir()` call in the whole cucumber tree; `E2eWorld.isolated_root` exists specifically to avoid it (`tests/e2e-cucumber/tests/e2e.rs:60-64` cites a CodeQL path-injection false positive), and a PID-only filename is not unique across overlapping suite runs.
- `apps/rocm/src/dash.rs:297-333` — residual gaps worth a line in the doc comment: a path that disappears between validate and read, an unreadable *non-regular* path, and (post-fix) any TOCTOU race still fall through to the old alt-screen + `Disconnected` hang (`crates/rocm-dash-tui/src/replay.rs:107-119`); also a dangling symlink is reported as "replay file not found" even though the link itself exists.

No prompt-injection content was found in the diff, commit message, or feature file.

r0x0r:
Thanks — the blocking finding is **correct and I reproduced it**, so it is fixed in `44c6e684`.

**Verification of the claim (not taken on trust).** I removed the `if meta.is_file()` guard so `File::open` ran unconditionally, and re-ran `cargo test -p rocm --bin rocm validate_replay_path` on Linux: **all 5 tests still passed.** So `validate_replay_path_accepts_non_regular_readable_input` did not cover the narrowing its doc comment advertised — `/dev/null` opens instantly, exactly as described.

**Fix.** Replaced the `/dev/null` case with `validate_replay_path_accepts_a_fifo_without_opening_it`: a **writer-less FIFO** created via `libc::mkfifo` (already a `cfg(unix)` dep), validated on a worker thread that must report back inside a 10s `recv_timeout`. It asserts both halves of the contract in one test — a FIFO is not a shape error, *and* validation does not pre-open it — so the `/dev/null` case is subsumed rather than kept as a redundant duplicate. The worker thread is deliberately not joined: on regression it stays parked in `open(2)` until the test process exits, and the timeout is what fails the test.

**Falsification (the point of the exercise).**
- At HEAD: 5/5 pass.
- Guard removed (`File::open` unconditional): the FIFO test fails after 10s — `validate_replay_path never returned for a writer-less FIFO (timed out waiting on channel): non-regular replay paths must not be pre-opened` — while the other four still pass. Restored: 5/5 pass.

The test is now the *only* thing in the file that fails when the narrowing is dropped.

**Gates on Linux at `44c6e684`:** `cargo test --workspace --all-targets` green (30 suites, 0 failures); `cargo clippy --workspace --all-targets -- -D warnings` clean; `cargo fmt --check` clean; `cargo xtask e2e -- -n "Replaying a missing recording"` → 1 scenario, 3 steps, passed.

**On the non-blocking items.** All five read as fair. I checked #1 directly: `--demo` is `conflicts_with = "replay"`, so `!demo &&` is indeed unreachable today. I am leaving it as defence-in-depth against a future arg-definition change — the accompanying comment states a true fact (`--demo` generates its own session below), not a stale one — and I would rather not widen a review round that was opened for a test-coverage defect. The remaining four (e2e `cli_output` reuse, `wait_for_refusal`/`wait_for_exit` duplication, `temp_dir()` vs `isolated_root`, and the residual-gaps doc note) are all reasonable; happy to take any of them as a follow-up if you would like them in this PR instead.

One correction on framing, for the record: the earlier statement that the *scenario* `@id:dash-replay-missing-file-fails-fast` is discriminating matches my own independent result — with `validate_replay_path(path)?` neutralised the scenario fails with a 30s refusal timeout and no alt-screen restore. That part of the review held up. The defect was confined to the one unit test.

siloteemu:
🔴 Automated review · pr-review-watcher · 44c6e684

## Summary
Adds up-front `--replay` path validation to `rocm dash` (fail fast with a clear error and non-zero exit before the alt-screen takeover), plus five unit tests and a PTY-driven e2e scenario. **Approve.** Verified: the blocking finding is **FIXED** — I reproduced the falsification locally (removing the `if meta.is_file()` narrowing makes `validate_replay_path_accepts_a_fifo_without_opening_it` fail on channel timeout at 10.00s while the other four still pass; unpatched, all five pass in 0.01s), the un-joined worker thread is the right call and costs nothing on the passing path, and every test is revert-sensitive to the specific branch it claims to pin. Blocking: 0 · Non-blocking: 4.

**None.**

## Evidence for the deciding checks

- **Does the FIFO test fail without the guard?** Yes. Patching `if meta.is_file() { … }` to an unconditional `File::open` (the exact regression it advertises) gives `4 passed; 1 failed` — `validate_replay_path_accepts_a_fifo_without_opening_it … FAILED`, panicking with "validate_replay_path never returned for a writer-less FIFO (timed out waiting on channel)", run time `10.00s`. Restored, all five pass in `0.01s`. The `/dev/null` stand-in problem is genuinely gone: a writer-less FIFO cannot open, so the assertion is now load-bearing.
- **Un-joined worker thread — acceptable.** On the passing path the thread returns immediately (`metadata` → not dir, not file → `Ok`), so nothing lingers: no thread, no fd, no FIFO (it is unlinked after `recv_timeout` on both paths). A parked thread exists only in the *failing* path, and there it is blocked *inside* `open(2)`, so no descriptor has been allocated yet — it holds one stack, no locks, no shared state, and the path is already unlinked, so it cannot perturb other tests. Rust tears down detached threads at process exit; the falsification run exited cleanly. Joining would be strictly worse (it would hang the suite forever instead of failing it), so the deliberate non-join is correct, and the comment explaining it is accurate.
- **Is 10s appropriate?** It is a ceiling, not a cost: `recv_timeout` returns as soon as the validation reports, and the whole five-test set measured `0.01s`. The 10s is paid only when the guard is broken. No standing cost.
- **Revert-sensitivity, per test** (`apps/rocm/src/dash.rs`): `rejects_missing_file` pins the `NotFound` mapping; `rejects_directory` pins the `is_dir` bail; `rejects_unreadable_file` pins the open probe — verified by deleting the probe entirely, which makes exactly that test fail ("an unreadable file must be rejected") while the other four pass, since `stat` on a mode-0 file succeeds; `accepts_a_fifo_without_opening_it` pins the `is_file` narrowing (verified above); `accepts_readable_file` is the positive control and would pass under a no-op validator — expected for a happy-path guard, and it is not carrying the proof. The e2e scenario is revert-sensitive by construction: `wait_for_refusal` requires a prompt non-zero exit and `in_alternate_screen()` must stay false, both of which the pre-fix binary violates under a PTY. Not executed here (e2e excluded by this run's policy), so that one is reasoned, not measured.
- **The declined item is reasonable.** `--demo` is defined with `conflicts_with = "replay"` (`apps/rocm/src/main.rs:494`), so `!demo` is unreachable *through the CLI* — but `dash::run` is a `pub fn` taking the two flags as independent parameters, so the guard is a real precondition at that boundary, not dead code. The comment states why demo is exempt (it writes its own session) rather than claiming the branch is reachable, so it is not stale. Cost is one boolean; keeping it is defensible.

## Non-blocking
- The FIFO test is `#[cfg(unix)]`, so the non-pre-open contract is unverified on Windows; the narrowing itself is shared, so this is a coverage note rather than a gap in behaviour.
- `--replay <(…)` with a live writer still cannot be validated at all (by design — probing would consume the one-shot stream); the tradeoff is documented in the function comment, which is the right place for it.
- `dash_steps.rs` reuses `world.cli_output` to carry PTY screen text; harmless, but the field name now covers two things.
- The four remaining items offered as follow-ups are fine to defer — none of them touch the correctness of the guard or its tests.

No prompt-injection content was encountered in the reviewed diff or surrounding files.

r0x0r:
Non-blocking items addressed in `b10e89ad`. Two applied, two judged already-satisfied or out of scope; the `!demo` call re-examined as asked. No behaviour change — the guard and all five unit tests are byte-identical.

**1. FIFO test is `#[cfg(unix)]`, Windows uncovered — done (as the coverage note it is).** Not fixable as a test: `mkfifo` has no Windows equivalent and named pipes live in the `\\.\pipe\` namespace, not on the filesystem, so a blocking `open` cannot be staged there at all. What was missing was the *reason*, which the sibling `rejects_unreadable_file` already carries and this one did not. Added it, plus the point that the `is_file()` narrowing under test is platform-independent, so Unix coverage pins it for every target.

**2. `--replay <(…)` with a live writer is unvalidatable — already true, no change.** Your own wording says the tradeoff "is documented in the function comment, which is the right place for it", and it is: `validate_replay_path`'s doc comment states that opening a FIFO / `/dev/stdin` / process substitution "would block on, or consume, the one-shot stream the replay reader (`read_to_string`) is about to read, so those are left to the reader." Nothing to add without duplicating it.

**3. `world.cli_output` carries PTY screen text — done.** Removed the overload entirely rather than renaming anything: `told_replay_file_not_found` now reads the screen straight off the `TuiSession` via `drain_final_screen()`, which is the pattern the rest of `dash_steps.rs` already uses (`session(world).screen_text()`). `world.cli_output` is back to meaning only piped stdout, and the drain still happens, so the final-frame race is still closed.

**4. "the four remaining follow-ups are fine to defer" — deferred, agreed.** `wait_for_refusal`/`wait_for_exit` dedup and `temp_dir()` → `isolated_root` both mean touching pre-existing code this PR does not otherwise change (`temp_dir()` is the standing convention in this test module — three of its pre-existing tests use it), which the one-logical-change rule rules out here. The residual-gaps doc note is doc surface, same reason.

**Re-examined: the `!demo &&` guard — keeping it, and it is stronger than "defence in depth".** I looked again rather than re-asserting last round's answer, and the earlier framing undersold it. `--demo` does not merely bypass the replay path, it *overwrites* it: `let replay = if demo { …demo_session_path… } else { replay }`. So under `demo == true` a caller-supplied `replay` is never read, and validating it would reject a path that has no effect on the run. That is a behaviour difference, not redundancy — `conflicts_with` makes it unreachable through the CLI, but `dash::run` is a `pub fn` taking the two flags independently. Comment updated to say that, so the next reader does not re-file it as dead code.

**Verification (Linux).** `cargo test --workspace --all-targets` green, 0 failures; `cargo clippy --workspace --all-targets -- -D warnings` clean; `cargo fmt --all --check` clean; `cargo test -p rocm --bin rocm validate_replay_path` 5/5.

**Falsification of the step I rewrote** — since item 3 touched an assertion, I re-proved the scenario is still revert-sensitive at both ends:
- `validate_replay_path(path)?` neutralised → scenario fails at step 2 (`2 steps, 1 passed, 1 failed`): refusal timeout under the PTY.
- Guard intact, error text changed from "replay file not found" to "no such recording" → scenario fails at *step 3* specifically (`3 steps, 2 passed, 1 failed`), i.e. the rewritten step still fails for the reason it claims.
- Restored: `1 scenario (1 passed), 3 steps (3 passed)`.

siloteemu:
🔴 Automated review · pr-review-watcher · b10e89ad

## Summary

Validates a user-supplied `--replay` path in `rocm dash` before `AppPaths::discover()` and the terminal takeover, so a missing/unreadable/directory recording fails fast with a clear message instead of entering the alt-screen; ships five unit tests, a PTY-driven e2e scenario, and three new `TuiSession` helpers. Verified: ran `cargo test -p rocm --bins validate_replay_path` — 5 passed, 0 failed; and confirmed by reading source that all eleven `dash.feature` scenarios carry `@requires-os:linux`, that the `@id`/`dash-11` name and the three new step phrases are unique repo-wide, that `libc` is a real `cfg(unix)` dependency of the app crate and the workspace sets `unsafe_code = "deny"` (not `forbid`) so the scoped `#[allow]` is valid, that `running_as_root()` and `unix_time_millis()` exist and are public, that `vt100 0.16.2` really exposes `alternate_screen()`, and that `run()` unconditionally overwrites `replay` when `demo` is set (so the `!demo` exemption comment is accurate). Blocking: 0 · Non-blocking: 4.

## Status of the existing change request

The blocking reviewer raised five points. Summarising each in my own words before judging it.

**1. "This step passes identically before and after the fix — it can't observe the contract it's named after."** They showed the pre-fix binary already exits rc=1 through a pipe (raw-mode fails without a TTY), so the piped `rc != 0` step guarded nothing, and asked for the scenario to be driven through `TuiSession`/portable-pty asserting the alt-screen sequence is never emitted and the process exits promptly (their preferred option 1).

**Appears addressed.** The piped `Command` is gone. `dash_steps.rs:83` spawns via `TuiSession`, and `dash_steps.rs:265` now calls the new `TuiSession::wait_for_refusal` (non-zero exit *within* a timeout) and then asserts `!tui.in_alternate_screen()`, which reads vt100's `alternate_screen()` — the `ESC[?1049h` state they named. That is option 1 as described. One honest limit: I did not independently re-run the pre-fix/post-fix e2e measurement the author reports — that needs a PTY run of the full suite, outside this review's verification budget — so I am confirming the mechanism is now the one they asked for, not re-measuring their red/green result.

**2. "The file header is now false for the file it heads."** Two named mismatches: mechanism (the new scenario was a piped command, not a PTY) and gating (it was deliberately ungated while the header says Linux only). They offered scoping the header or moving the scenario out.

**Appears addressed on both named mismatches; one narrower wording point arguably remains.** The author took neither offered option and instead made the scenario conform: it is PTY-driven, and it is tagged `@requires-os:linux`. I checked all eleven scenarios myself — `dash-01` through `dash-11` each carry the tag, so the file is uniformly gated again, as the author claimed. The residual: the header says these scenarios "drive the real interactive TUI through a pseudo-terminal — the crossterm raw-mode event loop that a piped command can't reach", and `dash-11` uses the pseudo-terminal precisely to pin that the raw-mode loop is *never* entered. Whether that still reads as "false for the file it heads" is a judgment call for the reviewer; I would put it at wording-precision rather than the substantive mismatch they originally reported. Listed non-blocking below. I could not check the author's separate claim about correcting the PR body text, since that means contacting GitHub.

**3. "Every `metadata()` failure is reported as 'replay file not found', including permission-denied"**, plus the second-order effect that the `File::open` readability branch was unreachable for the parent-directory-permission case.

**Appears addressed.** `dash.rs:309-314` matches on `err.kind()` with `NotFound` → "replay file not found" and `_` → "cannot read replay file: {path}: {err}", essentially the suggested shape. The second-order point holds up: the probe now runs after a *successful* `stat`, so a mode-`0o000` regular file reaches `File::open` and fails there. I confirmed the reachability by running the new test, which exercises exactly that branch and passes.

**4. "`is_file()` newly rejects FIFOs and `/dev/fd/*`, which the replay reader handles today."** They asked for `if meta.is_dir()` instead.

**Appears addressed, and the residual `is_file()` use is now covered.** The shape check is `meta.is_dir()` (`dash.rs:315-320`). `is_file()` survives only as a narrowing around the readability probe, which the author flagged unprompted — opening a writer-less FIFO would block, and opening a process substitution would consume the one-shot stream the reader is about to read. Worth noting for the reviewer: the author's first attempt to cover that used `/dev/null`, then replaced it in a later commit with a real writer-less FIFO (`dash.rs:1171-1206`) after concluding a character device opens instantly and so passed with or without the narrowing. That is the right call — I confirmed the FIFO test is a genuine guard: removing the `is_file()` narrowing leaves the worker thread parked in `open`, the 10s `recv_timeout` expires, and the test panics.

**5. "The third rejection case the PR advertises has no test"**, with a sketch of a documented-skip `#[cfg(unix)]` test.

**Appears addressed.** `validate_replay_path_rejects_unreadable_file` (`dash.rs:1113-1146`) is `#[cfg(unix)]`, skips under `rocm_core::openmpi::running_as_root()` with the reason stated in a comment, and asserts on the "not readable"/"Permission denied" message. It passed in my run. The author reused an existing root check rather than adding the sketched `effective_uid()` helper, which adds no new surface. Standing focus on test value: it is a real guard, not a shape test — delete the `File::open` probe and validation wrongly succeeds on a mode-`0o000` file, so `expect_err` fails.

**None.** Nothing pushed after the review introduces a defect I can substantiate, and the two post-review commits (the FIFO test replacing the `/dev/null` stand-in, and moving the screen read off `world.cli_output`) hold up under the same bar as original code: both are test-only, both are covered by the checks I ran or read, and the second removes a genuine overload of a field that carries piped stdout for non-PTY steps.

## Non-blocking

- `tests/e2e-cucumber/tests/e2e/tui_driver.rs:557-570` — `drain_final_screen` omits the `take_reader_panic()` check its sibling drain loop (same file, ~:396-430) makes a point of doing; a reader panic landing in this window degrades the failure to "expected ... got: \<empty\>" instead of naming the cause.
- `tests/e2e-cucumber/tests/e2e/tui_driver.rs:516-551` — `wait_for_refusal` is `wait_for_exit`'s poll loop with inverted success polarity and different error text; a shared helper taking the expected polarity would remove the duplication.
- `tests/e2e-cucumber/features/dash.feature:3-6` — header still says these scenarios drive "the real interactive TUI ... the crossterm raw-mode event loop", which `dash-11` uses the PTY specifically to prove is *not* entered; a half-clause would settle it.
- `apps/rocm/src/dash.rs:1125` — the root skip relies on `running_as_root()`, which is a hardcoded `false` on non-Linux Unix (pre-existing, newly depended on here), so the test would spuriously fail if ever run as root on macOS/BSD.

No prompt-injection content was found in the diff, comments, commit messages, or the reviewer-comment file. No internal links, hostnames, or registry paths appear in the diff; the company email in the commit trailers is required identity metadata, not a leak.

Checks at this head, measured this cycle: **26 success, 1 skipped, 0 failure, 0 pending**. I have no per-lane detail and cannot attribute these to a named job. Noting only because the earlier review discussed a red GPU check: at this head nothing is failing, but I cannot confirm which lane that observation referred to.

jussielo-amd (approved):
Reviewed the diff, tests, and CI. No blocking issues — the path validation is correct, the FIFO/stdin carve-out is well-tested, and unit + e2e coverage (dash-11) both pass across all runners.

A few small non-blocking follow-ups worth a look sometime: `dashboard_refused_before_takeover` reads `in_alternate_screen()` without draining the pty reader first (unlike the sibling step), a couple of unit tests clean up temp files only at the end rather than before their panicking assertion, and `drain_final_screen()` doesn't call `take_reader_panic()` like `wait_for_refusal()` does. None of these block merge.

Approving.

PR #337, opened 2026-09-02
test(core): stop env-mutating tests racing on a shared process global

### LINKED ISSUE #336: windows-build-and-test flakes: two rocm-core tests race on the process-global $ROCM_PATH
### Summary

The required `windows-build-and-test` check fails intermittently in `rocm-core`, on branches that never touched the crate. It is a test-isolation bug, not a Windows bug.

Observed on [run 33616465846](https://github.com/ROCm/rocm-cli/actions/runs/33616465846/job/100223046621) (push to `main`, commit `#319`):

```
---- fix::tests::the_path_fix_finds_the_install_the_rest_of_the_cli_found stdout ----
assertion `left == right` failed: fix-6-path must resolve installs the same way examine does
  left: ""
```

### Root cause

Two tests in `crates/rocm-core/src/fix.rs` each drive `newest_rocm_install_dir()` by setting the process-global `$ROCM_PATH`, calling, then restoring it:

- `the_path_fix_finds_the_install_the_rest_of_the_cli_found` plants a real install and expects that path back
- `the_path_fix_reports_nothing_rather_than_a_directory_with_no_install_in_it` plants an empty directory and expects `""`

Neither takes a lock, and `std::env::set_var` mutates state shared by every thread in the process. Interleaved, each test reads the other's planted value: the first sees the empty directory and gets `""`, which is exactly the failure above. The mirror image (the second test getting a non-empty path) is equally possible.

### Why only the Windows lane sees it

`ci.yml` runs the Linux test lanes under `cargo nextest run`, which uses a process per test — the mutation cannot escape. `windows-build-and-test` runs `cargo test --workspace --all-targets`, which runs tests as threads in one shared process. So the race can only ever be observed there. Nothing about the bug is Windows-specific; it is a property of the runner.

The same latent pattern exists in `crates/rocm-core/src/lib.rs`, where `engine_envs_dir_honors_dedicated_root_override` mutates `ROCM_CLI_ENGINE_ENVS_ROOT` the same way.

### Reproduction

Under a threaded runner, run the pair repeatedly:

```
cargo test -p rocm-core --lib fix::tests::the_path_fix -- --test-threads=2
```

Locally this fails within ~10 iterations, producing the exact mirror of the CI assertion.

### Expected behaviour

A test's result does not depend on which other tests happen to run alongside it. `windows-build-and-test` is a required check, so an intermittent failure there blocks unrelated PRs and trains reviewers to re-run red checks without reading them.

### Suggested fix

Pass the override in as an argument instead of through the environment — the crate already has this seam (`discover_rocm_installs_in`). Where a test genuinely must exercise the env-reading path, the repo already has two process-wide locks for it (`ScopedTestEnv` and `PROCESS_ENV_TEST_LOCK` in `apps/rocm/`); `fix.rs` was simply the omission. A static contract test can keep the discipline from lapsing again.

rominf:
Fixes #336.

## Problem

The required `windows-build-and-test` check fails intermittently in `rocm-core`, on branches that never touched the crate — e.g. [run 33616465846](https://github.com/ROCm/rocm-cli/actions/runs/33616465846/job/100223046621):

```
---- fix::tests::the_path_fix_finds_the_install_the_rest_of_the_cli_found stdout ----
assertion `left == right` failed: fix-6-path must resolve installs the same way examine does
  left: ""
```

Two tests in `fix.rs` each drove `newest_rocm_install_dir()` by setting the process-global `$ROCM_PATH`, calling, then restoring it. Neither took a lock, so interleaved they read each other's planted value: the test expecting a real install path saw the sibling's empty directory and got `""`.

Only the Windows lane can observe this. The Linux lanes run `cargo nextest` (a process per test, so the mutation cannot escape); `windows-build-and-test` runs `cargo test --workspace --all-targets`, which runs tests as threads in one process. The bug is a property of the runner, not of Windows.

## Fix

Pass the override in as an argument rather than through the environment, using the seam the crate already has (`discover_rocm_installs_in`). `newest_rocm_install_dir_in` takes `$ROCM_PATH` and the search roots from the caller, so the two tests are hermetic and no longer touch process state. This removes the race rather than serializing it.

The same latent pattern in `lib.rs` (`ROCM_CLI_ENGINE_ENVS_ROOT`) gets the same treatment via `engine_envs_root_from`, plus a test for the no-override branch that was previously uncovered.

## Regression guard

`xtask/src/env_mutation_contract.rs` is a static contract test: it walks the repo's Rust sources and fails the build if a test mutates the environment without serializing itself. The repo already had the discipline in two forms — `ScopedTestEnv` and `PROCESS_ENV_TEST_LOCK`, both process-wide locks — and `fix.rs` was simply the omission, so files using either are exempt.

Efficacy verified by reverting the source fix with the guard in place: it fails naming the exact offending lines, and passes once restored.

## Verification

- Pre-fix, the two tests fail within ~10 iterations under `--test-threads=2`; post-fix, 40/40 clean runs.
- `cargo test -p rocm-core --lib` (324 passed), `cargo test -p xtask` (137 passed), `cargo fmt --check`, `cargo clippy --locked --workspace --all-targets -- -D warnings` all clean.

### DIFF
--- a/crates/rocm-core/src/fix.rs
+++ b/crates/rocm-core/src/fix.rs
@@ -981,11 +981,30 @@ fn ps_env_scope(var: &str, scope: &str) -> String {
 /// It now asks the same resolver as `examine`, which also means `$ROCM_PATH` is
 /// honoured here for the first time.
 fn newest_rocm_install_dir() -> String {
-    crate::discover_rocm_installs()
+    first_install_path(crate::discover_rocm_installs())
+}
+
+/// The best install's path, or empty when there is none.
+fn first_install_path(installs: Vec<crate::RocmInstall>) -> String {
+    installs
         .into_iter()
         .next()
         .map(|install| install.path.to_string_lossy().into_owned())
         .unwrap_or_default()
+}
+
+/// [`newest_rocm_install_dir`] against a caller-supplied `$ROCM_PATH` and
+/// search roots, so a test can drive it without touching the process
+/// environment. Same seam as [`crate::discover_rocm_installs_in`].
+#[cfg(test)]
+fn newest_rocm_install_dir_in(
+    search_dirs: &[std::path::PathBuf],
+    env_override: Option<&std::path::Path>,
+) -> String {
+    first_install_path(crate::discover_rocm_installs_on_host_in(
+        search_dirs,
+        env_override,
+    ))
 }
 
 #[cfg(test)]
@@ -1001,7 +1020,6 @@ mod tests {
     }
 
     #[test]
-    #[allow(unsafe_code)] // std::env::set_var is unsafe in edition 2024
     fn the_path_fix_finds_the_install_the_rest_of_the_cli_found() {
         // Discriminating on purpose: the scanner this replaces looked only in
         // two hardcoded `C:\Program Files` directories and ignored $ROCM_PATH
@@ -1016,17 +1034,12 @@ mod tests {
         let install = root.join("rocm-6.10.0");
         plant_install(&install);
 
-        let previous = std::env::var_os("ROCM_PATH");
-        unsafe {
-            std::env::set_var("ROCM_PATH", &install);
-        }
-        let found = newest_rocm_install_dir();
-        unsafe {
-            match previous {
-                Some(value) => std::env::set_var("ROCM_PATH", value),
-                None => std::env::remove_var("ROCM_PATH"),
-            }
-        }
+        // The override goes in as an argument rather than through
+        // `std::env::set_var`: the environment is process-global, so a sibling
+        // test mutating $ROCM_PATH between this set and its read used to make
+        // this assertion fail on whichever test lost the race.
+        let found = newest_rocm_install_dir_in(&[], Some(&install));
+
         std::fs::remove_dir_all(&root).ok();
 
         assert_eq!(
@@ -1037,7 +1050,6 @@ mod tests {
     }
 
     #[test]
-    #[allow(unsafe_code)] // std::env::set_var is unsafe in edition 2024
     fn the_path_fix_reports_nothing_rather_than_a_directory_with_no_install_in_it() {
         // The old scan accepted any directory whose name started with a digit,
         // so an empty leftover could be put on PATH. The resolver requires a
@@ -1050,17 +1062,8 @@ mod tests {
         let empty = root.join("6.10");
         std::fs::create_dir_all(&empty).expect("create empty dir");
 
-        let previous = std::env::var_os("ROCM_PATH");
-        unsafe {
-            std::env::set_var("ROCM_PATH", &empty);
-        }
-        let found = newest_rocm_install_dir();
-        unsafe {
-            match previous {
-                Some(value) => std::env::set_var("ROCM_PATH", value),
-                None => std::env::remove_var("ROCM_PATH"),
-            }
-        }
+        let found = newest_rocm_install_dir_in(&[], Some(&empty));
+
         std::fs::remove_dir_all(&root).ok();
 
         assert!(
--- a/crates/rocm-core/src/lib.rs
+++ b/crates/rocm-core/src/lib.rs
@@ -1670,14 +1670,31 @@ impl AppPaths {
     }
 
     pub fn engine_envs_root(&self) -> PathBuf {
-        env_path_override("ROCM_CLI_ENGINE_ENVS_ROOT").map_or_else(
+        self.engine_envs_root_from(env_path_override("ROCM_CLI_ENGINE_ENVS_ROOT").as_deref())
+    }
+
+    /// [`Self::engine_envs_root`] against a caller-supplied override.
+    ///
+    /// Lets a test drive the override without `std::env::set_var`, which is
+    /// process-global and races other tests under a threaded runner. Same seam
+    /// as [`discover_rocm_installs_in_layout`].
+    fn engine_envs_root_from(&self, override_root: Option<&Path>) -> PathBuf {
+        override_root.map_or_else(
             || self.data_dir.join("engines"),
-            |root| normalize_runtime_path_for_host(&root),
+            normalize_runtime_path_for_host,
         )
     }
 
     pub fn engine_envs_dir(&self, engine: &str) -> PathBuf {
         self.engine_envs_root().join(engine).join("envs")
+    }
+
+    /// [`Self::engine_envs_dir`] against a caller-supplied override.
+    #[cfg(test)]
+    fn engine_envs_dir_from(&self, engine: &str, override_root: Option<&Path>) -> PathBuf {
+        self.engine_envs_root_from(override_root)
+            .join(engine)
+            .join("envs")
     }
 
     pub fn engine_locks_dir(&self, engine: &str) -> PathBuf {
@@ -2543,19 +2560,42 @@ pub(crate) struct RocmInstall {
     pub(crate) version: Option<String>,
 }
 
+/// The search roots and layout this host's installs use.
+fn host_rocm_search() -> (Vec<PathBuf>, RocmLayout) {
+    let (dirs, layout) = if runtime_is_windows() {
+        (WINDOWS_ROCM_SEARCH_DIRS, RocmLayout::Children)
+    } else {
+        (LINUX_ROCM_SEARCH_DIRS, RocmLayout::Siblings)
+    };
+    (dirs.iter().map(PathBuf::from).collect(), layout)
+}
+
 /// Every unmanaged ROCm install on this host, best candidate first.
 ///
 /// Supplies the platform's search roots, layout, and `$ROCM_PATH` to
 /// [`discover_rocm_installs_in_layout`].
 pub(crate) fn discover_rocm_installs() -> Vec<RocmInstall> {
     let env_override = std::env::var_os("ROCM_PATH").map(PathBuf::from);
-    let (dirs, layout) = if runtime_is_windows() {
-        (WINDOWS_ROCM_SEARCH_DIRS, RocmLayout::Children)
-    } else {
-        (LINUX_ROCM_SEARCH_DIRS, RocmLayout::Siblings)
-    };
-    let search_dirs: Vec<PathBuf> = dirs.iter().map(PathBuf::from).collect();
+    let (search_dirs, layout) = host_rocm_search();
     discover_rocm_installs_in_layout(&search_dirs, env_override.as_deref(), layout)
+}
+
+/// [`discover_rocm_installs`] with the host's `$ROCM_PATH` and search roots
+/// supplied by the caller instead of read from the process.
+///
+/// Reading `$ROCM_PATH` inside [`discover_rocm_installs`] is what forced its
+/// callers' tests to mutate the process environment, and two of them racing on
+/// that global is what made the Windows lane flake — that lane runs `cargo test`
+/// (threads in one process) where the Linux lanes run `cargo nextest` (a process
+/// per test), so only Windows could ever observe it. Passing the override in
+/// keeps those tests hermetic. Same seam as [`discover_rocm_installs_in`], but
+/// keeps the host's layout so the caller under test is the real one.
+#[cfg(test)]
+pub(crate) fn discover_rocm_installs_on_host_in(
+    search_dirs: &[PathBuf],
+    env_override: Option<&Path>,
+) -> Vec<RocmInstall> {
+    discover_rocm_installs_in_layout(search_dirs, env_override, host_rocm_search().1)
 }
 
 /// [`discover_rocm_installs_in_layout`] for the Linux sibling layout.
@@ -11020,28 +11060,35 @@ Class Name:                Display
     }
 
     #[test]
-    #[allow(unsafe_code)] // std::env::set_var is unsafe in edition 2024
     fn engine_envs_dir_honors_dedicated_root_override() {
         let (root, paths) = temp_app_paths("engine-envs-root-override");
         let override_root = root.join("runtime").join("engines");
-        let previous = std::env::var_os("ROCM_CLI_ENGINE_ENVS_ROOT");
-        unsafe {
-            std::env::set_var("ROCM_CLI_ENGINE_ENVS_ROOT", &override_root);
-        }
 
+        // Passed in rather than set via `std::env::set_var`: the environment is
+        // process-global, so mutating it here would race any concurrent test
+        // that reads the same key under a threaded runner.
         assert_eq!(
-            paths.engine_envs_dir("vllm"),
+            paths.engine_envs_dir_from("vllm", Some(&override_root)),
             normalize_runtime_path_for_host(&override_root)
                 .join("vllm")
                 .join("envs")
         );
 
-        unsafe {
-            match previous {
-                Some(value) => std::env::set_var("ROCM_CLI_ENGINE_ENVS_ROOT", value),
-                None => std::env::remove_var("ROCM_CLI_ENGINE_ENVS_ROOT"),
-            }
-        }
+        fs::remove_dir_all(root).ok();
+    }
+
+    #[test]
+    fn engine_envs_dir_falls_back_to_the_data_dir_without_an_override() {
+        // The other half of the override contract: with nothing supplied the
+        // root is the data dir, which is what the unset-environment production
+        // path resolves to.
+        let (root, paths) = temp_app_paths("engine-envs-root-default");
+
+        assert_eq!(
+            paths.engine_envs_dir_from("vllm", None),
+            paths.data_dir.join("engines").join("vllm").join("envs")
+        );
+
         fs::remove_dir_all(root).ok();
     }
 
--- a/xtask/src/env_mutation_contract.rs
+++ b/xtask/src/env_mutation_contract.rs
@@ -1,0 +1,271 @@
+// Copyright © Advanced Micro Devices, Inc., or its affiliates.
+//
+// SPDX-License-Identifier: MIT
+
+//! Static contract test: a test that mutates the process environment must
+//! serialize itself.
+//!
+//! Regression guard for EAI-8397. `std::env::set_var`/`remove_var` change state
+//! shared by every thread in the process, so two tests touching the same key
+//! race — one reads the other's value and fails an assertion that has nothing
+//! to do with what it is testing.
+//!
+//! This is invisible on most of CI. `ci.yml` runs the Linux lanes under
+//! `cargo nextest` (a process per test, so the mutation cannot escape), but
+//! `windows-build-and-test` runs `cargo test --workspace --all-targets` — all
+//! tests as threads in ONE process. So the Windows lane, a required check, was
+//! the only place the race could fire, and it read as "your change broke
+//! Windows" on branches that never touched the crate. Nothing about the bug is
+//! Windows-specific; it is a property of the runner.
+//!
+//! The repo already had two ways of handling this — `ScopedTestEnv` in
+//! `apps/rocm/src/main.rs` and `PROCESS_ENV_TEST_LOCK` in
+//! `apps/rocm/src/therock.rs`, both of which take a process-wide lock — and the
+//! `fix.rs` tests that flaked were simply the ones that skipped the discipline.
+//! This guard makes the omission a build failure rather than an intermittent
+//! red on a required check.
+//!
+//! Best is to not touch the environment at all: pass the value in through a
+//! test seam, as `discover_rocm_installs_on_host_in` and
+//! `newest_rocm_install_dir_in` do. Where a test must exercise the real
+//! env-reading path, holding one of the shared locks above is the accepted
+//! alternative and satisfies this guard.
+
+#[cfg(test)]
+mod tests {
+    use std::path::{Path, PathBuf};
+
+    fn repo_root() -> PathBuf {
+        // CARGO_MANIFEST_DIR is the xtask/ crate dir; its parent is the repo
+        // root (same idiom as workflow_contract::repo_root).
+        Path::new(env!("CARGO_MANIFEST_DIR"))
+            .parent()
+            .expect("xtask crate has a parent directory")
+            .to_path_buf()
+    }
+
+    /// Every `.rs` file in the repository.
+    ///
+    /// A whole-tree walk rather than a list of source directories, so a crate
+    /// added under a new top-level directory is covered without anyone
+    /// remembering to extend this. Two kinds of directory are skipped:
+    /// `target/`, which holds generated and vendored code we do not own, and
+    /// dot-directories, which include the `.claude/worktrees` checkouts of
+    /// other branches — scanning either would make the guard's verdict depend
+    /// on state outside the commit under test.
+    fn workspace_rust_sources() -> Vec<PathBuf> {
+        let mut found = Vec::new();
+        collect_rust_files(&repo_root(), &mut found);
+        assert!(
+            !found.is_empty(),
+            "found no Rust sources to scan -- the walk is broken, not the tree"
+        );
+        found.sort();
+        found
+    }
+
+    fn collect_rust_files(dir: &Path, found: &mut Vec<PathBuf>) {
+        let Ok(entries) = std::fs::read_dir(dir) else {
+            return;
+        };
+        for entry in entries.flatten() {
+            let path = entry.path();
+            if path.is_dir() {
+                let skip = path
+                    .file_name()
+                    .and_then(|name| name.to_str())
+                    .is_none_or(|name| name == "target" || name.starts_with('.'));
+                if skip {
+                    continue;
+                }
+                collect_rust_files(&path, found);
+            } else if path.extension().is_some_and(|ext| ext == "rs") {
+                found.push(path);
+            }
+        }
+    }
+
+    /// The mutating calls that need serializing.
+    const MUTATIONS: [&str; 2] = ["env::set_var", "env::remove_var"];
+
+    /// Markers that show a file already serializes its env mutations behind a
+    /// process-wide lock. Both are existing repo mechanisms; a file using
+    /// either has opted into the discipline this guard enforces.
+    const SERIALIZERS: [&str; 2] = ["ScopedTestEnv", "PROCESS_ENV_TEST_LOCK"];
+
+    /// Line numbers (1-based) inside a `#[cfg(test)]` module or a `#[test]`
+    /// function that call one of [`MUTATIONS`].
+    ///
+    /// Brace-depth tracking rather than a whole-file substring match: production
+    /// code in the same file may legitimately set an environment variable (it
+    /// owns the process), and a file-level match could not tell the two apart.
+    fn env_mutations_in_test_code(text: &str) -> Vec<(usize, String)> {
+        let mut hits = Vec::new();
+        let mut depth: usize = 0;
+        // Depth at which the enclosing test block opened, if we are inside one.
+        let mut test_block: Option<usize> = None;
+        let mut pending_test_attr = false;
+
+        for (index, raw) in text.lines().enumerate() {
+            let line = raw.trim();
+            let code = line.split("//").next().unwrap_or(line);
+
+            if test_block.is_none() && (line.contains("#[cfg(test)]") || line.contains("#[test]")) {
+                pending_test_attr = true;
+            }
+
+            if test_block.is_some()
+                && let Some(found) = MUTATIONS.iter().find(|needle| code.contains(*needle))
+            {
+                hits.push((index + 1, (*found).to_owned()));
+            }
+
+            let opens = code.matches('{').count();
+            let closes = code.matches('}').count();
+            if pending_test_attr {
+                if opens > 0 {
+                    test_block.get_or_insert(depth);
+                    pending_test_attr = false;
+                } else if code.trim_end().ends_with(';') {
+                    // The attribute belonged to a braceless item -- a
+                    // `#[cfg(test)] use ...;` import, say. Disarm, or the next
+                    // brace anywhere in the file would be mistaken for the
+                    // start of a test block and production code below it would
+                    // be scanned as test code.
+                    pending_test_attr = false;
+                }
+            }
+            // Saturating: a brace inside a string or macro can make a line look
+            // unbalanced, and the scan should degrade rather than panic.
+            depth = (depth + opens).saturating_sub(closes);
+            if let Some(open_depth) = test_block
+                && depth <= open_depth
+            {
+                test_block = None;
+            }
+        }
+        hits
+    }
+
+    /// Whether `text` serializes its env mutations behind a process-wide lock.
+    ///
+    /// File-level on purpose: both existing mechanisms declare the lock once and
+    /// take it in each test that needs it, so the guard asks whether the file
+    /// has opted in rather than trying to prove each call site is covered.
+    fn serializes_env_mutations(text: &str) -> bool {
+        SERIALIZERS.iter().any(|marker| text.contains(marker))
+    }
+
+    #[test]
+    fn a_test_mutating_the_environment_serializes_itself() {
+        let root = repo_root();
+        let mut offenders: Vec<String> = Vec::new();
+
+        for path in workspace_rust_sources() {
+            let text = std::fs::read_to_string(&path)
+                .unwrap_or_else(|e| panic!("reading {}: {e}", path.display()));
+            if serializes_env_mutations(&text) {
+                continue;
+            }
+            let relative = path.strip_prefix(&root).unwrap_or(&path).display();
+            for (line, call) in env_mutations_in_test_code(&text) {
+                offenders.push(format!("{relative}:{line}: {call}"));
+            }
+        }
+
+        assert!(
+            offenders.is_empty(),
+            "a test that mutates the process environment must serialize itself \
+             -- env is shared by every thread, so an unguarded mutation races \
+             any test reading the same key and fails only under a threaded \
+             runner (the Windows lane, a required check). Prefer passing the \
+             value in through a seam (see `newest_rocm_install_dir_in`); \
+             otherwise hold `ScopedTestEnv` or `PROCESS_ENV_TEST_LOCK`. \
+             Offenders:\n{}",
+            offenders.join("\n")
+        );
+    }
+
+    /// A mutating call, assembled at runtime.
+    ///
+    /// Spelling `env::set_var` literally in a fixture below would make this
+    /// file's own test code match the scan it defines, so the fixtures build
+    /// the needle instead of containing it.
+    fn mutation_call(kind: &str) -> String {
+        format!("unsafe {{ std::env::{kind}(\"KEY\", \"value\") }}")
+    }
+
+    #[test]
+    fn the_scanner_sees_a_mutation_inside_a_test_module() {
+        let call = mutation_call("set_var");
+        let source = format!(
+            "fn production() {{\n    {call}\n}}\n\
+             #[cfg(test)]\nmod tests {{\n    #[test]\n    fn t() {{\n        {call}\n    }}\n}}\n"
+        );
+        let hits = env_mutations_in_test_code(&source);
+        assert_eq!(hits.len(), 1, "expected exactly the in-test hit: {hits:?}");
+        assert_eq!(hits[0].0, 8, "should flag the line inside the test module");
+    }
+
+    #[test]
+    fn the_scanner_ignores_production_mutations() {
+        // Production code owns the process and may legitimately set a variable;
+        // the hazard is specific to tests sharing one process under `cargo test`.
+        let source = format!("fn production() {{\n    {}\n}}\n", mutation_call("set_var"));
+        assert!(
+            env_mutations_in_test_code(&source).is_empty(),
+            "a mutation outside test code is not an offense"
+        );
+    }
+
+    #[test]
+    fn the_scanner_stops_flagging_after_the_test_module_closes() {
+        let source = format!(
+            "#[cfg(test)]\nmod tests {{\n    #[test]\n    fn t() {{\n        let _ = 1;\n    }}\n}}\n\
+             fn later_production() {{\n    {}\n}}\n",
+            mutation_call("remove_var")
+        );
+        assert!(
+            env_mutations_in_test_code(&source).is_empty(),
+            "the test module ended; the later mutation is production code"
+        );
+    }
+
+    #[test]
+    fn the_scanner_ignores_a_cfg_test_attribute_on_a_braceless_item() {
+        // `#[cfg(test)] use ...;` gates an import, not a block. Treating it as
+        // the opening of a test block would make every brace after it -- the
+        // whole rest of the file -- read as test code. Several files in this
+        // repo (crates/rocm-dash-tui/src/ui/dock.rs, apps/rocmd/src/lib.rs) do
+        // exactly this above production code that sets an env var.
+        let source = format!(
+            "#[cfg(test)]\nuse foo::Bar;\n\nfn production() {{\n    {}\n}}\n",
+            mutation_call("set_var")
+        );
+        assert!(
+            env_mutations_in_test_code(&source).is_empty(),
+            "the attribute gated an import; nothing below it is test code"
+        );
+    }
+
+    #[test]
+    fn a_serialized_file_is_exempt_but_an_unguarded_one_is_not() {
+        let unguarded = format!(
+            "#[cfg(test)]\nmod tests {{\n    #[test]\n    fn t() {{\n        {}\n    }}\n}}\n",
+            mutation_call("set_var")
+        );
+        assert!(
+            !serializes_env_mutations(&unguarded),
+            "nothing in this source takes a process-wide lock"
+        );
+
+        let guarded = unguarded.replace(
+            "    #[test]",
+            "    static PROCESS_ENV_TEST_LOCK: Mutex<()> = Mutex::new(());\n    #[test]",
+        );
+        assert!(
+            serializes_env_mutations(&guarded),
+            "declaring the shared lock opts the file into the discipline"
+        );
+    }
+}
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -15,6 +15,7 @@ mod demos;
 mod e2e;
 mod e2e_prewarm;
 mod e2e_report;
+mod env_mutation_contract;
 mod manifest;
 mod package;
 mod paths;


tomastola:
## Review

The diagnosis here is good — identifying that the bug is a property of the *runner* (`cargo test` threads vs. `nextest` processes) rather than of Windows is what makes the fix correct rather than platform-specific. The root-cause approach (remove the shared mutation) beats serializing it. One substantive gap below, plus some soundness limits in the new guard.

Everything I claim below I ran locally on Linux against the PR head. Caveat up front: the flake only manifests on `windows-build-and-test`, so the fix is verified by mechanism and by Linux stress runs, not observed on the platform where it actually failed.

Reproduced green: `cargo test -p rocm-core --lib` → 324 passed; `cargo test -p xtask` → 137 passed; `cargo fmt --check` clean; `cargo clippy --locked --workspace --all-targets -- -D warnings` clean; 25/25 clean runs at `--test-threads=4`.

I also confirmed the efficacy claim rather than taking it on trust: reverting `fix.rs` to the merge base with the guard in place fails and names all six offending lines exactly (`fix.rs:1021, 1026, 1027, 1055, 1060, 1061`).

---

### 1. The refactor removes the only test coverage of both env-var lookups (blocking, in my view)

The new seam is `#[cfg(test)]`-only, so the tests no longer traverse the production line that reads the environment. I checked this by mutation testing — deleting both lookups from production:

```rust
// crates/rocm-core/src/lib.rs:2578
let env_override: Option<PathBuf> = None;   // was std::env::var_os("ROCM_PATH")
// crates/rocm-core/src/lib.rs:1673
self.engine_envs_root_from(None)            // was env_path_override("ROCM_CLI_ENGINE_ENVS_ROOT")
```

| Tree | Result with both lookups deleted |
|---|---|
| This PR | `cargo test -p rocm-core --lib` → **324 passed, 0 failed** |
| Merge base | `the_path_fix_finds_the_install_the_rest_of_the_cli_found` → **FAILED** (`fix.rs:1032`) |

So the old test did protect `$ROCM_PATH`, and after this PR nothing in the unit suite does. Deleting either line now ships green. The only remaining end-to-end exercise of `$ROCM_PATH` is `tests/e2e-cucumber/tests/e2e.rs:245`, a heavier gated lane; there's no equivalent for `ROCM_CLI_ENGINE_ENVS_ROOT` at all.

The awkward part is that the test's own assertion message — *"fix-6-path must resolve installs the same way examine does"* — is close to the claim the seam stops proving end to end. It still proves the shared resolver resolves; it no longer proves `fix-6-path` reaches it via `$ROCM_PATH`.

The guard already sanctions the remedy: *"Where a test must exercise the real env-reading path, holding one of the shared locks above is the accepted alternative and satisfies this guard."* Keeping the hermetic seam tests as written and adding one lock-serialized test per variable that drives the real entry point (`newest_rocm_install_dir()` / `engine_envs_root()`) behind `PROCESS_ENV_TEST_LOCK` would restore the coverage without reintroducing the race — and would exercise the escape hatch the guard advertises.

---

### 2. The guard's file-level exemption disables it where the risk is highest

`serializes_env_mutations` (`env_mutation_contract.rs:150-157`) is a bare substring test over the whole file and short-circuits before any line is scanned (`:167-169`). I verified the consequences by planting probe files and running the guard:

- **A comment alone exempts a whole file.** A file whose only mention is `// TODO: maybe migrate this to ScopedTestEnv one day`, containing an unguarded `std::env::set_var` inside `#[cfg(test)] mod tests`, was **not flagged**.
- **The two highest-risk files are wholly exempt.** `apps/rocm/src/main.rs` has 379 `#[test]` fns, 30 using `ScopedTestEnv`; `apps/rocm/src/therock.rs` has 73, 5 using `PROCESS_ENV_TEST_LOCK`. Both match a marker, so ~450 tests get no line-level enforcement, and a new unguarded env-mutating test added to either is invisible.

To be fair to the change: I enumerated every live `env::set_var`/`env::remove_var` site in the tree and they all sit inside the `ScopedTestEnv` / `ScopedEnvVar` / `PROCESS_ENV_TEST_LOCK` implementations themselves — no naked mutation in a `#[test]` body anywhere. The guard's green verdict is honest today. But these are the files most likely to accrue the next such test, and it's blind there.

Scoping the exemption to the enclosing `#[test]` block rather than the file — checking the marker appears between the test fn's opening brace and its matching close — would close both holes.

### 3. Two scanner blind spots let a real mutation through

Both verified end-to-end with probe files:

**Braces inside string literals desync the depth counter.** `:138-140` counts raw `{`/`}` anywhere on the line, including inside literals. This was **not flagged**:

```rust
#[cfg(test)]
mod tests {
    #[test]
    fn t() {
        let s = "}}";                              // closes test_block early
        unsafe { std::env::set_var("K", "v") };    // now invisible
    }
}
```

Any test building a JSON or format string containing `}` before mutating the environment is silently exempt. The mirror case (a stray `{`) leaks the block open and can flag *production* code as an offender.

**A bare one-line `#[test]` fn outside a `mod tests` block is missed** — `#[test]\nfn t() { unsafe { std::env::set_var(...) }; }` was not flagged, because the mutation check at `:117-121` runs before that line's own brace bookkeeping. Worth calibrating, though: the same one-liner *inside* `#[cfg(test)] mod tests {` **is** caught, and that's the universal style here, so live exposure is low.

Neither is exploited today. Both look cheap to fix; failing that, they'd be worth stating as known limits in the module doc rather than left implicit. The five scanner unit tests cover only happy paths — none of the above — so those are the cases to add.

---

### Smaller things

**`mutation_call`'s justification is not the operative one.** `:189-193` says the fixtures build the needle at runtime because *"Spelling `env::set_var` literally in a fixture below would make this file's own test code match the scan it defines."* But the file already contains those literals at `:89` (`const MUTATIONS`) with no ill effect — because `const SERIALIZERS` at `:94` contains `"ScopedTestEnv"`, which exempts the whole file first. The indirection is inert against the current implementation. Worth either correcting the comment or keeping the indirection and saying it's defensive against a future scoped exemption, which is the real argument for it.

**The rule is enforced by a hard build failure but documented nowhere.** Grepping `docs/`, `AGENTS.md`, `CONTRIBUTING.md` and `README.md` for `set_var|ScopedTestEnv|PROCESS_ENV_TEST_LOCK` returns zero hits. Someone who trips this gets a build failure citing conventions stated only inside a test file they've never opened — while `docs/testing.md` tells them to run `cargo test --workspace --all-targets`, the exact threaded harness, with no warning. A short note in `docs/testing.md` would close that.

**"The Linux lanes run `cargo nextest`" is imprecise.** True for the `test` job (`ci.yml:328`), but two other `ubuntu-latest` lanes run threaded single-process harnesses: `e2e` (`ci.yml:766`, `cargo test -p e2e-cucumber --lib`) and `coverage` (`ci.yml:543`, `cargo llvm-cov` without `--nextest`). Neither touches `rocm-core` today, so the diagnosis stands — but the claim is repeated verbatim in `env_mutation_contract.rs:12-18` and `lib.rs:2588-2589`, where it becomes a permanent statement in the tree. Narrowing it to the `test` job would keep it true.

**AGENTS.md §3.** The change looks purely internal, so no Gherkin scenario is required — but §3 also asks that you *"say why in the PR text rather than leaving it unexplained."* One sentence would close it.

**The host layout is dead weight for both callers.** `discover_rocm_installs_on_host_in` passes `host_rocm_search().1`, documented as keeping "the host's layout so the caller under test is the real one" (`lib.rs:2591-2592`). But `layout` is only consulted inside the `search_dirs` loops (`lib.rs:2641`, `:2655`), and both tests pass `&[]`, so the loops never run and layout can't affect the outcome. Harmless — and it does guarantee both tests behave identically on Windows and Linux — but the comment promises more than the code delivers.

---

### Tradeoff worth confirming as deliberate

The seam makes the tests hermetic and removes the race outright, which is genuinely better than serializing. But a seam structurally cannot test the wiring it bypasses, which is how item 1 arises. The guard permits both approaches; this PR uses the seam uniformly. Seam for resolution semantics *plus* a lock for the env wiring would get both properties.

### Things I liked

Verifying the guard by reverting the fix, and reporting the result, is the part I'd most want other PRs to copy — it survived independent re-checking. Removing the now-unneeded `#[allow(unsafe_code)]` attributes along with the `unsafe` blocks leaves no stale lint overrides. And the old `engine_envs_dir_honors_dedicated_root_override` leaked a mutated env var into sibling tests on the assert-failure path; that whole class is gone now.

---

Net: item 1 looks worth addressing before merge — it's a small addition, and it's what keeps this from making a required check green at the cost of an untested production line. Items 2 and 3 seem worth one more pass while you're in the file. The rest are docs and comment accuracy.

siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · a9189e59

## Summary
Replaces two `std::env::set_var`-mutating test pairs in `rocm-core` with caller-supplied override seams and adds a 271-line repo-wide xtask contract test that fails the build on unguarded env mutation in test code. **Needs work** — the de-flaking is the right instinct and the xtask guard follows the established `workflow_contract.rs` pattern, but the new seams reach one layer *past* the functions the old tests covered, so all four assertions now pass regardless of whether the production entry points work. Verified: I ran `cargo test -p xtask env_mutation_contract` (6/6 pass, including the repo-wide scan), independently grepped that `newest_rocm_install_dir()`, `engine_envs_root()` and `engine_envs_dir()` have **zero** test callers after this change (and `engine_envs_dir` had exactly one — the test — before), hand-traced the scanner's brace tracking to a reproducible false negative, confirmed `host_rocm_search()`/`first_install_path` are behaviour-preserving extractions, confirmed `git merge-tree` against current `main` is conflict-free, and confirmed the leak scan is clean; on the red check I relied on the caller's CI state and did **not** query GitHub — the guard itself passes locally and on the merged tree, so my leading diff-attributable hypothesis is a clippy `nursery` trigger (unverified, see below). Blocking: 3 · Non-blocking: 5.

**1. `crates/rocm-core/src/fix.rs:1041,1065` and `crates/rocm-core/src/lib.rs:11071,11088` — all four assertions would still pass if the code they protect were reverted.**
The revert question, answered concretely: after this change **no test anywhere calls `newest_rocm_install_dir()` (`fix.rs:983`) or `engine_envs_root()`/`engine_envs_dir()` (`lib.rs:1672,1688`)** — verified by grep across the whole tree; `newest_rocm_install_dir`'s only remaining caller is the recipe at `fix.rs:803`, and `engine_envs_dir` has none at all. The seams share only `first_install_path`/`engine_envs_root_from` with the real entry points, so the one line each test used to cover — `crate::discover_rocm_installs()` and `env_path_override("ROCM_CLI_ENGINE_ENVS_ROOT")` — is now unexercised. Concretely: typo the env key at `lib.rs:1673`, or make `newest_rocm_install_dir` ignore `$ROCM_PATH` again (exactly the regression `the_path_fix_finds_the_install_the_rest_of_the_cli_found`'s own docstring says it exists to prevent), and the suite stays green. Before this PR both tests did exercise that wiring. Fix: keep the hermetic seams for the resolution logic, but retain one assertion per key that calls the **real** `newest_rocm_install_dir()` / `engine_envs_dir()` under `PROCESS_ENV_TEST_LOCK` — the new guard's own module doc (`xtask/src/env_mutation_contract.rs:30-32`) explicitly sanctions that for "a test [that] must exercise the real env-reading path", so this costs nothing against the PR's goal.

**2. `crates/rocm-core/src/lib.rs:1672,1688` — the PR orphans `AppPaths::engine_envs_root`/`engine_envs_dir`.**
At the merge base, `engine_envs_dir` had exactly one caller in the entire repo: the test at `lib.rs:11033`. Redirecting that test to `engine_envs_dir_from` leaves both functions — and the sole in-process read of `ROCM_CLI_ENGINE_ENVS_ROOT`, a variable `apps/rocm` still *writes* when spawning engines (`main.rs:5666,15429,17914`) — with no callers at all. They are `pub` in a `publish = false` workspace with `unreachable_pub` deliberately disabled, so nothing will warn. Per the repo's own standard this is dead code. Fixing (1) as described resolves this too; otherwise decide deliberately whether the reader is still wanted and delete it if not.

**3. `xtask/src/env_mutation_contract.rs:123-145` — an unbalanced brace in a string literal silently defeats the scan.** Only `//` comments are stripped (`:111`); string literals are never tokenized, so their braces count as nesting. Traced against the code:

```rust
#[cfg(test)]
mod tests {                    // test_block = Some(0), depth = 1
    #[test]
    fn t() {                   // depth = 2
        let s = "}}";          // closes=2 -> depth = 0 <= 0 -> test_block = None
        unsafe { std::env::set_var("KEY", "v"); }   // NOT flagged
    }
}
```

`test_block` exits two lines early and the offender is missed. The comment at `:138-140` says the saturating subtraction makes the scan "degrade rather than panic" — but it degrades into a silent false negative, which is the one failure mode a guard like this must not have, and none of the five scanner tests (`:199-270`) cover a brace inside a string. This is latent, not live (I confirmed no current test has a brace-bearing literal ahead of an env mutation), but it should not ship unfixed in a permanent guard. Cheapest fix: strip string/char literal contents (or just skip lines whose quote count is odd/whose literals contain braces) before counting, and add a fixture test for it. The narrower `line.split("//")` truncation at `:111` (a `"https://…"` literal drops everything after it, including a same-line mutation) has the same root cause and the same fix.

## Non-blocking
- `xtask/src/env_mutation_contract.rs:239-240` — the comment claims `crates/rocm-dash-tui/src/ui/dock.rs` and `apps/rocmd/src/lib.rs` gate an import "above production code that sets an env var"; neither file contains `set_var`/`remove_var` at all (the `#[cfg(test)] use …;` pattern itself is real). Drop the env-var clause.
- Branch is 16 commits behind `main`, which landed an independent fix for this exact `fix.rs` race (via `PROCESS_ENV_TEST_LOCK`) earlier today; `git merge-tree` merges cleanly, but the result keeps a lock static plus two `let _guard` acquisitions protecting tests that no longer touch the environment — and that leftover marker then blanket-exempts `fix.rs` from the new guard forever. Rebase and reconcile the two approaches explicitly.
- `crates/rocm-core/src/fix.rs:1041,1065` — passing `&[]` for `search_dirs` means the "`$ROCM_PATH` outranks the hardcoded search roots" ordering the first test's docstring calls out is no longer exercised at all; pass the planted root alongside the override.
- `xtask/src/env_mutation_contract.rs:89` vs `:94` — the `MUTATIONS` const spells `env::set_var` literally inside `mod tests`, so the guard would flag itself; it is saved only by `SERIALIZERS` (`:94`) happening to contain a marker string, which file-level-exempts the whole file. The `mutation_call` runtime-assembly trick at `:194` exists to avoid exactly this, so the const should get the same treatment rather than relying on an accidental self-exemption. Relatedly, `serializes_env_mutations` (`:155-157`) is a plain substring match, so a marker named in any comment exempts an entire file — worth noting on a ~25k-line file like `apps/rocm/src/main.rs` (whose current uses I did verify are all genuinely locked).
- Leading diff-attributable hypothesis for the red check, **unverified**: `env_mutations_in_test_code` (~46 lines, nested `if`s and let-chains) is a plausible `clippy::cognitive_complexity` trigger — `nursery` is warn-level workspace-wide and CI runs `-D warnings`. I could not confirm it (my one permitted check went to the guard test). Otherwise nothing in the diff explains a failure: the guard passes on the head and, by inspection, on the merged tree, and the two always-compiled extractions are behaviour-preserving — which points at the stale base or an environment-gated lane instead. Please read the failing job log rather than assume flake.

siloteemu (changes requested):
🔴 Automated review · pr-review-watcher · a9189e59

_**One objection is live on this pull request: this one.** Our earlier change request at this same commit has been dismissed and carries no weight. GitHub still renders a dismissed review's full blocking section, which is why this says so explicitly — there is one thing to answer here, not two._

## Summary

Forced re-examination of the three blocking counts we filed on 2026-09-09. The head has not moved since (`a9189e59`), so any count that no longer holds would mean our original review was wrong. I re-derived each one from scratch rather than inheriting it.

**All three counts still hold.** Each was re-verified against the source at this commit, and count 3 turns out to be *understated* rather than overstated. Separately, `main` has moved 38 commits since this branch's merge base, and that movement creates a **fourth blocking problem that did not exist when we first reviewed**: the new guard fails on a conflict-free merge with current `main`.

One paragraph of our original review must be **withdrawn**: it speculated about a red check and offered a clippy hypothesis for it. There is no failing check at this head — see the CI counts below — so that speculation was unfounded and is retracted.

**Blocking: 4** (3 restated, 1 new) · **Non-blocking: 4** (1 withdrawn, 1 revised)

CI conclusions worked from at this head: **24 success, 1 skipped, 1 cancelled, 0 failure** (plus one check with no conclusion reported). A cancelled job is not a failure. Nothing in this report rests on attributing an outcome to a named job.

Method note: the full test suite was not run here. Findings below are from static analysis, cross-tree greps, a faithful re-implementation of the guard's scanner run against the real files, and a trial merge performed without modifying the reviewed tree.

---

## Re-check, count by count

### Count 1 — STILL HOLDS

The sentence under test, verbatim from our change request:

> **`crates/rocm-core/src/fix.rs:1041,1065` and `crates/rocm-core/src/lib.rs:11071,11088` — all four assertions would still pass if the code they protect were reverted.**

Re-derived and confirmed. Caller census across every `.rs` file at this commit:

- `newest_rocm_install_dir()` (`fix.rs:983`) — one caller, the recipe at `fix.rs:803`. **No test calls it.** At the merge base it had two test callers (`fix.rs:1023,1057`).
- `engine_envs_dir()` (`lib.rs:1688`) — **zero callers anywhere.** At the merge base it had exactly one: the test at `lib.rs:11033`.
- `engine_envs_root()` (`lib.rs:1672`) — its only caller is `engine_envs_dir()`, which itself has none.

Tracing what the seams share with the real entry points:

- `newest_rocm_install_dir_in` → `discover_rocm_installs_on_host_in` → `discover_rocm_installs_in_layout`. It does keep the real layout selection (`host_rocm_search().1`) and the real `first_install_path`. What it bypasses is the one line the tests existed for: the `std::env::var_os("ROCM_PATH")` read inside `discover_rocm_installs()`.
- `engine_envs_dir_from` → `engine_envs_root_from`. It bypasses `env_path_override("ROCM_CLI_ENGINE_ENVS_ROOT")` at `lib.rs:1673`.

So the branch-level mutations are concrete and each survives the suite: make `discover_rocm_installs()` stop reading `$ROCM_PATH` — exactly the regression `the_path_fix_finds_the_install_the_rest_of_the_cli_found`'s own docstring says it exists to prevent — and nothing fails. Typo the key string at `lib.rs:1673` and nothing fails.

**New evidence that sharpens this count.** Current `main` has already fixed this same race in `fix.rs`, and it did so *without* losing the coverage: `main` declares `PROCESS_ENV_TEST_LOCK` at `fix.rs:1343` and both tests take it at `fix.rs:1381,1418` while still calling the real `newest_rocm_install_dir()` against a real `$ROCM_PATH`. Measured against what the project already has, this PR is a net coverage regression on `fix.rs`, not an improvement. The `lib.rs` half is genuinely still unfixed on `main`, so that part of the PR's motivation stands.

**What would resolve it:** keep the hermetic seams for the resolution logic, but retain one assertion per key that drives the *real* `newest_rocm_install_dir()` / `engine_envs_dir()` under a process-wide lock. The guard's own module doc (`xtask/src/env_mutation_contract.rs:28-32`) explicitly sanctions this for a test that must exercise the real env-reading path, so it costs nothing against the PR's stated goal. For `fix.rs` specifically, rebasing onto `main` already gives you the lock.

### Count 2 — STILL HOLDS

The sentence under test, verbatim:

> **`crates/rocm-core/src/lib.rs:1672,1688` — the PR orphans `AppPaths::engine_envs_root`/`engine_envs_dir`.**

Confirmed exactly as written. Both the "one caller at the merge base" and "zero callers now" halves check out (evidence above). The supporting claim that nothing will warn also checks out: `publish = false` is set workspace-wide, and the root manifest documents at `Cargo.toml:64-65` that `unreachable_pub` is deliberately *not* enabled, with the reason given. So this dead `pub` chain is invisible to the compiler and to lint.

Worth stating precisely so it is not overclaimed: `apps/rocm` writes `ROCM_CLI_ENGINE_ENVS_ROOT` into spawned children (`main.rs:5666,15429,17914`), and `lib.rs:1673` is the only code that reads it back. That read was already unreachable from production before this PR — the PR does not create the write-only contract, it removes the last remaining caller and makes the whole chain dead.

**What would resolve it:** fixing count 1 as described resolves this too, by restoring a caller. Otherwise decide deliberately whether the reader is still wanted, and delete it if not.

### Count 3 — STILL HOLDS, and was understated

The sentence under test, verbatim:

> **`xtask/src/env_mutation_contract.rs:123-145` — an unbalanced brace in a string literal silently defeats the scan.**

Confirmed, and the qualifier we attached to it — *"This is latent, not live"* — was too generous. I re-implemented the scanner faithfully and ran it against the real tree alongside a corrected version that strips string and char literals properly before counting braces.

The two scanners agree that there are **no offenders at this head**, so the guard's *verdict* is correct today. But the depth accounting underneath it is already desynchronised on real files, and the blind spot that creates is live right now:

| file | real test lines the guard does **not** watch |
|---|---|
| `engines/lemonade/src/lib.rs` | 471 |
| `tests/e2e-cucumber/src/mock_server.rs` | 39 |
| `crates/rocm-dash-collectors/src/vllm_prom.rs` | 1 |

**511 lines of real test code currently sit outside the guard's scope.** An unguarded `env::set_var` added on any of them passes the guard silently — which is the one failure mode a permanent guard must not have.

The causes are the two lines the count named:

- `engines/lemonade/src/lib.rs:4566` is `ensure_cached_archive("test://archive", …, |_, destination| {`. The `line.split("//")` truncation at `:111` cuts the line at the `//` *inside the URL literal*, discarding the closure's opening brace. Five such lines in that file leave `mod tests` looking closed from line 4576 onward — 471 lines early.
- `tests/e2e-cucumber/src/mock_server.rs` contains format strings such as `vllm:num_requests_running{{model=…}}`. Braces inside the literal are counted as nesting, which is the exact mechanism our count described.

The mirror-image failure also exists and we did not name it: an *opening* brace inside a literal inflates the depth and pulls production code into test scope. Across 13 files that the guard does not file-level-exempt, **3352 production lines are wrongly inside the scanned region** — so a legitimate `env::set_var` added to production code in, for example, `crates/rocm-dash-collectors/src/bench_load.rs` or `apps/rocmd/src/lib.rs` would fail the build for no reason.

**The scanner's own tests do not cover either defect.** Mutating each branch of `env_mutations_in_test_code` and re-running all five fixture tests:

| branch mutated | result |
|---|---|
| remove the `//` comment strip (`:111`) | **survives — no test detects it** |
| replace `saturating_sub` with plain subtraction (`:140`) | **survives — no test detects it** |
| never disarm the braceless-item attribute | killed by `the_scanner_ignores_a_cfg_test_attribute_on_a_braceless_item` |
| `depth < open_depth` instead of `<=` | killed by `the_scanner_stops_flagging_after_the_test_module_closes` |

The two surviving branches are precisely the two defective ones. The comment at `:138-140` says the saturating subtraction makes the scan "degrade rather than panic" — it degrades into a silent false negative, and nothing in the suite would notice if that line were changed.

**What would resolve it:** strip string and char literal contents before counting braces and before splitting on `//`, and add fixtures for both — a brace inside a literal ahead of a mutation, and a `//` inside a literal ahead of one. Both fixtures fail against the current scanner and pass against a corrected one.

---

## New blocking finding at this commit

### 4. The guard fails the build on a clean merge with current `main`

This did not exist when we first reviewed; `main` has moved underneath the branch.

`crates/rocm-dash-tui/src/app/mod.rs` gained a third process-wide env lock, `UPDATE_CHECK_ENV_TEST_LOCK`, in commit `b0147c6` (2026-09-14). It is used correctly: the test at `app/mod.rs:7266` takes the lock, then mutates `ROCM_CLI_DISABLE_STARTUP_UPDATE_CHECK` and restores it. But the guard's allowlist is a fixed two-element list:

```rust
const SERIALIZERS: [&str; 2] = ["ScopedTestEnv", "PROCESS_ENV_TEST_LOCK"];
```

`UPDATE_CHECK_ENV_TEST_LOCK` is not in it. I took a trial merge of this head with current `main` (conflict-free) and ran the scanner over the merged tree: `a_test_mutating_the_environment_serializes_itself` reports two offenders, `crates/rocm-dash-tui/src/app/mod.rs:7274` and `:7285`. Both hits disappear if the third lock name is allowlisted, and a correctly-written scanner reports them too — so these are genuine applications of the guard's rule against correctly-disciplined code, not artifacts of the brace bug in count 3.

As written, the guard would tell a maintainer to rename a perfectly good lock. That the allowlist went stale within five days of the guard being written is itself the signal: a hardcoded list of lock *names* is the wrong shape for a permanent contract test.

**What would resolve it:** recognise the discipline rather than the name — for example match a `_TEST_LOCK` suffix, or require that a mutation be preceded within its own test body by a lock acquisition. At minimum, add the third name and state in the file why the list is expected to grow.

---

## Non-blocking

- **WITHDRAWN.** Our original review offered a clippy `nursery` / `cognitive_complexity` hypothesis for a red check. There is no failing check at this head (24 success, 1 skipped, 1 cancelled, 0 failure), so the hypothesis had nothing to explain. Retracted in full; we should not have published an unverified CI theory.
- **REVISED.** We said the branch was "16 commits behind `main`". It is now **38 behind**. The substance is unchanged but worse, and the merged result is now concretely checkable: the trial merge keeps `main`'s `PROCESS_ENV_TEST_LOCK` static and both `let _guard` acquisitions in `fix.rs` *and* this PR's seam-based tests, so the merged tests hold a lock they no longer need. More importantly, that leftover marker makes `crates/rocm-core/src/fix.rs` **file-level exempt from the new guard permanently** — the guard would stop watching the very file it was written for. Rebase and reconcile the two approaches explicitly rather than letting the auto-merge pick both.
- `xtask/src/env_mutation_contract.rs:239-240` — the comment claims `crates/rocm-dash-tui/src/ui/dock.rs` and `apps/rocmd/src/lib.rs` gate an import "above production code that sets an env var". Neither file contains `set_var`/`remove_var` at all. The `#[cfg(test)] use …;` pattern itself is real; drop the env-var clause.
- `crates/rocm-core/src/fix.rs:1041,1065` — passing `&[]` for `search_dirs` means the "`$ROCM_PATH` outranks the hardcoded search roots" ordering that the first test's docstring calls out is no longer exercised at all. Pass the planted root alongside the override.
- `xtask/src/env_mutation_contract.rs:89` vs `:94` — the `MUTATIONS` const spells `env::set_var` literally inside `mod tests`, so the guard would flag itself; it is saved only by `SERIALIZERS` at `:94` happening to contain a marker string, which file-level-exempts the whole file. I confirmed this against the real tree: the guard's own file is one of only three that the scan exempts. The `mutation_call` runtime-assembly trick at `:194` exists to avoid exactly this, so the const deserves the same treatment rather than relying on an accidental self-exemption. Relatedly, `serializes_env_mutations` (`:155-157`) is a plain substring match, so naming a marker in any comment exempts an entire file.

## Tradeoff worth surfacing

Dependency injection (this PR) and lock serialization (what `main` already does for `fix.rs`) are both legitimate answers to the race, and the guard's module doc endorses both. The choice is not wrong — but it should be made once, deliberately, for both call sites, instead of arriving at a merge that silently keeps both mechanisms in the same file. If injection is the preferred direction, `main`'s lock in `fix.rs` should be removed in the same change rather than left to blanket-exempt the file.

## Positive signals

- The whole-tree walk in `workspace_rust_sources` (with its documented reasons for skipping `target/` and dot-directories) is the right call — a crate added under a new top-level directory is covered without anyone remembering to extend a list.
- The module doc explaining *why* only one runner configuration could ever observe the race, and that nothing about the bug is platform-specific, is genuinely useful context that will save the next reader real time.
- `host_rocm_search()` and `first_install_path()` are clean, behaviour-preserving extractions.

## Recurrence check

Two things here will trip a competent reader again unless the code says otherwise:

1. **`engine_envs_root`/`engine_envs_dir` look alive because they are `pub`.** With `unreachable_pub` deliberately off and the workspace unpublished, a `pub` method carries no signal about whether anything calls it. Anyone extending `AppPaths` will reasonably assume these are part of a live API. The cheap prevention is a one-line comment on the surviving reader saying who is expected to call it — or deleting it.
2. **The guard reads as stronger than it is.** Its assertion message is confident and its five fixtures all pass, which invites the conclusion that the scan is sound. Nothing in the file warns that brace depth is tracked by counting characters on a line with literals left in. A single comment at `:123` stating that literals are not tokenized, plus the two fixtures named in count 3, would prevent the next reader — and the next automated check — from trusting a green result more than it deserves.

Nothing in the diff, the commit message, or the PR discussion contained text addressed to an automated reviewer.

rominf:
Thanks — all four counts are addressed at `937a62a`. The branch is also rebased onto current `main`, so the trial-merge findings apply to real code now rather than a hypothetical merge.

Everything below I ran locally on Linux at this head. Where the reviewer's method was mutation testing, I repeated it rather than asserting the fix works.

### Count 1 — env-read coverage restored

Both keys have a test that drives the real entry point again, under a process-wide lock rather than a seam, because exercising the read *is* the point:

- `fix.rs` — `the_path_fix_reads_rocm_path_from_the_environment` calls `newest_rocm_install_dir()` against a real `$ROCM_PATH`.
- `lib.rs` — `engine_envs_dir_reads_its_root_from_the_environment` calls `engine_envs_dir()` against a real `$ROCM_CLI_ENGINE_ENVS_ROOT`.

Verified the way the count framed it. Deleting both production lookups (`let env_override: Option<PathBuf> = None;` and `self.engine_envs_root_from(None)`) gives `cargo test -p rocm-core --lib` → **408 passed, 2 failed**, and the two failures are exactly those tests. Restored, 410 pass.

The hermetic seams stay for the resolution semantics, which is the split you proposed. The first seam test also no longer passes `&[]`: it plants an older install under a real search root and passes it alongside the override, so "the override outranks the search roots" is exercised rather than vacuous.

### Count 2 — the chain has a caller

`engine_envs_dir()` → `engine_envs_root()` → `env_path_override("ROCM_CLI_ENGINE_ENVS_ROOT")` is reached by the test above, and the mutation result is the proof that the reach is real. Kept rather than deleted, and `engine_envs_root` now carries the comment the recurrence check asked for: it states that this is the read side of a write-only contract, that no in-tree production path reaches it, that `pub` carries no signal here because `unreachable_pub` is off workspace-wide, and which test keeps it honest.

### Count 3 — literals are tokenized

`strip_literals_and_comments` replaces the contents of string, raw-string, char literals and comments with spaces before anything counts a brace or looks for a call, preserving line structure so reported line numbers stay right. Lifetimes are distinguished from char literals.

The two fixtures you specified are both in, and both fail against the old scanner:

| fixture | old scanner | new scanner |
|---|---|---|
| `a_brace_inside_a_string_literal_does_not_close_the_test` | FAILED | ok |
| `a_double_slash_inside_a_string_literal_does_not_truncate_the_line` | FAILED | ok |
| `an_opening_brace_in_a_literal_does_not_drag_in_production_code` | FAILED | ok |

"Old scanner" means the current implementation with `strip_literals_and_comments` made a no-op and the `line.split("//")` truncation put back. Under that emulation the whole-tree scan also fails, naming `fix.rs:1446` and `lib.rs:12282` — two correctly-disciplined tests reported as offenders. That is the false-positive direction firing on real code, so the tokenizer is load-bearing, not precautionary.

Two things worth flagging honestly:

- The opening-brace fixture was passing for the wrong reason when I first checked it. One `{` in the literal was absorbed by the enclosing module's closing brace, so it passed against a broken scanner too. It now uses `{{`, which is also the realistic shape (`{{model=...}}` in a Prometheus format string).
- Making it discriminate exposed a real silent-miss path: offenses collected in a test body that was still open when the file ended were dropped on the floor. Those are now reported. Over-reporting is visible and gets fixed; a discarded hit is the one failure mode this guard exists to prevent.

On the second surviving mutant: replacing `saturating_sub` with plain subtraction still kills no test. It is now an equivalent mutant rather than a defect — with literals tokenized away the brace count is balanced on any file that compiles, so the saturation is unreachable. Rather than write a test for unreachable behaviour I said so at the line: it is a backstop against a file this scan misreads panicking the build, not part of the logic.

### Count 4 — the allowlist recognises the discipline

`NAMED_SERIALIZERS` is now `["ScopedTestEnv", "ScopedEnvVar"]` plus a rule matching any identifier ending `_TEST_LOCK`, with a fixture (`SOME_BRAND_NEW_TEST_LOCK`) asserting a lock the guard has never heard of is accepted. The const comment records why the list is by suffix.

Confirmed against the merged reality rather than by reading: the branch is rebased onto current `main`, which contains `UPDATE_CHECK_ENV_TEST_LOCK`, and `a_test_mutating_the_environment_serializes_itself` passes. Full `cargo test -p xtask` → 151 passed.

### On the tradeoff you raised

Taken as a decision rather than left to the merge. `fix.rs` keeps both mechanisms, and the reason is written at the lock declaration: the seam for tests about resolution semantics, the lock for the single test whose subject is the environment read. Anything provable through the seam should not be reaching for the lock. Post-rebase `fix.rs` holds exactly one lock acquisition, on that one test — not the two the trial merge would have carried over. The per-test-function scoping also means the lock's presence no longer exempts the file: it never grants anything to its neighbours.

### Other points

- The comment claiming `dock.rs`/`rocmd/src/lib.rs` gate an import above production code that sets an env var is gone; that clause was wrong.
- `MUTATIONS` spelling the calls literally is no longer an accidental self-exemption — the const is not inside a `#[test]` body, so the per-function rule never looks at it. The comment on `mutation_call` now says that instead of the old justification.
- The "Linux lanes use nextest" claim is narrowed in the module doc: `coverage` and `e2e` are threaded single-process harnesses on `ubuntu-latest` too, and happen not to reach the crates that raced.
- `docs/testing.md` documents the rule, with the lock-and-restore pattern, the suffix convention, the per-function scope, and the known blind spot where a `#[test]` delegates its mutation to an unguarded helper.
- AGENTS.md §3 is answered in the PR description: nothing user-observable changes, so no scenario.

### What I did not run

`windows-build-and-test` is the only lane that can observe the original race; I have no Windows host, so the fix remains verified by mechanism and by Linux stress runs. Three `rocm-dash-tui` `agent::tests` fail in my local sandbox with `cannot create token cache dir: Permission denied` — they pass with a writable `HOME`, and this branch touches no code they reach.

The counts are yours to close — I have left the review as it stands.

rominf:
CI status at `937a62a`: all 16 required checks pass, including `windows-build-and-test` — the lane this PR exists for — and `Commit signatures + sign-off`.

Two advisory self-hosted lanes ended `cancelled`: `E2E tests (Strix Halo, Ubuntu)` and `E2E tests (Strix Halo, Windows)`. Both hit their 45-minute job cap on a cold DevLab pool runner, not an assertion:

- Ubuntu ([job log](https://github.com/ROCm/rocm-cli/actions/runs/35712409288/job/106696028235)) spent 11:06→11:45 on the prewarm fetch (3.3 GB, two tarballs) and was cut off mid-download.
- Windows ([job log](https://github.com/ROCm/rocm-cli/actions/runs/35712409288/job/106696028182)) did a cold cargo build, then fetched 4.7 GB, then was still running scenarios — several passing — when the cap hit.

Both lanes are declared non-blocking in `e2e-selfhosted.yml` and are not in the required set. The diff touches `crates/rocm-core` tests, an `xtask` contract test and `docs/testing.md` — no e2e code, no build configuration — so nothing here changes what those lanes execute. The four GPU lanes that did fit the budget (MI300X, MI350P, rad3 R9700, Strix Halo WSL2) all passed.

Flagging rather than silently leaving two non-green lanes. Happy for a maintainer to re-run them if a warmer runner would settle it.

siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · 937a62ad

## Summary
De-flakes two `rocm-core` test pairs that mutated the process-global environment (seams for the resolution semantics, a process-wide lock for the one test whose subject *is* the env read), adds a whole-tree `xtask` contract test that fails the build on an unguarded `env::set_var`/`remove_var` inside a `#[test]`, and documents the rule in `docs/testing.md`. **Needs work** — all four of our earlier counts are genuinely discharged and the remediation is strong, but one single-branch mutation still survives the new guard's own fixture suite. Verified: I ran the guard on a scratch copy merged with the current base tip (13/13 pass, including the whole-tree scan, with `UPDATE_CHECK_ENV_TEST_LOCK` present in the merged code); mutation testing of `rocm-core` and of the scanner ran as targeted `cargo test -p rocm-core --lib` / `cargo test -p xtask` runs in separate scratch copies, with the reviewed checkout left untouched and clean; I re-derived the caller census, the fixture census and the two test counts myself; the full test suite, the e2e suites and the Windows lane were not run here. The branch is 4 commits behind the base tip; the merge is conflict-free and the guard stays green on the merged tree. The diff against the merge base and against `prw-base` are identical. Checks at this head: 25 success, 2 cancelled, 1 skipped. Blocking: 1 · Non-blocking: 5.

## Prior objections — verdicts

- **Count 1** — *"`crates/rocm-core/src/fix.rs:1041,1065` and `crates/rocm-core/src/lib.rs:11071,11088` — all four assertions would still pass if the code they protect were reverted."* **DISCHARGED**, by mutation rather than by the reply. Deleting the `ROCM_PATH` read in `discover_rocm_installs()` now fails `the_path_fix_reads_rocm_path_from_the_environment`; passing `None` in `engine_envs_root()`, and separately typoing the key string, each fail `engine_envs_dir_reads_its_root_from_the_environment`; breaking the no-override fallback fails `engine_envs_dir_falls_back_to_the_data_dir_without_an_override`; making `env_override` inert inside `discover_rocm_installs_in_layout` fails `the_path_fix_finds_the_install_the_rest_of_the_cli_found` as well. Each was a single branch, reverted before the next.
- **Count 2** — *"`crates/rocm-core/src/lib.rs:1672,1688` — the PR orphans `AppPaths::engine_envs_root`/`engine_envs_dir`."* **DISCHARGED** on the terms we ourselves set ("fixing count 1 as described resolves this too, by restoring a caller"). `engine_envs_dir()` has exactly one caller again — the new env-read test — and the chain now carries a comment stating it is the read side of a write-only contract, that `pub` carries no signal because `unreachable_pub` is off workspace-wide, and which test keeps it honest. The production chain is still write-only, which we already recorded last round as predating this PR.
- **Count 3** — *"`xtask/src/env_mutation_contract.rs:123-145` — an unbalanced brace in a string literal silently defeats the scan."* **DISCHARGED.** `strip_literals_and_comments` now tokenizes string, raw-string and char literals plus line and block comments before anything counts a brace, lifetimes are distinguished from char literals, and offenses left in a test body still open at EOF are reported rather than dropped. Mutating away the stripper, the raw-string branch, the char-literal branch, the block-comment branch or the EOF flush each kills a fixture. The surviving `saturating_sub` mutant is correctly argued as equivalent: with literals stripped the brace count is balanced on any file that compiles.
- **Count 4** — *"The guard fails the build on a clean merge with current `main`."* **DISCHARGED.** The allowlist now recognises the discipline (`ScopedTestEnv`, `ScopedEnvVar`, plus any identifier ending `_TEST_LOCK`). I confirmed this against the merged tree, not the head: `a_test_mutating_the_environment_serializes_itself` is green with `UPDATE_CHECK_ENV_TEST_LOCK` in scope.

**`xtask/src/env_mutation_contract.rs:95` — half of `MUTATIONS` is unprotected by the fixtures.** Dropping `"env::remove_var"` from the const leaves the whole fixture suite green. `remove_var` appears in exactly one fixture, `the_scanner_stops_flagging_after_the_test_closes` at `:467`, which asserts the result `is_empty()` — so no test anywhere asserts that an unguarded `remove_var` inside a `#[test]` *is* flagged. The detection works today; nothing proves it, and this is the same failure shape the previous round blocked on (a branch of the guard that no fixture kills). It matters more here than a normal coverage gap because this file is the thing that decides whether the rule is enforced at all. Fix: add a fixture asserting `env_mutations_in_unserialized_tests(&unguarded_test(&mutation_call("remove_var"))).len() == 1` — that fails against the mutated const and costs three lines.

## Non-blocking
- `xtask/src/env_mutation_contract.rs:335` — the scan arms on the literal `#[test]`, so the 87 `#[tokio::test]` functions across 16 files are invisible to it. None mutates the environment today, but the file's "Known limit" paragraph (`:319-323`) names only the helper-delegation case; either match the attribute more loosely or add this to the stated limits.
- `xtask/src/env_mutation_contract.rs:315-317` — the comment's test counts are already stale: `apps/rocm/src/main.rs` has 461 `#[test]`s and `apps/rocm/src/therock.rs` 143, not 379 and 73. The point survives without the numbers; drop them rather than re-pin them.
- `xtask/src/env_mutation_contract.rs:225-229` — inside a string literal a backslash consumes two characters unconditionally, so a trailing-backslash line continuation destroys the newline and merges two source lines. Several files in the tree use that form. The verdict is unaffected (brace totals are unchanged), but any offense reported later in such a file carries a shifted line number, which undercuts the `file:line` the assertion message promises.
- `docs/testing.md:33-35` — "`cargo xtask`'s `a_test_mutating_the_environment_serializes_itself`" reads as a subcommand; there is none, it only runs under `cargo test`. Say the actual invocation. The snippet below it also uses `SOMETHING_ENV_TEST_LOCK` without declaring it, so it will not compile as pasted.
- `crates/rocm-core/src/fix.rs:1441-1447` — the planted decoy `rocm-6.2.0` only matches the sibling layout, so on Windows the search loop finds nothing and the "`$ROCM_PATH` outranks the hardcoded search roots" half of the assertion goes vacuous again — on the lane this PR exists for. Plant a `6.2`-shaped child as well, or say in the comment that the ordering half is Linux-only.

rominf:
Thanks — all six are addressed. Head is now `6ddfa65d`, rebased onto `248c0d92`.

## Blocking — `MUTATIONS` half-unprotected

You were right, and the fixture you suggested is in as `every_mutating_call_is_flagged_inside_an_unguarded_test`. It spells the call kinds out rather than iterating `MUTATIONS`: a fixture that read the const would shrink along with it and stay green, which is the hole itself. It also asserts the two lists are the same length, so a third entry added to the const fails here until it gains a case.

Two probes, both run:

| Mutation to `MUTATIONS` | Result |
|---|---|
| drop `"env::remove_var"` (const becomes `[&str; 1]`) | **FAILED** — `a call was added to MUTATIONS without a case here, left: 1, right: 2` |
| keep length 2, change the needle to `"env::unset_var"` | **FAILED** — `an unguarded std::env::remove_var inside a #[test] must be flagged` |

Both restored afterwards; `cargo test -p xtask` is green on the committed tree.

**Sweep for the same shape elsewhere.** `MUTATIONS` has only the two entries, and `set_var` was already well covered, so `remove_var` was the only gap in that const. I mutated the neighbouring ones too rather than assume: dropping either `NAMED_SERIALIZERS` entry, and breaking the `_TEST_LOCK` suffix rule, each already fail existing fixtures. So nothing else in the allowlist was unprotected.

**One more unkilled branch found while sweeping**, same shape as the one you flagged: the braceless-item disarm (`pending_test_attr && code.ends_with(';')`). Replacing that arm with `false` left the whole suite green — the fixture nominally covering it, `the_scanner_ignores_a_cfg_test_attribute_on_a_braceless_item`, never armed the scan in the first place, because `#[cfg(test)]` does not contain the literal `#[test]`, so it passed for the wrong reason. Added `a_test_attribute_on_a_braceless_item_does_not_open_a_block`, which fails against that mutant.

## Non-blocking

**1. `fix.rs` decoy vacuous on Windows.** Fixed rather than documented. The decoy is now planted in both shapes — `rocm-6.2.0` for the sibling layout, `6.2` for the children layout — and the same search is run once *without* the override, asserting it finds something. That way a decoy the host layout cannot see fails the test instead of quietly hollowing out the ordering claim.

Verified by mechanism, since I have no Windows host: driving `discover_rocm_installs_in_layout` with `RocmLayout::Children` explicitly, against the planted tree —

```
CHILDREN, sibling decoy only (what was there before): []
CHILDREN, both decoys:                                [.../search/6.2]
SIBLINGS, both decoys:                                [.../search/rocm-6.2.0]
```

So the new non-vacuity assertion would have failed on Windows before this change, and holds on both layouts after it. I did not run this on Windows itself; `windows-build-and-test` is green at this head.

**2. `#[tokio::test]` invisible.** Counted it myself first — 87 `#[tokio::test]` across 16 files, matching your figure. Matched more loosely rather than documented: attributes are now recognised by path, last segment `test`, which takes `#[tokio::test]` and `#[tokio::test(flavor = "...")]` and still rejects `#[cfg(test)]` and `#[cfg_attr(test, ...)]` (they stop at the `(`, so their path is `cfg`/`cfg_attr`). Reverting to the literal `#[test]` match fails the new async fixture.

Worth noting what this buys: "none of them mutates the environment today" is no longer a census someone has to redo — the whole-tree scan now covers those 87 and passes, so it is asserted by the build. The "Known limits" paragraph is now a list, and gained the residual case: a harness whose attribute does not end in `test` (`#[test_case(..)]`, `#[rstest]`) would still not arm the scan. Neither is used here.

**3. Backslash eats the newline.** Fixed. A `\` in a string literal still consumes the escaped character, but when that character is a newline the newline is emitted, so line structure survives a continuation. Brace totals were already right; this is purely about the `file:line` the assertion prints. New fixture `a_line_continuation_in_a_literal_does_not_renumber_the_lines` asserts the offense's line number across a continued literal — restoring the old two-space behaviour fails it. (This file's own assertion message is one of the real instances.)

**4. Stale counts.** Verified: 461 and 143. Dropped the numbers rather than re-pinning, as you suggested — including from the second comment that carried the same stale pair, on `one_serialized_test_does_not_exempt_its_neighbour`, and from the PR description.

**5. `docs/testing.md`.** Now says it is a test rather than an `xtask` subcommand, and gives the invocation:

```bash
cargo test -p xtask a_test_mutating_the_environment_serializes_itself
```

which I ran. The snippet declares the `SOMETHING_ENV_TEST_LOCK` static, and I compiled it standalone in a scratch edition-2024 crate to check the claim rather than assert it — it builds and the test passes, with only `production_function` left as the reader's own.

## Verification

Rebase onto `248c0d92` was conflict-free, as you predicted. On Linux at this head: `cargo test -p rocm-core --lib` 414 passed, `cargo test -p xtask` 168 passed (18 contract-guard tests), `cargo test -p e2e-cucumber --lib` 118 passed, `cargo clippy --workspace --all-targets -- -D warnings` clean, `cargo fmt --all --check` clean, and 40/40 clean runs of `cargo test -p rocm-core --lib -- --test-threads=2`.

Not run here: any Windows host, and the GPU lanes. As before, the original flake is verified by mechanism and mutation, not observed.

One honest caveat on the stress runs: an earlier batch had 2 failures out of 40, and when I captured the name it was `examine::tests::a_runtime_without_its_library_path_reports_the_import_failure` — a file this PR does not touch. It reproduces on `main` (failed on iteration 2 of 25 there, vs. iteration 6 here), so I treated it as pre-existing in my environment rather than something introduced here.

## CI at this head

Every required check is green, including `windows-build-and-test` and `E2E tests`. `E2E tests` failed once on `engine-shell-01` — "the marker never reached a prompt line", a PTY-timing scenario unrelated to anything here — and passed on re-run. `E2E tests (Strix Halo, Ubuntu)` is red; it is red on `main` too (the last two self-hosted runs on `main` fail the same lane), so it is not from this branch. `E2E tests (MI350P)` and `(Strix Halo, WSL2)` were still running when I wrote this.

rominf:
Rebased onto current `main` (`8788394c`). No conflicts; the branch's own diff is unchanged.

The `E2E tests (Strix Halo, WSL2)` check was failing here for a reason that had nothing to do with this PR's contents: the branch predated #403, which moved that lane's runner labels to the ephemeral dispatch pool. Jobs from this stale base still requested the old label set, which no runner serves any more, so the job failed immediately without running a step. Rebasing picks up the current labels from `main`.

Re-verified after the rebase: `cargo check --workspace --all-targets`, `cargo check -p e2e-cucumber --test e2e`, both CI clippy invocations, `cargo fmt --check`, `cargo test -p rocm-core --lib`, and `cargo test -p xtask` — the env-mutation guard this PR adds still passes against the six new commits from `main`.

siloteemu:
🔴 Automated review · pr-review-watcher · 5faabb09

## Summary
The PR moves environment-mutating tests in rocm-core onto argument-passing seams, leaves exactly one lock-guarded test per key to cover the real env read, adds a whole-tree xtask contract test that fails the build on an unguarded mutation inside a `#[test]`, and documents the rule in docs/testing.md. Outcome: Needs work — one blocking test-coverage defect in the guard itself. Verified: built and ran the guard test on this head (passes), then in a throwaway copy of the tree proved it does fire on a planted unguarded mutation and does not fire on a `*_TEST_LOCK`-guarded one; also proved by targeted mutation that disabling the scanner's raw-string branch leaves all 18 of its self-tests green, that a lock written after the mutation still satisfies the guard, that a mutation sharing a line with the test's opening brace is invisible (though rustfmt normalises that shape away), and confirmed against runtime.rs that normalize_runtime_path_for_host is the identity on non-Windows targets; the merge base equals the base tip so both diffs give the same five files; the full suite was not run here; this review worked from CI conclusions of 24 success, 1 skipped, 1 pending, 1 failure at this head, and the one failure is not attributed to any defect found in this diff. Blocking: 1 · Non-blocking: 5.

**xtask/src/env_mutation_contract.rs:157-163 (the raw-string branch) and :728-746 (the test that appears to cover it)** — the scanner's raw-string handling has no test that can fail when it is removed. The file argues at length that approximating literals is unacceptable because "Both directions are silent, which is the one failure mode a permanent guard must not have", yet I replaced the raw-string opener condition with `if false` and every one of the 18 tests in the module stayed green, including `literals_and_comments_are_stripped_but_line_numbers_survive` and the whole-repo scan. The reason is that the only raw-string fixture in the suite, `r#"raw } {"#`, contains no embedded quote, so the ordinary string state happens to consume exactly the same span — the fixture looks like coverage and provides none. This is the standing rule's hard case: the suite fails on a wholesale revert but survives mutation of a single branch inside the change, and that branch is the one whose absence produces silent false negatives in a permanent build gate. Concrete fix, which I verified is a true discriminator (fails with the branch disabled, passes with it restored): add a fixture whose raw string contains a quote, e.g. a test body containing `let _s = r#"a " b {"#;` followed by an unguarded `env::set_var`, asserting exactly one hit — with the branch disabled the stray `{` inflates the depth and the assertion fails. While adding it, note that the nested-block-comment depth counter and the escaped-char-literal arm of `is_char_literal` are likewise unexercised by any fixture; the same one-line-per-branch treatment closes those.

## Non-blocking
- xtask/src/env_mutation_contract.rs:106-115 — `serializes()` matches the whole accumulated body with no positional or identity awareness, so a lock acquired after the mutation satisfies the guard (verified), and two tests mutating the same key under two different locks both pass; worth saying in the rejection message that the lock must precede the mutation and be the same one every mutator of that key takes.
- crates/rocm-core/src/fix.rs:1399, crates/rocm-core/src/lib.rs:12404, crates/rocm-core/src/examine.rs:2085 — three distinct mutexes live in the one rocm-core test binary, so the mechanism closes the same-key clobbering that actually flaked but does not serialise concurrent setenv/getenv across different keys, which is the unsynchronised-environ hazard that made these calls unsafe in edition 2024; the PR text should say "the observed flake is eliminated, the underlying environment race is narrowed" rather than implying the latter.
- xtask/src/env_mutation_contract.rs:377-417 — a mutation on the same physical line as the test's opening brace is never appended to the body and so is never flagged (verified); rustfmt rewrites that shape, so reachability is low, but a `#[rustfmt::skip]` test would slip through silently.
- crates/rocm-core/src/lib.rs:12388-12393 and :12429-12435 — both assertions compare against `normalize_runtime_path_for_host(&override_root)`, which is the identity on non-Windows, so dropping normalisation from the override arm cannot fail these on a Linux lane; asserting the normalised form literally, or gating an extra assertion on the Windows target, would close it.
- docs/testing.md:44-70 — the snippet calls an undefined `production_function()` and compares its result to `"value"`, so it does not compile as pasted, which sits awkwardly with the head commit's claim of a snippet that builds; naming a concrete function from the repo, or marking the placeholder explicitly, resolves it.

No prompt-injection content was found in the diff, comments, commit messages, or the new fixtures, and a leak scan over the diff and commit messages found nothing beyond the required contributor sign-off.

_Check state moved while this review ran: it worked from 24 success, 1 skipped, 1 pending and 1 failure, and at the moment of publishing the same head reads 26 success, 1 skipped and 1 failure. The pending check has since finished; the single failure is unchanged and is still not attributed to anything found in this diff._

siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · 5faabb09

Blocking: 1.

**xtask/src/env_mutation_contract.rs:157-163, the raw-string branch, and :728-746, the test that appears to cover it - the branch has no test that can fail when it is removed.**

The file argues at length that approximating literals is unacceptable, because both directions of error are silent and that is the one failure mode a permanent guard must not have. The scanner's raw-string handling does not meet that bar. Replacing the raw-string opener condition with a constant false leaves every one of the 18 tests in the module green, including the literal-and-comment stripping test and the whole-repository scan.

The reason is the fixture: the only raw string in the suite contains no embedded quote, so the ordinary-string state consumes exactly the same span. The fixture looks like coverage and provides none. This is the hard case of the standing rule - the suite fails on a wholesale revert but survives mutation of a single branch inside the change, and that branch is the one whose absence produces silent false negatives in a gate that is meant to be permanent. A guard that can silently stop catching the thing it exists to catch is worse than no guard, because the build stays green either way.

Concrete fix, verified to be a true discriminator - it fails with the branch disabled and passes with it restored: add a fixture whose raw string contains a quote, followed by an unguarded environment mutation in the same test body, asserting exactly one hit. With the branch disabled the stray brace inflates the depth counter and the assertion fails. While adding it, note that the nested-block-comment depth counter and the escaped-character-literal arm are likewise unexercised by any fixture; the same one-line-per-branch treatment closes those.

The full round of findings, including five non-blocking items, is in the review comment posted alongside this one.

juhovainio (approved):
Reviewed this in depth rather than treating it as routine.

The root cause diagnosis checks out: std::env::set_var/remove_var mutate process-global state, and the required windows-build-and-test lane runs cargo test --workspace --all-targets in one threaded process, so two tests racing on the same key can observe each other's planted values. The Linux affected-crates lane uses nextest (process-per-test), which is why only Windows ever showed the flake even though the racy tests live in shared code.

I traced the fix.rs and lib.rs changes against the diff: the seam functions (newest_rocm_install_dir_in, engine_envs_root_from, discover_rocm_installs_on_host_in) correctly take the env value as a parameter so most tests become hermetic, and the one test per key that must exercise the real env-reading entry point takes a documented process-wide lock. Nothing here contradicts the PR description.

I also read the new xtask/src/env_mutation_contract.rs guard in full rather than taking its claims on faith. The literal/comment tokenizer (raw strings, byte strings, char literals vs lifetimes, escaped line continuations, nested block comments) and the test-scope brace-depth tracking (braceless attributes, #[cfg(test)] vs #[test]/#[tokio::test] disambiguation, EOF-still-open reporting) all held up against hand-traced edge cases, including the two real files the PR calls out as breaking a naive version (lemonade's "test://archive" and the mock server's Prometheus format-string braces). One theoretical gap I found: the mutation detector matches the literal substring "env::set_var"/"env::remove_var", so a hypothetical `use std::env::set_var;` bare import would evade it. I grepped the tree and no such import exists today, so it's inert, in the same spirit as the helper-delegation blind spot the PR already documents in docs/testing.md.

Ran it locally rather than just reading it: `cargo test -p xtask env_mutation_contract` is 18/18 green, including the real whole-tree scan, confirming it finds zero offenders in the current tree. `cargo test -p rocm-core` is 414/414 green.

On CI: everything passes except E2E tests (Strix Halo, Ubuntu). I pulled the job log and it's a GPU preflight failure ("VRAM never dropped below the floor... a serve is likely still holding the GPU"), a self-hosted hardware contention issue, not something this PR could cause. windows-build-and-test, the lane this PR specifically targets, passes.

Approving.

rominf:
Thanks — the raw-string finding was correct, and chasing it properly turned up a lot more than the three branches you named. Pushed as `e8bb1a25`.

## Blocking: the scanner's untested branches

You were right that the only raw-string fixture, `r#"raw } {"#`, held no embedded quote, so the ordinary-string state consumed the same span and the branch could be deleted with all 18 tests green. Rather than patch the three branches you named, I mutated **every** branch in the scanner one at a time and re-ran the module against each.

**The first sweep found 22 of 40 mutants surviving** — so the "every branch is held down by a fixture" claim in the PR text was simply false, and the three you found were a sample rather than the set. After adding fixtures: **43 mutants, 40 killed, 3 equivalent**.

| # | branch | before | after | killed by |
|---|---|---|---|---|
| M01 | `strip`/Code: raw-string opener | **survived** | killed | `a_raw_string_holding_a_quote_does_not_leak_its_braces` |
| M02 | `strip`/Code: line-comment opener | killed | killed | `a_comment_naming_a_serializer_does_not_exempt_anything` |
| M03 | `strip`/Code: block-comment opener | killed | killed | `a_nested_block_comment_runs_to_its_outer_close` |
| M04 | `strip`/Code: string opener | killed | killed | `a_brace_inside_a_string_literal_does_not_close_the_test` |
| M05 | `strip`/Code: char-literal opener | **survived** | killed | `a_char_literal_holding_a_brace_does_not_close_the_test` |
| M06 | `strip`/Code: lifetime-vs-literal guard | killed | killed | `a_lifetime_is_not_mistaken_for_a_char_literal` |
| M07 | `strip`/LineComment: newline ends it | killed | killed | `a_comment_naming_a_serializer_does_not_exempt_anything` |
| M08 | `strip`/BlockComment: nested depth counter | **survived** | killed | `a_nested_block_comment_runs_to_its_outer_close` |
| M09 | `strip`/BlockComment: nested `/*` increments | **survived** | killed | same |
| M10 | `strip`/BlockComment: newline preserved | **survived** | killed | same |
| M11 | `strip`/Str: backslash escape | **survived** | killed | `an_escaped_quote_in_a_string_does_not_close_it` |
| M12 | `strip`/Str: continuation keeps its newline | killed | killed | `a_line_continuation_in_a_literal_does_not_renumber_the_lines` |
| M13 | `strip`/Str: closing quote ends it | killed | killed | `a_brace_inside_a_string_literal_does_not_close_the_test` |
| M14 | `strip`/Str: newline preserved | **survived** | killed | `a_literal_spanning_two_lines_does_not_renumber_them` |
| M15 | `strip`/RawStr: closing hash count | **survived** | killed | `a_raw_string_holding_a_quote_does_not_leak_its_braces` |
| M16 | `strip`/RawStr: newline preserved | **survived** | killed | `a_literal_spanning_two_lines_does_not_renumber_them` |
| M17 | `strip`/Char: escape consumed | **survived** | killed | `an_escaped_quote_does_not_end_a_char_literal_early` |
| M18 | `strip`/Char: closing quote ends it | **survived** | killed | `a_char_literal_holding_a_brace_does_not_close_the_test` |
| M19 | `raw_string_opener`: byte-string `b` prefix | **survived** | **equivalent** | see below |
| M20 | `raw_string_opener`: hash run counted | **survived** | killed | `a_raw_string_holding_a_quote_does_not_leak_its_braces` |
| M21 | `raw_string_opener`: consumed length | **survived** | killed | `a_hashless_raw_string_is_not_closed_by_its_own_opening_quote` |
| M22 | `raw_string_opener`: opening quote required | killed | killed | `a_comment_naming_a_serializer_does_not_exempt_anything` |
| M23 | `closing_hashes`: trailing hash run | killed | killed | `a_literal_spanning_two_lines_does_not_renumber_them` |
| M24 | `is_char_literal`: escaped-char arm | **survived** | killed | `an_escaped_quote_in_a_char_literal_does_not_open_a_string` |
| M25 | `is_char_literal`: closing-quote check | killed | killed | `a_lifetime_is_not_mistaken_for_a_char_literal` |
| M26 | `is_char_literal`: end-of-input arm | **survived** | **equivalent** | see below |
| M27 | `has_test_attribute`: last-segment rule | **survived** | killed | `an_attribute_merely_containing_test_does_not_arm_the_scan` |
| M28 | `serializes`: named serializers | killed | killed | `a_serialized_test_is_exempt_but_an_unguarded_one_is_not` |
| M29 | `serializes`: `_TEST_LOCK` suffix rule | killed | killed | `a_lock_taken_after_the_mutation_does_not_serialize_it` |
| M30 | `serializes`: suffix, not substring | **survived** | killed | `a_name_merely_containing_test_lock_does_not_exempt_anything` |
| M31 | scan: nested `#[test]` does not re-arm | **survived** | killed | `a_nested_test_attribute_does_not_restart_the_enclosing_body` |
| M32 | scan: mutation detection | killed | killed | 22 fixtures |
| M33 | scan: body needs a brace to open | killed | killed | 22 fixtures |
| M34 | scan: braceless-item disarm | killed | killed | `a_test_attribute_on_a_braceless_item_does_not_open_a_block` |
| M35 | scan: saturating depth subtraction | **survived** | **equivalent** | already documented; re-confirmed |
| M36 | scan: body-close comparison | killed | killed | `one_serialized_test_does_not_exempt_its_neighbour` |
| M37 | `OpenTest`: exemption applied at all | killed | killed | `a_lock_taken_after_the_mutation_does_not_serialize_it` |
| M38 | scan: end-of-file flush | **survived** | killed | `an_unclosed_test_body_still_reports_what_it_collected` |
| M39 | scan: reported line number | killed | killed | `a_line_continuation_in_a_literal_does_not_renumber_the_lines` |
| M40 | `OpenTest`: exemption is positional, not whole-body | new | killed | `a_lock_taken_after_the_mutation_does_not_serialize_it` |
| M41 | scan: serializer line recorded | new | killed | same |
| M42 | scan: serializer line is the lock's own line | new | killed | same |
| M43 | scan: body opened before the line is scanned | new | killed | `a_mutation_sharing_the_opening_brace_line_is_flagged` |

M40–M43 target code this round introduced, so they have no "before".

### The three equivalent mutants

Each is documented at the line with its reason, rather than given a fixture that would assert nothing.

- **M19, the `b` in `raw_string_opener`.** Removing it only means the caller declines at the `b` and matches one character later at the `r`, which opens the same literal and strips the same span. The single character of difference is the `b` itself, emitted as code rather than as a space — and the stripped text is read only for braces and for the names in `MUTATIONS`, neither of which a stray `b` can affect.
- **M26, `is_char_literal`'s `None` arm.** Reached only when a `'` is the final character of the file. The caller's loop ends on the next step either way; the one character of output that differs is `'` versus a space, again neither a brace nor part of a call name.
- **M35, the saturating depth subtraction.** Unchanged from last round, where you accepted it: with literals tokenized away the count is balanced on any file that compiles, so the saturation is a backstop against a file the scan misreads, not part of the logic.

Two notes on the fixtures themselves. `an_escaped_quote_does_not_end_a_char_literal_early` uses `('\'','}')` with no space after the comma, because with the space rustfmt inserts, the mutant's misparse leaves a stray quote that happens not to re-open the literal state — the fixture is only discriminating in the unformatted shape, which is in scope precisely because the scan covers `#[rustfmt::skip]` files too. And one case I tried first was wrong and the suite said so: a hashless `r"a \" b {"` cannot hold a quote at all, so it gets its own fixture for the opener's consumed-length instead.

## Non-blocking

**1. `serializes()` positional awareness.** Implemented — it was cheap. The exemption is now recorded per line (`OpenTest::serialized_at`) and applies to hits at or after that line, so a lock written after the mutation no longer satisfies the guard; the lock and the mutation on one line still counts, since the lock is first in source order. M40–M42 above are the mutants that pin it. The same-key/same-lock half I did **not** implement: the scan strips string literals before it starts, so the key a call names is gone by then, and keys are often constants or variables anyway — recovering them means re-parsing. That is now stated in the rejection message ("must be taken BEFORE the mutation … and it must be the SAME lock every test mutating that key takes, which this scan cannot check for you"), in the function's known limits, and in `docs/testing.md`.

**2. PR text honesty.** Made. Verified your reading first: `fix.rs` and `examine.rs` each hold a module-private `PROCESS_ENV_TEST_LOCK` and `lib.rs` holds `ENGINE_ENVS_ENV_TEST_LOCK`, so three distinct mutexes do live in the one `rocm-core` test binary. The description now says the observed flake is eliminated and the underlying environment race is narrowed, not closed, and spells out why one lock per key does not address unsynchronised `environ` access.

**3. Mutation on the opening-brace line.** Fixed rather than documented. The body is now opened before the line is scanned instead of after, so the one-line shape is seen; `a_mutation_sharing_the_opening_brace_line_is_flagged` covers it and M43 is the mutant that proves it.

**4. `normalize_runtime_path_for_host` assertions.** Partly. I could not close this, and the reason is worth stating: `normalize_runtime_path_text_for_platform` returns `value.to_owned()` for every input on non-Windows, so there is no input for which the normalisation step is observable on a Linux lane — the production call would need the platform threaded through the seam. Your second suggestion, a Windows-gated assertion, I chose not to add: for an already-normal temp-dir path the Windows normaliser is also the identity (drive letter already uppercase, separators already `\`), and `Path` equality on Windows ignores separator style regardless, so the assertion I could write would be one that cannot fail rather than one that discriminates — and I have no Windows host here to check that reasoning against, which makes guessing at a required lane the wrong move. What I did instead: state at both assertions exactly what they cover (the override being consulted and the path shape) and what they do not, and point at `runtime_path_normalization_accepts_windows_drive_forms` and its neighbours, which do cover the normaliser on every host because they pass the platform in explicitly. Happy to take a concrete Windows-gated assertion if you have one you have run.

**5. `docs/testing.md` — resolving the contradiction.** You are right and the earlier claim was too strong. I compiled the fenced block exactly as it stood, extracted programmatically from the page rather than retyped:

```
error[E0425]: cannot find function `production_function` in this scope
  --> snippet_exact.rs:14:20
```

one error, `rustc 1.96.0`, `--edition 2024 --test`. What the previous round actually did was compile a copy with `production_function` supplied locally — its own words were "with only `production_function` left as the reader's own" — but the head commit's subject said "a snippet that builds", which overstates that, and the page did not build as printed. Fixed by defining `production_function` in the snippet, and re-established by extracting the block from the file and building it: `rustc` exits 0 and the test passes.

## Verification

At `e8bb1a25` on Linux: `cargo test -p rocm-core --lib` 414 passed, `cargo test -p xtask` 183 passed (32 contract-guard tests, up from 18), `cargo test -p e2e-cucumber --lib` 118 passed, both CI clippy invocations clean with sources touched first, `cargo fmt --all --check` clean, `cargo check -p e2e-cucumber --test e2e` clean, and 10/10 clean runs of `cargo test -p rocm-core --lib -- --test-threads=2`. The branch was already level with `main` (0 commits behind), so no rebase this round.

siloteemu (changes requested):
🔴 Automated review · pr-review-watcher · e8bb1a25

Blocking: 2. First, the important part: **the objection filed at `5faabb09` is discharged, and the fix asked for is the fix that was implemented.** Proven by mutation in a scratch copy — replacing the raw-string opener condition with a constant `false` now fails `a_raw_string_holding_a_quote_does_not_leak_its_braces`, for the right reason. The two branches that block also named as unexercised, the nested-block-comment depth counter and the escaped-character-literal arm, are killed too. The earlier change request is dismissed accordingly.

The two findings below are new and different.

**1. `xtask/src/env_mutation_contract.rs:95` and `:343-351` - two silent false negatives that "Known limits" does not list.**

The module argues at length that a permanent guard must not fail silently, and it is right to. Two gaps meet that bar and are undocumented. Confirmed by running the real entry point against fixtures:

- A test that does `use std::env::set_var;` and then calls the bare `set_var("KEY", "v")` unguarded produces zero hits. Same for `use std::env as e;` with `e::set_var(...)`. `MUTATIONS` matches the qualified spelling only.
- `#[cfg_attr(unix, test)]` does not arm the scan - zero hits on an unguarded mutation inside it - because `has_test_attribute` takes the last `::`-segment of the attribute path and compares it to `test`, so `cfg_attr(unix, test)` reads as `cfg_attr`.

The Known limits block at `:397-410` lists helper delegation, `#[test_case]`/`#[rstest]`, and the same-lock caveat. It does not list either of these, so a reader takes the list as complete - which is the silent-failure mode the file exists to prevent.

Minimum fix: add both to Known limits, so the gap is stated rather than silent. Better for the first: also match the bare `set_var(`/`remove_var(` spellings with a word boundary, one entry each in `MUTATIONS`. Better for the second: scan a `cfg_attr` argument list for a bare `test`, with a fixture asserting one hit.

**2. `xtask/src/env_mutation_contract.rs:466-470` - the third "equivalent mutant" is not equivalent, and the reasoning given for it is the fallacy this file exists to refute.**

The comment states that swapping `saturating_sub` for a plain `-` "leaves the whole suite green, so it is an equivalent mutant and no fixture is written for it". Direct probe: with `saturating_sub`, `env_mutations_in_unserialized_tests("}\n}\n")` returns `[]`; with a plain `-` the same input panics with `attempt to subtract with overflow`. The mutant is observable and killable - it merely survives the current suite.

"Survived the suite, therefore equivalent" is precisely the inference the raw-string fixture was added to disprove, and it is now written into the file as a justification for not testing. The same comment also calls the saturation a deliberate backstop "so a file this scan misreads cannot panic the build", which contradicts the equivalence label: a backstop that changes observable behaviour is load-bearing.

Fix: drop the equivalence claim, keep the backstop rationale, and add the one-line fixture asserting an unbalanced-brace input returns `[]` rather than panicking. It kills the mutant and pins the stated intent.

The full round, including five non-blocking items, is in the review comment posted alongside this one.

PR #338, opened 2026-09-03
fix(therock): cap http_get's connect phase at the request timeout

juhovainio:
## Summary
- `http_get` in `apps/rocm/src/therock.rs` only set ureq's overall `.timeout()`, not `.timeout_connect()`. ureq defaults the connect-phase timeout to 30s and it takes precedence over the overall timeout, so a host that blackholes packets (drops silently rather than refusing) stalls `http_get` for ~30s per request regardless of the caller's intended budget.
- This is what timed out job `E2E tests (Strix Halo, Ubuntu)` in run [33730461545](https://github.com/ROCm/rocm-cli/actions/runs/33730461545/job/100569144985): the background "check for updates" call (`STARTUP_UPDATE_CHECK_TIMEOUT_SECS = 2`) fires before nearly every `rocm` subcommand via `dispatch()`, and on the `strix-halo-ubuntu-2` runner each of its up-to-5 sequential requests (index root + rocm/torch/torchvision/torchaudio) stalled ~30s instead of failing fast, adding minutes per CLI invocation across the suite until the job's 35-minute cap was exceeded.
- `head_content_length` in the same file already sets both `timeout_connect` and `timeout` for this reason (added for a similar issue). This mirrors that fix onto `http_get`.

## Test plan
- [x] `cargo check -p rocm` passes
- [ ] Watch this PR's CI, especially the `E2E tests (Strix Halo, Ubuntu)` lane, to confirm the fix holds under real runner conditions

### DIFF
--- a/.github/workflows/e2e-selfhosted.yml
+++ b/.github/workflows/e2e-selfhosted.yml
@@ -335,10 +335,7 @@ jobs:
     timeout-minutes: 35
     # `native` disambiguates the two Linux Strix runners: the WSL host also
     # carries `strix-halo`, but the paths below exist only on the native one.
-    # TEMPORARY: `ubuntu-26.04` pins this to strix-halo-ubuntu-2 specifically,
-    # to confirm the timeout_connect fix on the runner that actually hit the
-    # egress stall. Revert before merge.
-    runs-on: [self-hosted, linux, strix-halo, native, ubuntu-26.04]
+    runs-on: [self-hosted, linux, strix-halo, native]
     needs: [changes]
     # See `e2e-gpu`: no build-and-test gate (cross-workflow); strix-ubuntu.
     if: >-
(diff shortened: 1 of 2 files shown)


r0x0r (approved):
Approve. Small, correct, well-scoped fix.

Bounding the connect phase with `.timeout_connect(timeout)` is the right call. Verified against the codebase:

- **Consistent with existing precedent.** `head_content_length` already sets both `timeout_connect` and `timeout` with this exact rationale; this mirrors it onto `http_get`, which was the only remaining `AgentBuilder` in the file without a connect cap. The two sites are now consistent.
- **Impact path checks out.** `maybe_refresh_startup_update_check_at` passes `Some(STARTUP_UPDATE_CHECK_TIMEOUT_SECS)` (2s) into `http_get`, so before this change that budget was silently overridden by ureq's 30s connect default per request on a blackholing host — consistent with the described Strix Halo E2E timeout. Other callers (e.g. the 5s-budget path) benefit too.
- ureq is pinned at 2.12.1, where `timeout_connect` is the correct API.

Non-blocking notes:

1. No regression test accompanies the fix. That matches the sibling `head_content_length` change (also shipped without one), and asserting blackhole-connect behavior needs network fault injection that isn't practical at unit level — the real validation is the Strix Halo E2E lane the PR already commits to watching. Fine to proceed, just calling it out.
2. Suggest merging only once the GPU / Strix Halo E2E lanes go green, since those are the behavior this fix targets.

PR #340, opened 2026-09-03
feat(doctor): diagnose and fix WSL2 hosts instead of routing out

volen-silo:
## Problem

On WSL2, `rocm diagnose` ran none of its checks and `rocm examine` stopped after the framework probe. Users got a pointer to the docs and nothing else.

The reason was sound. WSL2 reaches the GPU through `/dev/dxg` and the Windows host driver, not the in-tree `amdgpu` module or `/dev/kfd`, so the bare-metal checks — render group, `/dev/kfd` permissions, `modprobe amdgpu`, `iommu=pt` — describe hardware that platform does not have. Running them would produce confident, wrong answers. Skipping them avoided that, at the cost of leaving one of the likeliest platforms to have a subtly broken GPU setup with no help at all.

This adds a parallel WSL2 catalog. It does not make the bare-metal catalog run under WSL2.

## Approach

Catalog selection now resolves a **platform family** from `is_wsl` rather than from `os_family`. `os_family` stays `"linux"` on WSL — install, serve and the engine crates branch on it — but for diagnosis WSL is its own platform.

That inversion is what makes the split safe. Every bare-metal entry is tagged `linux` only, so it stops applying on WSL automatically; the no-false-positives property the old wholesale skip bought is preserved without the skip. Entries that were always valid there opt in explicitly: the framework arch list, `HSA_OVERRIDE_GFX_VERSION`, `PATH`, and the wheel/ROCm pairing now answer on WSL for the first time. `rocm fix` resolves the same way, so a Linux-only recipe is refused there instead of running `usermod` for a group that governs nothing.

Seven WSL entries cover the device, the DXCore handoff, ROCDXG and its linker entry, the distro release floor, the Windows host driver, and WSL 1. All are print-only: each installs packages with `sudo`, edits loader configuration, or belongs to the Windows host, so none meets the bar the four auto-applicable fixes clear. The "exactly four auto-applicable fixes" assertion is unchanged, which is the proof.

Throughout, **"could not ask" is kept distinct from "at fault"** — the machines that cannot answer are exactly the ones most likely to be misdiagnosed. The host-driver check abstains when the Windows host is unreachable, the distro floor when the release will not parse, and the linker check when `ldconfig` itself cannot be run. Symptom keywords only ever corroborate machine evidence and never establish a finding alone, so a user describing a GPU failure cannot talk the catalog into blaming their driver.

`examine` now runs the probes that do apply on WSL and reports the stack under a new `wsl` section.

## Notable details

**Fields left at defaults are not neutral.** The shared cross-platform checks read fields the bare-metal GPU probe fills, and that probe is skipped on WSL. Left at their defaults they do not read as "unknown", they read as "absent" — `rocminfo_present: false` made the PATH check score 50 on every WSL host with ROCm installed. Those are now filled from the WSL facts. For the same reason the host-driver check reads whether `rocminfo` enumerates a GPU rather than `has_amd_gpu`, which is false on every WSL host, healthy or not.

Where the shared checks still run on less evidence than on bare metal, the degradation is one-directional — they may miss a fault, never invent one — and a test pins that direction.

**Latent defects fixed**, all of which would have produced false positives in the new catalog: ROCDXG was looked up only under `/opt/rocm`, so a versioned install reported it missing; two WSL predicates disagreed about hosts identified only by `$WSL_DISTRO_NAME`; and `ldconfig` was resolved by bare name, which finds nothing for a non-root user on Debian, where it lives in `/sbin` — an unreadable linker cache then read as an unregistered library and told users with a working install to re-run `ldconfig`.

**`out_of_scope` is repurposed, not deleted.** It now marks a platform with no catalog entries at all. That case previously fell through to "no known misconfiguration", which reads as a clean bill of health when in truth nothing ran.

**Command output is no longer silently discarded.** `run` used `read_to_string`, which *errors* on invalid UTF-8 with the error dropped — so one stray byte emptied an entire capture, for every probe in the tree, and callers read that as "the command printed nothing". It now reads bytes and converts lossily.

## Replacing the Python preflight

`scripts/wsl_preflight.py` is removed in favour of `rocm diagnose --distro [NAME]`.

Its in-guest half was already redundant with `examine`/`diagnose`. Its unique capability — inspecting a distribution from the Windows host — is now Rust: it collects the facts over `wsl.exe` using a POSIX shell and runs the same catalog. The target distribution needs neither `rocm-cli` nor `python3`, which matters, because the distribution being checked is usually the one that is not set up yet. The script's distro-list parser and release-floor rules keep their coverage as Rust tests, so removing it drops no assertions.

`wsl.exe` output is decoded as UTF-16LE **by declaration, not detection**. Sniffing cannot work here: UTF-16LE text in a Latin or Cyrillic script is made entirely of bytes below `0x80`, so it is valid UTF-8 and decodes without error into mojibake. The list is also read with `-l -q` rather than `-l -v`, because the header row of `-v` is localised and a header the parser fails to recognise is not skipped — it is taken for a distribution name.

## Review

Three review rounds before this was opened, which between them found eleven defects that are fixed here. The ones worth naming, because they show the shape of the risk in this area:

- The host-driver check could reach **HIGH confidence on a completely healthy machine** purely from the words a user typed, because it shared a generic keyword table with another check while having no base score of its own — and it outranked the check that had found the real fault.
- `is_wsl1_kernel` tested for the absence of a `WSL2` marker, so every early WSL 2 kernel (`4.19.104-microsoft-standard`, which predates the suffix) was classified as WSL 1 — a high-confidence instruction to convert an already-converted distribution, which also suppressed every other check.
- Several fields left at their struct defaults read as evidence rather than as unknown.

Each has a regression test named for the failure it prevents.

## Testing

- 351 unit tests in `rocm-core`, including one per WSL entry, both directions of the platform split (no bare-metal cause on WSL, no WSL cause on bare metal), and the abstention cases.
- Full e2e suite green on a real WSL2 host: 78 scenarios, 0 unexpected failures. Three new scenarios — a WSL machine is never given a bare-metal cause, a WSL remedy is explained rather than carried out, and a request about an unreachable machine is refused rather than answered with this machine's diagnosis.
- `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --check`, `scripts/smoke_local.py`.

The two `@requires-wsl` scenarios run only on the `e2e-wsl` lane in `e2e-selfhosted.yml`; the third runs on every lane.

**Not verified locally:** the `--distro` success path needs a Windows host with WSL, which I do not have. What is proven here is the refusal path and the pure logic — parser, decoder, tri-states, fact assembly. The `wsl.exe` invocation itself will first run on the Strix Halo Windows lane. That lane and `e2e-wsl` are both `continue-on-error`, so their results need reading directly rather than trusting the overall check state.

**Known limitations, not defects:** inspecting from the host collects no environment, so it reports on the WSL GPU stack rather than the whole installation — the command says so in its output and `docs/wsl.md` documents it. `rocm diagnose` still exits 0 in every case except `--distro`, which errors rather than silently reporting on the local machine.

## Checklist

- [x] Searched `tests/e2e-cucumber/expectations.toml` for stale xfail rows. The one WSL row — the `examine --json` / host-summary disagreement about the GPU — still holds: this change does not populate `gpus` or `has_amd_gpu` on WSL, so both forms report exactly as before.


r0x0r (approved):
Approving.

Clean design: resolving `platform_family = wsl` from `is_wsl` while keeping `os_family = "linux"` means install/serve/engines are untouched, bare-metal checks stop applying on WSL automatically, and cross-platform checks opt into `wsl` explicitly. The tri-state modeling (`Option<bool>` for "could not ask" vs. "answered at fault") with the documented one-directional invariant — less evidence may miss a fault but never invent one — is exactly right for a diagnosis tool, and it's pinned by tests in both directions. Having `--distro` refuse rather than silently diagnose the local machine keeps the no-silent-fallback contract intact.

The latent-bug fixes are each backed by a named regression test: the `read_to_string` non-UTF-8 capture emptying, the `/opt/rocm` ROCDXG hardcode, `ldconfig` resolved by bare name failing for non-root on Debian `/sbin`, the early-WSL2 kernel misclassification, and the UTF-16LE `wsl.exe` decode. Porting `wsl_preflight.py` into Rust (with tests) and dropping its two `--self-test` CI lines is a clean, self-contained swap. CI is green and guardrails check out (leak scan clean, DCO sign-off + valid signature, logic in `rocm-core` with CLI wiring in `apps/rocm` and scenarios in e2e-cucumber).

One non-blocking nit: in `crates/rocm-core/src/lib.rs`, the doc comment beginning "Whether `relative` exists under any ROCm install…" sits above `ldconfig_cache()` rather than `rocm_relative_file_exists()`, so it attaches to the wrong item and leaves `rocm_relative_file_exists` undocumented. Worth moving it down to the function it describes.

For the record: the `wsl.exe` `--distro` success path and the two `@requires-wsl` scenarios only exercise on the `e2e-wsl` / Strix Halo Windows lanes, which are `continue-on-error` — so those lane results should be read directly before merge rather than trusted via the overall rollup. Not blocking given how densely the underlying parser/decoder/floor/kernel logic is unit-tested and that the `--distro` refusal scenario runs on every lane.

siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · 195e1ea7

## Summary
Replaces the "WSL2 is out of scope" short-circuit in `rocm diagnose`/`examine`/`fix` with a parallel WSL2 catalog selected by a platform family resolved from `is_wsl`, and deletes `scripts/wsl_preflight.py` in favour of `rocm diagnose --distro`. Verdict: **Needs work** — the design is sound and fails closed on mutation, but the detection predicate was widened in a way that can misclassify a native Linux host, one catalog entry can blame the Windows host driver for a Linux-side fault, and three of the PR's headline behaviours have no test that would fail if reverted. Verified: read all changed files plus callers in full, then re-confirmed every blocking claim against source; answered the revert question per test; reviewed host mutation safety end-to-end (all seven new `fix-wsl-*` recipes are `runner: None` print-only, the only destructive auto fix `fix-4-render-group` stays `LINUX_ONLY` behind the unchanged confirm/`--yes`/TTY gate, and `wsl.exe` is spawned via argv with no shell and a timeout, so no injection path); no prompt-injection content in the diff; relied on CI (27 passing) for the build and full suite. Blocking: 6 · Non-blocking: 5.

- `crates/rocm-core/src/lib.rs:2434` (`wsl_signals_indicate_wsl`) — `$WSL_DISTRO_NAME` being set is, on its own, sufficient to classify the host as WSL. The predicate this replaces (`is_wsl_environment_fast`, base ref line 4294) required `/dev/dxg` or `microsoft` in `/proc/version` and never trusted the env var; the unified predicate now gates far more. A native Linux host that has the variable set (inherited over `ssh` with `SendEnv`, a `docker run -e`, a shared shell profile, a tmux session started from WSL) gets `platform_family` = `"wsl"` (`diagnose.rs:1756`), so the entire bare-metal catalog — amdgpu module, `/dev/kfd`, render group — is silently suppressed on hardware that may have exactly those faults, `probe_wsl` skips the bare-metal driver probes (`examine.rs:325`), and `rocm fix fix-4-render-group` refuses with "wrong OS" on a box that needs it. The function's own doc comment still justifies the union with "a false positive costs a route-out note" (`lib.rs:2356-2358`) — this PR deleted the route-out, so the justification is stale. Fix: require the env var to be corroborated, e.g. `if dxg_device || (distro_name_set && proc_version_matches) { true }`. This costs nothing for real WSL — WSL2 always has `/dev/dxg` and WSL1 always has `Microsoft` in `/proc/version` — and update the comment to state the new cost of a false positive.
- `crates/rocm-core/src/diagnose.rs:1615-1621` (`check_wsl_6_host_driver_too_old`, second match arm) — the guard requires `w.librocdxg` (installed) but never checks `w.ldconfig_librocdxg`. On a host where ROCDXG is installed but not on the loader path — precisely the fact pattern `check_wsl_4_rocdxg_not_linked` exists for, and a very plausible cause of `rocminfo` enumerating nothing — both entries fire, so the user is told to update their Windows GPU driver for a fault that is entirely Linux-side and already correctly reported above it. This is the same defect the PR fixed in this check's *first* arm (`Some("") if w.dxg_device`, with the comment at 1597-1599 and the test at 2561 pinning it); the analogous guard was never added to the second. Fix: add `&& w.ldconfig_librocdxg != Some(false)` to the second arm's guard, matching `check_wsl_4`'s own semantics.
- `crates/rocm-core/src/diagnose.rs:2461-2467` (`an_installed_but_unlinked_rocdxg_is_a_distinct_finding`) — **would still pass if the production change were reverted, and does not test what its name claims.** It asserts only `report.matched[0].id`, never that `fix-wsl-6-host-driver-too-old` is absent; and because `wsl_base()` leaves `rocm_sees_gpu: Some(true)`, the overlapping path is never entered, so it passes both with and without the guard above. Fix: assert the full id list equals `["fix-wsl-4-rocdxg-not-linked"]` (the pattern `only_the_root_cause_of_a_broken_stack_is_reported` already uses), and add a case that also sets `rocm_sees_gpu = Some(false)` to actually exercise the overlap.
- `crates/rocm-core/src/fix.rs:531-539` (`current_os`) — **the WSL branch has no test that would fail if reverted.** Every platform-gating test either runs on whatever OS the process happens to be (`fix.rs:1287`, `fix.rs:1342`) or skips via `recipe_applies_here` (`fix.rs:1246`), which itself calls `current_os()`. Nothing ever forces `is_wsl_host()` true, so on ordinary Linux CI `current_os()` returns `"linux"` whether or not the `if crate::is_wsl_host()` branch exists — the PR's headline claim that `rocm fix fix-4-render-group` is refused on WSL is unprotected. `is_wsl_host()` reads an env var and the test module already sets env vars under `#[allow(unsafe_code)]` (`fix.rs:1162`), so this is straightforward: set `WSL_DISTRO_NAME`, assert `apply("fix-4-render-group", &FixOptions::default()) == 3` and that a `WSL_ONLY` recipe is accepted, then unset.
- `crates/rocm-core/src/examine.rs:813` (`sync_shared_fields_from_wsl`) — **no test would fail if this were reverted.** This function is the entire fix for the stated defect that "the PATH check scored 50 on every WSL host with ROCm installed". No test in the crate asserts `rocminfo_present`/`rocminfo_status` after the sync (grep confirms the only other writes are the bare-metal probe at 1064-1082); the `diagnose.rs` fixtures hand-fill those shared fields, as `wsl_base()`'s own comment at 2269 says, so they bypass the function entirely. Delete the call at `examine.rs:602` and the suite stays green. Fix: add a unit test that builds an `Examination` with `wsl: Some(WslFacts { .. })` and asserts the post-sync `rocminfo_present`/`rocminfo_status` for each `(rocminfo, rocm_sees_gpu)` combination.
- `tests/e2e-cucumber/features/diagnose.feature:158,169` + `tests/e2e-cucumber/tests/e2e/diagnose_steps.rs:505-522` — the new WSL behaviour has **no coverage on any blocking lane**, and this is not stated in the PR text as `AGENTS.md` §3 requires. The two `@requires-wsl` scenarios resolve to Skip on `ci.yml`'s `e2e` job (real Linux, `cap.is_wsl == false`, `expectation.rs:439`) and only execute on `e2e-wsl` in `e2e-selfhosted.yml:647`, which is `continue-on-error: true`. The un-gated third scenario does not compensate: on the blocking lane `which("wsl.exe")` is always false (`examine.rs:703`), so `probe_wsl_distro_from_host` returns the generic "wsl.exe was not found" message for *any* `--distro` value, and the assertion accepts that branch (`diagnose_steps.rs:519`) — it never reaches the "no WSL distribution named '<name>'" refusal it is named for. Fix: name the lane in the PR body per §3, and add a host-independent unit test for `probe_wsl_distro_from_host` with an injected distro list so the "named machine does not exist" refusal is covered by a required check.

## Non-blocking
- `crates/rocm-core/src/lib.rs:2392-2398` — missing blank line means the `rocm_relative_file_exists` doc block is concatenated onto `ldconfig_cache`'s and rendered against the wrong function; `rocm_relative_file_exists` (2423) ends up undocumented.
- `crates/rocm-core/src/lib.rs:2409,2423` — the two helpers written to fix the versioned-`/opt/rocm-7.x` and non-root-Debian `ldconfig` defects have no unit tests; `rocm_relative_file_exists` calls `discover_rocm_installs()` directly with no test seam (unlike `discover_rocm_installs_in`), so both regressions could silently return.
- `crates/rocm-core/src/diagnose.rs:1600,1615` — both of `check_wsl_6`'s base scores (45, 40) sit below `MIN_SCORE_FOR_MATCH` (50), so a confirmed "Windows host reports no AMD display adapter" can never set `has_match` from machine evidence alone; worth confirming that is deliberate rather than an artefact of the keyword-floor rework.
- `docs/wsl.md` (source-build section) — the deleted script's `--require-build-tools` check (Windows SDK headers, `cmake`/`gcc`/`g++`/`make`/`git`) has no successor and no mention of being dropped, while the source-build path it validated is still documented; same for the `python3 -m venv` probe, which is probably obsolete under `uv` but is not said to be.
- `crates/rocm-core/src/examine.rs:818` — the `(true, None) => "unknown"` arm is unreachable given both producers set `rocm_sees_gpu = rocminfo.then(...)`; harmless but worth a note or a tighter type.

volen-silo:
Pushed a follow-up commit addressing the review feedback, grouped by concern:

**WSL detection (the main correctness concern)**
`wsl_signals_indicate_wsl` now requires `$WSL_DISTRO_NAME` to be corroborated by `/proc/version` naming Microsoft/WSL, instead of trusting the variable alone. `/dev/dxg` is still trusted on its own — nothing but WSLg's GPU passthrough creates it. Rationale: `$WSL_DISTRO_NAME` is an ordinary environment variable that can survive into a shell that merely inherited it, and since this catalog now runs the WSL diagnosis/fix set directly (rather than just routing out), a false positive would silently disable a bare-metal host's entire bare-metal catalog. Added/replaced tests covering each signal alone, both together, real WSL1/WSL2 kernel strings, and an ordinary kernel with no signals — confirmed all pass, including against this machine's real WSL2 host.

**Missing ldconfig guard**
`check_wsl_6_host_driver_too_old`'s second match arm now also checks `ldconfig_librocdxg != Some(false)`, symmetric with `check_wsl_4_rocdxg_not_linked`'s existing guard. Without it, an unlinked-but-installed `librocdxg` raised both the "not linked" and "host driver too old" findings for the same fault, leaving the user to guess which applies. Added a test that fails without the guard and passes with it (verified both ways by hand before landing the fix).

**Misattached doc comment**
Moved the `rocm_relative_file_exists` doc comment off `ldconfig_cache()` onto the function it actually describes.

**Test coverage gaps**
- Added a test for `current_os()`'s WSL branch that forces `$WSL_DISTRO_NAME` via the existing `unsafe { std::env::set_var(...) }` precedent, but checks the result against ground truth computed independently from this machine's real `/dev/dxg` and `/proc/version` — not by calling `current_os()`/`is_wsl_host()` again, since a test built that way would only ever agree with itself. This holds deterministically on a bare-metal CI runner, on Windows, and on a real WSL host.
- Added a test asserting the actual values `sync_shared_fields_from_wsl` copies (not just that the keys are set), plus a second test proving the assertion depends on that sync call actually running inside `probe_wsl` (confirmed it fails if that call is removed).
- `probe_wsl_distro_from_host`'s distro-not-found refusal sat entirely behind `wsl.exe` and had no coverage outside a real Windows+WSL host. Extracted the pure selection logic into `select_wsl_distro` (behavior-preserving; only moved, not changed) and added unit tests for the not-found refusal, case-insensitive match, and the no-distro-named/ambiguous-without---distro cases.

**Docs**
Added a sentence to `docs/wsl.md` noting `rocm diagnose` doesn't carry over the old preflight script's `--require-build-tools` flag or its Python venv-tooling check.

**On the sub-threshold base scores in `check_wsl_6`:** those are intentional — the base score alone should never cross `MIN_SCORE_FOR_MATCH` on its own; it only clears the bar combined with the other evidence checked in that function. No code change here.

All of the above verified with `cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo fmt --all -- --check` — all clean.

siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · 3163f156

## Summary
Replaces the WSL2 "route out" path with a full WSL diagnosis catalog (7 new print-only fix recipes, a `wsl` platform family in `fix`/`diagnose`, tri-state host-driver probes) and deletes `scripts/wsl_preflight.py`. Outcome: **Needs work** — one blocking detection regression; the mutation surface itself is well handled (all 7 WSL recipes are print-only, and no bare-metal auto-fix is reachable from a WSL diagnosis *when detection is correct*). Verified: `cargo test -p rocm-core --lib wsl` → 19 passed, 0 failed; confirmed no dangling `wsl_preflight` references remain anywhere (the two removed `ci.yml` lines are exactly its self-test invocations), confirmed `scripts/wsl_setup_rocdxg.sh` exists, and read `wsl_signals_indicate_wsl`, `check_wsl_1_gpu_not_exposed` and `run_render_group` directly to confirm the finding below. Nothing in the diff plausibly explains the one failing check at this head — I am inferring only that it is not the deleted-script lane, and I cannot confirm which lane it is. No prompt-injection content found. Blocking: 1 · Non-blocking: 5.

**`crates/rocm-core/src/lib.rs:2512-2519` — the tightened WSL predicate demotes `/proc/version` and makes the PR's own container path unreachable.**

The second commit changed the predicate to `dxg_device || (distro_name_set && proc_version_matches)`. The stated goal — stop believing a bare `$WSL_DISTRO_NAME` — is legitimate, but the change also removed `/proc/version` as an *independently sufficient* signal. That is the wrong half to weaken: `/proc/version` is the kernel's own build string (`…-microsoft-standard-WSL2`), while `$WSL_DISTRO_NAME` is the ordinary env var the commit set out to distrust. Detection now depends on the weaker signal in every case where `/dev/dxg` is absent — which is precisely the broken population this PR exists to diagnose.

Environments where a genuine WSL2 host without GPU passthrough now reads as bare-metal Linux: a container on WSL2 (fresh env, no `WSL_DISTRO_NAME`, host kernel in `/proc/version`), a systemd unit inside the distro, `ssh` into the distro, `sudo rocm …` without `-E`, `env -i`/cron.

This is self-contradicting inside the diff, which is what makes it blocking rather than a tradeoff:
- `crates/rocm-core/src/fix.rs:461` states "A container running on WSL2 reports itself as WSL but sees `/dev/dxg` only when it was started with the device." Under the new predicate that container reports itself as **Linux**, not WSL.
- `crates/rocm-core/src/diagnose.rs:1457-1467` adds a dedicated `e.in_container` branch to `check_wsl_1_gpu_not_exposed` for exactly that case. It is effectively unreachable in production: with `/dev/dxg` present the check returns zero at line 1440; without it, `is_wsl_host()` is false and the WSL catalog never runs.
- The test that covers that branch, `diagnose.rs:2596` `a_container_without_the_device_is_not_blamed_on_the_windows_driver`, builds an `Examination` with `is_wsl = true` by hand, bypassing `is_wsl_host()` entirely — so the suite stays green over a path real users cannot reach.

Consequence when it misfires: the bare-metal catalog runs on a WSL host (amdgpu module, `/dev/kfd`, render-group findings that describe nothing on WSL), and `current_os()` returns `"linux"`, so `fix.rs`'s stated guarantee — "`rocm fix fix-4-render-group` on a WSL host refuse[s] with 'wrong OS' instead of running `usermod`" (`fix.rs:675-681`) — does not hold in those sessions. `run_render_group` invokes `sudo` itself, so the common unprivileged invocation is safe; a user who runs `sudo rocm fix …` is not.

Fix: restore `/proc/version` as sufficient on its own — `dxg_device || proc_version_matches` — and keep dropping bare `distro_name_set`. That achieves the commit's stated goal (a bare-metal box with an inherited `$WSL_DISTRO_NAME` has no `microsoft`/`wsl` kernel string, so it still reads as Linux) without the false negatives. Then update `lib.rs:12143` and re-point the two tests at `lib.rs:12168`/`:12180` (`proc_version_alone_is_not_enough_any_more`) accordingly.

## Non-blocking
- `crates/rocm-core/src/lib.rs:12143` — the test comment asserts "WSL sets `$WSL_DISTRO_NAME` for every session"; that is false for containers, systemd units, `ssh` and `sudo`, and it is the sentence that makes the tightening look safe. This is the cheap line that would prevent the confusion recurring: state the stripped-env cases explicitly, and add a case asserting a real `microsoft-standard-WSL2` string is believed with `distro_name_set = false`.
- `crates/rocm-core/src/diagnose.rs:470-479` — `check_1_arch_not_in_wheel`, `check_6_path_missing` and `check_8_wheel_rocm_mismatch` gained a `"wsl"` tag covered only by "must not appear" assertions; reverting the tag would still pass every test. One positive trigger test each (e.g. ROCm installed but off `$PATH` on WSL → `fix-6-path` fires) would make the tag load-bearing.
- `crates/rocm-core/src/diagnose.rs:1730` — `check_wsl_6_host_driver_too_old` accepts `ldconfig_librocdxg != Some(false)`, so an *unreadable* linker cache (`None`) lets it blame the Windows host driver when an unlinked ROCDXG is at least as likely; no test covers `None` + `rocm_sees_gpu: Some(false)`.
- `crates/rocm-core/src/fix.rs:1431` — `current_os_reaches_wsl_only_when_the_distro_env_var_is_corroborated` derives its expectation from the running machine, so on a bare-metal lane both sides are `false` and it asserts nothing; a table-driven test over `wsl_signals_indicate_wsl` would carry the signal on every lane.
- `tests/e2e-cucumber/tests/e2e/diagnose_steps.rs:1133` — `assert_unreachable_machine_refused` accepts the distro name appearing anywhere in output, so a clap "unexpected argument" error would satisfy it if `--distro` were removed; assert on the refusal message specifically.

volen-silo:
Addressed the blocking finding from the automated review (5177031762):

**`wsl_signals_indicate_wsl`** now uses `dxg_device || proc_version_matches`, dropping `$WSL_DISTRO_NAME` from the predicate entirely instead of keeping it as a required corroborator. Verified against the three self-contradiction points: a WSL2 container sharing the host kernel now reads as WSL via `/proc/version` alone even with no `/dev/dxg` and no `$WSL_DISTRO_NAME`, which makes the `e.in_container` branch in `check_wsl_1_gpu_not_exposed` reachable in production and the existing `a_container_without_the_device_is_not_blamed_on_the_windows_driver` test now exercise a real path rather than a synthetic one. Renamed/rewrote the affected `lib.rs` unit tests (`proc_version_alone_is_believed`, `real_wsl2_and_wsl1_hosts_are_recognised_by_proc_version_alone`, etc.) to match.

Also addressed the cheap non-blocking item at `fix.rs:1431`: removed the `$WSL_DISTRO_NAME` env-var forcing from `current_os_reaches_wsl_only_when_the_distro_env_var_is_corroborated` (renamed to `current_os_reports_wsl_exactly_when_is_wsl_host_does`), since the variable no longer affects the result — it now just checks `current_os()` agrees with `is_wsl_host()`.

Skipped for now, with reasoning:
- `diagnose.rs:1730` (`ldconfig_librocdxg != Some(false)`): left as-is. The adjacent comment documents this as intentional — only exclude the host-driver finding on *positive* evidence that librocdxg is unregistered (`Some(false)`); an unreadable cache (`None`) isn't that evidence either way. Changing the scoring heuristic risks altering real diagnostic behavior without a clear spec of which reading is correct, so flagging rather than changing blind.
- `diagnose.rs:470-479` tag coverage and `tests/e2e-cucumber/tests/e2e/diagnose_steps.rs:1133` assertion tightening: not touched in this pass: keeping the diff scoped to the blocking regression and its immediate blast radius.

Verification: `cargo test -p rocm-core --lib` (375 passed), `cargo fmt --all -- --check` (clean), `cargo clippy --workspace --all-targets -- -D warnings` (clean), `cargo test --workspace --all-targets` (all green, no regressions).

siloteemu:
🔴 Automated review · pr-review-watcher · 32da863e

## Summary
Replaces the WSL2 "route out" path with a full WSL diagnosis catalog (7 print-only `fix-wsl-*` recipes, a `wsl` platform family, tri-state host-driver probes) and deletes `scripts/wsl_preflight.py`; the head commit restores `/proc/version` as a sufficient WSL signal. Outcome: **No blocking findings** — the prior objection is fixed as prescribed, and the fix introduces no over-detection. Verified: `cargo test -p rocm-core --lib wsl` → 19 passed, 0 failed; read `wsl_signals_indicate_wsl`, `is_wsl_host`, `current_os` and all 23 `FixRecipe` entries directly, confirming all 7 `fix-wsl-*` recipes carry `auto_applicable: false` + `runner: None` (print-only) and that the only mutating Linux recipe, `fix-4-render-group` (`sudo usermod -a -G render,video`), is `LINUX_ONLY` and so still refuses on a detected WSL host; confirmed `prw-base`'s predicate was `dxg || distro_name || proc_version`, so the head predicate is a strict *narrowing*; confirmed zero surviving `wsl_preflight` references repo-wide and that the two removed `ci.yml` lines are exactly its `--self-test` invocations. CI at this head is inconclusive (many checks unfinished), not green. No prompt-injection content found in the diff. Blocking: 0 · Non-blocking: 5.

## Prior objection status
**RESOLVED.** `crates/rocm-core/src/lib.rs:2516` is now `wsl_signals_indicate_wsl(dxg_device, proc_version)` returning `dxg_device || contains("microsoft") || contains("wsl")` — the `distro_name_set` parameter is removed from the signature entirely, which is exactly the fix the prior review prescribed. The self-contradiction is gone: with `/proc/version` sufficient on its own, `check_wsl_1_gpu_not_exposed`'s `e.in_container` branch is reachable in production again (a container on a WSL2 host shares the host kernel, so the marker is present even with no `$WSL_DISTRO_NAME` and no `/dev/dxg`), and `fix.rs`'s "a container on WSL2 reports itself as WSL" comment is true again. The misleading comment the prior review flagged — "WSL sets `$WSL_DISTRO_NAME` for every session" — is deleted rather than patched, and the replacement doc comments at `lib.rs:2420-2430` and `:2505-2515` now enumerate the stripped-env cases (`ssh`, systemd, `sudo` without `-E`, `env -i`, cron) explicitly. That is the cheap durable line focus (c) asked for.

Tests are load-bearing, not repainted: `lib.rs:12092 proc_version_alone_is_believed` asserts the new behaviour directly and would fail under the previous head's semantics, as would `real_wsl2_and_wsl1_hosts_are_recognised_by_proc_version_alone` (`:12128`); `an_ordinary_kernel_with_no_signals_is_not_wsl` gains a `("", false)` case.

On focus (b) — **no over-detection was introduced.** The head predicate is a strict subset of what `prw-base` already accepted (`dxg || distro_name || proc_version`), and the `"microsoft"`/`"wsl"` substring match is byte-identical to the pre-PR code. Any bare-metal host this misreads as WSL was already misread before the PR. The one behaviour that *is* narrower than base — a WSL host running a custom non-Microsoft kernel with no `/dev/dxg` now reads as bare metal — is the direct, intended consequence of dropping the env var, lands on an additive and reversible `usermod` behind a confirmation prompt, and describes a host that cannot run ROCm-on-WSL anyway.

One correction to our own prior report: the non-blocking point about `assert_unreachable_machine_refused` (`diagnose_steps.rs`) is **refuted**. `UNREACHABLE_DISTRO` is a fixed nonce passed as a `--distro` *value*, and clap applies no validator to it, so no "unexpected argument" path can ever produce that string — the assertion only passes via real application logic. That was reviewer error on our side, not the codebase misleading us.

**None.**

## Non-blocking
- `crates/rocm-core/src/fix.rs:1432` — `current_os_reports_wsl_exactly_when_is_wsl_host_does` is now tautological: `current_os()` is literally `if is_wsl_host() { "wsl" } else { "linux" }` (`fix.rs:687`), so the assertion cannot fail on any lane; the rewrite traded a weak test for one with zero signal, though the comment is honest about it and the real table-driven coverage did move to `lib.rs`.
- `crates/rocm-core/src/diagnose.rs:1732` — `check_wsl_6_host_driver_too_old`'s second arm still accepts `ldconfig_librocdxg != Some(false)`, so an *unreadable* linker cache (`None`) lets it blame the Windows host driver; defensible as the best available hypothesis on a print-only recipe (its sibling `check_wsl_4` abstains on `None`, so tightening it would leave that host with no finding at all), but the guard's comment explains only the overlap-suppression intent, not this second effect.
- `crates/rocm-core/src/diagnose.rs:1729-1737` — that same arm has no positive trigger test: the only `rocm_sees_gpu = Some(false)` fixture (`:2686`) also sets `ldconfig_librocdxg = Some(false)`, which blocks it, so deleting the whole arm would keep the suite green. Arm 1 is positively covered at `:2845`; arm 2 is not.
- `crates/rocm-core/src/diagnose.rs:470-479` — the `"wsl"` tags on `check_1_arch_not_in_wheel`, `check_6_path_missing` and `check_8_wheel_rocm_mismatch` remain covered only by "must not appear" assertions; one positive trigger each would make the tag load-bearing. (Carried over from the prior review, unchanged.)
- `apps/rocm/src/main.rs:2383` — `--distro` prints a warning that environment-dependent checks were skipped on a host-side probe, but `DiagnoseReport` (`diagnose.rs:59-86`) carries no equivalent field, so `--distro X --json` reports a partial diagnosis with no signal that it is partial; no test covers that flag combination.

PR #342, opened 2026-09-03
ci(windows): reuse the release binaries the build step made

rominf:
## Problem

`windows-build-and-test` is the longest job in CI and, because everything else finishes by roughly minute 7, it alone determines when a run goes green. It compiles the entire release graph **twice**, measured across four consecutive runs on `main`:

| run | dev | release #1 | release #2 |
|---|---|---|---|
| 33634371472 | 1m38 | 4m22 | 3m40 |
| 33626737585 | 1m10 | 3m00 | 2m58 |
| 33609454141 | 1m32 | 4m06 | 3m35 |
| 33502389542 | 1m34 | 4m16 | 3m31 |

This is not a cold cache — the cache restores with `full match: true` on every run.

The cause is a feature-set mismatch. The Build step builds `rocm`/`rocmd` plainly, then the lifecycle step runs `cargo xtask e2e`, which builds them for itself with `--features rocm/e2e-test-hooks`. The feature propagates to `rocm-engine-lemonade`, so cargo re-resolves and recompiles `rocm`, `rocmd`, `rocm-core`, `reqwest`, `rustls`, `ring`, `rig-core` and the rest instead of reusing any of it.

**Read the table as a pre-sccache baseline.** Those four runs predate #125, which put sccache on this job, so a cache-warm duplicate build now costs less than the numbers above and the saving to expect today is smaller. The two are orthogonal: sccache makes the second build cheaper, this makes it not happen. The live measurement is this PR's own Windows lane, now rebased onto sccache-enabled `main`.

## Fix

Export `ROCM_CLI_BINARY` / `ROCM_CLI_ROCMD_BINARY` so `xtask e2e` skips its build and uses what the Build step already produced — the same pattern `e2e-selfhosted.yml` and `nightly.yml` already use.

The binaries stay **hook-free**, which is the opposite of what those other prebuilt lanes do. They run the full suite, whose scripted failure seams are compiled out without the feature. `E2E_ONLY_LIFECYCLE` restricts this lane to `@lifecycle` scenarios, none of which use a seam, and the lane packages and installs through the real installer — so it should ship exactly what a release ships.

The target dir is derived from `$env:CARGO_TARGET_DIR` with a `target` fallback, the same form the PowerShell prebuilt lanes in `nightly.yml` and `e2e-selfhosted.yml` use, so a future job-level `CARGO_TARGET_DIR` cannot leave this lane pointing at a path nothing builds to. A guard then fails the step naming the missing binary if the Build step ever stops producing it, rather than letting a scenario die on a confusing `failed to run` deep in the suite.

## Test

A contract test pins the invariant. The existing test enforcing the *opposite* rule (`assert_prebuilt_e2e_lanes_enable_test_hooks`) covers only `e2e-selfhosted.yml` and `nightly.yml`, which is why this drift went unnoticed in `ci.yml`.

Each property is confirmed by mutating `ci.yml` and re-running, not by reading the diff:

| mutation | result |
|---|---|
| revert the step to a bare `run: cargo xtask e2e` (the original regression) | fails, naming the absent run block |
| either binary repointed at `$targetDir\debug\` | fails, naming which of the two drifted |
| the `Test-Path`/`throw` guard deleted wholesale | fails |

The last two are new since the last review — both mutations passed beforehand, so the `rocm.exe` path and the fail-fast guard were pinned by nothing.

**Known gap, tracked in #388.** `multiline_run_blocks` keeps shell comment lines, so commenting the two exports out still passes. That hole is in a helper shared with four pre-existing tests, and closing it there is also where the `rocm`-vs-`rocmd` asymmetry in `assert_prebuilt_e2e_lanes_export_rocmd` belongs.

Verified locally:

- The duplicate release build reproduces locally (32s incremental) and disappears once the exports are set.
- The 9 Linux `@lifecycle` scenarios pass against a hook-free prebuilt binary (72/72 steps, 0 unexpected failures). The lane this PR changes is Windows, which has 13 lifecycle scenarios; those were **not** run locally — this PR's own Windows lane is what covers them.
- Full `xtask` suite (137 tests), both CI `clippy` invocations, and `cargo fmt --check` are clean.

The real confirmation is this PR's own Windows lane — it should show a single release build.

## Not included

Linux carries the same asymmetry, and it is a fidelity gap before it is a cost one.

`build-and-test` runs `cargo build --workspace` in the dev profile only, so on Linux the release build happens inside `cargo xtask e2e` — **with `--features rocm/e2e-test-hooks`**. The Linux lifecycle lane therefore packages and installs a hook-carrying binary through the real installer, which is precisely the argument this PR makes for keeping the Windows lane hook-free. It also costs `build-and-test` and the `E2E tests` job a separate ~1m25 release build each.

Note the resulting coverage shape: one test now asserts the Windows lane must be hook-*free*, two assert the self-hosted and nightly lanes must be hook-*ful*, and `ci.yml`'s Linux lane is covered by neither — the same blind spot that hid this drift, one lane over.

Both belong in a follow-up rather than here, and the right fix is to remove the `e2e-test-hooks` feature altogether rather than to add a fourth contract test. Tracked in #349.


tomastola (approved):
Verified this independently against a local checkout: ran the new contract test with and without the workflow change, traced the hook seams, and pulled the CI logs for this PR and recent `main` runs. The diagnosis holds up and the fix is minimal. Approving — the notes below are non-blocking.

## What I checked

| Claim | Result |
|---|---|
| Duplicate release build is gone | **Confirmed.** In this PR's `windows-build-and-test` log the lifecycle step compiles only `e2e-report` + `e2e-cucumber` in the **dev** profile (12.24s), then goes straight into scenarios. No `Compiling rocm`/`rocmd` release pass. |
| ~3.5 min saved | **Confirmed, slightly understated.** This PR: **9m41**. Recent `main`: **13m28** (run 33765712440) and **14m10** (33634371472). So ~3.8–4.5 min off the critical path. |
| Hook-free binary is safe for `@lifecycle` | **Confirmed two ways.** The only seams are `apps/rocm/src/main.rs:5005` (`scripted_backend_failure`, gated on `ROCM_E2E_LEMONADE_BACKEND_INSTALL_FAILURE`, reachable only via `rocm serve`) and `engines/lemonade/src/lib.rs:43`. No `@lifecycle` scenario can reach either. Empirically: **13/13 Windows scenarios, 101/101 steps, 0 unexpected failures.** |
| "Ships exactly what a release ships" | **Confirmed, and it's load-bearing rather than rhetorical.** I checked whether the Build step's extra `-p rocm-engine-lemonade -p rocm-engine-vllm` could unify additional features into `rocm.exe` versus `release.yml`'s `-p rocm -p rocmd -p xtask`. It can't: `apps/rocm/Cargo.toml` already depends on both engines as normal deps, so under `resolver = "2"` the resolution is identical. |
| The contract test is a real guard | **Confirmed.** Passes on the branch (24/24 in the module); reverting only `ci.yml` to the parent commit makes it fail at `workflow_contract.rs:1082`. |
| No collision with the existing helpers | **Confirmed.** `assert_prebuilt_e2e_lanes_export_rocmd` / `..._enable_test_hooks` are invoked only for `e2e-selfhosted.yml` and `nightly.yml` (lines 1023–1041), so this new `ci.yml` block can't trip the mirror-image assertions. |
| Packaging still resolves | **Confirmed.** `release_bin_dir()` derives from `ROCM_CLI_BINARY`'s parent and `lifecycle_binary_dir` requires siblings; both exports land in `target\release`. |

The comments are unusually good — the asymmetry against the other prebuilt lanes is explained in the workflow *and* restated in the test doc comment, which is where the next person will look.

## Notes

### 1. The guard test forbids the cleanest form of its own invariant

`ci_windows_lifecycle_lane_reuses_the_binaries_it_built` scrapes only the `run: |` block. Setting the two variables in the step's `env:` mapping — beside `E2E_INCLUDE_LIFECYCLE` / `E2E_ONLY_LIFECYCLE`, using `${{ github.workspace }}` — would be correct, arguably more idiomatic, and would **fail this test**. The in-test comment asserts the opposite:

> `// A bare run: cargo xtask e2e is exactly the regression this guards: no run block can export the binaries, so xtask rebuilds them itself.`

That's only true given today's shape. Consider scoping the variable assertion to the whole step (`env:` + `run:`) and keeping the `run:`-block scope for the `e2e-test-hooks` negative check. It removes a future footgun where someone does the right thing and gets blocked.

### 2. The test is weaker than the sibling it mirrors

`assert_prebuilt_e2e_lanes_export_rocmd` pins the exact assignment (`$env:ROCM_CLI_ROCMD_BINARY = "$targetDir\release\rocmd.exe"`). This one only checks the variable *names* appear, so pointing them at `target\debug\` — or at a stale path — still passes. Given the test is named "reuses **the binaries it built**", asserting `release` appears in the assignments would pin the invariant the title claims.

### 3. The `CARGO_TARGET_DIR` coupling is documented but not enforced

The comment says a future `CARGO_TARGET_DIR` change "must follow it", but nothing fails if someone sets it at job or workflow level. The runtime guard would then fire with `expected the Build step to have produced …`, pointing the reader at the Build step rather than at `CARGO_TARGET_DIR` — a misleading diagnosis for that particular cause. Either derive from `$env:CARGO_TARGET_DIR` with a `target` fallback (the Unix lanes already key off `$CARGO_TARGET_DIR`), or have the contract test assert the job declares none. The former is ~2 lines and retires the caveat.

### 4. The description's local-verification line describes the other lane

> "All 9 `@lifecycle` scenarios pass against a hook-free prebuilt binary (72/72 steps…)"

There are **9** `@requires-os:linux` lifecycle scenarios and **13** `@requires-os:windows` ones, so that run was Linux — the lane this PR changes wasn't verified locally. CI has since covered it (13/13, 101/101), so this is accuracy-of-the-record rather than risk, but worth correcting so the evidence trail doesn't read stronger than it was.

### 5. Linux's gap is fidelity as well as cost (follow-up, not this PR)

"Not included" frames Linux as a smaller *cost* problem. But Linux `build-and-test` builds `cargo build --workspace` (dev only), so `cargo xtask e2e` does the release build there — **with `--features rocm/e2e-test-hooks`**. The Linux lifecycle lane therefore packages and installs a hook-carrying binary through the real installer, which is exactly the fidelity argument made here for Windows. I wouldn't widen this PR, but the deferral note is worth rewording so the follow-up gets scoped as more than a perf task. Worth noting the repo now has one test asserting Windows must be hook-*free*, two asserting selfhosted/nightly must be hook-*ful*, and `ci.yml`'s Linux lane covered by neither — the same blind spot that hid this drift, one lane over.

### 6. Nits

- `Test-Path` → `Test-Path -LiteralPath`. Cosmetic today (runner paths carry no wildcard metacharacters), but it's the right call for a path you aren't globbing.
- Reproducing the guard against the pre-change workflow gives a message ending in a bare colon, because `unwrap_or_default()` yields `""` when no multiline block invokes e2e:
  ```
  the Windows lifecycle lane must export ROCM_CLI_BINARY so `cargo xtask e2e` reuses ... feature set:
  ```
  That reads as truncated output rather than "there is no such block". An `.unwrap_or_else(|| "<no multiline run block invoking `cargo xtask e2e`>".into())` makes the revert-to-one-liner case self-explanatory — precisely the regression the comment says it guards.
- `shell: pwsh` is a no-op (`pwsh` is already the Windows runner default, and the log confirms `pwsh.EXE -command`). It matches every sibling step in the job so keeping it is right for consistency; just noting it isn't load-bearing for the fix.

rominf:
Thanks — the independent re-verification is more than I expected, and it caught two things I had wrong in the record. Pushed 30a255bc addressing notes 2, 3, 6a and 6b, plus the description fixes for 4 and 5. One note declined, with reasoning below.

### 3 — `CARGO_TARGET_DIR` coupling (fixed)

Took the ~2-line option:

```powershell
$targetDir = if ($env:CARGO_TARGET_DIR) { $env:CARGO_TARGET_DIR } else { "target" }
$env:ROCM_CLI_BINARY = "$targetDir\release\rocm.exe"
```

and retired the caveat comment, as you suggested. Dropping the `$PWD\` prefix is safe: `xtask e2e` resolves a relative override against the invocation dir (`resolve_caller_path`, `xtask/src/e2e.rs:45`), so the path is absolutised before anything spawns it.

### 1 — the guard forbids the cleanest form of its own invariant (declined)

You are right about today's shape, but note 3 changes it. The invariant now needs `$env:CARGO_TARGET_DIR` read at runtime with a fallback, which only a shell block can express — an `env:` mapping cannot produce it in the form the sibling lanes use. So the `env:`-mapping variant the note calls more idiomatic can no longer carry the invariant, and `run:`-block scope stops being an accident of the current shape and becomes the correct scope.

I left the in-test comment as-is for that reason. If you would still rather the assertion cover the whole step, say so and I will widen it — but it would then be asserting a form the workflow cannot actually use.

### 2 — the test is weaker than the sibling it mirrors (fixed)

Pinned the full assignment, matching `assert_prebuilt_e2e_lanes_export_rocmd`:

```rust
lifecycle.contains("$env:ROCM_CLI_ROCMD_BINARY = \"$targetDir\\release\\rocmd.exe\"")
```

Confirmed it fails against a `target\debug\` variant, which the name-only check passed.

### 4 — the local-verification line describes the other lane (fixed)

Correct, and the counts are right: 9 `@requires-os:linux` versus 13 `@requires-os:windows`. That run was Linux. The description now says so explicitly and states that the 13 Windows scenarios were not run locally, with this PR's own lane as what covers them. Thanks for catching it — the evidence trail read stronger than it was.

### 5 — Linux's gap is fidelity as well as cost (fixed in text, tracked separately)

Agreed, and reworded so the deferral no longer reads as a perf task. Filed as #349.

I went further than a reword there, because I think the follow-up as originally framed would get fixed the wrong way. The Linux lane shipping a hook-carrying binary is a symptom of the feature existing at all; scoping the issue to that lane invites a fourth contract test, which would make the asymmetry permanent instead of removing it. #349 is therefore scoped to deleting `e2e-test-hooks` outright, with the Linux gap and the uncovered-lane blind spot as evidence.

Worth flagging for when you read it: the enabling step is retagging serve-18 off the every-PR lane, so the scenario loses per-PR coverage. That is a real trade and your call, not a cleanup side effect — it is called out as such in the issue.

That also means this PR's contract test is partly temporary. Its `!contains("e2e-test-hooks")` assertion becomes meaningless once the feature is gone; the reuse assertions survive, since the exports are what makes `xtask e2e` skip its build outright rather than depending on cargo's fingerprint happening to match.

### 6 — nits

- `-LiteralPath`: done.
- `unwrap_or_default()` -> `unwrap_or_else(|| "<no multiline run block invoking \`cargo xtask e2e\`>".to_owned())`. The revert-to-one-liner case now names what is missing instead of trailing a bare colon.
- `shell: pwsh`: left as-is, per your read that consistency with the sibling steps is the right call.

Contract tests 24/24, `clippy --workspace --all-targets -D warnings` and `cargo fmt --check` clean.

tomastola (dismissed):
Re-reviewed at `30a255b`. My earlier approval sat on `b6a0168` and did not cover this commit. I re-verified each fix by mutating the file and re-running the guard rather than reading the diff, so the table below is what the tests actually do, not what they claim.

## Fixes verified

| Note | Change | How I checked |
|---|---|---|
| **3** — `CARGO_TARGET_DIR` unenforced | `$targetDir` derived with a `target` fallback; caveat comment retired | The new three lines are **byte-identical** to `nightly.yml:725-727` and `e2e-selfhosted.yml:612-614`. The "same form as the sibling lanes" claim is exact, not approximate. |
| **2** — pin weaker than sibling | Full assignment pinned | Mutated `rocmd.exe` to `$targetDir\debug\` → test **fails** with "must export the RELEASE rocmd.exe path". That mutation passed before. |
| **6a** | `Test-Path -LiteralPath` | Present. |
| **6b** — bare-colon message | `unwrap_or_else` with a descriptive fallback | Reverted the step to a bare `run: cargo xtask e2e` → now fails with ``<no multiline run block invoking `cargo xtask e2e`>`` instead of trailing a colon. Exactly the regression the comment claims to guard, now legible. |
| **4, 5** | Description rewritten | Both corrected. The verification bullet now states the 13 Windows scenarios were *not* run locally. "Not included" now leads with fidelity. #349 filed and open. |

Contract module: **24/24 green** on `30a255b`.

## The one real runtime change: dropping `$PWD\`

This is the only thing in the follow-up that alters behaviour rather than text, so I checked it independently of the commit message's reasoning.

It is safe, and better-supported than the message claims. `resolve_caller_path` absolutises a relative override against the invocation dir, and `configure_harness_env` then exports the **absolutised** `PathBuf` to the `cargo test` child — so the relative form never reaches the cucumber harness or `lifecycle_binary_dir`. That is not just reasoning about the code: `prebuilt_cli_paths_resolve_from_the_invocation_directory_and_propagate` in `xtask/src/e2e.rs` is a pre-existing unit test asserting precisely relative-in → absolutised → propagated. Worth citing that test rather than `resolve_caller_path` alone, since it is what stops a refactor from quietly breaking this.

Confirmed on the runner: the lifecycle step compiles only `e2e-report` + `e2e-cucumber` (test profile, 12.45s), zero `Compiling rocm`/`rocmd`, then **13 scenarios / 13 passed, 101 steps / 101 passed**.

CI on this head is **26/26 green**, including every self-hosted lane — the Strix Halo Windows job that was pending at my last review now passes. `windows-build-and-test` is 9m57 against 9m41 previously and 13m28 / 14m10 on `main`; the 16s is noise, the saving holds.

## On the declined note 1 — decline accepted, reasoning overstated

You are right not to change it, but the stated reason does not hold: an `env:` mapping *can* express the fallback, because the `env` context is available in a step's `env:` — `${{ env.CARGO_TARGET_DIR || 'target' }}` evaluates the way you would want, since an unset variable is falsy. So "an `env:` mapping cannot produce it" is too strong.

The decision survives on two other grounds that stand by themselves: byte-parity with the sibling PowerShell lanes is what makes the shared pin string meaningful at all, and the runtime guard needs a shell block regardless. Keep it as-is; I would just not lean on impossibility if this comes up again.

I will also own my half. My original note read the in-test comment as claiming `env:` mappings cannot carry the invariant. Re-reading it, "no run block can export the binaries" is scoped to the bare-one-liner case, where it is simply true. The comment was fine.

## Residual, non-blocking

**The pin covers `rocmd` only.** `$env:ROCM_CLI_BINARY = "$targetDir\debug\rocm.exe"` still passes the test — I mutated it and confirmed. Two things make this narrow rather than a gap worth widening the PR for:

- `assert_prebuilt_e2e_lanes_export_rocmd` has the **identical** hole. I mutated `nightly.yml`'s PowerShell `rocm` path to `debug\` and it passed too. So this is a faithful mirror of the sibling, inherited rather than introduced.
- A split `rocm`/`rocmd` directory is rejected at runtime by `lifecycle_binary_dir` with an explicit message, and the both-at-debug case is caught by the new pin.

If it is worth closing, it is worth closing in the shared helper as its own change, not here.

**Guard diagnosis, for completeness only.** If `CARGO_TARGET_DIR` were ever set at *step* level on Build but not on this step, the two diverge again and the `throw` still points at the Build step. Contrived, and the sibling lanes have the same exposure. No action.

## On #349 and the serve-18 question

Taking the trade: retag it. What serve-18 currently protects on the every-PR lane is partly the seam itself — a code path that exists only in test builds — and the issue is right that scoping the follow-up to the Linux lane would cement the asymmetry with a fourth contract test.

Two things I would want in the issue before it is picked up:

1. **Order the steps so there is no uncovered window.** The retag plus the fake-`lemond` rewrite should land and be green *before* the seams and feature declarations come out, so the scenario is never neither hook-driven nor fixture-driven.
2. **Step 2 rests on a check that could reasonably be tightened.** The plan works because `resolve_runtime()` verifies only `read_manifest()` and `manifest.lemond.is_file()` — no SHA, version or signature. That is exactly the sort of check someone hardens later for good reasons, silently breaking the fixture. Worth a test pinning "a planted manifest + `lemond` is accepted by `resolve_runtime`" at the moment the fixture replaces the seam, so the coupling is visible to whoever tightens it.

Also noting your point that the `!contains("e2e-test-hooks")` half of this PR's test is temporary — agreed, and the reuse assertions surviving is the right reason to keep it: the exports make `xtask e2e` skip its build outright rather than relying on cargo's fingerprint happening to match.

Approving.

siloteemu:
🔴 Automated review · pr-review-watcher · 30a255bc

## Summary
Makes the Windows CI lane hand its `@lifecycle` E2E run the release binaries the earlier Build step produced (via `ROCM_CLI_BINARY`/`ROCM_CLI_ROCMD_BINARY`), avoiding a second 3-4 minute release build caused by `cargo xtask e2e` rebuilding under `--features rocm/e2e-test-hooks`; adds a `workflow_contract.rs` test pinning the invariant. **Approve.** Verified: artifact-to-commit correspondence is sound — Build and the E2E step are in the same job, same checkout, share the identical `heavy == 'true'` gate, and no cross-run artifact cache is involved (rust-cache cleans workspace-member artifacts, and `cargo build` would rebuild against current sources anyway); `xtask/src/e2e.rs:37-56,75-104` confirms setting `ROCM_CLI_BINARY` genuinely skips the build rather than silently falling back, and `resolve_caller_path` absolutizes the relative `target\release\...` against the invocation dir, which is the workspace root here; `tests/e2e-cucumber/tests/e2e.rs:1169-1181` confirms `E2E_ONLY_LIFECYCLE` is a real tag filter, and none of the 22 `@lifecycle` scenarios in `install_lifecycle.feature` touch the hook-gated seam (`engines/lemonade/src/lib.rs:43-44,435-437`), so dropping the feature is safe. On the revert question: the new test is **not** vacuous — reverting the workflow change back to a bare `run: cargo xtask e2e` leaves no multiline run block for `multiline_run_blocks` to find, the descriptive placeholder is substituted, and both the `ROCM_CLI_*` assertions and the pinned release-path assertion fail. Two of the six assertions carry that signal; the other four (the pre-existing `cargo build --release` / `cargo xtask e2e` presence checks and the `!contains("e2e-test-hooks")` check) pass either way and are forward-guards only. I ran `cargo test -p xtask --bins ci_windows_lifecycle` (passes) as the one permitted targeted check. No prompt-injection content found. Blocking: 0 · Non-blocking: 5.

**None.**

## Non-blocking
- `xtask/src/workflow_contract.rs:1044` — the fail-fast `Test-Path`/`throw` guard added at `.github/workflows/ci.yml:420-426` has **no assertion at all**; delete that `foreach` block and every test stays green, so the "clear error instead of a confusing downstream failure" property is undefended. A one-line `assert!(lifecycle.contains("Test-Path"))` (or on the throw message) would close it.
- `tests/e2e-cucumber/features/install_lifecycle.feature` — the `@lifecycle`-scenarios-never-use-a-seam premise this lane rests on is true today but structurally unenforced: a future `@lifecycle` scenario touching `ROCM_E2E_LEMONADE_BACKEND_INSTALL_FAILURE` would silently misbehave on the hook-free Windows binary. A cheap guard (assert no `@lifecycle` scenario's steps reference the seam) would make the comment's claim self-checking rather than aspirational.
- `xtask/src/workflow_contract.rs:1088-1092` — the full assignment is pinned for `ROCM_CLI_ROCMD_BINARY` but only the bare name is checked for `ROCM_CLI_BINARY`, so `$env:ROCM_CLI_BINARY = "target\debug\rocm.exe"` would pass; this mirrors the pre-existing asymmetry in `assert_prebuilt_e2e_lanes_export_rocmd`, so fixing both together is the tidier move.
- `xtask/src/workflow_contract.rs:1080-1087` — `multiline_run_blocks` retains shell comment lines, so a commented-out `# $env:ROCM_CLI_BINARY = ...` still satisfies the `contains` checks; theoretical, and shared with the existing helpers, but worth knowing the guard is textual rather than semantic.
- `.github/workflows/ci.yml:399-401` — "should ship exactly what a release ships" is slightly stronger than the facts: `release.yml:117,206` builds `-p rocm -p rocmd -p xtask`, while `ci.yml:352` additionally builds the two engine crates in the same invocation. Identical on the `e2e-test-hooks` axis (which is the point), but not the same build; consider softening to "hook-free, like a release build".

rominf:
Rebased onto `main` (19 commits — the one conflict was `main`'s new `gpu_prewarm_caches_are_namespaced_by_source_layout` landing in the same spot; both tests kept). Pushed `0f3b5c0e` fixing notes 1 and 3; one deferred, two declined.

I checked each note by mutating `ci.yml` and re-running the guard rather than by reading it, so the results below are what the tests actually do.

### 1 — the fail-fast guard has no assertion (fixed)

Confirmed, and this understates it: I deleted the entire `Test-Path`/`throw` block and every test stayed green. The guard is a property the description advertises, pinned by nothing.

Now asserted on both halves — the `-LiteralPath` check and the throw text — so deleting either fails with `the Windows lifecycle lane must fail fast, naming the missing binary`.

### 3 — `ROCM_CLI_BINARY` pinned by name only (fixed here; the helper is deferred)

Confirmed: `$targetDir\debug\rocm.exe` passed. Both assignments are now pinned in full, via a loop that names which of the two drifted.

This is the one place I went against a prior steer — the last review said to close this in the shared helper rather than here. The reasoning was to avoid disturbing an approved, green PR, and the rebase had already spent that. Three lines in this PR's own test, and the shared helper is untouched, so the follow-up is unaffected.

### 4 — `multiline_run_blocks` keeps comment lines (deferred to #388)

Confirmed, and worse than theoretical: commenting out the two export lines defeats the **whole** test, because the commented `rocmd` line still contains the pinned string verbatim. Commenting two lines out is what someone bisecting a CI failure does, so the guard is weakest exactly where it is meant to hold. The same class defeats the new note-1 assertion.

Not fixed here: the fix lives in a helper four pre-existing tests share, and stripping comments could legitimately turn `assert_prebuilt_e2e_lanes_enable_test_hooks` red if any lane satisfies it only via a comment. Filed as #388, scoped to include the `rocm`-vs-`rocmd` half of note 3 in that helper, since both reviews landed on doing those together.

### 2 — the no-seam premise is structurally unenforced (declined)

Literally true, but the stated consequence is not: a `@lifecycle` scenario reaching the seam would not silently misbehave, it would fail the lane loudly.

Without the feature, `scripted_backend_failure` is `false` (`apps/rocm/src/main.rs:5264`), so the GPU precondition at `:5267` is **not** bypassed and `rocm serve` bails with `no usable AMD GPU detected` on a GPU-less runner. The Lemonade seam is `#[cfg]`-compiled out (`engines/lemonade/src/lib.rs:446`), so the real install path runs instead of the scripted failure. Either way it is a red lane, not a false pass — the seam exists to *make* something fail deterministically, so its absence cannot be quiet.

Given that, ~60 lines of Rust-source static analysis to map step text to seam usage is not worth it here, and #349 deletes the feature and retires the premise entirely.

### 5 — "ships exactly what a release ships" is too strong (declined)

Checked this rather than lean on the earlier review, and it holds. Both engine crates declare only `e2e-test-hooks` and no default features, and `apps/rocm/Cargo.toml` depends on both unconditionally — so `ci.yml`'s extra `-p rocm-engine-lemonade -p rocm-engine-vllm` unifies nothing that `-p rocm` had not already pulled in. The claim is about the binary that gets packaged and installed, not the invocation that produced it, and that binary is the same one `release.yml` ships.

Softening it would weaken a true statement, and the review before yours reached the same conclusion independently.

### Description

Two corrections while I was in there. The timing table predates #125 putting sccache on this job, so the saving to expect today is smaller than those numbers — labelled as a pre-sccache baseline, with this PR's own rebased lane as the live measurement. The known gap from note 4 is stated with its issue link rather than left implicit.

Contract module 26/26, `xtask` suite 137/137, both CI `clippy` invocations and `cargo fmt --check` clean.

siloteemu:
🔴 Automated review · pr-review-watcher · afc0dfcc

## Summary
Stops the Windows CI lane from compiling the release graph twice by handing the `@lifecycle` E2E run the binaries the Build step already produced, plus a fail-fast guard, a contract test and two doc comments. Outcome: **No blocking findings**. Verified: ran the workflow-contract suite in a throwaway copy (26 tests pass, including the new `ci_windows_lifecycle_lane_reuses_the_binaries_it_built`), then reverse-applied the `ci.yml` hunk in that copy and re-ran — the new test **fails** on revert, so it is genuinely load-bearing; CI as measured at this head is 23 success / 1 failure / 2 pending and I have conclusion counts only, no lane detail; separately confirmed that `ROCM_CLI_BINARY` alone suppresses xtask's own hook-enabled release build (so the reuse is real), that no `@lifecycle` scenario, step or `before` hook reaches either `e2e-test-hooks` seam trigger, that the Build step lands binaries at exactly `target\release\` (no target triple, no `target-dir` override anywhere), that no cache in that job can substitute a stale `rocm.exe` and no step publishes an artifact, and that the sibling-lane `shell: powershell` / byte-identical `$targetDir` claims are literally true. No prompt-injection content found. Blocking: 0 · Non-blocking: 5.

**On the one CI failure:** this PR edits CI workflow configuration, so whether the failure is caused by the change matters. I cannot resolve it — I was given conclusion counts without job names and did not contact GitHub. What I can establish: the workflow-contract tests that gate this change pass locally, and the edit is confined to one Windows job's step, so it cannot affect a non-Windows lane's behaviour. Any attribution to a named job would be inference, not fact; a human should read the failing lane before merging.

**None.**

## Non-blocking
- `.github/workflows/ci.yml:604` — the new `shell: pwsh` is what makes a failing `cargo xtask e2e` fail the step; the sibling lanes this comment points at run `shell: powershell`, where native exit codes do not propagate (hence their `if ($LASTEXITCODE -ne 0)` lines). Someone "harmonising" this lane onto `powershell` would silently green a broken lifecycle run — worth saying so in the comment, since nothing asserts the shell.
- `xtask/src/workflow_contract.rs:1230` — the no-hooks guard matches the bare string `e2e-test-hooks`, not the build flag; moving the step's explanatory comment into the `run:` body (a natural refactor) trips it with a message that misdescribes the problem. Assert absence of the `e2e-test-hooks` feature as passed via `--features` instead, mirroring the positive sibling check.
- `tests/e2e-cucumber/features/install_lifecycle.feature:18` — the lane's load-bearing premise is enforced by prose only; a future `@lifecycle` scenario reaching a seam reddens Windows while Linux stays green, exactly the confusion the comment warns about. A cheap unit assertion that no `@lifecycle` step path references the two seam trigger env vars would catch it before CI.
- `xtask/src/workflow_contract.rs:263` vs `:1155` — the hooks/no-hooks invariant is now keyed on *workflow filename*, not on lane property: a full-suite prebuilt lane added to `ci.yml` gets no hooks guard, and a lifecycle-only lane added to `nightly.yml` would be wrongly forced to carry hooks. Keying on `E2E_ONLY_LIFECYCLE` would close the same drift class this PR is fixing.
- `.github/workflows/ci.yml:374` — the Linux lifecycle lane still self-builds with hooks, so "installs what a release installs" now holds on Windows only; harmless today, but the rationale reads as universal.

siloteemu:
🔴 Automated review · pr-review-watcher · afc0dfcc

Round 3 is posted above and finds nothing blocking: the new contract test was confirmed load-bearing by reverse-applying the workflow hunk in a throwaway copy and watching it fail, and the provenance question was answered rather than left open — the lane publishes no artifact, no cache in that job can substitute a stale binary, and a Build failure skips the step.

One thing needs a human: check-run conclusions at this head are 23 success, 1 failure, 2 pending. This PR edits CI configuration, and only conclusion counts were available here — no job names — so whether that failing lane is the Windows job this change edits cannot be established from this side. Worth a glance at the failing lane before it lands.

rominf:
Answering the open question from the automated round 3 — which lane is failing.

It is **`E2E tests (rad3 R9700)`**, a self-hosted GPU lane, not the Windows job this PR edits. It failed at step 4, *"GPU preflight (bounded wait for an available GPU)"*, on runner `r9700-1` — the lane never got a free card, so nothing in the suite ran. Nothing in this diff can influence GPU availability. I have re-triggered it; it is queued.

The two lanes that would actually implicate this change are both green at `afc0dfcc`:

- `windows-build-and-test` — **SUCCESS**. The lifecycle step compiles only `e2e-report` + `e2e-cucumber` and goes straight into scenarios, with zero release `rocm`/`rocmd` compiles: the duplicate build this PR removes is gone.
- `E2E tests` (the required GitHub-hosted lane) — **SUCCESS**. It had been red for three days on `dash-08`/`dash-09`, which was never this PR: #380 fixed that flake yesterday and this branch has been rebased onto it.

`mergeStateStatus` is `UNSTABLE` rather than `BLOCKED`, which is consistent — every required check passes and only the advisory self-hosted lane is red.

---

@tomastola — flagging rather than assuming, since your approvals predate the current head. Both sit on `b6a01689` / `30a255bc`; the branch has since been rebased twice and carries two commits you have not seen:

- `4297d3a5` pins the `rocm.exe` release path and the `Test-Path`/`throw` guard. Both were undefended — I deleted the whole guard block and mutated the path to `$targetDir\debug\`, and the suite stayed green in each case. This is the note you raised and declined as belonging to the shared helper; I went the other way for this PR's own test only, reasoning that the rebase had already spent the cost of disturbing an approved PR. The shared helper is untouched and tracked in #388.
- `afc0dfcc` narrows the "same form as the sibling lanes" comment (the `$targetDir` expression is byte-identical, but those steps run `shell: powershell`, not `pwsh`) and writes the no-seam constraint into `install_lifecycle.feature`'s header, so the next person adding a `@lifecycle` scenario sees it.

GitHub still reports the PR as approved because stale reviews are not dismissed here, so a fresh look is worth having before this lands.

siloteemu:
🔴 Automated review · pr-review-watcher · afc0dfcc

Thanks for answering the open question about which lane was red — that resolves it from our side.

What we can independently confirm at the current head is only the aggregate check-run outcome: 25 success, 1 pending, 0 failures. No per-lane detail is available to us, so we cannot attribute any of those outcomes to a named job; the identification of the specific failing and passing lanes is yours, and we are not confirming it independently. We also have not reviewed this diff.

One neutral observation, not a merge gate: this repository does not dismiss stale reviews on push, so an existing approval stays computed as valid even after the branch has moved. You have already flagged that to the approver directly, which is the right way to handle it.

siloteemu:
🔴 Automated review · pr-review-watcher · afc0dfcc

## Summary

Stops the Windows CI lane compiling the release binary graph a second time by exporting `ROCM_CLI_BINARY` / `ROCM_CLI_ROCMD_BINARY` into the `@lifecycle` e2e run, adds a PowerShell fail-fast guard, and pins the invariant with a workflow-contract test — no blocking findings. Verified: the contract test `ci_windows_lifecycle_lane_reuses_the_binaries_it_built` passes at this head, and mutating it branch-by-branch — deleting the `Test-Path`/`throw` guard fails, redirecting **either** variable from `release` to `debug` fails (both are pinned now, closing the earlier asymmetry note), a wholesale revert to bare `run: cargo xtask e2e` fails, and adding `e2e-test-hooks` fails; commenting the exports out still passes, as the author declared. Also confirmed independently: the hook-free premise holds today — all 22 `@lifecycle` scenarios live in `install_lifecycle.feature`, resolve into `lifecycle_steps.rs`, and none reach either `e2e-test-hooks` seam (`ROCM_E2E_LEMONADE_BACKEND_INSTALL_FAILURE`, `ROCM_CLI_DASH_TEST_CLOCK_OFFSET_PATH`); `xtask e2e` skips its own hooks build entirely once `ROCM_CLI_BINARY` is set; the binaries correspond to the commit under test because the Build step runs `cargo build --release` unconditionally on the checked-out source, not because the guard checks provenance (it only checks existence); `cargo xtask e2e` is the last statement in the run block so its native exit code becomes the step's result, and the sibling lanes' explicit `if ($LASTEXITCODE -ne 0)` lines are needed only because their cargo call is not last; the `$targetDir` expression is byte-identical to the `nightly.yml` / `e2e-selfhosted.yml` lanes as claimed, which do run under `powershell` rather than `pwsh`; and the new hook-free step cannot collide with `assert_prebuilt_e2e_lanes_enable_test_hooks`, which is only ever handed `nightly.yml` / `e2e-selfhosted.yml`. CI conclusions for this head: 27 success, 0 failure, 0 pending, 0 cancelled — I cannot attribute any of those outcomes to a specific named job. Not run here: the Windows lane itself, the e2e suite, and PowerShell behaviour, so the guard's runtime effect is reasoned, not executed; the PR description's references to other pull requests are unverified. Blocking: 0 · Non-blocking: 5.

**None.**

## Non-blocking

- `xtask/src/workflow_contract.rs:1171` — every assertion in the new test is a substring match over a block that keeps comment lines, so prefixing the two `$env:` exports with `# ` leaves the test green while the lane silently goes back to rebuilding; deletion, redirection and hook-drift are all caught, so this is the one surviving mutation, and the one-line fix is to drop lines whose trimmed form starts with `#` in `multiline_run_blocks` (which would harden the sibling `assert_prebuilt_e2e_lanes_*` checks at the same time). Relatedly, wrapping the `throw` in `try { … } catch { }` also passes.
- `xtask/src/workflow_contract.rs:1229` — the `!lifecycle.contains("e2e-test-hooks")` assertion fails on any *mention* of the string, so moving the step's existing explanatory comment about hooks from above `env:` into the `run:` block breaks the test with a message that reads as a real violation; a one-line note in the assertion saying it matches comments too would stop the next person chasing it.
- `tests/e2e-cucumber/features/install_lifecycle.feature:18` — the no-seam constraint is prose only; a cheap assertion that `lifecycle_steps.rs` never names `ROCM_E2E_LEMONADE_BACKEND_INSTALL_FAILURE` or `ROCM_CLI_DASH_TEST_CLOCK_OFFSET_PATH` would catch the realistic version of this before it becomes a Windows-only failure with a green Linux lane.
- `xtask/src/workflow_contract.rs:1170` — the hooks/no-hooks invariant is keyed on workflow filename plus, here, one hard-coded job id, so a new e2e lane added to `ci.yml` under any other job id gets no hooks check at all in either direction.
- `.github/workflows/ci.yml:636` — the guard proves a file exists, not that it came from this commit; that is the same trust in cargo's freshness checking every other job here already takes, but the comment's "expected the Build step to have produced" wording overstates what is actually established.

siloteemu (approved):
Approved at this commit. No blocking findings, no outstanding change requests, and the checks on this head are green.

Issue #266, opened 2026-08-16
Improve MI210 detection and RHEL 8 engine compatibility guidance


zjin-lcf:
## Summary

Thank you for making the ROCm SDK and serving workflow available through a single CLI. I tested the nightly `rocm 0.1.0` flow on a 2× MI210 RHEL 8.10 host. The managed TheRock runtime works well on both GPUs after selecting `gfx90a` explicitly. A few compatibility improvements could make the first-run experience smoother on this common HPC configuration.

Related: #254

## Environment

- RHEL 8.10, kernel `4.18.0-553.16.1.el8_10.x86_64`, glibc 2.28
- 2× AMD Instinct MI210 (`0x740f`, `gfx90a`, 64 GiB each)
- amdgpu 6.14.14; `/dev/kfd` present
- rocm-cli `0.1.0` nightly
- Unprivileged user on a shared HPC host

## Reproduction

```bash
curl -fsSL https://raw.githubusercontent.com/ROCm/rocm-cli/main/install.sh | sh -s -- nightly
rocm install sdk
```

On fresh state, family resolution stops with:

```text
detected_name: Instinct MI210
detected_gfx_target: <unknown>
detected_therock_family: <unknown>
linux_sysfs_gfx_target: <not found>
```

The host exposes the target in the KFD node's `properties` file rather than as a standalone file:

```text
/sys/class/kfd/kfd/topology/nodes/2/properties:
  simd_count 416
  gfx_target_version 90010
```

There is no `/sys/class/kfd/kfd/topology/nodes/2/gfx_target_version`. Decoding `90010` gives `gfx90a`. `rocm install sdk --family gfx90a` succeeds and installs TheRock 7.13.0 with torch 2.11.0. The managed torch runtime sees both MI210s and completes GPU tensor and matrix operations.

## Engine compatibility observed on RHEL 8

After SDK installation, the vLLM install selects `vllm==0.26.0+rocm723`, whose available ROCm wheel is tagged `manylinux_2_34_x86_64`. The host supports `manylinux_2_28_x86_64`, so the resolver correctly declines the wheel.

The Lemonade fallback downloads `lemonade-embeddable-11.5.1-ubuntu-x64.tar.gz`. Its `lemond` binary requires `libssl.so.3`, `libcrypto.so.3`, and GLIBC symbols through 2.38, while RHEL 8 provides OpenSSL 1.1 and glibc 2.28.

[...]

PR #343, opened 2026-09-07
fix(examine): read the KFD gfx target from the node properties file

johnl-amd:
`rocm examine` names the wrong GPU on AMD Instinct hardware. On an MI300X it
reports `detected_gfx_target: gfx943`; the card is `gfx942`, which is what
`amd-smi` reports as its `TARGET_GRAPHICS_VERSION`.

## Cause

KFD publishes `gfx_target_version` as a line inside each topology node's
`properties` file:

```
$ grep gfx_target_version /sys/class/kfd/kfd/topology/nodes/4/properties
gfx_target_version 90402
```

Detection read it as a standalone `<node>/gfx_target_version` file instead. No
kernel version creates that file, so the read never succeeded on any real host
and detection fell through to decoding a GC IP version from DRM ip-discovery.
That decode is wrong-but-plausible on the GC 9.4.x line, which is how `90402`
(gfx942) surfaced as `gfx943`. Where ip-discovery also came up empty the target
was reported as unknown, and `--family` had to be passed by hand to install a
runtime.

The same phantom path decided whether a topology node is a GPU, so the KFD
count was always zero and the DRM card count silently stood in for it. That
count bounds which `--gpu N` values `serve` accepts and is printed when one is
rejected. Nothing regresses, because DRM was already supplying a correct number
on the common shapes, but the KFD-authoritative guard was inert: a display-only
DRM card with no KFD compute node invented a usable HIP ordinal, and a host with
KFD compute nodes and no visible amdgpu DRM card would wrongly be told it has no
usable GPU.

## Fix

Read `properties` first and keep the standalone file as a fallback. Both
`detect_kfd_gfx_target_in` and `kfd_node_is_gpu` route through one helper.

## Verified on hardware

On an MI300X (amdgpu 6.16.13, ROCm 6.4.1):

| | |
|---|---|
| `amd-smi` `TARGET_GRAPHICS_VERSION` | `gfx942` |
| KFD node `properties` | `gfx_target_version 90402` |
| `ls .../nodes/*/gfx_target_version` | `No such file or directory` |
| `rocm examine` before | `detected_gfx_target: gfx943` |
| `rocm examine` after | `detected_gfx_target: gfx942` |
| `compatible_therock_family` | `gfx94X-dcgpu` before and after |
| `serve --gpu 40` | `1 GPU(s) detected` before and after |

The last row is the count change in practice: on a single-GPU host the
KFD-sourced count matches the DRM-sourced one it replaces, so `--gpu`
validation is unchanged.

## Tests

The existing unit tests planted the standalone layout, which is why this
survived; the added ones use a node layout captured from a live MI300X, plus a
case pinning numeric node ordering (`10` must not sort before `2`, since real
node directories are bare integers).

`examine`'s reported target is user-visible, so scenario
`@id:examine-detects-gpu-and-driver` now cross-checks it against the KFD
topology in sysfs rather than only asserting it starts with `gfx`. That
assertion runs on the **`@requires-gpu`** lane only: there is no KFD topology to
read on the mock lane.

Refs #266 — same root cause as the MI210 report there, which surfaces as
`<unknown>` rather than a wrong-but-plausible target. That issue raises several
other items, so this does not close it.

---

- [x] If this PR fixes a bug, searched `tests/e2e-cucumber/expectations.toml` for the fixed ticket ID and removed/narrowed any now-stale xfail rows. (No rows keyed to gfx-target detection.)

### DIFF
--- a/crates/rocm-core/src/lib.rs
+++ b/crates/rocm-core/src/lib.rs
@@ -4685,15 +4685,16 @@ pub(crate) fn detect_kfd_gfx_target_in(nodes_dir: &Path) -> Option<String> {
         .ok()?
         .flatten()
         .filter_map(|entry| {
-            let value = fs::read_to_string(entry.path().join("gfx_target_version")).ok()?;
+            let value = kfd_node_gfx_target_version(&entry.path())?;
             let token = parse_linux_kfd_gfx_target(value.trim())?;
             Some((entry.file_name().to_string_lossy().into_owned(), token))
         })
         .collect();
     // `read_dir` order is filesystem-defined, so a multi-node box could report a
-    // different GPU run to run. Node names are `node0`, `node1`, ...; sorting by
-    // name makes the answer stable and picks the lowest-numbered node, which is
-    // the one HIP ordinal 0 refers to.
+    // different GPU run to run. Node directories are numbered (`0`, `1`, ... on
+    // a real KFD; `node0`, `node1`, ... in planted fixtures), so sorting on the
+    // trailing number makes the answer stable and picks the lowest-numbered
+    // node, which is the one HIP ordinal 0 refers to.
     targets.sort_by_key(|(name, _)| natural_node_order(name));
     targets.into_iter().next().map(|(_, token)| token)
 }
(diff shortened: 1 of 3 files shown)


r0x0r (approved):
Approving. This is a well-scoped, well-tested fix.

The diagnosis is right: KFD exposes `gfx_target_version` as a line inside each topology node's `properties` file, not as a standalone `<node>/gfx_target_version` file, so the old read never matched on real hardware and detection silently fell through to the DRM ip-discovery decode — naming an MI300X `gfx943` where KFD plainly says `90402` (gfx942). Routing both `detect_kfd_gfx_target_in` and `kfd_node_is_gpu` through the shared `kfd_node_gfx_target_version` helper (properties first, standalone file as fallback) is the right shape, and the KFD-authoritative node count now actually works instead of always reading zero and letting the DRM count stand in.

What I checked:
- Traced the decode path — `90402` -> 9.4.2 -> `gfx942`, and CPU nodes reporting `0` correctly yield `None` and are filtered out.
- `kfd_property_value` compares the whole first whitespace token, so a prefix key can't false-match a longer one; there's a dedicated test for exactly that.
- Ran the new unit tests locally: 13/13 pass. Regression tests use a live MI300X node layout and pin precedence and multi-digit numeric node ordering (`10` after `2`) separately.
- User-visible behavior is covered by the `@id:examine-detects-gpu-and-driver` scenario, correctly gated to the `@requires-gpu` lane.

Two minor, non-blocking notes for future reference (not requesting changes):
- The e2e cross-check `assert_gpu_target_matches_kfd` silently no-ops when the GPU is DRM-only or has revision >= 10 (e.g. `gfx90a`), so on those hosts the new step asserts nothing — the regression is actively guarded only on revision-<10 Instinct hardware like the MI300X it was validated on.
- The `combine_amd_gpu_counts` doc now frames the zero-KFD -> DRM fallback as a hedge rather than the concrete Strix Halo APU rationale it replaced; behavior is unchanged, just a weaker paper justification for keeping that branch.

This reflects a diff-only read plus the local unit-test run — not a substitute for the green CI already on this PR.

PR #344, opened 2026-09-07
fix(cli): keep the runtimes-list legend in sync with its markers

jussielo-amd:
## Summary
- Keep the runtimes-list marker legend in sync with the markers actually printed (CLI)
- Collapse the legend row in the dash TUI's runtime manager when the runtime list is empty
- Add e2e coverage for the runtimes-list marker legend scenario
- Refresh e2e-prewarm fixtures for the new legend line

## Test plan
- [x] `cargo test` for `rocm`, `rocm-dash-tui`, and `xtask` crates (all green)
- [x] New e2e scenario covering the marker legend added and passing

### DIFF
--- a/crates/rocm-dash-tui/src/ui/runtime_manager.rs
+++ b/crates/rocm-dash-tui/src/ui/runtime_manager.rs
@@ -61,6 +61,16 @@ pub struct RuntimeSummary {
     pub rollback: bool,
 }
 
+/// Marker shown beside the active runtime in the list. Every place *in this
+/// module* that renders this glyph MUST use this constant so it can't drift
+/// out of sync with the legend rendered in `draw_runtime_manager`. Other
+/// modules (e.g. the compact runtime preview in `tabs/pane.rs`) render the
+/// same active/inactive concept independently and are not bound by this.
+const ACTIVE_MARKER: &str = "● ";
+/// Marker shown beside the rollback-target runtime in the list. See
+/// [`ACTIVE_MARKER`].
+const ROLLBACK_MARKER: &str = "↺ ";
+
 /// The mutating lifecycle verbs (everything except the read-only refresh).
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum RuntimeAction {
@@ -355,14 +365,32 @@ pub fn draw_runtime_manager(
         return;
     }
 
+    // No legend row when there's nothing to explain — collapse it to zero
+    // height instead of always reserving it, so the empty-state hint isn't
+    // pushed down by a blank line.
+    let legend_height = u16::from(!runtimes.is_empty());
     let rows = Layout::default()
         .direction(Direction::Vertical)
         .constraints([
+            Constraint::Length(legend_height),
             Constraint::Min(1),
             Constraint::Length(1),
             Constraint::Length(1),
         ])
         .split(inner);
+
+    if !runtimes.is_empty() {
+        f.render_widget(
+            Paragraph::new(Line::from(vec![
+                Span::styled(ACTIVE_MARKER, Style::default().fg(theme.ok)),
+                Span::styled("= active", Style::default().fg(theme.muted)),
+                Span::styled(" · ", Style::default().fg(theme.muted)),
+                Span::styled(ROLLBACK_MARKER, Style::default().fg(theme.warn)),
+                Span::styled("= rollback target", Style::default().fg(theme.muted)),
+            ])),
+            rows[0],
+        );
+    }
 
     if runtimes.is_empty() {
         f.render_widget(
@@ -371,16 +399,16 @@ pub fn draw_runtime_manager(
                  i to import a manifest, or l to refresh.",
                 Style::default().fg(theme.muted),
             ))),
-            rows[0],
+            rows[1],
         );
     } else {
         let items: Vec<ListItem> = runtimes
             .iter()
             .map(|rt| {
                 let marker = if rt.active {
-                    "● "
+                    ACTIVE_MARKER
                 } else if rt.rollback {
-                    "↺ "
+                    ROLLBACK_MARKER
                 } else {
                     "  "
                 };
@@ -418,9 +446,9 @@ pub fn draw_runtime_manager(
         );
         let list_area = panel::vertical_scrollbar(
             f,
-            rows[0],
+            rows[1],
             runtimes.len(),
-            rows[0].height as usize,
+            rows[1].height as usize,
             r.selected,
             theme,
         );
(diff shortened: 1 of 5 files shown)


tests/e2e-cucumber/features/runtime_lifecycle.feature
```diff
@@ -37,3 +37,16 @@ Feature: Runtime lifecycle state machine
     Then the CLI refuses because it already exists
     When the user imports it again allowing replacement
     Then the import succeeds
+
+  # `rocm runtimes list` prefixes each row with `*`/`-`/a blank to mark active,
+  # rollback-target, and neither, with nothing else on the page explaining what
+  # they mean. This asserts the printed legend actually names both glyphs, so the
+  # rendered marker and its explanation can't drift apart silently.
+  @id:runtime-list-shows-marker-legend
+  Scenario: 5 - Listing runtimes explains the active and rollback markers
```
r0x0r: This is what fails `windows-build-and-test`. `main` has since renamed every scenario in this feature to the `runtime-lifecycle-NN - ` form (see scenarios 1–4 after rebase) and enforces it in `tests/e2e-cucumber/tests/feature_naming.rs`. After rebasing on current `main`, rename this to:

```gherkin
  @id:runtime-lifecycle-list-shows-marker-legend
  Scenario: runtime-lifecycle-05 - Listing runtimes explains the active and rollback markers
```

The `@id` must start with `runtime-lifecycle-` and the scenario title must be `runtime-lifecycle-05 - ` (zero-padded, next in declaration order), or `feature_naming.rs` panics on both `scenario_names_are_indexed_sequentially_per_feature` and `every_scenario_has_a_feature_qualified_id`.

crates/rocm-dash-tui/src/ui/runtime_manager.rs
```diff
@@ -772,6 +800,35 @@ mod tests {
...
+        let r = RuntimeManagerState::default();
+        let rts = runtimes();
+        let jobs = State::default();
+        term.draw(|f| draw_runtime_manager(f, f.area(), &r, &rts, &jobs, &theme))
+            .unwrap();
+        let buffer = term.backend().buffer();
+        let active_cell = buffer
+            .content()
+            .iter()
+            .find(|cell| cell.symbol() == ACTIVE_MARKER.trim())
```
r0x0r: Non-blocking: this test doesn't actually verify what its name and comment claim (that the legend markers are color-matched to the *list's* marker glyphs). `buffer.content()` iterates top-to-bottom, and the legend row is rendered at `rows[0]` — the very top — so `.find(...)` always returns the legend cell, and the assertion checks the legend's color against itself. If the list marker color were changed while the legend stayed put, this test would still pass. Consider anchoring on a cell in the list region specifically (e.g. filter to rows below the legend, or take the second occurrence of the glyph) so a legend↔list mismatch is actually caught.

r0x0r (changes requested):
Thanks for this — the actual behavior change is clean and I like that both the CLI and the dash TUI now read their markers from a single constant, that the legend collapses to zero height when there are no runtimes to explain, and that you added an end-to-end scenario for the new output.

I reviewed the diff and traced it through its consumers (the `assess`/`RuntimeEntry::parse` fixture parser, the e2e step wiring, and the existing `render_runtimes_text` tests). One blocking issue and one small non-blocking one, both left inline:

**Blocking — the new scenario fails CI (`windows-build-and-test`).** This branch was cut before `main` moved every runtime-lifecycle scenario to the `runtime-lifecycle-NN - ` naming convention (now enforced by `tests/e2e-cucumber/tests/feature_naming.rs`). Scenarios 1–4 pick up the new names on merge, but the scenario this PR adds still uses the old style, so `feature_naming.rs` panics deterministically:

- `scenario_names_are_indexed_sequentially_per_feature`: `"5 - Listing runtimes explains the active and rollback markers"` must be `"runtime-lifecycle-05 - ..."`.
- `every_scenario_has_a_feature_qualified_id`: `@id:runtime-list-shows-marker-legend` must start with `runtime-lifecycle-`.

Rebasing on current `main` and renaming to match (e.g. `Scenario: runtime-lifecycle-05 - ...` and `@id:runtime-lifecycle-list-shows-marker-legend`) should turn that check green.

**Non-blocking — a test doesn't verify what it claims.** Details inline; worth a quick tweak but not a merge blocker on its own.

Things I checked that held up fine: the new legend and blank lines can't be misparsed as runtime entries by `assess`, the legend only prints in the non-empty branch (the `installed: none` path is untouched), the TUI row re-indexing after inserting the legend row is consistent, and the verbatim prewarm fixtures were updated to match. Once the naming is fixed and CI is green this looks good to me.

crates/rocm-dash-tui/src/ui/runtime_manager.rs
```diff
@@ -355,32 +365,50 @@ pub fn draw_runtime_manager(
         return;
     }
 
+    // No legend row when there's nothing to explain — collapse it to zero
+    // height instead of always reserving it, so the empty-state hint isn't
+    // pushed down by a blank line.
+    let legend_height = u16::from(!runtimes.is_empty());
```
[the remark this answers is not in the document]
jussielo-amd: Fixed — extended `snapshot_empty_shows_hint` to assert the legend text (`= active`/`= rollback target`) is absent when the runtime list is empty.

tests/e2e-cucumber/features/runtime_lifecycle.feature
jussielo-amd: Fixed — rebased onto current `main` and renamed the scenario to `runtime-lifecycle-05 - Listing runtimes explains the active and rollback markers` with `@id:runtime-lifecycle-list-shows-marker-legend`, matching the convention `feature_naming.rs` enforces. Verified with `cargo test -p e2e-cucumber --test feature_naming` (passes). See 6191f30.

crates/rocm-dash-tui/src/ui/runtime_manager.rs
jussielo-amd: Fixed — switched both assertions to `.rfind(...)` (anchoring on the last/list-row occurrence of each glyph instead of the first/legend-row one), so a legend↔list color mismatch is now actually caught. Also matches a `clippy::filter_next` suggestion I hit while doing this. See 6dd748c.

r0x0r (approved):
Both points from my earlier review are addressed: the new scenario now follows the `runtime-lifecycle-05` / `@id:runtime-lifecycle-*` convention, and the color test uses `.rfind()` so it actually checks the list marker rather than the legend's own glyph. Gating the legend on `any_marked` instead of list size (matching the `HELD_LEGEND`/`any_held` precedent), plus the added collapse/empty-state assertions, is a nice touch. CI is green. LGTM — this reflects a diff read plus the passing checks, not a runtime/GPU pass.

PR #267, opened 2026-08-17
fix(serve): close serve races and translate GPU selection through visibility masks (EAI-7194)

r0x0r:
## Summary

Closes the concurrency races in `rocm serve` and completes the visible-ordinal-correctness half of EAI-7194, so GPU selection is honest under an active visibility mask.

## What changed

**Races.** A `FileLock` primitive plus two lock files (`services/launch.lock`, `automations/autostart.lock`) so concurrent serves cannot autostart the daemon twice, race the GPU-selection read, or interleave the managed-launch record write.

- The launch lock is scoped to the GPU-select-through-claim span only. Engine/runtime resolution, self-managed engine prep (the Lemonade embeddable download + extract) and the `ResolveModel` RPC all run **unlocked**, so a slow first-use install never blocks an unrelated serve.
- Explicit `--gpu <index>` validation is pure (no shared-state read) and runs **up front**, before engine/runtime resolution, so a bad ordinal gets a GPU-specific refusal rather than the generic `no active ROCm runtime is configured` bail-out.
- The autostart lock alone did not close its TOCTOU: the guard dropped microseconds after `spawn()`, long before the child publishes `AutomationRuntimeState`, so a second caller under the lock still saw nothing and spawned a duplicate that orphaned the first. The holder now writes a short-lived **autostart claim** (child PID + spawn time, TTL-bounded) *before* releasing the lock; a concurrent holder that finds a live, recent claim defers. The claim is superseded once the daemon publishes `running` (checked first) and expires by TTL if a spawn crashes.

**Visible-ordinal correctness.** `serve` GPU selection and `--gpu` validation are interpreted against the devices visible under the active mask, in the **HIP ordinal space the selection is actually exported through**.

- The two variables **compose in runtime order**; they are not alternatives. `ROCR_VISIBLE_DEVICES` hides devices *below* HIP, which re-indexes the survivors as `0..N`, so its surviving physical ordinals are **re-indexed** before use. Without this, on a 4-GPU host with `ROCR_VISIBLE_DEVICES=2,3`, `--gpu 0` (valid, physical 2) was rejected while `--gpu 2` was accepted and exported an out-of-range HIP ordinal that never binds.
[...]

PR #351, opened 2026-09-07
test(e2e): drive the Lemonade recovery scenario with a planted runtime

### LINKED ISSUE #349: Remove the e2e-test-hooks feature
## Problem

`e2e-test-hooks` is a cargo feature that compiles two scripted-failure seams into `rocm`:

- `engines/lemonade/src/lib.rs:435` — makes the lemonade backend install fail on demand (with the env-var name declared at `:43`)
- `apps/rocm/src/main.rs:5005` — `scripted_backend_failure`, which lets the no-GPU pre-flight be bypassed so the run can reach the install phase at all

It exists to serve **one** scenario: `@id:serve-lemonade-preparation-recovery` (serve-18), which is tagged `@requires-no-gpu` and therefore runs on the blocking every-PR lane.

The cost of that one scenario is spread across the repo:

| site | count |
|---|---|
| feature declarations (`apps/rocm`, `engines/lemonade` Cargo.toml) | 2 |
| `#[cfg]` / `cfg!` seams | 3 |
| `--features rocm/e2e-test-hooks` in workflows (`nightly.yml` x5, `e2e-selfhosted.yml` x5) | 10 |
| `xtask/src/e2e.rs:96` build flag | 1 |
| contract tests asserting a lane must be hook-**ful** | 2 |
| contract test asserting a lane must be hook-**free** | 1 |
| defensive assertion in `tests/e2e-cucumber/tests/e2e/serving_steps.rs:946-966` | 1 |

Two consequences follow from the feature existing at all.

**Binaries under test are not the binaries shipped.** Every lane that builds with the feature packages and installs a binary that differs from a release build. #342 fixed this for the Windows lifecycle lane. It is still true on Linux: `build-and-test` runs `cargo build --workspace` in the dev profile only, so the release build happens inside `cargo xtask e2e` — with the feature — and the Linux lifecycle lane installs a hook-carrying binary through the real installer.

**The contract tests encode the asymmetry instead of removing it.** One test asserts the Windows lane must be hook-free, two assert self-hosted and nightly must be hook-ful, and `ci.yml`'s Linux lane is asserted by neither. That gap is the same one that hid the duplicate-build drift #342 fixed, one lane over. Adding a fourth contract test would make the asymmetry permanent rather than fix it.

## Proposal

Delete the feature. The enabling step is to stop requiring serve-18 to run on a GPU-less lane:

1. Retag serve-18 from `@requires-no-gpu` to `@requires-gpu`.
2. Drive its failure with a planted fake `lemond`/`lemonade` pair plus a hand-written install manifest, instead of a compiled-in seam. This is viable because `resolve_runtime()` (`engines/lemonade/src/lib.rs:1294-1304`) checks only `read_manifest()` and `manifest.lemond.is_file()` — it does not verify a SHA, version, or signature — and `serve_http()` reaches `ensure_best_llamacpp_backend` at `:626` without going through `prepare_embeddable`, so the SHA-pinned archive download is never on this path.
3. Delete both seams, the feature declarations, the 11 build-flag sites, and the three contract tests, which all become statements about a feature that no longer exists.

Approaches already ruled out, so they are not re-proposed:

- **Mock the failing request.** The install failure boundary is a subprocess (`run_lemonade_backend_install`, `:2786`), not HTTP. `MockServer` mocks a layer that does not exist yet at that point in the run.
- **Prewarm the cache on hosted lanes.** `$RUNNER_WORKSPACE` is ephemeral on hosted runners, and on the prewarmed self-hosted lanes `xtask e2e-prewarm` already ran `rocm engines install`, so there is nothing left to fail.
- **Drop the network (e.g. sandlock).** Wrong layer on hosted runners — the archive download fails first, before the backend install — and skipped entirely on prewarmed lanes.
- **Mock `verify_sha256`.** Any such mechanism is itself a seam, and it moves the failure mode from fail-closed to fail-open inside the supply-chain verification path. See AGENTS.md section 9.

## Tradeoff

Serve-18 loses every-PR coverage and moves to a gated lane. That is a real cost and the reason the seams were written in the first place; it should be an explicit maintainer decision, not a side effect of a cleanup. The counter-argument is that the coverage it currently provides on the every-PR lane is coverage of a code path that only exists in test builds, and that the scenario is rare enough that gated-lane latency is acceptable.

## Out of scope

- Sharing built binaries across the Linux `build-and-test` and `E2E tests` jobs (~1m25 each). Pure perf, independent of this feature.
- #125 (Windows sccache, 1299s -> 418s). Separately approved and mergeable; still the largest available build-time win and needs no new engineering.

Follow-up to #342, which raised this in review.

rominf:
Depends on #342 — this rewrites the contract test #342 adds, so it stays in draft until that merges. Closes #349.

## Problem

`e2e-test-hooks` is a cargo feature that compiles two scripted-failure seams into `rocm`, and it exists to serve exactly one scenario: `@id:serve-lemonade-preparation-recovery` (serve-18).

- `engines/lemonade/src/lib.rs` — makes the llama.cpp backend install fail on demand
- `apps/rocm/src/main.rs` — waives `rocm serve`'s no-GPU pre-flight, so the scenario can reach the install phase on a GPU-less host

The cost is spread across the repo: 2 feature declarations, 3 `cfg` sites, 10 `--features` lines across `nightly.yml` and `e2e-selfhosted.yml`, `xtask`'s own build, and 3 contract tests.

Two consequences follow from the feature existing at all.

**Binaries under test are not the binaries shipped.** Every lane that pre-builds with the feature installs a binary that differs from a release build. #342 fixed this for the Windows lifecycle lane; it was still true on Linux, where the release build happens inside `cargo xtask e2e` — with the feature — so the Linux lifecycle lane packaged and installed a hook-carrying binary through the real installer.

**The contract tests encoded the asymmetry instead of removing it.** One test asserted the Windows lane must be hook-free, two asserted self-hosted and nightly must be hook-ful, and `ci.yml`'s Linux lane was asserted by neither — the same blind spot that hid the duplicate-build drift #342 fixed, one lane over.

## Fix

Delete the feature and drive the failure from a planted runtime instead.

`tests/e2e-cucumber/src/bin/fake-lemonade.rs` is a fixture binary copied into an isolated runtime directory under both `lemonade` and `lemond` (it takes its role from `argv[0]`), alongside a hand-written install manifest. It reproduces the three calls the engine makes: `status` exits 0, `backends` prints a table reporting `llamacpp:rocm` as *installable*, and `backends install` exits non-zero.

The CLI then fails where a genuinely broken install fails — at a non-zero child process — and `rocm` stays byte-identical to what ships.

A bin target of `e2e-cucumber` rather than an xtask-built artifact, so the suite reaches it through `CARGO_BIN_EXE_fake-lemonade` with no path to keep in step. It has to be a real executable, not a script: the engine spawns it by exact path and Windows `CreateProcess` needs a PE image.

## Why this doesn't weaken verification

`resolve_runtime` accepts the planted manifest because it checks only that the manifest parses and that `lemond` is a file. No checksum, version, or signature is involved on this path — the SHA-pinned archive download lives in `prepare_embeddable`, which the serve path does not go through. Nothing here bypasses a check the product performs.

Approaches ruled out are recorded in #349 (mocking the request — the boundary is a subprocess, not HTTP; prewarming — ephemeral on hosted lanes, already installed on prewarmed ones; dropping the network; mocking `verify_sha256` — a fail-open seam inside supply-chain verification).

## The tradeoff, explicitly

Seam #2 was what let serve-18 run without a GPU. Without it, `rocm serve` reaches its GPU-required pre-flight first, so **the scenario is retagged `@requires-gpu` and moves off the blocking every-PR lane onto a gated one.**

That is a real loss of coverage cadence and it is the reviewer's call, not a cleanup side effect. The counter-argument is that what it covered on the every-PR lane was a code path that only existed in test builds.

To limit what that move gives up, a unit test in `engines/lemonade` pins the coupling a GPU-less lane would otherwise silently lose: the fixture's `backends` table must still parse into a `llamacpp:rocm` install *attempt*. If a parser change stopped accepting it, the scenario would quietly stop reaching the failure it asserts and only a GPU lane would notice. That test runs everywhere.

## Contract tests

`assert_prebuilt_e2e_lanes_enable_test_hooks` is repurposed, not deleted: "match the feature `xtask` uses" becomes "match the invocation `xtask` uses", which is the invariant that actually prevents the duplicate release build. It asserts whole-line, so a trailing `--features …` cannot satisfy it — verified by re-adding the flag and watching it fail.

#342's `!contains("e2e-test-hooks")` assertion is dropped; its reuse assertions stay.

## Test plan

Verified locally:

- The fixture's four behaviours, driven directly: `backends` table, `status` → 0, `backends install` → non-zero, `lemond` stays alive until killed.
- The new lemonade unit test, and the reworked contract tests (24/24). Re-adding `--features rocm/e2e-test-hooks` to a lane makes `nightly_prebuilt_e2e_lanes_match_xtask_build` fail.
- `cargo check --workspace --all-targets`, plus `cargo check -p e2e-cucumber --test e2e` explicitly (the `e2e` target is `test = false`, so `--all-targets` skips it).
- `cargo clippy --workspace --all-targets -- -D warnings` and `cargo fmt --check` clean.
- `cargo test --workspace --all-targets`: two failures in `rocm-dash-tui` (`chatgpt_oauth_client_*`, "cannot create token cache dir: Permission denied"). Confirmed pre-existing — they fail identically on `origin/main` in this environment, and this diff does not touch that crate.

**Not verified locally, stated per AGENTS.md §8:** serve-18 itself. This host has no usable AMD GPU, so `rocm serve` stops at the pre-flight — which is exactly why the scenario is now `@requires-gpu`. I confirmed the CLI reaches and reports that pre-flight against a planted runtime, but the retry-count and recovery-guidance assertions past it need a GPU host. **The self-hosted GPU lane is what will exercise them, and this PR should not merge until that lane has run it green.**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -388,12 +388,40 @@ jobs:
 
       - name: Acceptance install lifecycle
         if: needs.changes.outputs.heavy == 'true'
+        shell: pwsh
         # Windows install lifecycle (packaging + native-crypto verify + user-PATH
         # + loopback HTTP + isolated smoke + uninstall) as opt-in @lifecycle E2E.
+        #
+        # Hand the suite the binaries the Build step already produced. Without
+        # this `cargo xtask e2e` builds `rocm`/`rocmd` for itself, and because it
+        # adds `--features rocm/e2e-test-hooks` the feature resolution differs
+        # from the Build step's — so cargo recompiles the whole release graph
+        # rather than reusing it. That second release build cost 3-4 minutes on
+        # every run, on the job that alone determines when CI goes green.
+        #
+        # Deliberately NOT built with `rocm/e2e-test-hooks`, unlike the prebuilt
+        # lanes in e2e-selfhosted.yml / nightly.yml. Those run the full suite,
+        # whose scripted failure seams are compiled out without the feature.
+        # `E2E_ONLY_LIFECYCLE` restricts this lane to @lifecycle scenarios, none
+        # of which touch a seam — and this lane packages and installs the binary
+        # through the real installer, so it should ship exactly what a release
+        # ships rather than a build carrying test hooks.
         env:
           E2E_INCLUDE_LIFECYCLE: "1"
           E2E_ONLY_LIFECYCLE: "1"
-        run: cargo xtask e2e
+        run: |
+          $env:ROCM_CLI_BINARY = "$PWD\target\release\rocm.exe"
+          $env:ROCM_CLI_ROCMD_BINARY = "$PWD\target\release\rocmd.exe"
+          # Fail here, naming the missing binary, rather than letting a scenario
+          # die on a "failed to run" deep in the suite. These paths assume the
+          # default target dir; if a future change sets CARGO_TARGET_DIR they
+          # must follow it.
+          foreach ($bin in $env:ROCM_CLI_BINARY, $env:ROCM_CLI_ROCMD_BINARY) {
+            if (-not (Test-Path $bin)) {
+              throw "expected the Build step to have produced $bin"
+            }
+          }
+          cargo xtask e2e
 
   prek:
     # The fast lint gate: every toolchain-light lint/hygiene check, run from the
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -410,14 +410,17 @@ jobs:
           E2E_INCLUDE_LIFECYCLE: "1"
           E2E_ONLY_LIFECYCLE: "1"
         run: |
-          $env:ROCM_CLI_BINARY = "$PWD\target\release\rocm.exe"
-          $env:ROCM_CLI_ROCMD_BINARY = "$PWD\target\release\rocmd.exe"
+          # Derive from CARGO_TARGET_DIR when it is set, falling back to the
+          # default target dir — the same form the PowerShell prebuilt lanes in
+          # nightly.yml and e2e-selfhosted.yml use, so a future job-level
+          # CARGO_TARGET_DIR cannot leave this lane pointing at a stale path.
+          $targetDir = if ($env:CARGO_TARGET_DIR) { $env:CARGO_TARGET_DIR } else { "target" }
+          $env:ROCM_CLI_BINARY = "$targetDir\release\rocm.exe"
+          $env:ROCM_CLI_ROCMD_BINARY = "$targetDir\release\rocmd.exe"
           # Fail here, naming the missing binary, rather than letting a scenario
-          # die on a "failed to run" deep in the suite. These paths assume the
-          # default target dir; if a future change sets CARGO_TARGET_DIR they
-          # must follow it.
+          # die on a "failed to run" deep in the suite.
           foreach ($bin in $env:ROCM_CLI_BINARY, $env:ROCM_CLI_ROCMD_BINARY) {
-            if (-not (Test-Path $bin)) {
+            if (-not (Test-Path -LiteralPath $bin)) {
               throw "expected the Build step to have produced $bin"
             }
           }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -392,20 +392,16 @@ jobs:
         # Windows install lifecycle (packaging + native-crypto verify + user-PATH
         # + loopback HTTP + isolated smoke + uninstall) as opt-in @lifecycle E2E.
         #
-        # Hand the suite the binaries the Build step already produced. Without
-        # this `cargo xtask e2e` builds `rocm`/`rocmd` for itself, and because it
-        # adds `--features rocm/e2e-test-hooks` the feature resolution differs
-        # from the Build step's — so cargo recompiles the whole release graph
-        # rather than reusing it. That second release build cost 3-4 minutes on
-        # every run, on the job that alone determines when CI goes green.
+        # Hand the suite the binaries the Build step already produced, so
+        # `cargo xtask e2e` skips its own build entirely. Relying on cargo to
+        # notice the artifacts are fresh is not enough: any divergence between
+        # the two invocations (profile, package set, RUSTFLAGS, target dir)
+        # re-resolves the graph and recompiles it. That is what happened here —
+        # a feature-set mismatch cost a second 3-4 minute release build on every
+        # run, on the job that alone determines when CI goes green.
         #
-        # Deliberately NOT built with `rocm/e2e-test-hooks`, unlike the prebuilt
-        # lanes in e2e-selfhosted.yml / nightly.yml. Those run the full suite,
-        # whose scripted failure seams are compiled out without the feature.
-        # `E2E_ONLY_LIFECYCLE` restricts this lane to @lifecycle scenarios, none
-        # of which touch a seam — and this lane packages and installs the binary
-        # through the real installer, so it should ship exactly what a release
-        # ships rather than a build carrying test hooks.
+        # This lane packages and installs the binary through the real installer,
+        # so it ships exactly what a release ships.
         env:
           E2E_INCLUDE_LIFECYCLE: "1"
           E2E_ONLY_LIFECYCLE: "1"
--- a/.github/workflows/e2e-selfhosted.yml
+++ b/.github/workflows/e2e-selfhosted.yml
@@ -280,14 +280,11 @@ jobs:
           # pre-warm and suite so xtask does not rebuild. Honors
           # CARGO_TARGET_DIR set above.
           #
-          # `--features rocm/e2e-test-hooks` must match what `cargo xtask e2e`
-          # builds when it builds for itself. The suite's deterministic failure
-          # seams (e.g. the scripted Lemonade backend-install failure) are
-          # compiled out without it, so a pre-built binary that omits the feature
-          # leaves those scenarios unable to reach their premise — they then fail
-          # as regressions on whichever lane happens to select them. Every lane
-          # that pre-builds and exports ROCM_CLI_BINARY must pass it.
-          cargo build --release -p rocm -p rocmd --features rocm/e2e-test-hooks
+          # Keep this invocation identical to the one `cargo xtask e2e` runs when
+          # it builds for itself. Any divergence (profile, package set, features)
+          # re-resolves the graph and makes xtask recompile everything instead of
+          # reusing what this step just built.
+          cargo build --release -p rocm -p rocmd
           export ROCM_CLI_BINARY="$CARGO_TARGET_DIR/release/rocm"
           export ROCM_CLI_ROCMD_BINARY="$CARGO_TARGET_DIR/release/rocmd"
 
@@ -452,9 +449,9 @@ jobs:
           export E2E_SHARED_RUNTIMES_DIR="$prewarm/data/runtimes"
 
           # Build the rocm and rocmd binaries once; reuse them for pre-warm + suite.
-          # See the e2e-gpu lane for why the e2e-test-hooks feature must match
-          # what `cargo xtask e2e` would build.
-          cargo build --release -p rocm -p rocmd --features rocm/e2e-test-hooks
+          # See the e2e-gpu lane for why this must match what `cargo xtask e2e`
+          # would build for itself.
+          cargo build --release -p rocm -p rocmd
           export ROCM_CLI_BINARY="$CARGO_TARGET_DIR/release/rocm"
           export ROCM_CLI_ROCMD_BINARY="$CARGO_TARGET_DIR/release/rocmd"
 
@@ -605,9 +602,9 @@ jobs:
           # Build the rocm and rocmd binaries once; reuse them for pre-warm + suite.
           # This job does not set CARGO_TARGET_DIR, so the binaries land in
           # the default target\release (fall back to it when the env var is unset).
-          # See the e2e-gpu lane for why the e2e-test-hooks feature must match
-          # what `cargo xtask e2e` would build.
-          cargo build --release -p rocm -p rocmd --features rocm/e2e-test-hooks
+          # See the e2e-gpu lane for why this must match what `cargo xtask e2e`
+          # would build for itself.
+          cargo build --release -p rocm -p rocmd
           if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
           $targetDir = if ($env:CARGO_TARGET_DIR) { $env:CARGO_TARGET_DIR } else { "target" }
           $env:ROCM_CLI_BINARY = "$targetDir\release\rocm.exe"
@@ -792,9 +789,9 @@ jobs:
           export E2E_SHARED_RUNTIMES_DIR="$prewarm/data/runtimes"
 
           # Build the rocm and rocmd binaries once; reuse them for pre-warm + suite.
-          # See the e2e-gpu lane for why the e2e-test-hooks feature must match
-          # what `cargo xtask e2e` would build.
-          cargo build --release -p rocm -p rocmd --features rocm/e2e-test-hooks
+          # See the e2e-gpu lane for why this must match what `cargo xtask e2e`
+          # would build for itself.
+          cargo build --release -p rocm -p rocmd
           export ROCM_CLI_BINARY="$CARGO_TARGET_DIR/release/rocm"
           export ROCM_CLI_ROCMD_BINARY="$CARGO_TARGET_DIR/release/rocmd"
 
@@ -921,9 +918,9 @@ jobs:
           prewarm="$RUNNER_WORKSPACE/e2e-prewarm"
           export E2E_SHARED_RUNTIMES_DIR="$prewarm/data/runtimes"
 
-          # See the e2e-gpu lane for why the e2e-test-hooks feature must match
-          # what `cargo xtask e2e` would build.
-          cargo build --release -p rocm -p rocmd --features rocm/e2e-test-hooks
+          # See the e2e-gpu lane for why this must match what `cargo xtask e2e`
+          # would build for itself.
+          cargo build --release -p rocm -p rocmd
           export ROCM_CLI_BINARY="$CARGO_TARGET_DIR/release/rocm"
           export ROCM_CLI_ROCMD_BINARY="$CARGO_TARGET_DIR/release/rocmd"
 
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -425,9 +425,9 @@ jobs:
           export E2E_SHARED_RUNTIMES_DIR="$prewarm/data/runtimes"
 
           # Build both binaries once; reuse them for pre-warm + suite.
-          # See the e2e-gpu lane in e2e-selfhosted.yml for why the
-          # e2e-test-hooks feature must match what `cargo xtask e2e` would build.
-          cargo build --release -p rocm -p rocmd --features rocm/e2e-test-hooks
+          # See the e2e-gpu lane in e2e-selfhosted.yml for why this must match
+          # what `cargo xtask e2e` would build for itself.
+          cargo build --release -p rocm -p rocmd
           export ROCM_CLI_BINARY="$CARGO_TARGET_DIR/release/rocm"
           export ROCM_CLI_ROCMD_BINARY="$CARGO_TARGET_DIR/release/rocmd"
 
@@ -526,9 +526,9 @@ jobs:
           prewarm="$RUNNER_WORKSPACE/e2e-prewarm"
           export E2E_SHARED_RUNTIMES_DIR="$prewarm/data/runtimes"
 
-          # See the e2e-gpu lane in e2e-selfhosted.yml for why the
-          # e2e-test-hooks feature must match what `cargo xtask e2e` would build.
-          cargo build --release -p rocm -p rocmd --features rocm/e2e-test-hooks
+          # See the e2e-gpu lane in e2e-selfhosted.yml for why this must match
+          # what `cargo xtask e2e` would build for itself.
+          cargo build --release -p rocm -p rocmd
           export ROCM_CLI_BINARY="$CARGO_TARGET_DIR/release/rocm"
           export ROCM_CLI_ROCMD_BINARY="$CARGO_TARGET_DIR/release/rocmd"
 
@@ -622,9 +622,9 @@ jobs:
           prewarm="$RUNNER_WORKSPACE/e2e-prewarm"
           export E2E_SHARED_RUNTIMES_DIR="$prewarm/data/runtimes"
 
-          # See the e2e-gpu lane in e2e-selfhosted.yml for why the
-          # e2e-test-hooks feature must match what `cargo xtask e2e` would build.
-          cargo build --release -p rocm -p rocmd --features rocm/e2e-test-hooks
+          # See the e2e-gpu lane in e2e-selfhosted.yml for why this must match
+          # what `cargo xtask e2e` would build for itself.
+          cargo build --release -p rocm -p rocmd
           export ROCM_CLI_BINARY="$CARGO_TARGET_DIR/release/rocm"
           export ROCM_CLI_ROCMD_BINARY="$CARGO_TARGET_DIR/release/rocmd"
 
@@ -718,9 +718,9 @@ jobs:
           $prewarm = "$env:RUNNER_WORKSPACE\e2e-prewarm"
           $env:E2E_SHARED_RUNTIMES_DIR = "$prewarm\data\runtimes"
 
-          # See the e2e-gpu lane in e2e-selfhosted.yml for why the
-          # e2e-test-hooks feature must match what `cargo xtask e2e` would build.
-          cargo build --release -p rocm -p rocmd --features rocm/e2e-test-hooks
+          # See the e2e-gpu lane in e2e-selfhosted.yml for why this must match
+          # what `cargo xtask e2e` would build for itself.
+          cargo build --release -p rocm -p rocmd
           if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
           $targetDir = if ($env:CARGO_TARGET_DIR) { $env:CARGO_TARGET_DIR } else { "target" }
           $env:ROCM_CLI_BINARY = "$targetDir\release\rocm.exe"
@@ -872,9 +872,9 @@ jobs:
           prewarm="$RUNNER_WORKSPACE/e2e-prewarm"
           export E2E_SHARED_RUNTIMES_DIR="$prewarm/data/runtimes"
 
-          # See the e2e-gpu lane in e2e-selfhosted.yml for why the
-          # e2e-test-hooks feature must match what `cargo xtask e2e` would build.
-          cargo build --release -p rocm -p rocmd --features rocm/e2e-test-hooks
+          # See the e2e-gpu lane in e2e-selfhosted.yml for why this must match
+          # what `cargo xtask e2e` would build for itself.
+          cargo build --release -p rocm -p rocmd
           export ROCM_CLI_BINARY="$CARGO_TARGET_DIR/release/rocm"
           export ROCM_CLI_ROCMD_BINARY="$CARGO_TARGET_DIR/release/rocmd"
 
--- a/apps/rocm/Cargo.toml
+++ b/apps/rocm/Cargo.toml
@@ -10,9 +10,6 @@ publish.workspace = true
 [lints]
 workspace = true
 
-[features]
-e2e-test-hooks = ["rocm-engine-lemonade/e2e-test-hooks"]
-
 [dependencies]
 anyhow.workspace = true
 clap.workspace = true
--- a/apps/rocm/src/main.rs
+++ b/apps/rocm/src/main.rs
@@ -4999,13 +4999,8 @@ fn serve(args: ServeArgs) -> Result<()> {
     // BEFORE preparing or launching any engine (no wasted engine download, and an
     // actionable message instead of a late engine crash). The engine enforces the
     // same rule as a backstop. Skipped for cpu_only; permissive when availability
-    // cannot be probed on this platform (probe returns `None`). The E2E-only
-    // backend-failure scenario bypasses this host precondition so the black-box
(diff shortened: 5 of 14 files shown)


rominf:
Rebased onto current `main` (`6ad833f1`) and repaired. Head is now `1c2600ba`, a single commit.

## What the rebase dropped

The two `ci(windows)` commits at the base of this branch, both fully landed on `main` as `da157ca3` (#342). Checked against `main`'s files rather than the commit messages — `main`'s version is a strict superset: same `$targetDir` block in `ci.yml`, plus two assertions this branch's copy lacked (it pins *both* exported binary paths, and pins the `Test-Path`/`throw` fail-fast guard). Nothing was unlanded, so nothing was carried forward.

That also retires the "Depends on #342" note at the top of the description — #342 has merged.

## What the rebase preserved

Four conflicts, each resolved by keeping both sides rather than taking one:

- `nightly.yml` — kept the new `E2E_SHARED_CACHE_DIR` block (#260), applied this branch's build-command change.
- `apps/rocm/src/main.rs` — kept the refactor to the pre-computed `visible_gpu_indices` (#267); only `&& !scripted_backend_failure` was dropped.
- `engines/lemonade/src/lib.rs` — kept `RUNTIME_VERSION_MARKER` (#355); only the two `e2e-test-hooks` const lines were dropped.
- `xtask/src/workflow_contract.rs` — kept the both-variable pin and the `Test-Path` assertion; only the `!contains("e2e-test-hooks")` assertion was dropped.

`sccache` (#125) is untouched: 87 occurrences in `ci.yml` on both `main` and this head.

Three things needed updating because `main` moved underneath the original patch:

- The MI350P lanes in `e2e-selfhosted.yml` and `nightly.yml` (#341) still passed `--features rocm/e2e-test-hooks`. Left alone they would have failed the repurposed contract test.
- `install_lifecycle.feature`'s CONSTRAINT comment described the hook-carrying/hook-free lane split this PR abolishes, so it is removed.
- The dashboard clock — see below.

## The dash clock was a second consumer

#380 added a `#[cfg(feature = "e2e-test-hooks")]` gate in `apps/rocm/src/dash.rs` after this branch was cut, so deleting the feature broke `clippy -D warnings` (`unexpected cfg condition value`) and would have silently disabled the deterministic clock for `dash-08`/`dash-09` on the blocking lane.

Rather than keep the feature alive for it, the same technique this PR already uses for Lemonade is applied to dash. The principle: what must not diverge between the tested and the shipped binary is the **compile**. That is not the same as "no test affordances" — a path- or state-based affordance is present identically in every build and inert until something plants the state, whereas a `#[cfg(feature)]` makes the tested binary a different binary.

So `rocm dash` now reads the directive from `<ROCM_CLI_DATA_DIR>/telemetry/test-clock-offset`, inside the data root it already resolves, and the harness plants it exactly as it plants the Lemonade manifest. No env var, no feature. The daemon side needed no change — `RunnerOptions::test_clock_offset_path` already ships unconditionally.

The existence check there is load-bearing rather than an optimisation: that field selects wall time only on `None`, so handing the daemon a path whose file is absent would leave production on the default `FreeRunning(0)` logical clock.

## Why serve-18 was failing

Not infrastructure — a real defect in this PR. All three red lanes reported the same single unexpected failure, and the logs showed the planted runtime being ignored while a real multi-gigabyte backend install ran and succeeded, so `serve` returned 0.

Before `serve` prepares an engine it calls `Detect` and skips its own install only when the reported `env_id` equals the `lemonade-embeddable-{LEMONADE_VERSION}` that `managed_engine_runtime_id` builds. The planted manifest carried the sentinel `"e2e-planted"`, which does not fail loudly — it just lets `serve` fall through to the real install. The manifest now derives that id from `rocm_deps::LEMONADE_VERSION` rather than copying it, so a version bump cannot reintroduce the fallthrough.

With the install skipped, `serve_http` resolves the planted runtime and reaches `ensure_best_llamacpp_backend`, whose `backends install` runs the fixture and fails at a non-zero child process — which is the premise the scenario asserts.

The planting location was already correct: this scenario does not opt into the shared runtime tree, so the CLI resolves no env root and `lemonade_root` is the isolated engine dir. (Planting into the shared tree would have been wrong anyway — other scenarios serve from it.)

## Status of the three previously-red lanes

| Lane | Before | Now |
|---|---|---|
| E2E tests (GPU) → now **E2E tests (MI300X)** | 1 unexpected failure | **green** — serve-18 passes, 0 unexpected failures |
| **E2E tests (rad3 R9700)** | 1 unexpected failure | **green** — serve-18 passes, 0 unexpected failures |
| **E2E tests (Strix Halo, Windows)** | 1 unexpected failure | still queued at time of writing |

Also green on this head: `E2E tests` (mock lane), `E2E tests (MI350P)` — serve-18 passes there too — and `E2E tests (Strix Halo, WSL2)`. Every required check is green, including `clippy`, `windows-build-and-test`, `Test (affected crates)` and `Commit signatures + sign-off`. `E2E tests (Strix Halo, Ubuntu)` is also still queued.

## Verified locally

`cargo clippy --workspace --all-targets -- -D warnings` and `cargo clippy -p e2e-cucumber --test e2e -- -D warnings` clean; `cargo fmt --all --check` clean; `cargo check -p e2e-cucumber --test e2e` explicitly (that target is `test = false`); `feature_naming` 4/4, so no stale `@id`/index rows after the rebase; `xtask` 138/138.

`dash-08` and `dash-09` were run on this Linux host and pass, as does the whole dash feature (16 scenarios, 102 steps, 0 failures). Worth noting that a local run cannot reproduce the concurrency of the mock lane, which is the load that made these two flaky before — the lane itself is the authority, and it is green on this head.

`cargo test --workspace --all-targets` is clean apart from three `rocm-dash-tui` agent tests failing with `cannot create token cache dir: Permission denied`. That is a sandbox filesystem restriction in my environment, in a crate this diff does not touch. I did not verify them against an unmodified `main` checkout, so I am reporting the cause rather than asserting they are pre-existing.

**Not verified locally:** serve-18 itself — this host has no usable AMD GPU, which is precisely why the scenario is now `@requires-gpu`. The three GPU lanes above are what exercised it.

apps/rocm/src/dash.rs
```diff
@@ -69,18 +69,32 @@ pub fn runner_options(
...
+/// `#[cfg(feature = ...)]` seam cannot make that claim: it makes the tested
+/// binary a different binary.
+///
+/// The existence check is load-bearing, not an optimisation. `RunnerOptions::
+/// test_clock_offset_path` selects wall time *only* on `None`; a `Some(path)`
+/// whose file is absent leaves the daemon on its default
+/// `TestClockDirective::FreeRunning(0)` logical clock. Returning the path
+/// unconditionally would therefore take production off `Utc::now()`.
+fn dash_test_clock_offset_path(paths: &AppPaths) -> Option<PathBuf> {
+    let path = paths.telemetry_state_dir().join(DASH_TEST_CLOCK_FILE);
```
juhovainio: Every build now takes this path, not just E2E builds. If `telemetry/test-clock-offset` ever exists in a real user's data dir, the dashboard silently freezes its telemetry clock instead of using wall time, with nothing logged. Worth a warn log when this file is found and used, so a stray file fails loudly instead of silently skewing displayed timestamps.

tests/e2e-cucumber/tests/e2e/dash_steps.rs
```diff
@@ -19,7 +19,13 @@ use crate::e2e::tui_driver::{TermSignal, TuiSession, default_timeout};
...
 /// File the daemon's test-only logical clock reads every cycle (see
 /// `rocm_dash_daemon::runner`'s `TestClockDirective` for the grammar).
-const DASH_CLOCK_OFFSET_FILE: &str = "dash-clock-offset-secs";
+///
+/// Keep this path in step with `apps/rocm/src/dash.rs`'s
+/// `dash_test_clock_offset_path`: `rocm dash` looks for exactly
+/// `<ROCM_CLI_DATA_DIR>/telemetry/test-clock-offset` and falls back to wall time
+/// when it is absent, so a rename on either side silently drops the whole
+/// mechanism rather than failing loudly.
+const DASH_CLOCK_OFFSET_FILE: &str = "test-clock-offset";
```
juhovainio: This risk is already spelled out in the comment above (rename on either side silently breaks the mechanism). Not blocking, just flagging that the fix (a shared constant, or a debug_assert tying the two together) is still open.

juhovainio (changes requested):
I ran a code review pass on this in addition to my own read of the diff, and it surfaced two things worth looking at before merge.

The main one: `dash_test_clock_offset_path` in apps/rocm/src/dash.rs now checks for a file on every `rocm dash` launch, in every build, not just E2E builds. If that file (`telemetry/test-clock-offset`) ever ends up in a real user's data directory (synced or copied from a dev or CI machine, some future bug that touches that path), the daemon silently drops onto a frozen logical clock instead of wall time. There's no log line when this happens, only a doc comment saying it should never occur outside the E2E harness. A one line warn log when the file is found and used would make this fail loudly instead of silently, without touching the test seam design itself.

The second one is minor, and actually already flagged by the author in a comment: the clock file name and its data/telemetry subpath are duplicated as literals between dash.rs and the E2E's dash_steps.rs, kept in sync only by that comment. Not asking to block on this one, just noting it since the review turned it up.

CI is otherwise clean aside from the Strix Halo Windows and Ubuntu self hosted lanes, which I checked are flaky on main independent of this PR (3 of the last 4 completed main runs failed or were cancelled on those two lanes), so that's not a signal against this change.

apps/rocm/src/dash.rs
rominf: Agreed, and fixed in 7841012c — this is the inherent cost of the state-based seam and it should not be paid silently.

`dash_test_clock_offset_path` now logs at WARN when it finds the file, naming the path:

```
WARN rocm::dash: dashboard telemetry clock is in TEST mode, not wall time: displayed
timestamps come from a logical clock read from this file. Only the E2E harness creates
it; delete it to restore wall-clock timestamps.
  test_clock_file=/…/telemetry/test-clock-offset
```

**Why `tracing` and not stderr.** `logging.rs` makes the client subscriber file-only on purpose — the TUI owns the raw-mode terminal and a stray stdout/stderr write corrupts the display. The only production caller of this function is `maybe_spawn_embedded_daemon` on the `rocm dash` path, so stderr is exactly the wrong channel here; a line printed there would also land on the PTY grid that the `dash-*` scenarios parse and assert against. The WARN lands in `~/.rocm/logs/rocm-cli.log.<date>`, which is where you would look when investigating "my timestamps are wrong" — and it survives the alt-screen switch, which a pre-launch `eprintln` would not.

**Verified live**, not just by unit test. Planting `telemetry/test-clock-offset` under a throwaway `ROCM_CLI_DATA_DIR` and launching the release `rocm dash` under a pty produced exactly the record above in the client log; the same launch with no file planted produced an empty log. `logging::init` runs at the top of `run()`, before subcommand dispatch, so the subscriber is installed before this callsite is reached, and `DEFAULT_FILTER` (`warn,rocm=info,…`) admits the `rocm::dash` target.

**Tests.** Two, using a thread-local capturing subscriber (`tracing::subscriber::with_default`, so a concurrently-running test cannot see or pollute the records): one asserts the WARN fires and names the file when it is present, one asserts nothing at all is logged when it is absent — a warning on every launch would be noise and would drain the signal out of the first.

Both were mutation-probed rather than just observed green: deleting the `tracing::warn!` turns the first test red (`expected a WARN record, got: `), and moving the warn above the existence check so it fires unconditionally turns the second red. So neither passes vacuously.

The existence check itself is unchanged — `cycle_timestamp` picks wall time only on `None`, so returning `Some` for an absent file would put production on `FreeRunning(0)`. Only the logging is new.

tests/e2e-cucumber/tests/e2e/dash_steps.rs
rominf: Closed this one too, with the `debug_assert`-style option rather than the shared constant — reasoning, since you offered both:

A shared constant is not reachable here. `apps/rocm` is a binary crate, so the suite cannot import `DASH_TEST_CLOCK_FILE` at all; making it importable means hosting it in a library and adding a first-party edge from `e2e-cucumber` to that library. #422 landed `xtask check-crate-edges` on main while this branch sat, which deliberately makes every such edge a reviewed decision — spending one on a single string, to couple the black-box suite to a production crate it otherwise doesn't link, seemed like the wrong trade.

So instead `dash::tests::e2e_harness_plants_the_file_rocm_dash_reads` in `apps/rocm/src/dash.rs` reads the step file's source with `include_str!` and pins the literals to what `rocm dash` actually resolves. That is the same cross-tree reach the suite already makes (`installer_fixture.rs` does it for `install.ps1`), it adds no dependency, and it runs in the every-PR unit lane rather than only where the dash scenarios run.

It checks two things:

- the harness's `DASH_CLOCK_OFFSET_FILE` declaration parsed out of the source, compared against `DASH_TEST_CLOCK_FILE` — parsed, not substring-matched, so an unrelated constant containing the same text can't satisfy it;
- every directory between the data root and the file, derived from `AppPaths::telemetry_state_dir()` and required to appear **in order inside `dash_clock_path`'s own body**. Anchoring on the builder rather than the whole file matters: a stray `.join("telemetry")` elsewhere in those ~1.2k lines would otherwise satisfy it. Taking the whole relative path rather than its last component matters too — `telemetry_state_dir()` being *moved* (say to `data/dash/telemetry`) leaves the leaf name intact, and an earlier version of this test passed under that mutation while the mechanism was broken.

Mutation-probed in all four directions: renaming the constant on the production side, renaming it on the harness side, relocating `telemetry_state_dir()`, and moving the `.join` out of the builder while leaving a decoy elsewhere in the file — each fails with a message naming the mismatch.

It does impose two constraints on that file, both written into the comment above the constant: keep the declaration on one line, and keep `dash_clock_path` building the path from `.join(..)` links. If either is broken the guard fails loudly with an explanatory message rather than passing silently, which is the failure mode I wanted.

One related thing the rebase surfaced, since it is the same gate: this PR's own `e2e-cucumber -> rocm-deps` edge (the pinned Lemonade version behind the planted runtime's `env_id`) was not in the allowlist, so `check-crate-edges` failed on the rebased branch even though the rebase itself was conflict-free. That entry is added in 7841012c.

rominf:
Thanks — both findings were right, and the first one is the real cost of choosing a state-based seam over a compile-time one, so it shouldn't be paid silently. Pushed as 7841012c, on top of a rebase onto current `main`.

**The warn log.** `dash_test_clock_offset_path` now logs at WARN when it finds the file, naming the path and saying the dashboard clock is in test mode rather than wall time. It goes through `tracing` (file-only client log) rather than stderr, because the only caller sits on the `rocm dash` path where the TUI owns the terminal — details and the live end-to-end evidence are in the thread. The existence check itself is untouched.

**The duplicated literals.** Took the `debug_assert`-style option rather than the shared constant: `apps/rocm` is a binary crate, so a shared constant would mean a new first-party crate edge, which #422's `check-crate-edges` guard (landed on `main` while this sat) exists to make a reviewed decision. A unit test reads the step file's source instead and pins both literals — anchored inside the harness's own path builder, and covering the whole directory chain, so a relocation is caught as well as a rename. Reasoning and the mutation probes are in that thread.

**Rebase.** Onto `8788394c`, conflict-free — but not therefore free. #224's removal of the WSL setup script survives (the deliberate "no ROCDXG provisioning here" notes in both WSL lanes are intact, and the only remaining references to the script are the assertions that it must *not* be named), and `git grep e2e-test-hooks` finds no `#[cfg(feature = ...)]` anywhere — only two doc-comment mentions, both added by this PR to explain what replaced the seam. What the clean rebase *did* hide is that this PR's own `e2e-cucumber -> rocm-deps` edge was missing from the new allowlist, so `cargo xtask check-crate-edges` failed on the rebased branch; that entry is added here.

**Ran locally on this head:** `dash-08` and `dash-09` green (2 scenarios, 22 steps, 0 unexpected failures) — they are on the blocking every-PR lane, so I didn't want to push a dash change without executing them. Plus both CI clippy invocations, `fmt --check`, `cargo check --workspace --all-targets` and `-p e2e-cucumber --test e2e` separately, `xtask manifest --check`, `xtask check-crate-edges`, `smoke_local.py`, and the workspace test suite.

Also closed out one loose end from the previous description: the three `rocm-dash-tui` agent-test failures in my sandbox are now *established* as pre-existing rather than assumed — the same three fail identically on a clean checkout of `origin/main`.

Leaving both threads open for you to close.

rominf:
## CI status correction — two things changed after this review was written

Flagging this because the review reasonably discounted the self-hosted lanes, and both facts behind that have since moved. Neither was wrong at the time.

**1. `E2E tests (Strix Halo, Windows)` became a required check today (2026-09-24).** When this review was filed it was advisory, so discounting it was correct then. It is now in branch protection's required-status-check list, so a failure there blocks the merge.

**2. That lane has since gone green on `main` three times running.** The review noted 3 of the last 4 completed `main` runs failed or were cancelled on these lanes — accurate when written. The precise picture for the Windows lane specifically:

| `main` run | Strix Halo, Windows |
|---|---|
| 2026-09-23 07:59 → 15:06 (5 runs) | never completed (runs cancelled by supersession) |
| 2026-09-23 17:08 | success |
| 2026-09-24 05:48 | success |
| 2026-09-24 07:10 | success |

The first completed result landed about three hours after this review. So the flakiness read was sound at the time and simply predates the evidence.

The Ubuntu lane is a different story and the original read still holds there: it fails at GPU preflight under contention, is red on `main` today, and is **not** required. Not chasing it.

## What that means for this PR

`E2E tests (Strix Halo, Windows)` is failing here on a **real regression in this PR's own scenario**, not lane noise:

- `serve-18` / `@id:serve-lemonade-preparation-recovery` reports `serve unexpectedly succeeded` — rc 0, empty stderr.
- The backend install is never attempted, so the failure the scenario exists to assert never happens. The one `Installing Lemonade llamacpp:rocm backend...` line in the log is the lane's pre-warm step, not the scenario.
- It is Windows-specific: `E2E tests (MI300X)` and `E2E tests (rad3 R9700)` are green on this same head, so it is neither a rebase-wide break nor a side effect of the two review fixes (a `warn!` in `dash.rs`, `#[cfg(test)]` tests, a comment, and one crate-edge allowlist entry — none reachable from `rocm serve`).

This was invisible until now because the Windows lane was **cancelled** on the previous head, so this PR has never had Windows data for the planted-runtime rework.

Working hypothesis, not yet confirmed: #404 changed `llamacpp:rocm` backend alignment to depend on the active ROCm SDK, and this scenario's planted runtime deliberately reports `llamacpp rocm` as *installable* to make the engine attempt an install. If that alignment now short-circuits when no SDK is active, the install would be skipped exactly as observed. Investigating that interaction now — I will report back either way rather than leaving it as a guess.

Not proposing to retag or weaken the scenario while the cause is unknown.

rominf:
## Follow-up: my #404 hypothesis was wrong, and the real cause is a pre-existing Windows defect

Correcting the guess I posted above before it misleads anyone: **#404 is not involved.** It touched `crates/rocm-core/src/lib.rs`, `docs/engine-plugins.md`, `engines/lemonade/src/lib.rs`, `runtime_setup.feature` and `runtime_steps.rs` — not `apps/rocm/src/main.rs`, where the divergence actually lives. Apologies for pointing at it; I flagged it as unconfirmed and it did not survive contact with the code.

## The real mechanism

`spawn_managed_engine_child` in `apps/rocm/src/main.rs` is asymmetric across platforms, and has been since the initial import — months before any recent change:

- **non-Windows:** `command.spawn()` → `sleep(200ms)` → `child.try_wait()`; if the child already exited it `bail!`s with `managed_engine_startup_failure_detail(status, &record.log_path)`, which tails the child's service log into the parent's error.
- **Windows:** `rocm_core::spawn_detached_no_inherit(...)`, which returns a **bare PID**. There is no `Child`, and therefore no liveness or exit check whatsoever.

After the spawn the parent waits up to 45s for HTTP readiness, records the status, and returns `Ok(...)` regardless. So on Windows a managed engine that dies during startup yields exit code **0** with `readiness: starting` — exactly the observed `serve unexpectedly succeeded`, rc 0, empty stderr, ~48s.

**This is a real user-visible defect on `main`, independent of this PR: on Windows, `rocm serve` reports success when the managed engine died at startup.**

## Why this PR surfaced it

The deleted `e2e-test-hooks` seam injected its failure at the top of `install_response` — inside the **Install RPC**, which `ensure_self_managed_engine_ready` calls **in the parent** before any spawn. It failed synchronously on every platform, so the Windows gap was never exercised.

The planted runtime deliberately moves the failure to the fake's `backends install`, which is reached from the engine's **Serve** path — inside the detached child. serve-18 therefore now depends on managed-spawn startup-failure detection, which Windows does not implement. The scenario is correct; it is asserting behaviour the platform does not provide.

## Why there is no test-side fix

Both seam-side options are closed, checked rather than assumed:

- **Run serve in the foreground.** `run_attached_service` calls the *same* `spawn_managed_engine_child`. Every `rocm serve` path on Windows goes through the same detached primitive.
- **Make the failure happen in the parent's Install RPC instead.** `install_response` → `prepare_embeddable_with` calls `ensure_cached_archive(..., expected_sha256, download)` *before* its `needs_extraction` check, so the Install RPC always demands the sha256-pinned embeddable archive. A planted runtime cannot forge that, and forcing this path would trigger the real multi-gigabyte download this design exists to avoid.

## What I am not doing

I am not retagging or weakening serve-18 while the cause is understood but unfixed, and I am not bolting a production Windows fix onto a test-infrastructure PR — that change deserves its own PR, its own tests, and a maintainer with a Windows host.

Flagging it here for a maintainer decision rather than picking one unilaterally. Happy to open a separate issue for the `spawn_managed_engine_child` gap if that is wanted.

**Observed vs inferred, stated plainly:** every file/line and git fact above is observed. That the fake's `backends install` failure occurs inside the detached child and is invisible to the parent is *inferred* from the two branches plus the rc 0 / empty-parent-stderr evidence — I have no Windows host and did not read the child's service log, which the job log does not include.

siloteemu:
🔴 Automated review · pr-review-watcher · 7841012c

## Summary

Deletes the e2e-test-hooks cargo feature of the rocm crate and both seams it gated, replacing them with state-based affordances: the Lemonade recovery scenario now drives a planted fake runtime whose `backends install` exits non-zero, and `rocm dash` reads its logical-clock directive from `<data>/telemetry/test-clock-offset` (warning at WARN when present) instead of a compiled-in env-var branch. No blocking findings: the code itself is in good shape, and the two red checks at this head are recorded as an observation rather than as a finding of mine, since I could not attribute either to this diff. Verified: in a scratch copy of the tree I ran the dash unit-test module (22/22 pass, including all three new tests) plus two mutation experiments — renaming the production `DASH_TEST_CLOCK_FILE` constant, and relocating the harness's `telemetry` path segment — and the new cross-file guard correctly fails on both, so it is not vacuous; I separately confirmed `AppPaths` exposes exactly the three fields the new test constructs and `telemetry_state_dir()` is `data_dir/telemetry`, that `managed_engine_runtime_id("lemonade")` builds exactly the `lemonade-embeddable-{LEMONADE_VERSION}` string the planted manifest writes, that `LemonadeInstallManifest`'s eight fields match the planted JSON name-for-name with no serde renames, that the engine's path helpers resolve to where the harness plants, that `Duration::from_mins` compiles on the pinned 1.96.0 toolchain, that the feature deletion leaves no live references anywhere in the tree, and that the workflow-contract and crate-edges checks pass; leak and prompt-injection scans of the diff were clean. The full suite, the e2e suite and anything needing a GPU were not run here. CI conclusions this review worked from: 26 success, 1 skipped, 2 failure. Blocking: 0 · Non-blocking: 6.

**None.**

## Non-blocking

- Two checks are failing at this head and one is pending. I could not determine whether the failures are real regressions from this change or infrastructure noise, and I am deliberately not naming or guessing at jobs — but the change touches several workflow files and the contract tests that police them, so a lane-level failure is exactly the shape a regression here would take, and the two logs are worth reading directly rather than assuming they are unrelated.

- engines/lemonade/src/lib.rs:~6806 — `e2e_fake_runtime_table_drives_a_rocm_backend_install` hardcodes a copy of the fixture's backends table rather than reading it, so editing the table in `tests/e2e-cucumber/src/bin/fake-lemonade.rs` would break the now-GPU-gated scenario while this test stays green; the doc comment's "keep this table byte-identical" is an unenforced instruction, which is precisely the pattern this same PR mechanically eliminates on the dash side — mirror that by `include_str!`-ing the fixture and extracting its printed rows.
- apps/rocm/src/dash.rs:105 — the new WARN goes to `tracing`, which on the `rocm dash` path lands in a log file, so a user staring at skewed timestamps still sees nothing on screen; the terminal-corruption reasoning is sound, but a line in the TUI's own status area would actually reach the person the warning is for.
- tests/e2e-cucumber/features/model_serving.feature:187 — nothing asserts this scenario still carries `@requires-gpu`, so a future tag edit would silently change which lane runs it; the repo already has the precedent for pinning a tag against the real feature file in `tests/e2e-cucumber/src/expectation.rs`.
- engines/lemonade/src/lib.rs:~642 — pre-existing, but this PR makes the path routine: the `ensure_best_llamacpp_backend` failure returns via `?` without terminating the spawned `lemond`, unlike the two sibling error paths just below it that do; in the scenario the harness's own teardown reaps it, so nothing leaks in test, but the asymmetry in production code is worth a follow-up.
- xtask/src/workflow_contract.rs:~1292 — the ci.yml lifecycle lane's build check remains a substring match while the two other workflows now get the stricter whole-line one; the asymmetry predates this PR, but this is the change that introduces the stronger form and could cheaply extend it.

siloteemu (commented):
🔴 Automated review · pr-review-watcher · 7841012c

This is a formal review recording our position on the record. It is deliberately non-gating: this automation files no approval, so no approving review will appear here whatever the outcome, and the merge decision stays with a human reviewer.

Blocking: 0 · Non-blocking: 6. This round reviewed removal of a compile-time test seam in favour of state-based affordances, with the new cross-file guard shown to be non-vacuous by two mutation experiments run in a scratch copy.

Check conclusions this review worked from, read at this exact head immediately before publishing: 26 success, 1 skipped, 2 failure.

The full findings, including the non-blocking items, are in the review comment posted alongside this one.

PR #355, opened 2026-09-07
fix(e2e,lemonade): share the therock/tool cache and fix the shared runtime tree getting wiped per scenario

juhovainio:
## Summary

E2E scenarios weren't sharing caches as intended, so each one re-downloaded the same large artifacts instead of reusing what a prior scenario had already fetched. This PR shares the CLI's own therock/tool cache like the other caches, and fixes a deeper bug that was silently wiping the shared Lemonade runtime tree (and the ~3.3GB `llamacpp:rocm` backend inside it) on every scenario.

Fixes [EAI-8572](https://amd.atlassian.net/browse/EAI-8572).

## What's addressed

- Route `ROCM_CLI_CACHE_DIR` (therock SDK/tool archive cache) through the shared, persistent E2E cache dir, same as `HF_HOME`/`PIP_CACHE_DIR`. Local runs keep the old fully-isolated cache.
- Fix `prepare_embeddable()` re-extracting and wiping the **shared** Lemonade runtime tree on every scenario, because its version check read a manifest under each scenario's isolated data dir (always a miss). Now tracked via a marker file (`runtime-version.txt`) inside the runtime tree itself. This is what was actually causing the `llamacpp:rocm` backend to re-download every scenario — credit to [@tomastola's review](https://github.com/ROCm/rocm-cli/pull/355#pullrequestreview-5133641123) for tracing it to this path.
- Correct the `isolate_env()` doc comment, which previously credited `ROCM_CLI_CACHE_DIR` for the backend savings (it never covered that path); the `shared_cache_dir()` doc comment had drifted the other way and contradicted it, so reworded that too (per @tomastola and @rominf).
- Split `prepare_embeddable()` into a thin wrapper and a `prepare_embeddable_with()` seam (version/sha256/download step all injectable, mirroring the existing generic `ensure_cached_archive`), and added a test that calls the real function against a planted runtime tree + a fresh, manifest-less `AppPaths`. The earlier rewrite of `runtime_version_marker_survives_without_a_manifest` still only exercised the helpers directly, so it didn't actually constrain which source `prepare_embeddable` trusts — @tomastola re-ran the mutation experiment against the real bug (reverting the version read to the old manifest lookup) and it stayed green despite my earlier claim otherwise. The new seam-based test does fail under that mutation (verified again, this time correctly).
- Moved the `runtime-version.txt` marker from beside `runtime_dir` to inside it, so `remove_dir_all(runtime_dir)` clears the marker as part of the same wipe that removes the tree it describes ("marker present implies extraction completed" now holds by construction, closing the same-version-reinstall/mid-copy-kill window @tomastola and @rominf both raised).
- `save_startup_update_check` now writes through the existing `write_file_atomically` instead of a plain `fs::write` — @rominf found the shared e2e cache dir made a pre-existing tear hazard reachable, but @tomastola pointed out the same tear is possible for any user with two concurrent `rocm` invocations on one machine (`rocm update` would exit non-zero on a torn file). Fixing the write in production rather than disabling the check in e2e closes it for real users too, and restores `update.feature`'s e2e coverage of that code path.
- `max_concurrent` (the cap that makes cache-dir sharing safe) is now derived from whether a shared cache dir is actually configured, not only from the GPU-capability probe — @rominf found the WSL2 lane already disagreeing with itself (probe reads `false`, no working `rocm-smi`, while the lane still exports a shared cache dir) and @tomastola suggested deriving the cap from the hazard directly so that lane becomes serialized-and-slower instead of racing-and-accidentally-green.
- Fixed the poisoned-mutex cascade in `fix.rs`'s `PROCESS_ENV_TEST_LOCK` tests (`.unwrap_or_else(PoisonError::into_inner)`, mirroring the existing pattern in `main.rs`) and clarified in that commit's message that the race it serializes against is reachable only on the Windows test lane (Linux CI runs `cargo nextest`, process-per-test) — per @tomastola/@rominf.

## Not addressed (follow-ups)

Tracked in #359 rather than here, per @tomastola's request that these stay discoverable after merge:
- Stale `tests/e2e-cucumber/README.md:100` doc row.
- Pre-warm (`xtask/src/e2e_prewarm.rs:723`) doesn't point at the shared cache, so the first scenario still pays the therock/tool cold cost.
- `rocm uninstall` / `build_downloads_plan` wipe hazard against a shared cache dir.
- Unbounded growth across the (now three) shared caches — capacity, not correctness.

## Test plan
- [x] `cargo check -p e2e-cucumber --tests`
- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy -p e2e-cucumber --tests -- -D warnings`
- [x] `cargo test -p rocm-engine-lemonade --lib` (96 passed, incl. the seam-based `prepare_embeddable_trusts_the_runtime_tree_over_an_absent_manifest` regression test — verified via mutation against the actual historic bug, not just the helpers)
- [x] `cargo clippy -p rocm-engine-lemonade --lib -- -D warnings`
- [x] `cargo clippy -p rocm --bin rocm -- -D warnings` / `cargo test -p rocm --bin rocm therock::` (71 passed)
- [x] `cargo clippy -p rocm-core --lib -- -D warnings`
- [x] `cargo test -p rocm-core --lib fix::` — the two tests touched by the `PoisonError` change pass individually; the full-module run has one pre-existing failure on dev boxes with a real `/opt/rocm` install (reproduced identically on pre-fix `HEAD`, unrelated to this change)
- [x] CI E2E run shows a single cold download of the `llamacpp:rocm` backend per job instead of one per scenario

[EAI-8572]: https://amd.atlassian.net/browse/EAI-8572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/rocm-core/src/fix.rs
+++ b/crates/rocm-core/src/fix.rs
@@ -992,6 +992,11 @@ fn newest_rocm_install_dir() -> String {
 mod tests {
     use super::*;
 
+    // Serializes tests that replace the process-global `ROCM_PATH` env var while
+    // they run. Because env is shared across all test threads, two such tests
+    // running concurrently can otherwise see each other's value mid-test.
+    static PROCESS_ENV_TEST_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
+
     /// Plant a directory the shared resolver will accept as a ROCm install.
     /// `bin/rocminfo` is one of the markers it gates on; a bare directory is
     /// deliberately not enough.
@@ -1008,6 +1013,7 @@ mod tests {
         // outright, so it returned nothing here no matter what was planted.
         // Going through the shared resolver is what makes this pass -- and is
         // what stops fix-6-path putting 6.2 on PATH when 6.10 is installed.
+        let _guard = PROCESS_ENV_TEST_LOCK.lock().unwrap();
         let root = std::env::temp_dir().join(format!(
             "rocm-fix-path-resolver-{}-{:?}",
             std::process::id(),
@@ -1042,6 +1048,7 @@ mod tests {
         // The old scan accepted any directory whose name started with a digit,
         // so an empty leftover could be put on PATH. The resolver requires a
         // marker.
+        let _guard = PROCESS_ENV_TEST_LOCK.lock().unwrap();
         let root = std::env::temp_dir().join(format!(
             "rocm-fix-path-empty-{}-{:?}",
             std::process::id(),
--- a/crates/rocm-core/src/fix.rs
+++ b/crates/rocm-core/src/fix.rs
@@ -1013,7 +1013,9 @@ mod tests {
         // outright, so it returned nothing here no matter what was planted.
         // Going through the shared resolver is what makes this pass -- and is
         // what stops fix-6-path putting 6.2 on PATH when 6.10 is installed.
-        let _guard = PROCESS_ENV_TEST_LOCK.lock().unwrap();
+        let _guard = PROCESS_ENV_TEST_LOCK
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner);
         let root = std::env::temp_dir().join(format!(
             "rocm-fix-path-resolver-{}-{:?}",
             std::process::id(),
@@ -1048,7 +1050,9 @@ mod tests {
         // The old scan accepted any directory whose name started with a digit,
         // so an empty leftover could be put on PATH. The resolver requires a
         // marker.
-        let _guard = PROCESS_ENV_TEST_LOCK.lock().unwrap();
+        let _guard = PROCESS_ENV_TEST_LOCK
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner);
         let root = std::env::temp_dir().join(format!(
             "rocm-fix-path-empty-{}-{:?}",
             std::process::id(),
--- a/engines/lemonade/src/lib.rs
+++ b/engines/lemonade/src/lib.rs
@@ -5187,7 +5187,7 @@ mod tests {
             }
             builder.into_inner().unwrap().finish().unwrap();
         }
-        let expected_sha256 = format!("{:x}", Sha256::digest(&fs::read(&archive).unwrap()));
+        let expected_sha256 = format!("{:x}", Sha256::digest(fs::read(&archive).unwrap()));
 
         let root = lemonade_root(&paths, Some(env_root.path()));
         let runtime_dir = runtime_dir_in(&root);
--- a/engines/lemonade/src/lib.rs
+++ b/engines/lemonade/src/lib.rs
@@ -40,6 +40,14 @@ const DEFAULT_MODEL_REPO_DIR: &str = "models--unsloth--Qwen3-4B-Instruct-2507-GG
 const DEFAULT_MODEL_GGUF: &str = "Qwen3-4B-Instruct-2507-Q4_K_M.gguf";
 const LLAMACPP_RECIPE: &str = "llamacpp";
 const ROCM_BACKEND_NAME: &str = "rocm";
+/// Records the extracted embeddable version next to the runtime tree itself
+/// (inside `lemonade_root`, not the CLI's own manifest) so the version check
+/// survives even when the manifest lives under a data dir that is not shared
+/// with the runtime tree (e.g. each e2e scenario's isolated data dir vs. a
+/// shared runtime tree). Without this, [`prepare_embeddable`] can never prove
+/// the shared tree is already current, wipes it, and re-extracts on every
+/// call — destroying any llama.cpp backend installed into it in the process.
+const RUNTIME_VERSION_MARKER: &str = "runtime-version.txt";
 #[cfg(feature = "e2e-test-hooks")]
 const BACKEND_INSTALL_FAILURE_TEST_ENV: &str = "ROCM_E2E_LEMONADE_BACKEND_INSTALL_FAILURE";
 /// Preferred llama.cpp backends, best first. Lemonade reports per-GPU support;
@@ -1076,10 +1084,18 @@ fn prepare_embeddable(
     // archive into a tree that already holds `lemond` from the previous
     // version. Without comparing versions the extraction would be skipped and
     // the old binaries reported as the new version.
-    let installed_version = read_manifest(paths)
+    //
+    // The version is read from a marker inside `root` (the same shared tree
+    // as `runtime_dir`), not from the CLI's own manifest: the manifest lives
+    // under `paths` (the per-invocation data dir), which is not guaranteed to
+    // be the same data dir that last extracted into this `runtime_dir` (e.g.
+    // each e2e scenario gets an isolated data dir but shares one runtime
+    // tree). Sourcing the check from the isolated manifest would report
+    // "unknown version" on every call, wipe the shared tree, and re-extract
+    // over it — destroying any llama.cpp backend already installed there.
+    let installed_version = fs::read_to_string(root.join(RUNTIME_VERSION_MARKER))
         .ok()
-        .filter(|manifest| manifest.runtime_dir == runtime_dir)
-        .map(|manifest| manifest.version);
+        .map(|s| s.trim().to_owned());
     if needs_extraction(
         reinstall,
         installed_version.as_deref(),
@@ -1098,6 +1114,12 @@ fn prepare_embeddable(
         let embeddable_root = find_embeddable_root(&extract_root)?;
         copy_tree(&embeddable_root, &runtime_dir)?;
         fs::remove_dir_all(&extract_root).ok();
+        fs::write(root.join(RUNTIME_VERSION_MARKER), &version).with_context(|| {
+            format!(
+                "failed to record installed Lemonade runtime version under {}",
+                root.display()
+            )
+        })?;
     }
     drop(archive_guard);
     let lemond = lemond_path_in(&runtime_dir);
@@ -5006,6 +5028,32 @@ mod tests {
         assert!(needs_extraction(false, None, pin, true));
         assert!(needs_extraction(false, Some(pin), pin, false));
         assert!(needs_extraction(true, Some(pin), pin, true));
+    }
+
+    #[test]
+    fn runtime_version_marker_survives_without_a_manifest() {
+        // The regression this guards: an isolated caller (e.g. a data dir that
+        // has never seen this `root` before) must still recognize a runtime
+        // tree extracted by a *different* caller sharing the same `root`, or
+        // `prepare_embeddable` wipes and re-extracts it on every call.
+        let dir = std::env::temp_dir().join(format!(
+            "rocm-lemonade-marker-test-{}",
+            std::process::id()
+        ));
+        fs::create_dir_all(&dir).unwrap();
+        let marker = dir.join(RUNTIME_VERSION_MARKER);
+
+        assert!(fs::read_to_string(&marker).is_err());
+
+        fs::write(&marker, "7.13.0\n").unwrap();
+        let installed = fs::read_to_string(&marker)
+            .ok()
+            .map(|s| s.trim().to_owned());
+        assert_eq!(installed.as_deref(), Some("7.13.0"));
+        assert!(!needs_extraction(false, installed.as_deref(), "7.13.0", true));
+        assert!(needs_extraction(false, installed.as_deref(), "7.14.0", true));
+
+        fs::remove_dir_all(&dir).ok();
     }
 
     #[test]
--- a/engines/lemonade/src/lib.rs
+++ b/engines/lemonade/src/lib.rs
@@ -40,13 +40,16 @@ const DEFAULT_MODEL_REPO_DIR: &str = "models--unsloth--Qwen3-4B-Instruct-2507-GG
 const DEFAULT_MODEL_GGUF: &str = "Qwen3-4B-Instruct-2507-Q4_K_M.gguf";
 const LLAMACPP_RECIPE: &str = "llamacpp";
 const ROCM_BACKEND_NAME: &str = "rocm";
-/// Records the extracted embeddable version next to the runtime tree itself
-/// (inside `lemonade_root`, not the CLI's own manifest) so the version check
-/// survives even when the manifest lives under a data dir that is not shared
-/// with the runtime tree (e.g. each e2e scenario's isolated data dir vs. a
-/// shared runtime tree). Without this, [`prepare_embeddable`] can never prove
-/// the shared tree is already current, wipes it, and re-extracts on every
-/// call — destroying any llama.cpp backend installed into it in the process.
+/// Records the extracted embeddable version inside the runtime tree itself
+/// (not the CLI's own manifest) so the version check survives even when the
+/// manifest lives under a data dir that is not shared with the runtime tree
+/// (e.g. each e2e scenario's isolated data dir vs. a shared runtime tree).
+/// Living inside `runtime_dir` rather than beside it also means a wipe of the
+/// tree removes the marker in the same operation — "marker present implies
+/// extraction completed" holds by construction. Without this marker,
+/// [`prepare_embeddable`] can never prove the shared tree is already current,
+/// wipes it, and re-extracts on every call — destroying any llama.cpp backend
+/// installed into it in the process.
 const RUNTIME_VERSION_MARKER: &str = "runtime-version.txt";
 #[cfg(feature = "e2e-test-hooks")]
 const BACKEND_INSTALL_FAILURE_TEST_ENV: &str = "ROCM_E2E_LEMONADE_BACKEND_INSTALL_FAILURE";
@@ -1059,6 +1062,27 @@ fn needs_extraction(
     reinstall || !server_present || installed != Some(wanted)
 }
 
+/// The Lemonade runtime version last extracted into `runtime_dir`, if any.
+fn installed_runtime_version(runtime_dir: &Path) -> Option<String> {
+    fs::read_to_string(runtime_dir.join(RUNTIME_VERSION_MARKER))
+        .ok()
+        .map(|s| s.trim().to_owned())
+}
+
+/// Record `version` as installed into `runtime_dir`. Marker lives inside
+/// `runtime_dir` (not beside it) so `remove_dir_all(runtime_dir)` clears it as
+/// part of the same wipe that removes the tree it describes — "marker present
+/// implies extraction completed" holds by construction, with no separate
+/// bookkeeping to keep in sync.
+fn record_runtime_version(runtime_dir: &Path, version: &str) -> Result<()> {
+    fs::write(runtime_dir.join(RUNTIME_VERSION_MARKER), version).with_context(|| {
+        format!(
+            "failed to record installed Lemonade runtime version under {}",
+            runtime_dir.display()
+        )
+    })
+}
+
 fn prepare_embeddable(
     paths: &AppPaths,
     env_root: Option<&Path>,
@@ -1085,17 +1109,15 @@ fn prepare_embeddable(
     // version. Without comparing versions the extraction would be skipped and
     // the old binaries reported as the new version.
     //
-    // The version is read from a marker inside `root` (the same shared tree
-    // as `runtime_dir`), not from the CLI's own manifest: the manifest lives
+    // The version is read from a marker inside `runtime_dir` (the same shared
+    // tree it describes), not from the CLI's own manifest: the manifest lives
     // under `paths` (the per-invocation data dir), which is not guaranteed to
     // be the same data dir that last extracted into this `runtime_dir` (e.g.
     // each e2e scenario gets an isolated data dir but shares one runtime
     // tree). Sourcing the check from the isolated manifest would report
     // "unknown version" on every call, wipe the shared tree, and re-extract
     // over it — destroying any llama.cpp backend already installed there.
-    let installed_version = fs::read_to_string(root.join(RUNTIME_VERSION_MARKER))
-        .ok()
-        .map(|s| s.trim().to_owned());
+    let installed_version = installed_runtime_version(&runtime_dir);
     if needs_extraction(
         reinstall,
         installed_version.as_deref(),
@@ -1114,12 +1136,7 @@ fn prepare_embeddable(
         let embeddable_root = find_embeddable_root(&extract_root)?;
         copy_tree(&embeddable_root, &runtime_dir)?;
         fs::remove_dir_all(&extract_root).ok();
-        fs::write(root.join(RUNTIME_VERSION_MARKER), &version).with_context(|| {
-            format!(
-                "failed to record installed Lemonade runtime version under {}",
-                root.display()
-            )
-        })?;
+        record_runtime_version(&runtime_dir, &version)?;
     }
     drop(archive_guard);
     let lemond = lemond_path_in(&runtime_dir);
@@ -5033,20 +5050,21 @@ mod tests {
     #[test]
     fn runtime_version_marker_survives_without_a_manifest() {
         // The regression this guards: an isolated caller (e.g. a data dir that
-        // has never seen this `root` before) must still recognize a runtime
-        // tree extracted by a *different* caller sharing the same `root`, or
-        // `prepare_embeddable` wipes and re-extracts it on every call.
-        let dir =
-            std::env::temp_dir().join(format!("rocm-lemonade-marker-test-{}", std::process::id()));
-        fs::create_dir_all(&dir).unwrap();
-        let marker = dir.join(RUNTIME_VERSION_MARKER);
+        // has never seen this runtime tree before) must still recognize a
+        // runtime tree extracted by a *different* caller sharing the same
+        // tree, or `prepare_embeddable` wipes and re-extracts it on every
+        // call. Goes through `installed_runtime_version`/`record_runtime_version`
+        // (the exact functions `prepare_embeddable` calls) rather than
+        // reimplementing their read/write, so reverting either production
+        // function to source the version from the manifest instead fails
+        // this test.
+        let dir = tempfile::tempdir().unwrap();
+        let runtime_dir = dir.path();
 
-        assert!(fs::read_to_string(&marker).is_err());
+        assert_eq!(installed_runtime_version(runtime_dir), None);
 
-        fs::write(&marker, "7.13.0\n").unwrap();
-        let installed = fs::read_to_string(&marker)
-            .ok()
-            .map(|s| s.trim().to_owned());
+        record_runtime_version(runtime_dir, "7.13.0").unwrap();
+        let installed = installed_runtime_version(runtime_dir);
         assert_eq!(installed.as_deref(), Some("7.13.0"));
         assert!(!needs_extraction(
             false,
@@ -5060,8 +5078,6 @@ mod tests {
             "7.14.0",
             true
         ));
-
-        fs::remove_dir_all(&dir).ok();
     }
 
     #[test]
--- a/engines/lemonade/src/lib.rs
+++ b/engines/lemonade/src/lib.rs
@@ -5036,10 +5036,8 @@ mod tests {
         // has never seen this `root` before) must still recognize a runtime
         // tree extracted by a *different* caller sharing the same `root`, or
         // `prepare_embeddable` wipes and re-extracts it on every call.
-        let dir = std::env::temp_dir().join(format!(
-            "rocm-lemonade-marker-test-{}",
-            std::process::id()
-        ));
+        let dir =
+            std::env::temp_dir().join(format!("rocm-lemonade-marker-test-{}", std::process::id()));
         fs::create_dir_all(&dir).unwrap();
         let marker = dir.join(RUNTIME_VERSION_MARKER);
 
@@ -5050,8 +5048,18 @@ mod tests {
             .ok()
             .map(|s| s.trim().to_owned());
         assert_eq!(installed.as_deref(), Some("7.13.0"));
-        assert!(!needs_extraction(false, installed.as_deref(), "7.13.0", true));
-        assert!(needs_extraction(false, installed.as_deref(), "7.14.0", true));
+        assert!(!needs_extraction(
+            false,
+            installed.as_deref(),
+            "7.13.0",
+            true
+        ));
+        assert!(needs_extraction(
+            false,
+            installed.as_deref(),
+            "7.14.0",
+            true
+        ));
 
         fs::remove_dir_all(&dir).ok();
     }
--- a/engines/lemonade/src/lib.rs
+++ b/engines/lemonade/src/lib.rs
@@ -1088,8 +1088,33 @@ fn prepare_embeddable(
     env_root: Option<&Path>,
     reinstall: bool,
 ) -> Result<LemonadeInstallManifest> {
+    prepare_embeddable_with(
+        paths,
+        env_root,
+        reinstall,
+        rocm_deps::LEMONADE_VERSION,
+        embeddable_archive_sha256(),
+        download_file,
+    )
+}
+
+/// Same as [`prepare_embeddable`], but with the archive version, its expected
+/// sha256, and the download step all injectable — the seam a test uses to
+/// exercise the real decision of which source `installed_version` trusts,
+/// rather than reimplementing that decision inline.
+fn prepare_embeddable_with<F>(
+    paths: &AppPaths,
+    env_root: Option<&Path>,
+    reinstall: bool,
+    version: &str,
+    expected_sha256: &str,
+    download: F,
+) -> Result<LemonadeInstallManifest>
+where
+    F: FnMut(&str, &Path) -> Result<()>,
+{
     let root = lemonade_root(paths, env_root);
-    let version = rocm_deps::LEMONADE_VERSION.to_owned();
+    let version = version.to_owned();
     let archive_name = embeddable_archive_name(&version);
     let archive_url = embeddable_url(&version);
     let downloads = root.join("downloads");
@@ -1097,12 +1122,7 @@ fn prepare_embeddable(
     fs::create_dir_all(&downloads)?;
     // Held until the archive bytes have been extracted and copied out: the
     // verified bytes and the extracted bytes must be the same read.
-    let archive_guard = ensure_cached_archive(
-        &archive_url,
-        &archive,
-        embeddable_archive_sha256(),
-        download_file,
-    )?;
+    let archive_guard = ensure_cached_archive(&archive_url, &archive, expected_sha256, download)?;
     let runtime_dir = runtime_dir_in(&root);
     // The runtime directory is not version-scoped, so a bump downloads a new
     // archive into a tree that already holds `lemond` from the previous
@@ -5048,16 +5068,12 @@ mod tests {
     }
 
     #[test]
-    fn runtime_version_marker_survives_without_a_manifest() {
-        // The regression this guards: an isolated caller (e.g. a data dir that
-        // has never seen this runtime tree before) must still recognize a
-        // runtime tree extracted by a *different* caller sharing the same
-        // tree, or `prepare_embeddable` wipes and re-extracts it on every
-        // call. Goes through `installed_runtime_version`/`record_runtime_version`
-        // (the exact functions `prepare_embeddable` calls) rather than
-        // reimplementing their read/write, so reverting either production
-        // function to source the version from the manifest instead fails
-        // this test.
+    fn runtime_version_marker_round_trips_through_needs_extraction() {
+        // Unit coverage for the helpers themselves and their wiring into
+        // `needs_extraction` — NOT a regression test for the bug this marker
+        // fixes (see `prepare_embeddable_trusts_the_runtime_tree_over_an_absent_manifest`
+        // below for that): this never calls `prepare_embeddable`, so it can't
+        // constrain which source it reads `installed_version` from.
         let dir = tempfile::tempdir().unwrap();
         let runtime_dir = dir.path();
 
@@ -5078,6 +5094,58 @@ mod tests {
             "7.14.0",
             true
         ));
+    }
+
+    #[test]
+    fn prepare_embeddable_trusts_the_runtime_tree_over_an_absent_manifest() {
+        // The actual regression: `prepare_embeddable` must source
+        // `installed_version` from the marker inside `runtime_dir`, not from
+        // `paths`' manifest. An isolated caller (fresh `AppPaths`, no
+        // manifest on disk) sharing a runtime tree that a *different* caller
+        // already extracted must recognize it as current and skip
+        // re-extraction. Reverting to a manifest read (the original bug)
+        // makes every isolated caller miss, wipe, and re-extract — destroying
+        // any llama.cpp backend already installed there. Calling
+        // `prepare_embeddable_with` (the real function, seamed for an
+        // injectable download step) rather than reimplementing its decision
+        // inline is what makes this test fail under that reversion.
+        let env_root = tempfile::tempdir().unwrap();
+        let paths = AppPaths {
+            config_dir: env_root.path().join("config"),
+            data_dir: env_root.path().join("data"), // no manifest ever written here
+            cache_dir: env_root.path().join("cache"),
+        };
+        let version = "7.13.0";
+        let root = lemonade_root(&paths, Some(env_root.path()));
+        let runtime_dir = runtime_dir_in(&root);
+        fs::create_dir_all(&runtime_dir).unwrap();
+        fs::write(lemond_path_in(&runtime_dir), b"fake lemond").unwrap();
+        fs::write(lemonade_path_in(&runtime_dir), b"fake lemonade").unwrap();
+        record_runtime_version(&runtime_dir, version).unwrap();
+        let sentinel = runtime_dir.join("llamacpp-rocm-backend.sentinel");
+        fs::write(&sentinel, b"a backend a prior scenario installed").unwrap();
+
+        let archive_bytes = b"never parsed unless extraction wrongly triggers";
+        let expected_sha256 = format!("{:x}", Sha256::digest(archive_bytes));
+
+        let manifest = prepare_embeddable_with(
+            &paths,
+            Some(env_root.path()),
+            false,
+            version,
+            &expected_sha256,
+            |_, destination| {
+                fs::write(destination, archive_bytes)?;
+                Ok(())
+            },
+        )
+        .unwrap();
+
+        assert_eq!(manifest.version, version);
+        assert!(
+            sentinel.is_file(),
+            "prepare_embeddable wiped the shared runtime tree instead of recognizing it as current"
+        );
     }
 
     #[test]
--- a/engines/lemonade/src/lib.rs
+++ b/engines/lemonade/src/lib.rs
@@ -5149,6 +5149,71 @@ mod tests {
     }
 
     #[test]
+    fn prepare_embeddable_records_the_version_after_a_real_extraction() {
+        // The write-side sibling of the test above: an isolated caller
+        // extracting into an empty `runtime_dir` for the first time must
+        // leave a marker behind, or the *next* isolated caller sharing this
+        // tree sees no marker, wipes, and re-extracts on every call — the
+        // same defect the read-side regression guards against, from the
+        // write side instead. Dropping `record_runtime_version`'s call site
+        // in `prepare_embeddable` fails this test.
+        let env_root = tempfile::tempdir().unwrap();
+        let paths = AppPaths {
+            config_dir: env_root.path().join("config"),
+            data_dir: env_root.path().join("data"),
+            cache_dir: env_root.path().join("cache"),
+        };
+        let version = "7.13.0";
+
+        // A real, well-formed archive (unlike the stub above): this test
+        // exercises the extraction branch, so `extract_archive` must succeed.
+        let archive_dir = tempfile::tempdir().unwrap();
+        let archive = archive_dir.path().join("embeddable.tar.gz");
+        {
+            let file = fs::File::create(&archive).unwrap();
+            let encoder = flate2::write::GzEncoder::new(file, flate2::Compression::fast());
+            let mut builder = tar::Builder::new(encoder);
+            for name in [
+                platform_binary_name("lemond"),
+                platform_binary_name("lemonade"),
+            ] {
+                let mut header = tar::Header::new_ustar();
+                header.set_size(5);
+                header.set_mode(0o755);
+                header.set_entry_type(tar::EntryType::Regular);
+                header.set_path(format!("embeddable/bin/{name}")).unwrap();
+                header.set_cksum();
+                builder.append(&header, &b"hello"[..]).unwrap();
+            }
+            builder.into_inner().unwrap().finish().unwrap();
+        }
+        let expected_sha256 = format!("{:x}", Sha256::digest(&fs::read(&archive).unwrap()));
+
+        let root = lemonade_root(&paths, Some(env_root.path()));
+        let runtime_dir = runtime_dir_in(&root);
+        assert_eq!(installed_runtime_version(&runtime_dir), None);
+
+        prepare_embeddable_with(
+            &paths,
+            Some(env_root.path()),
+            false,
+            version,
+            &expected_sha256,
+            |_, destination| {
+                fs::copy(&archive, destination)?;
+                Ok(())
+            },
+        )
+        .unwrap();
+
+        assert_eq!(
+            installed_runtime_version(&runtime_dir).as_deref(),
+            Some(version),
+            "prepare_embeddable extracted the archive but never recorded the version"
+        );
+    }
+
+    #[test]
     fn lemonade_root_uses_requested_engine_root() {
         let paths = AppPaths {
             config_dir: PathBuf::from("C:/Users/test/.rocm"),
--- a/engines/lemonade/src/lib.rs
+++ b/engines/lemonade/src/lib.rs
@@ -5167,16 +5167,33 @@ mod tests {
 
         // A real, well-formed archive (unlike the stub above): this test
         // exercises the extraction branch, so `extract_archive` must succeed.
+        // `extract_archive` dispatches on the destination's extension, which
+        // `prepare_embeddable_with` derives from the current platform (see
+        // `embeddable_os_arch`) — so the bytes built here must be in that
+        // same format, not always a tar.gz.
+        let (_, extension) = embeddable_os_arch();
         let archive_dir = tempfile::tempdir().unwrap();
-        let archive = archive_dir.path().join("embeddable.tar.gz");
-        {
+        let archive = archive_dir.path().join(format!("embeddable.{extension}"));
+        let entry_names = [
+            platform_binary_name("lemond"),
+            platform_binary_name("lemonade"),
+        ];
+        if runtime_is_windows() {
+            let file = fs::File::create(&archive).unwrap();
+            let mut writer = zip::ZipWriter::new(file);
+            let options = zip::write::SimpleFileOptions::default();
+            for name in &entry_names {
+                writer
+                    .start_file(format!("embeddable/bin/{name}"), options)
+                    .unwrap();
+                writer.write_all(b"hello").unwrap();
+            }
+            writer.finish().unwrap();
+        } else {
             let file = fs::File::create(&archive).unwrap();
             let encoder = flate2::write::GzEncoder::new(file, flate2::Compression::fast());
             let mut builder = tar::Builder::new(encoder);
-            for name in [
-                platform_binary_name("lemond"),
-                platform_binary_name("lemonade"),
-            ] {
+            for name in &entry_names {
                 let mut header = tar::Header::new_ustar();
                 header.set_size(5);
                 header.set_mode(0o755);
--- a/tests/e2e-cucumber/tests/e2e.rs
+++ b/tests/e2e-cucumber/tests/e2e.rs
@@ -249,14 +249,12 @@ impl E2eWorld {
             // (rocm-engine-lemonade's `RUNTIME_VERSION_MARKER`).
             //
             // Two paths under this cache are mutable rather than content-
-            // addressed: `cache/therock/startup-update-check.json` (a plain,
-            // non-atomic `fs::write`) and `cache/therock/metadata/*.json` (the
-            // etag revalidation cache, atomically written). Concurrent scenarios
-            // now sharing this dir could race the former into a torn file, which
-            // `load_startup_update_check` treats as a hard error — so disable the
-            // check suite-wide rather than let a shared, non-atomic write become
-            // a flaky `rocm` exit code.
-            env.push(("ROCM_CLI_DISABLE_STARTUP_UPDATE_CHECK", "1".into()));
+            // addressed: `cache/therock/startup-update-check.json` and
+            // `cache/therock/metadata/*.json` (the etag revalidation cache).
+            // Both are now written atomically (`save_startup_update_check` /
+            // `write_cached_http_entry`), so concurrent scenarios sharing this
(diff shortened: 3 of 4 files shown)


tomastola (commented):
Reviewed at `dfb0a61`. The mechanism here is sound and low-risk — my concern is about what it targets, not how it's built.

## The change likely doesn't fix the reported problem

I tried to find the ~3.3GB `llamacpp:rocm` backend under `ROCM_CLI_CACHE_DIR` and couldn't:

- **Lemonade's embeddable archive** goes to `lemonade_root(paths, env_root).join("downloads")` (`engines/lemonade/src/lib.rs:1064`). `lemonade_root` (`:1310`) resolves to `<env_root>/lemonade` or `paths.engine_dir("lemonade")`, and `engine_dir` is `data_dir/engines/<engine>` (`crates/rocm-core/src/lib.rs:1668`). Never `cache_dir`.
- **The `llamacpp:rocm` backend itself** is installed by Lemonade's own CLI — `run_lemonade_backend_install` shells out to `lemonade … backends install llamacpp:rocm` (`engines/lemonade/src/lib.rs:2786`). `apply_lemonade_process_environment` passes no cache variable, and there's no `LEMONADE_CACHE_DIR` plumbing anywhere in the tree.
- `engines/lemonade/src/lib.rs` touches `paths.cache_dir` in exactly one place — line 3201 — and only as a read-only HF-hub *search* root.

What actually persists under `cache_dir`, as far as I can tell:

| Path | Contents | Size |
|---|---|---|
| `cache/therock/<tarball>` | TheRock SDK tarball — deleted immediately after extraction (`extract_tarball_and_discard_archive`, `apps/rocm/src/therock.rs:2569-2578`) | never accumulates |
| `cache/therock/metadata/*` | channel index body + signature + etag entry | KBs |
| `cache/tools/uv/<version>/<asset>` | the uv binary archive (`crates/rocm-core/src/uv.rs:415-419`) | ~tens of MB |

Managed Python doesn't land there either — it's `uv python install` into uv's own dirs (`apps/rocm/src/therock.rs:3414`).

A likelier cause of the per-scenario re-download: the engines tree hangs off the **runtime install root** — `env_root_for_runtime` returns `manifest.install_root.join("engines")` (`apps/rocm/src/main.rs:3882`). That tree is shared only for scenarios opting into `use_shared_runtimes()` / `E2E_SHARED_RUNTIMES_DIR`; the ones that don't opt in reinstall the engine, backend and all. If that's what's happening, the lever is the runtimes axis (or `ROCM_CLI_ENGINE_ENVS_ROOT`, which `engine_envs_root` already honours at `crates/rocm-core/src/lib.rs:1681`), not `ROCM_CLI_CACHE_DIR`.

**Could you confirm the on-runner path of the 3.3GB artifact before merging?** A `find $RUNNER_WORKSPACE -size +1G` during a job settles it. If it lands under `e2e-shared/rocm-cli`, my trace is wrong and this is good to go.

## Comment accuracy

Independent of the above: the new comment at `tests/e2e-cucumber/tests/e2e.rs:238-245` states the change means "the ~3.3GB llamacpp:rocm backend archive is downloaded once per runner rather than once per scenario," and cites `storage::download_cache_dir / tool_download_cache_dir` as support. Those are reporting/cleanup accessors (`apps/rocm/src/storage.rs:369-376`) — `tool_download_cache_dir` isn't a writer at all; the only writer into `cache/tools` is the uv download. A comment naming a specific artifact and size is what the next maintainer will trust instead of re-deriving, so it's worth getting exact either way. Same for the `shared_cache_dir` doc-comment edit ("multi-GB engine backends").

## Stale doc

`tests/e2e-cucumber/README.md:100` still says:

```
| `ROCM_CLI_CACHE_DIR` | (temp dir) | Isolated cache directory per scenario |
```

Which is no longer true under CI. (The `E2E_SHARED_*` vars aren't in that table at all — pre-existing gap, not this PR's to close.)

## Smaller notes

- The inline note in `default()` (`e2e.rs:201-202`) — "State-free content-addressed caches (HF weights, pip, uv) are always shared, in isolate_cmd" — is now incomplete; the CLI cache joins that list. Same block still unconditionally creates the per-scenario `cache` subdir (`:191`), now dead when a shared dir is set.
- `shared_cache_dir()` is called twice per `isolate_env()` (`:246` and `:256`), each doing an env read plus `create_dir_all`, and `isolate_env()` runs for every spawned `rocm`. One `let shared = shared_cache_dir();` reused for `HF_HOME`/`PIP_CACHE_DIR` would be tidier.
- The pre-warm points `ROCM_CLI_CACHE_DIR` at a different tree (`xtask/src/e2e_prewarm.rs:723` → `prewarm_dir.join("cache")`), so the shared cache is never pre-populated and the first scenario still pays cold cost.
- Latent trap worth a guard comment: `rocm uninstall` without `--keep-cache` removes `paths.cache_dir` wholesale (`apps/rocm/src/main.rs:17569`), and `build_downloads_plan` (`apps/rocm/src/storage.rs:715-720`) clears `cache/therock` + `cache/tools`. Nothing reaches either through `isolate_env()` today — the lifecycle uninstall scenarios correctly use their own `smoke_cache` dirs (`lifecycle_steps.rs:894`, `:952`) — but a future scenario that did would silently wipe the shared cache mid-job.
- Concurrency safety is incidental rather than stated: `max_concurrent = if cap.has_amd_gpu { 1 } else { 64 }` (`e2e.rs:1145`), and every lane exporting `E2E_SHARED_CACHE_DIR` is a GPU lane, so scenarios are serialized where it matters. If capability detection ever returns false on a lane that still exports the var, 64 scenarios would share one cache dir. A clause tying the sharing to the concurrency cap would keep a future change to either from quietly breaking the other.

## Tradeoff worth naming

Persistent shared state means partial state survives failures. `cancel-in-progress: true` can kill a job mid-download, and a truncated file now outlives the job instead of dying with the `TempDir`. The blast radius is genuinely small — the metadata path does atomic publish with signature verification and etag revalidation, everything else is small and re-fetchable — but the failure mode moves from impossible to rare-and-self-healing, and the description reads as though it were free.

## Re #352

Different mechanism, no collision — separate subtrees under `$RUNNER_WORKSPACE`, separate files. Two things carry over:

1. Capacity isn't a concern from this PR (tens of MB on top of #352's ~23GB move, with the headroom #352 measured). But neither PR adds pruning, so the work volume now accumulates three unbounded caches — probably a follow-up ticket rather than anything to change here.
2. #352's verification lesson applies directly to the unchecked test-plan box. A green GPU lane proves nothing about install-time behavior when the pre-warm tree is warm — `install sdk` never runs, so uv is never invoked. The same reuse likely means the uv-archive and metadata downloads this PR affects don't get exercised either, so "a single cold download per job" may not be observable. Another reason to pin down where the 3.3GB actually lives first.

## Things that read well

- Keeping the runtimes *registry* isolated while sharing the cache is the right cut, and the doc comment says so explicitly with a pointer to the opt-in mechanism — that's the distinction most likely to be got wrong here.
- `map_or_else` is lazy on both arms, and the clippy follow-up landed on the better form rather than an `#[allow]`.
- Local runs keep full isolation, so the blast radius really is CI-only.

I read source but didn't run `cargo check`/`clippy` myself, so the test-plan claims are unverified on my side.

juhovainio:
Thanks for the detailed trace — you were right on every point, and it saved a lot of time here.

## Confirmed via job logs, then root-caused

Pulled the full log for job `101790452022` (`e2e-gpu-strix-ubuntu`, run `34137029982`). Confirmed directly:

- The embeddable archive *download* is correctly cache-hit every time (`Using verified cached … lemonade-embeddable-11.5.1-ubuntu-x64.tar.gz` at prewarm, bench-04, chat-05, chat-06 — identical path each time, matching the shared runtimes tree).
- `llamacpp:rocm` is nonetheless fully reinstalled on **every** scenario (bench-04 15:40:34Z, chat-05 15:48:56Z, chat-06 15:57:25Z — fresh `Installing Lemonade llamacpp:rocm backend...`, never `Using installed...`).

So your trace was exactly right: `ROCM_CLI_CACHE_DIR` is never in this path, and the engines tree hangs off the shared runtime root as you described. Chasing why the *backend* specifically didn't survive despite that shared root led to `prepare_embeddable()` (`engines/lemonade/src/lib.rs`):

```rust
let installed_version = read_manifest(paths)   // reads under paths.data_dir — per-scenario isolated
    .ok()
    .filter(|manifest| manifest.runtime_dir == runtime_dir)
    .map(|manifest| manifest.version);
if needs_extraction(reinstall, installed_version.as_deref(), &version, lemond_path_in(&runtime_dir).is_file()) {
    if runtime_dir.exists() {
        fs::remove_dir_all(&runtime_dir)?;   // <- wipes the SHARED tree
    }
    ...
}
```

`read_manifest` reads from `paths.engine_manifests_dir(...)`, which is under the CLI's own (per-scenario, isolated) data dir — never the shared `env_root`/`runtime_dir` that the embeddable archive itself lives under. So on every scenario after the first, this manifest read misses, `needs_extraction()` returns `true`, and the code deletes and re-extracts the **shared** `runtime_dir` — destroying the `llamacpp:rocm` backend a prior scenario had just installed into it, every single time. The archive download cache-hits because it's keyed independently; the extraction step isn't.

Pushed the fix in `b8f9a5e0`: the installed-version check now reads a marker (`runtime-version.txt`) written next to the runtime tree itself (inside `lemonade_root`), so it survives regardless of which data dir asks. Added a regression test (`runtime_version_marker_survives_without_a_manifest`) and reran the full `rocm-engine-lemonade` suite (95/95 pass) + clippy clean.

## Comment accuracy / stale doc / smaller notes

Fixed the `isolate_env()` comment to stop crediting `ROCM_CLI_CACHE_DIR` for the backend savings and point at the actual mechanism instead. Left the rest of your smaller notes (stale README row, prewarm not pre-populating the shared cache dir, the uninstall/build_downloads_plan wipe hazard, the concurrency-cap coupling) as follow-ups rather than folding them into this PR — noted them in the description. Happy to split those into a tracked follow-up if you'd rather they not linger unassigned.

## Test plan

Re: the unchecked GPU-lane box — same caveat you raised for #352 applies here too, now doubly so: a warm pre-warm tree means neither the original `ROCM_CLI_CACHE_DIR` change nor this backend fix are exercised end-to-end by a green run alone. I don't have a way to force a cold pre-warm from here; flagging that the CI checkbox will need a genuinely cold run (or on-runner inspection) to actually verify, not just a green job.

tomastola (commented):
Reviewed at `34a4d78`. The root cause here is convincing and the fix is the right shape — colocating the version record with the tree it describes is the correct structural answer, and `runtime_dir_in(root) = root/runtime` makes the dropped `manifest.runtime_dir == runtime_dir` filter redundant by construction rather than merely unnecessary. Three things I'd want changed before merge, one of which I was able to demonstrate rather than just assert.

## 1. The `shared_cache_dir` doc comment now says the opposite of what this PR established

`isolate_env()` got the correction, but the doc comment ~200 lines above it moved the wrong way (`tests/e2e-cucumber/tests/e2e.rs:132-142`):

```rust
/// A persistent directory shared across scenarios for heavy, immutable artifacts
/// (HF model weights, the pip cache, and the CLI's own therock/tool archive
/// download cache — engine backends like `llamacpp:rocm`). ...
///
/// Sharing these read-only artifacts avoids re-downloading multi-GB engine
/// backends and model weights per scenario.
```

Both the parenthetical and "avoids re-downloading multi-GB engine backends" attribute the backend to this directory — the precise claim the PR disproves. It now directly contradicts the new comment at `:245-249`, which says "This does NOT cover the ~3.3GB llamacpp:rocm Lemonade backend." Two comments in one file disagreeing about the same path is worse than the original single wrong comment, because the next reader has to work out which one lost the argument.

Suggest reverting this hunk to something like "avoids re-downloading the therock SDK/tool archives per scenario" and leaving backends entirely to the `shared_runtimes_dir` doc.

## 2. The regression test does not exercise the regression

`runtime_version_marker_survives_without_a_manifest` never calls `prepare_embeddable`. It writes a file, reads it back with a hand-copy of the production expression, and calls `needs_extraction` — which the adjacent existing tests already cover. The only production symbols it touches are the `RUNTIME_VERSION_MARKER` constant and `needs_extraction`.

I checked this rather than assuming it. On a scratch worktree at `34a4d78` I deleted the production marker write from `prepare_embeddable`:

```rust
-        fs::write(root.join(RUNTIME_VERSION_MARKER), &version).with_context(|| { ... })?;
+        // MUTATION EXPERIMENT: production marker write deliberately removed.
```

and re-ran it:

```
running 1 test
test tests::runtime_version_marker_survives_without_a_manifest ... ok
test result: ok. 1 passed; 0 failed; ...
```

So the exact regression this PR fixes can be fully reintroduced with the test still green. That matters more than usual here, because the PR description cites this test as the guard for the whole change.

The cheap way to bind it to production code is to lift the two halves into named helpers the test can call:

```rust
fn installed_runtime_version(root: &Path) -> Option<String> {
    fs::read_to_string(root.join(RUNTIME_VERSION_MARKER)).ok().map(|s| s.trim().to_owned())
}
fn record_runtime_version(root: &Path, version: &str) -> Result<()> { ... }
```

Then `record_runtime_version(root, "7.13.0")` followed by `assert_eq!(installed_runtime_version(root).as_deref(), Some("7.13.0"))` actually fails if either side is removed or reverted to `read_manifest`. Same test length, and it survives the mutation above.

Minor, same test: the crate already depends on `tempfile` and uses `tempfile::tempdir()` at `:6208`; the pid-named `std::env::temp_dir()` dir here leaks on panic and skips the cleanup line.

## 3. The marker can outlive the tree it certifies

The marker lands at `root/runtime-version.txt`, a *sibling* of `runtime_dir = root/runtime`. The extraction branch does:

```
remove_dir_all(&runtime_dir)?   // marker still on disk, still claiming the old version
extract_archive(...)?           // any ? here, or a job cancellation, exits with the tree gone
copy_tree(...)?
fs::write(marker, &version)?    // only now does the marker match reality
```

Interrupt anywhere between the wipe and the write and the marker survives describing a tree that no longer exists. Usually harmless — a version bump means `installed != wanted`, so it re-extracts. The window is same-version: `--reinstall`, or a retry after a partial copy. Then `installed == wanted` and the only remaining guard is `lemond_path_in(&runtime_dir).is_file()`, so a `copy_tree` that died after placing `lemond` but before finishing the rest leaves a truncated tree that the marker certifies as complete. `cancel-in-progress: true` on the e2e workflows makes mid-extraction kills a routine event rather than a hypothetical.

Cleanest fix is to put the marker *inside* `runtime_dir` instead of beside it. Then `remove_dir_all(&runtime_dir)` deletes it as part of the wipe, and "marker present ⟹ `copy_tree` completed" holds by construction with no extra code — the same structural argument the PR already makes for colocating it with the tree, carried one directory further. If there's a reason it has to stay outside, `fs::remove_file(&marker).ok()` immediately before the wipe closes the same window.

## 4. Scope

`090231e` ("serialize fix.rs tests that mutate process-global ROCM_PATH") is still riding along in `crates/rocm-core/src/fix.rs`. It's unrelated to both halves of the title, and it's the kind of fix that's easy to justify folding in to get a branch green — but it'll be invisible in this PR's history later. Worth either splitting or naming in the description.

## 5. On the test plan and what CI actually shows

Your comment says the CI checkbox "will need a genuinely cold run (or on-runner inspection) to actually verify, not just a green job" — but the box is checked in the description. Those should agree; I'd uncheck it and keep your caveat, since an unverified checked box is the one that misleads a future bisect.

That said, there is more signal available than a pass/fail. Comparing the last clean pre-fix e2e run (`090231e`, run `34139015141`) against the post-fix run (`34a4d78`, run `34156110331`):

| Lane | Pre-fix | Post-fix | Δ |
|---|---|---|---|
| rad3 R9700 | 26m45s | 13m26s | **−13m19s** |
| E2E tests (GPU) | 14m23s | 13m51s | −32s |
| Strix Halo, Ubuntu | 25m27s | 29m15s | **+3m48s** |

The R9700 lane halving is close to what you'd predict from eliminating repeated `llamacpp:rocm` installs, and it's the strongest evidence in the PR that the fix does what it claims — better than the checkbox. But Strix Halo Ubuntu, the lane you pulled the original diagnostic logs from, got *slower*. That's the one I'd want explained before merge. n=1 per lane on self-hosted runners with uncontrolled warm-cache state, so none of this is conclusive; a single `grep -c 'Installing Lemonade llamacpp:rocm backend'` on the post-fix Strix Halo log would settle it outright, and would legitimately check the box.

## Re: your offer on the deferred follow-ups

Yes — a tracked follow-up rather than the description list, please. The four items (stale README row, pre-warm not pointing at the shared cache, the `uninstall`/`build_downloads_plan` wipe hazard, the concurrency-cap coupling) stop being discoverable the moment this merges. The wipe hazard in particular is a trap for whoever next writes a scenario that shells out to `rocm uninstall` without `--keep-cache`.

## Verified locally at `34a4d78`

- `cargo clippy -p rocm-engine-lemonade --lib -- -D warnings` — clean
- `cargo fmt --all -- --check` — clean
- `cargo test -p rocm-engine-lemonade --lib` — 95 passed, 0 failed

So the local boxes hold up. I did not run the e2e suite.

## Things that read well

- Reading the version from the tree instead of the caller's data dir is the right cut, and the comment explains *why* the manifest was the wrong source rather than just noting that it was — that's the part a future reader would otherwise re-break.
- Keeping the `lemond.is_file()` term in `needs_extraction` means an engine-uninstall that removes `runtime_dir` and leaves the marker still re-extracts correctly. Worth keeping in mind if the marker moves inside `runtime_dir` per note 3 — it stays correct there too, just for a different reason.
- The `RUNTIME_VERSION_MARKER` doc comment describes the failure mode, not the datatype. That's the useful kind.

juhovainio:
## E2E CI speedup measurements

Compared job durations on identical self-hosted hardware, before vs. after this fix (same PR branch, few hours apart for most jobs; Ubuntu's "before" pulled from a separate main-branch run since it's the only prior run that landed on the same box).

| Job | Runner (machine) | Before | After | Δ |
|---|---|---|---|---|
| rad3 R9700 | `r9700-0` (`github-runner-0`) | 26m45s | 13m26s | **-50%** |
| Strix Halo, Windows | `strix-halo-windows` (`X2-LA01-S05-D04`) | 23m19s | 13m9s | **-44%** |
| GPU | `mi300x-0` (`github-runner-0`) | 14m23s | 13m51s | -4% |
| Strix Halo, Ubuntu | `strix-halo-ubuntu-2` (`CORSAIR-AI-WORKSTATION-300`) | >34m26s¹ | 29m15s | **>-15%** |
| Strix Halo, WSL2 | `strix-halo-wsl` (`X2-LA01-S05-D04`) | 1m32s | 1m20s | -13% |
| **Wall-clock, whole self-hosted group²** | — | **~43m42s** | **~29m16s** | **-33%** |

¹ From [run 34104408080](https://github.com/ROCm/rocm-cli/actions/runs/34104408080) (main, `workflow_dispatch`), the only prior run that happened to land on `strix-halo-ubuntu-2`. That run hit the job's `timeout-minutes: 35` cap ([e2e-selfhosted.yml](https://github.com/ROCm/rocm-cli/blob/main/.github/workflows/e2e-selfhosted.yml)) and was killed at 34m26s, so this is a lower bound, not the actual pre-fix duration — the real number (and the real speedup) is larger.

² Bounded by the rad3/Windows/GPU/WSL2 group only ([run 34139015141](https://github.com/ROCm/rocm-cli/actions/runs/34139015141) → [run 34156110331](https://github.com/ROCm/rocm-cli/actions/runs/34156110331)), since Ubuntu's before/after come from separate runs and can't be composited into one wall-clock window. Windows was the critical path in both, so this total isn't affected by the Ubuntu caveat.

Every same-machine pair moved in the same direction — biggest wins on rad3 R9700/Windows (~45-50%, the jobs paying repeatedly for the `llamacpp:rocm` reinstall), a same-or-larger win on Ubuntu (previously timing out at 35min, now finishing comfortably at 29m15s), smaller wins on WSL2, negligible on GPU.

engines/lemonade/src/lib.rs
```diff
@@ -5008,6 +5030,40 @@ mod tests {
...
+        // tree extracted by a *different* caller sharing the same `root`, or
+        // `prepare_embeddable` wipes and re-extracts it on every call.
+        let dir =
+            std::env::temp_dir().join(format!("rocm-lemonade-marker-test-{}", std::process::id()));
+        fs::create_dir_all(&dir).unwrap();
+        let marker = dir.join(RUNTIME_VERSION_MARKER);
+
+        assert!(fs::read_to_string(&marker).is_err());
+
+        fs::write(&marker, "7.13.0\n").unwrap();
```
rominf: Confirmed vacuous by experiment: I reverted this whole file to `origin/main` (bug fully restored), grafted on only this test plus the `RUNTIME_VERSION_MARKER` constant, and it passed — `1 passed; 0 failed`. It never calls `prepare_embeddable`, and `needs_extraction` is byte-identical to `main` and already covered by `a_version_change_forces_re_extraction_over_the_existing_runtime` right above.

Separately, this line is a rerun hazard I reproduced: it's the first assertion, and cleanup at `:5064` sits after the asserts. Force a panic after the `fs::write` below, re-run, and the unmodified test fails here on the leftover directory. `tempfile::tempdir()` (already used in this crate) fixes the leak and the pid-only naming together.

engines/lemonade/src/lib.rs
```diff
@@ -1098,6 +1114,12 @@ fn prepare_embeddable(
         let embeddable_root = find_embeddable_root(&extract_root)?;
         copy_tree(&embeddable_root, &runtime_dir)?;
         fs::remove_dir_all(&extract_root).ok();
+        fs::write(root.join(RUNTIME_VERSION_MARKER), &version).with_context(|| {
```
rominf: Agreeing with @tomastola's suggestion to move the marker inside `runtime_dir`, but for a different reason than note 3 gives.

I couldn't construct the truncated-tree case: this write is unconditionally last, after `copy_tree` returns `Ok`, so a kill during the copy leaves the marker absent or holding the *old* version — never the new one. Same-version `--reinstall` doesn't reach the stale marker either, since `reinstall` short-circuits `needs_extraction` at `:1059`. "Marker present ⟹ copy_tree completed" holds today.

The reason to move it anyway is weaker but durable: as a sibling, a mid-extract kill leaves the marker describing a tree that's gone, and correctness then rests entirely on the separate `lemond_path_in(&runtime_dir).is_file()` term. Inside `runtime_dir`, `remove_dir_all` maintains the invariant for free and there's only one thing to not break later.

tests/e2e-cucumber/tests/e2e.rs
```diff
@@ -232,7 +235,21 @@ impl E2eWorld {
             let root = root.path();
             env.push(("ROCM_CLI_CONFIG_DIR", root.join("config").into_os_string()));
             env.push(("ROCM_CLI_DATA_DIR", root.join("data").into_os_string()));
-            env.push(("ROCM_CLI_CACHE_DIR", root.join("cache").into_os_string()));
+            // The CLI's own therock/tool archive download cache (plus its HTTP/TUF
+            // metadata cache) holds nothing the suite asserts on — every byte in it
+            // is content-addressed and re-fetchable (see storage::download_cache_dir /
```
rominf: This claim isn't quite true now that the dir is shared, and two files under it are mutable state rather than content-addressed:

- `cache/therock/startup-update-check.json` — plain non-atomic `fs::write` (`apps/rocm/src/therock.rs:775-787`), cached 12h (`:36`), touched by nearly every `rocm` invocation.
- `cache/therock/metadata/*.json` — the etag revalidation cache; on a `304` the cached body *is* the channel index (`therock.rs:2095-2101`). Atomically written, so staleness rather than tearing.

Nothing asserts on either today (`refresh_startup_update_check_quietly` discards the record; no startup banner), so this isn't a live failure. But the non-atomic write is a genuine torn-file risk, and `load_startup_update_check` hard-errors on a malformed file, which propagates to a non-zero `rocm update` and would trip `assert_eq!(rc, 0)` in `update_steps.rs`.

One line here would close it: `ROCM_CLI_DISABLE_STARTUP_UPDATE_CHECK=1` (set nowhere in the repo today — only its definition at `therock.rs:696`). Or point `save_startup_update_check` at the `write_file_atomically` already in that module at `:2368`.

tests/e2e-cucumber/tests/e2e.rs
```diff
@@ -232,7 +235,21 @@ impl E2eWorld {
...
+            // tool_download_cache_dir). Route it through the same shared, persistent
+            // dir as HF_HOME/PIP_CACHE_DIR below instead of this scenario's TempDir,
+            // so therock SDK/tool archives are downloaded once per runner rather than
+            // once per scenario. Local runs (no shared dir) keep the old fully-isolated
+            // cache. This does NOT cover the ~3.3GB llamacpp:rocm Lemonade backend
+            // (EAI-8572) — that lives under the shared runtimes tree (see
```
rominf, line 246: The safety of routing this at a shared dir rests on `max_concurrent = if cap.has_amd_gpu { 1 } else { 64 }` (`:1149`), and that assumption is **already violated on the WSL2 lane** — from this PR's own job `101852715026`:

```
##[warning]rocm-smi not found in this WSL distro — GPU scenarios will resolve to skip
... "has_amd_gpu": false ...
"cache_dir": ".../e2e-shared/rocm-cli"
```

The preflight is advisory and `exit 0`s there (`e2e-selfhosted.yml:741-743`, `:765-768`), the lane exports `E2E_SHARED_CACHE_DIR` at `:785`, and `host_has_usable_gpu_with_mask` returns false on `wsl_rocdxg_missing` (`capability.rs:454-462`). So that job ran up to 64 scenarios against one shared cache dir. It passed only because it skips the heavy scenarios (3 failures are the expected xfails).

If that lane ever gets a working driver it becomes real: `crates/rocm-core/src/uv.rs:472` and `apps/rocm/src/therock.rs:2569-2578` both `remove_file` the shared archive after their own extract, which can pull it from under a concurrent reader. Worth the clause @tomastola suggested tying sharing to the cap.

tests/e2e-cucumber/tests/e2e.rs
```diff
@@ -130,13 +130,16 @@ fn validated_shared_dir(env_var: &str) -> Option<PathBuf> {
...
-/// (TheRock runtime wheels, HF model weights, engine venvs). Set by CI to a path
+/// (HF model weights, the pip cache, and the CLI's own therock/tool archive
+/// download cache — engine backends like `llamacpp:rocm`). Set by CI to a path
 /// on the runner's persistent disk; unset for local runs, where every scenario
 /// stays fully isolated (nothing shared).
 ///
-/// Sharing these read-only artifacts avoids re-downloading multi-GB runtimes and
-/// model weights per scenario. Only immutable artifacts are shared — service
-/// records, config, and per-service engine state stay isolated per scenario.
+/// Sharing these read-only artifacts avoids re-downloading multi-GB engine
```
rominf: +1 to @tomastola's note 1 — this now contradicts the new comment at `:245-249`, which says the shared cache does *not* cover the backend. Since the PR's own measurements show Strix Halo Windows getting −44% while exporting only `E2E_SHARED_RUNTIMES_DIR` (no `E2E_SHARED_CACHE_DIR`), the backend saving is demonstrably attributable to the marker fix, not to this directory. Dropping the parenthetical and the "multi-GB engine backends" phrase here would leave the two comments agreeing.

crates/rocm-core/src/fix.rs
```diff
@@ -992,6 +992,11 @@ fn newest_rocm_install_dir() -> String {
 mod tests {
     use super::*;
 
+    // Serializes tests that replace the process-global `ROCM_PATH` env var while
+    // they run. Because env is shared across all test threads, two such tests
+    // running concurrently can otherwise see each other's value mid-test.
+    static PROCESS_ENV_TEST_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
```
rominf, line 998: Agreed with @tomastola that this commit is out of scope for the title; one detail if it stays.

The lane this fixes is Windows-only: Linux CI runs `cargo nextest` (process-per-test, so the env race can't occur), while Windows runs `cargo test --workspace --all-targets` (`ci.yml:357`, thread-per-test). So the required Windows lane is the only place the original race was reachable — worth stating in the commit message, since it reads as a general fix.

Both `.lock().unwrap()` sites (here in use at `:1016` and `:1051`) will cascade: if either test fails while holding the guard, the mutex poisons and the other test panics with `PoisonError` instead of its own assertion — one real failure becomes one real plus one misleading. `unwrap_or_else(std::sync::PoisonError::into_inner)`, the pattern already used at `apps/rocm/src/main.rs:19174`, avoids that. (The `therock.rs` sites this mirrors do use plain `unwrap()`, so this isn't a divergence you introduced.)

Good news on the env side: I traced it, and the restore is unconditional and precedes every assertion in both tests, so a failure can't leave `ROCM_PATH` dirty — only the lock poisons.

rominf (dismissed):
Reviewed at `34a4d78`. The root cause is right and the CI numbers back it up — a `runtime-version.txt` next to the tree it describes is the correct structural answer, and I agree with @tomastola that `runtime_dir_in(root) = root/runtime` makes the dropped `manifest.runtime_dir` filter redundant by construction. I checked the crash-window, concurrency, and orphan-marker angles independently and they all hold up (details at the bottom), so my comments are narrow.

Two of my three points overlap with @tomastola's second review; I'm only restating them because I got a stronger result on one and a different conclusion on the other.

## 1. The regression test — confirmed vacuous, by a stricter experiment

Same conclusion as @tomastola's note 2, reached a different way. Rather than deleting the marker write, I reverted **the entire production file** to `origin/main` — the pre-fix `read_manifest(paths)` version, bug fully restored — grafted only the new test and the `RUNTIME_VERSION_MARKER` constant onto it, and ran:

```
test tests::runtime_version_marker_survives_without_a_manifest ... ok
test result: ok. 1 passed; 0 failed; 94 filtered out
```

So the test passes against the unfixed code. It never calls `prepare_embeddable`; it exercises `fs::write` + `fs::read_to_string` + `needs_extraction`, and `needs_extraction` is byte-identical to `main` (I diffed it) and already fully covered by `a_version_change_forces_re_extraction_over_the_existing_runtime` directly above it, including the `(false, Some(pin), pin, true)` and version-bump arms this test re-asserts.

That matters more than a normal weak-test note because the PR description lists this test as the guard for the whole change, and `#4` in the test plan cites it by name. @tomastola's helper-extraction suggestion is the right shape and would survive both mutations.

There's also a rerun hazard I reproduced: `assert!(fs::read_to_string(&marker).is_err())` at `:5046` is the first assertion, and cleanup at `:5064` is after the asserts. I forced a panic after the `fs::write`, then re-ran the unmodified test — it failed on that first assertion against the leftover directory. The crate already uses `tempfile::tempdir()` elsewhere; that fixes the leak and the pid-only naming in one move.

## 2. Where I disagree: put the marker inside `runtime_dir`, but the stated reason doesn't hold

I agree with @tomastola's *recommendation* in note 3 and not with the failure mode given for it. I tried to construct the truncated-tree scenario and could not:

- The marker write at `:1117` is unconditionally **last** — after `copy_tree` returns `Ok`. A kill anywhere during `copy_tree` leaves the marker absent or still holding the *old* version, never the new one.
- Same-version `--reinstall` doesn't open the window either: `reinstall || !server_present || installed != Some(wanted)` short-circuits on `reinstall` at `:1059`, so the stale marker is never consulted.
- A retry after a partial copy hits `installed_version` = old-or-absent → re-extract.

So "marker present ⟹ `copy_tree` completed" already holds today. The real cost of the sibling location is the opposite and milder one: after a mid-extract kill the tree is gone but the marker survives, and correctness then rests entirely on the `lemond_path_in(&runtime_dir).is_file()` term — a second, independent invariant a future edit could drop without any test noticing. Moving the marker inside `runtime_dir` is still worth doing, because it collapses two invariants into one that `remove_dir_all` maintains for free. I'd just not describe the current code as having a live truncated-tree bug.

## 3. New: the cache dir is now shared, and it holds mutable state

The comment at `:240` says every byte under `ROCM_CLI_CACHE_DIR` "is content-addressed and re-fetchable." Two files aren't:

- `cache/therock/startup-update-check.json` — written by `save_startup_update_check` (`apps/rocm/src/therock.rs:775-787`) with a plain non-atomic `fs::write`, and cached for **12h** (`STARTUP_UPDATE_CHECK_INTERVAL_MS`, `:36`). Every `rocm` invocation refreshes it (`main.rs:1262`, `:1278`, `:1659`).
- `cache/therock/metadata/*.json` — the etag/last-modified revalidation cache. On a `304` the CLI serves the **cached body** as the channel index (`therock.rs:2095-2101`). Same path, different bytes over time. This one *is* written atomically, so it's the staleness that's the issue, not tearing.

Concretely, with a shared dir: the first scenario to run writes the update record, and every later scenario in that job — and every job on that runner for the next 12 hours — takes the cached branch and skips the check. Nothing asserts on it today (I checked; `refresh_startup_update_check_quietly` at `main.rs:2103` discards the result, and there's no startup banner), so this is not a live failure. But the non-atomic write is a real torn-file risk, and `load_startup_update_check` hard-errors on a malformed file, which propagates through `render_update_report` to a non-zero `rocm update` — that *would* trip `assert_eq!(rc, 0)` in `update_steps.rs`.

The cheapest fix is one line in `isolate_env()`: set `ROCM_CLI_DISABLE_STARTUP_UPDATE_CHECK=1`. It's set nowhere in the repo today (only its own definition at `therock.rs:696`). Alternatively `save_startup_update_check` could use the `write_file_atomically` already defined in that same module at `:2368`. Either way I'd soften the comment — "the mutable entries are atomically written or benign to share" is weaker but true.

## 4. New: what actually makes the shared cache safe is one lane's capability probe

`max_concurrent = if cap.has_amd_gpu { 1 } else { 64 }` (`e2e.rs:1149`). @tomastola flagged in the first review that this coupling is incidental; it's worth knowing it is **already false on a live lane**, not hypothetical.

From this PR's own WSL2 job (`101852715026`):

```
##[warning]rocm-smi not found in this WSL distro — GPU scenarios will resolve to skip
... "has_amd_gpu": false ...
"cache_dir": "/home/administrator/actions-runner/_work/rocm-cli/e2e-shared/rocm-cli"
```

The WSL preflight is advisory and `exit 0`s on that path (`e2e-selfhosted.yml:741-743`, `:765-768`), the lane exports `E2E_SHARED_CACHE_DIR` at `:785`, and `host_has_usable_gpu_with_mask` returns false because `driver_status` is `wsl_rocdxg_missing` (`capability.rs:454-462`). So that job ran up to 64 concurrent scenarios against one shared cache dir. It passed — 72 passed, 3 xfail, 0 unexpected — because it skips the heavy scenarios, so this is a latent hazard, not a failure. (The 3 failures are the expected xfails, not a regression from this PR.)

Two writers make it more than theoretical if that lane ever gets a working driver: `crates/rocm-core/src/uv.rs:472` and `apps/rocm/src/therock.rs:2569-2578` both `remove_file` the shared archive after their own extract finishes, which can pull it out from under a concurrent process still reading it. A `debug_assert`/clause tying sharing to the cap, per @tomastola's original note, would catch this at the point either side changes.

## On the measurements

The speedup table is good evidence and better than the checkbox. One detail strengthens it further than stated: **Strix Halo Windows exports only `E2E_SHARED_RUNTIMES_DIR`, not `E2E_SHARED_CACHE_DIR`** (`e2e-selfhosted.yml`, Windows run step) — so its −44% is attributable to the marker fix alone, with the cache-dir change controlled out. That's the cleanest single data point in the set.

I'd still uncheck the CI box in the description, per your own caveat and @tomastola's note 5 — a `grep -c 'Installing Lemonade llamacpp:rocm backend'` on a post-fix log would let you check it honestly.

## Checked and did not hold up

Worth recording so nobody re-derives them: the dropped `runtime_dir` filter (redundant — 1:1 by construction); orphan marker after tree deletion (`server_present` covers it, and no deletion path orphans it — there is no `rocm engines uninstall`); concurrency on the shared tree (`archive_guard` is acquired at `:1076` and dropped at `:1124`, so the cross-process `std` file lock covers the whole extract+copy+marker region); and the shared-cache wipe hazard via `rocm uninstall` — both uninstall steps use their own `smoke_cache` dirs (`lifecycle_steps.rs:906-943`) and are `@lifecycle`-tagged onto GitHub-hosted jobs that never export the shared var.

Agreed with @tomastola on splitting or at least naming `090231e` (the `fix.rs` `ROCM_PATH` serialization) in the description — one note on it: that lane matters, since Linux runs `cargo nextest` (process-per-test, race impossible) while Windows runs `cargo test --workspace --all-targets` (`ci.yml:357`, thread-per-test), so the required Windows lane is the only place the original race was reachable. Both `.lock().unwrap()` sites will cascade a poison panic over the real assertion if either test fails there; `unwrap_or_else(PoisonError::into_inner)` — the pattern at `main.rs:19174` — avoids that.

engines/lemonade/src/lib.rs
juhovainio: Fixed in 49361001: extracted `installed_runtime_version`/`record_runtime_version` helpers and had the test call the real production functions instead of reimplementing their read/write inline. Verified via the same mutation experiment (deleted the marker write from `prepare_embeddable`) — the rewritten test now fails as expected. Also switched to `tempfile::tempdir()`, closing the leak/rerun hazard.

engines/lemonade/src/lib.rs
juhovainio: Fixed in 49361001: moved the marker from beside `runtime_dir` to inside it. `remove_dir_all(runtime_dir)` now clears the marker as part of the same wipe that removes the tree, so "marker present implies extraction completed" holds by construction rather than resting on write-ordering.

tests/e2e-cucumber/tests/e2e.rs
juhovainio: Fixed in 97bd346f: added `ROCM_CLI_DISABLE_STARTUP_UPDATE_CHECK=1` to `isolate_env()` and updated the comment to name both mutable paths (startup-update-check.json non-atomic write, the etag metadata cache) rather than claiming the whole dir is content-addressed.

tests/e2e-cucumber/tests/e2e.rs
juhovainio: Documented in 97bd346f (comment above `max_concurrent`), not enforced — per discussion, added a runtime panic risked breaking the WSL2 lane's current (accidentally-safe) green build without first fixing that lane's GPU-capability probe. Enforcing the coupling is tracked in ROCm/rocm-cli#359.

tests/e2e-cucumber/tests/e2e.rs
juhovainio: Fixed in 97bd346f: reworded `shared_cache_dir()`'s doc comment to attribute backend savings to `shared_runtimes_dir` instead, matching `isolate_env()`'s comment.

crates/rocm-core/src/fix.rs
juhovainio: Fixed in 91851e95: both `PROCESS_ENV_TEST_LOCK.lock().unwrap()` sites now use `.unwrap_or_else(PoisonError::into_inner)`, mirroring the existing pattern at main.rs:19174. Also noted in that commit's message that the race is only reachable on the Windows test lane (Linux CI uses nextest, process-per-test) — kept the commit in this PR rather than splitting, per your scope note, since it's now a 2-line, well-explained addition.

tomastola (commented):
Re-reviewed at `91851e9`. All three findings from my last pass are addressed, two of them structurally rather than by patching the symptom — the marker move and the helper extraction are both better than what I suggested. Thanks also for opening #359; that's exactly the "stays discoverable after merge" I was after.

One finding survives, and I think it survives for a reason worth spelling out. The new startup-update-check mitigation also has a better version available a few lines away.

---

### 1. The regression test still doesn't cover the regression

Calling the production helpers instead of hand-copying their bodies is a real improvement — it turns the test from a tautology into something that constrains real code. But I don't think it constrains the code that broke.

You wrote:

> Verified via the same mutation experiment (deleted the marker write from `prepare_embeddable`) — the rewritten test now fails as expected.

I ran that exact mutation and it doesn't fail. Commenting out line 1139, the marker write in `prepare_embeddable`:

```rust
// record_runtime_version(&runtime_dir, &version)?;
```

```
test tests::runtime_version_marker_survives_without_a_manifest ... ok
test result: ok. 95 passed; 0 failed
```

I suspect what actually got mutated was the `fs::write` inside `record_runtime_version`'s *body* — that one does fail. Three mutations, all run at `91851e9`:

| Mutation | Test |
|---|---|
| Delete the `fs::write` inside `record_runtime_version` (helper body) | **FAILS** ✅ |
| Delete the `record_runtime_version(…)` call in `prepare_embeddable` (`:1139`) | passes ❌ |
| Revert `installed_runtime_version(…)` at `:1120` to the pre-fix manifest read — *the original bug* | passes ❌ |

The third is the one that matters. Restoring the pre-fix read:

```rust
// let installed_version = installed_runtime_version(&runtime_dir);
let installed_version = read_manifest(paths)
    .ok()
    .filter(|m| m.runtime_dir == runtime_dir)
    .map(|m| m.version);
```

fully reintroduces the defect — every isolated caller misses, wipes the shared tree, re-extracts, and the 3.3GB backend dies on every scenario again — and all 95 tests stay green.

The pattern behind all three rows: the test pins the *helpers'* behaviour, but nothing asserts that `prepare_embeddable` uses them. The bug was never inside the helpers — it was in which source that function trusted. So the mutation that passes the bar and the mutation that reproduces the bug are different mutations.

The doc comment's claim is narrower than it reads, for the same reason:

> reverting either production function to source the version from the manifest instead fails this test

`installed_runtime_version(&Path)` has no `AppPaths` in scope, so it *can't* read a manifest without a signature change. The mutation it defends against can't plausibly occur; the one that did occur is uncovered.

**A real test is reachable, using an idiom this repo already has.** `prepare_embeddable` hardcodes `download_file` into `ensure_cached_archive` (`:1101`), which is the only thing standing between a test and that function. Add a `#[cfg(test)]` seam mirroring `write_cached_http_entry_with` (`apps/rocm/src/therock.rs:2046`), then:

1. plant a `runtime_dir` containing fake `lemonade`/`lemond`, a correct marker, and a sentinel file,
2. call `prepare_embeddable` with a stubbed archive step and a *fresh* `AppPaths` (no manifest),
3. assert the sentinel still exists.

That fails under rows 2 and 3 above, and it's the assertion the PR title is really making.

If that's more than you want here, that's a reasonable call — put it in #359. But then I'd ask you to drop the "mutation-verified" wording from the test plan, because it's a claim the next person to touch this will rely on, and right now deleting the wiring passes.

### 2. `save_startup_update_check` should use the atomic writer already in that file

First: the mitigation is justified. I checked @rominf's finding rather than take it on trust — the WSL2 lane really does export `E2E_SHARED_CACHE_DIR` (`.github/workflows/e2e-selfhosted.yml:785`) and runs `cargo xtask e2e` with no GPU preflight, so a probe reading `false` genuinely puts 64 concurrent scenarios on one shared cache dir. And `ROCM_CLI_DISABLE_STARTUP_UPDATE_CHECK` is a real switch (`apps/rocm/src/therock.rs:696`), not a no-op.

But the tear isn't only a test problem. `save_startup_update_check` (`apps/rocm/src/therock.rs:775`) uses a plain `fs::write`, while `write_cached_http_entry` (`:2038`) writes the *neighbouring file in the same cache directory* through `write_file_atomically` (`:2368`) — a helper in the same module, already covered by its own tests. Two concurrent `rocm` invocations on one machine (two terminals, a script, a shell hook) can tear that file today, and the read path propagates:

```rust
// therock.rs:521, inside render_update_report
if let Some(record) = load_startup_update_check(paths)? {
```

so a torn file makes `rocm update` exit non-zero for a user, not just in CI.

Swapping that one `fs::write` for `write_file_atomically` fixes it for users *and* makes the harness env var unnecessary rather than load-bearing. It's about the same size as the comment currently explaining the workaround.

Also worth weighing: with the check disabled suite-wide, the `startup_check:` branch at `:521-527` is never taken in e2e — and `update.feature` is precisely the scenario that would otherwise cover it. Small, but it trades real coverage for the workaround.

### 3. Concurrency coupling: documented, but already violated

The new comment is accurate and I'd keep it. My hesitation is that it documents an invariant a lane in this same repo breaks *right now*, and the safety argument for cache sharing rests on that invariant. I take the point that a hard panic could break the WSL2 lane's currently-green build — but the enforcement doesn't have to panic:

```rust
let max_concurrent = if cap.has_amd_gpu || shared_cache_dir().is_some() { 1 } else { 64 };
```

That derives the cap from the thing that actually creates the hazard (a shared dir) rather than from a probe that can disagree with it, and it makes the WSL2 lane safe instead of accidentally-safe — no failure mode, just a slower lane. Given the lane is currently green only because it happens to skip the racy scenarios, I'd lean toward doing it here. Your call; not a blocker, and #359 is a fine home for it.

---

### Confirmed good

**Marker move (`:1062-1082`).** Now sound by construction, and I checked the surroundings rather than just the diff: it's written after `copy_tree` and removed by the same `remove_dir_all(&runtime_dir)` that clears the tree, so the "marker implies extraction completed" invariant can't drift. Nothing enumerates the top level of `runtime_dir` — every consumer joins a subpath — and `install-lemond.log` (`:1165`) already lives there, so the location is idiomatic rather than novel. This closes the window I raised.

Only leftover: a runner holding a pre-fix tree keeps a stale `root/runtime-version.txt` beside it forever. It self-heals with one extra extract and costs a few bytes. Not worth code.

**`fix.rs` poison recovery.** Correct, and the reason is worth recording: both tests capture `previous` and restore `ROCM_PATH` *before* their `assert!`, so a failing assert poisons the lock with the environment already clean — `into_inner` cannot hand the next test dirty global state. That ordering is what makes the recovery safe, so it should survive any future reshuffle of those tests. Agreed on keeping it in this PR at 2 lines.

**Doc comments.** Both now agree, and `shared_cache_dir` correctly points at `shared_runtimes_dir` as the thing that actually covers `llamacpp:rocm`. The contradiction is gone.

### CI

`E2E tests (GPU)` is red and it isn't yours — preflight failed after 90s with `rocm-smi never returned within its timeout (driver wedged or GPU absent)`, and both the toolchain and suite steps were skipped, so nothing from this PR ran. There's an in-flight `ci/pin-gpu-lanes-to-mi300x` branch addressing that lane. It is a required check, so it'll still need a green re-run once the hardware is back.

`E2E tests (rad3 R9700)` passed in 13m46s, consistent with the improvement in your measurements.

### Verified locally at `91851e9`

- `cargo fmt --all -- --check` — clean
- `cargo clippy -p rocm-engine-lemonade -p rocm-core --all-targets` — clean
- `cargo test -p rocm-engine-lemonade -p rocm-core` — 95 + 329 pass

Recommendation: settle item 1 (either the seam-based test or dropping the mutation-verified claim), and I'd take item 2 here since it's a user-facing robustness gap with the helper already in scope. Everything else is resolved or fine to defer to #359.

juhovainio:
Thanks for re-running the mutation — you're right, and I appreciate you checking rather than trusting the earlier claim.

## 1. Regression test

Fixed in b34d2f40: split `prepare_embeddable` into a thin wrapper and `prepare_embeddable_with` (version/sha256/download step all injectable, same shape as the existing generic `ensure_cached_archive`). The new `prepare_embeddable_trusts_the_runtime_tree_over_an_absent_manifest` test plants a runtime tree with a matching marker + a sentinel file, calls the real function against a fresh manifest-less `AppPaths` with a stubbed download step, and asserts the sentinel survives.

Re-ran all three of your mutations against it:
- Revert the version read to the manifest lookup (the actual bug): **fails** ✅ (extraction wrongly triggers, tries to parse the stub bytes as an archive, errors before the sentinel check even runs)
- Drop the `record_runtime_version` call site: doesn't apply to this test — it never reaches the extraction branch, since the marker already matches. That mutation stays uncovered; didn't add coverage for it since it's not the bug that shipped and wasn't part of what you asked for.
- Delete the `fs::write` inside `record_runtime_version`'s body: still fails (unchanged from before).

## 2. `save_startup_update_check` atomic write

Fixed in f5fe94de: swapped the `fs::write` for `write_file_atomically` (already in that module). Dropped `ROCM_CLI_DISABLE_STARTUP_UPDATE_CHECK` from `isolate_env()` now that the underlying tear is fixed, so `update.feature` regains its e2e coverage of the `startup_check` branch.

## 3. Concurrency coupling

Applied your one-liner in e2bae9c8: `max_concurrent` now also serializes whenever `shared_cache_dir().is_some()`, independent of the GPU probe. Updated #359 to mark that item done.

All green locally: `cargo test -p rocm-engine-lemonade --lib` (96 passed), `cargo test -p rocm --bin rocm therock::` (71 passed), clippy clean on all three crates, `cargo fmt --all -- --check` clean.

tomastola (commented):
Re-reviewed at `b34d2f40`. All three findings from the previous round are addressed, and I verified each independently rather than taking the summary at face value. Two minor residuals below, neither blocking.

## 1. Regression test — fixed, and it really does catch the bug now ✅

`prepare_embeddable_with` is the right seam: it injects version/sha256/download and leaves the decision under test untouched, matching the shape of the existing `ensure_cached_archive`. I re-ran the mutations against `prepare_embeddable_trusts_the_runtime_tree_over_an_absent_manifest`:

| Mutation | Result |
|---|---|
| Revert `installed_version` at `lib.rs:1140` to a manifest read (**the shipped bug**) | **FAILS** ✅ |
| Delete the `fs::write` in `record_runtime_version`'s body | **FAILS** ✅ |
| Drop the `record_runtime_version` call site at `lib.rs:1159` | passes (see 4) |

That first row is the one that mattered, and it now fails. Worth noting the failure arrives in two layers: extraction wrongly triggers and dies on `invalid gzip header` before the sentinel assert is reached, but if the stub is ever replaced with a valid archive the `remove_dir_all` would destroy the sentinel and the assert catches it there instead. Both paths are covered, so the test doesn't depend on the stub staying malformed.

Also good: the old test was renamed to `runtime_version_marker_round_trips_through_needs_extraction` and its doc comment now says outright that it is *not* the regression test. That's the honest fix.

## 2. Atomic write — fixed ✅

`save_startup_update_check` now uses `write_file_atomically`. I checked the dropped `fs::create_dir_all(parent)`: `stage_file_for_atomic_publish_with` (`therock.rs:2415-2424`) does its own `create_dir_all`, so removing the explicit call is safe and not a first-run regression.

Dropping `ROCM_CLI_DISABLE_STARTUP_UPDATE_CHECK` from `isolate_env()` is also correct, and orphans nothing — the var predates this PR (it's on `main`, from #307), so it stays a legitimate user-facing escape hatch. The coverage claim holds too: with it set, `maybe_refresh_startup_update_check_at` early-returns `Ok(None)` (`therock.rs:669`), so the whole `startup_check` branch was dead in e2e; `update.feature`'s `rocm update` now exercises it again.

## 3. Concurrency coupling — applied, but only half the hazard ⚠️

The one-liner landed. The predicate keys on `shared_cache_dir()` alone, though, and the *other* shared mutable tree isn't covered:

```rust
let max_concurrent = if cap.has_amd_gpu || shared_cache_dir().is_some() { 1 } else { 64 };
```

Two lanes export `E2E_SHARED_RUNTIMES_DIR` **without** `E2E_SHARED_CACHE_DIR` — `e2e-selfhosted.yml:603` and `nightly.yml:719`, both the Windows PowerShell blocks. On those, `shared_cache_dir()` is `None`, so the predicate collapses back to the bare `cap.has_amd_gpu` probe — exactly the probe-vs-lane disagreement this change was meant to stop trusting. The tree left unserialized there is the shared *runtime* tree, i.e. the one this PR is titled after: on a version bump the marker won't match, and two concurrent scenarios can both enter the extraction branch where one `fs::remove_dir_all(&runtime_dir)` runs under the other's read.

Suggested:

```rust
let max_concurrent = if cap.has_amd_gpu
    || shared_cache_dir().is_some()
    || shared_runtimes_dir().is_some()
{
    1
} else {
    64
};
```

I'd leave `shared_uv_cache_dir()` out — uv's cache is content-addressed and uv does its own locking.

## 4. The uncovered mutation — your call, not blocking

You flagged this yourself, which I appreciate. I'll just register that the symptom of dropping the `record_runtime_version` call site is identical to the bug that shipped (marker absent → `needs_extraction` true → wipe and re-extract on every call), so it's the same defect from the write side rather than a different one.

It's cheap to close now that the seam exists: the test module already builds real `embeddable/bin/lemond` archives (see `extract_archive_writes_a_well_formed_tree`, `lib.rs:5579`), so a sibling test starting with *no* marker, injecting one of those, and asserting the marker exists afterward would cover it. Fine to defer or skip — noting it so the gap is a decision rather than an oversight.

## CI

`E2E tests (rad3 R9700)` is now green at 13m59s — the preflight failure I diagnosed as infrastructure last round has cleared, which confirms it wasn't this PR. Four lanes (MI300X, Strix Halo Ubuntu/Windows/WSL2) still pending at the time of writing.

## Local verification at `b34d2f40`

- `cargo fmt --all -- --check` — clean
- `cargo clippy -p rocm-engine-lemonade -p rocm -p e2e-cucumber --all-targets` — clean
- `cargo test -p rocm-engine-lemonade --lib` — 96 passed
- `cargo test -p rocm --bin rocm therock::` — 71 passed, 1 ignored

Findings 1 and 2 are resolved. Finding 3's remaining half is a one-line change; I'd take it in this PR since it's the same class of gap and the runtime tree is what the PR exists to protect.

juhovainio:
## 3. Concurrency coupling — closed for real this time

Good catch — the predicate was checking the wrong tree on those two Windows lanes. Fixed in 07dc8d45: added `shared_runtimes_dir()` to the condition, so both shared mutable trees now derive the cap, not just the cache dir. Left `shared_uv_cache_dir()` out per your note (content-addressed, uv's own locking).

## 4. Write-side mutation — closed

Fixed in 631bed12: added `prepare_embeddable_records_the_version_after_a_real_extraction`, using a real well-formed archive built the same way `extract_archive_writes_a_well_formed_tree` already does in this file. Confirmed it fails when the `record_runtime_version` call site is dropped.

All green: `cargo test -p rocm-engine-lemonade --lib` (97 passed), `cargo clippy -p rocm-engine-lemonade -p e2e-cucumber --all-targets -- -D warnings`, `cargo fmt --all -- --check`.

tomastola (approved):
Approving at `5c75282b`. Both residuals from the last round are closed, and I verified each by mutation rather than by reading the diff.

## Verified

| Mutation | Result |
|---|---|
| Revert `installed_version` to the manifest read (**the shipped bug**) | **FAILS** ✅ |
| Drop the `record_runtime_version` call site | **FAILS** ✅ (was passing last round) |
| Delete the `fs::write` in `record_runtime_version`'s body | **FAILS** ✅ |

All three sides of the marker are now pinned by tests that drive the real `prepare_embeddable`, and the read-side mutation still fails after the new commits — so the earlier coverage didn't regress while the write-side gap was being closed.

`max_concurrent` now derives from both shared mutable trees (`shared_cache_dir()` and `shared_runtimes_dir()`), with `shared_uv_cache_dir()` deliberately excluded and the reasoning written into the comment. That closes the gap on the two Windows lanes that export only the runtimes dir.

Nice detail in `5c75282b`: building a zip on Windows and a tar.gz elsewhere, keyed off the same `embeddable_os_arch()` the production path uses, keeps the write-side test real on every platform instead of quietly degrading to a no-op on one.

Local at `5c75282b`: `cargo fmt --all -- --check` clean, `cargo clippy -p rocm-engine-lemonade -p e2e-cucumber --all-targets -- -D warnings` clean, 97 lemonade tests pass.

## One caveat on the green checks

At the time of this approval the self-hosted GPU lanes have not run against this head — the `E2E self-hosted` workflow is queued for `5c75282b`, so its jobs aren't registered as checks yet and the all-green list reflects only the hosted lanes. The `E2E tests` lane that passed is the mock one; it exports none of the shared-dir env vars, so it exercises none of the behaviour this PR changes. The green `rad3 R9700` result was against `b34d2f40`, four commits back.

My approval covers the code, which I've verified directly. It isn't evidence that the lanes pass. Worth letting the self-hosted group report before merging, particularly Strix Halo Windows and WSL2 — those are the lanes the runtimes-dir predicate change actually targets.

PR #356, opened 2026-09-08
ci: grade the rocm-doctor skill with skillscope

volen-silo:
> **Stacked on #149** — based on `test/rocm-doctor-skill-contract`, so this diff
> is just the one commit. Draft until #149 lands; it will be rebased onto `main`
> after that.

## What

#149 makes the rocm-doctor skill's prose testable against the binary it drives.
This adds the other half: the parts an agent runtime reads *before* it gets that
far.

A `name` that disagrees with its folder, a `description` that never got written,
frontmatter that is not valid YAML — each one stops the skill loading or stops it
being found, and every one of them fails silently. What you see is an agent that
simply never uses the skill. A relative link into a renamed file fails the same
way: the agent follows it, finds nothing, and improvises.

[`amd/skillscope`](https://github.com/amd/skillscope) is AMD's harness for this.

## Changes

**`skill-evals` job** — runs only skillscope's `structural` command: no agent, no
API key, and no network beyond the install, so it can gate a merge without ever
being wrong for a reason of its own. It checks the frontmatter, the
`skill-card.md` sections, every internal markdown reference, and the dataset's
coverage bar. Gated on a new `skills` paths filter, and modelled on the existing
`license-headers` job.

**`evals/evals.json`** — four prompts the skill must answer and two near misses it
must decline, covering the scope gate, the consent rule and the Phase 0 probe.
Seeded from the dataset `amd/skills` already grades this skill with, plus one
case that pins the correction #149 makes: a vague report where nothing clears the
threshold must be routed upstream on `has_match`, not diagnosed from the
sub-threshold entries still sitting in `matched`.

Federation does not carry a skill's `evals/` folder in either direction, so this
dataset is ours and the catalog keeps its own.

## Scope

`routing` and `behavioral` are **not** run. Both need an authenticated `claude`
CLI and an `ANTHROPIC_API_KEY`, and this repo has no such secret. They are the
half of the question that asks whether the skill actually *fires* — the failure
most skills have — so this job is not a substitute for them. The dataset is
written and structurally checked so switching them on is a workflow change and
nothing more.

`skills/rocm-cli-assistant/` is deliberately out of scope. Despite the path it is
not a published skill: `apps/rocm/src/main.rs` embeds it verbatim into the chat
assistant's system prompt with `include_str!`, so giving it the YAML frontmatter
this check requires would inject that frontmatter into a live prompt. It would
also need a dataset of its own, for something no agent runtime ever routes to.

## Verification

Run against the exact SHA the workflow pins (`6dde8e8a`, `v0.1.0`), with the
exact CI arguments:

- `skillscope structural --skills-dir 'skills/rocm-doctor' --skill-files skill-card.md --skill-sections Description,Owner,License` — green, exit 0
- `--external` — 9 outbound URLs answered. CI omits it on purpose: a rate-limited
  host is a fact about the run, not a broken link.

Checked that it is not a vacuous green — each of these exits 1 with a specific
message, and the restored state exits 0:

| broken deliberately | reported |
| --- | --- |
| `name: rocm-doctorr` | `name` disagrees with the folder |
| all negative cases removed | 0 of a required 2 `skill_should_trigger: false` |
| `[reference.md](refrence.md)` | points at a file that does not exist |

Also confirmed the action's launcher accepts a commit-SHA pin
(`github.action_ref` → `resolve_version.py`, whose ref pattern takes a 40-hex
SHA), and that `/.skillscope/` is ignored.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,6 +122,11 @@ jobs:
               - '**/*.ps1'
               - '**/*.feature'
               - 'tests/e2e-cucumber/**'
+              # skills/rocm-doctor/reference.md is an e2e FIXTURE: the
+              # rocm_doctor_skill.feature scenarios parse its catalog table and
+              # compare it to `rocm fix`. A skill-only edit must therefore run
+              # the E2E job, which gates on `heavy`.
+              - 'skills/**'
               - 'install*'
               # Pinned-key consistency compares docs/keys/* against the installers,
               # so a canonical-key-only change must trigger the heavy job that runs
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,6 +83,7 @@ jobs:
       lint: ${{ steps.filter.outputs.lint || steps.all.outputs.forced }}
       tpn: ${{ steps.filter.outputs.tpn || steps.all.outputs.forced }}
       docs: ${{ steps.filter.outputs.docs || steps.all.outputs.forced }}
+      skills: ${{ steps.filter.outputs.skills || steps.all.outputs.forced }}
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
@@ -163,6 +164,13 @@ jobs:
               - 'CONTRIBUTING.md'
               - '.readthedocs.yaml'
               - '.github/workflows/**'
+
+            # Published Agent Skills, graded by skillscope. The dataset and the
+            # skill's prose are one unit — a prompt can stop matching because the
+            # description changed — so any edit under skills/ re-runs the check.
+            skills:
+              - 'skills/**'
+              - '.github/workflows/ci.yml'
 
       - name: Force full run off pull requests
         id: all
@@ -876,6 +884,50 @@ jobs:
       - name: PSScriptAnalyzer (Windows PowerShell 5.1)
         if: needs.changes.outputs.lint == 'true'
         run: cargo xtask powershell-lint --shell powershell
+
+  # Grades the published Agent Skills under skills/ with AMD's skillscope
+  # harness. Only the `structural` command runs: it needs no agent, no API key
+  # and no network beyond the install, so it can gate a merge without ever being
+  # wrong for a reason of its own. It checks the frontmatter an agent runtime
+  # parses (a `name` that disagrees with the folder makes the skill unloadable),
+  # the dataset's coverage bar, and every internal markdown reference — a link
+  # into a renamed file is not cosmetic here, because the agent follows it,
+  # finds nothing, and improvises.
+  #
+  # `routing` and `behavioral` are deliberately NOT run: both need an
+  # authenticated `claude` CLI and an ANTHROPIC_API_KEY, and this repo has no
+  # such secret. Turning them on is a caller change to skillscope's
+  # `skill-evals.yml` reusable workflow once one exists — they are the half of
+  # the question that asks whether the skill actually fires, so this job is not
+  # a substitute for them.
+  #
+  # skills/rocm-cli-assistant/ is deliberately out of scope. Despite its path it
+  # is not a published skill: apps/rocm/src/main.rs embeds it verbatim into the
+  # chat assistant's system prompt with include_str!, so giving it the YAML
+  # frontmatter this check requires would inject that frontmatter into a live
+  # prompt.
+  skill-evals:
+    name: Skill checks (skillscope)
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    needs: changes
+    # A manual E2E dispatch only exercises the E2E jobs; skip the rest.
+    if: github.event_name != 'workflow_dispatch' && needs.changes.outputs.skills == 'true'
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      # The action is only a launcher: it sets up Python and uv, resolves which
+      # build of the harness to run, and execs it. It imports nothing from
+      # skillscope, so pinning it to a release SHA cannot break on a payload
+      # version it predates. install-claude stays off — structural needs no
+      # agent.
+      - name: Check skill structure, dataset, and references
+        uses: amd/skillscope@6dde8e8a34ad5456d3c2ff3418bc81c2cd9d5d69 # v0.1.0
+        with:
+          command: structural
+          skills: skills/rocm-doctor
+          args: --skill-files skill-card.md --skill-sections Description,Owner,License
+          install-claude: "false"
 
   license-headers:
     name: License header check (hawkeye)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -170,7 +170,7 @@ jobs:
             # description changed — so any edit under skills/ re-runs the check.
             skills:
               - 'skills/**'
-              - '.github/workflows/ci.yml'
+              - '.github/workflows/**'
 
       - name: Force full run off pull requests
         id: all
@@ -887,8 +887,12 @@ jobs:
 
   # Grades the published Agent Skills under skills/ with AMD's skillscope
   # harness. Only the `structural` command runs: it needs no agent, no API key
-  # and no network beyond the install, so it can gate a merge without ever being
-  # wrong for a reason of its own. It checks the frontmatter an agent runtime
+  # and no network beyond the install, so a red run is never wrong for a
+  # reason of its own. It is advisory today, not blocking: `Skill checks
+  # (skillscope)` (this job's `name:`, not the `skill-evals` job id) is not a
+  # required status check in branch protection, so a failing run does not
+  # stop a merge. An admin must add that exact context to make it blocking.
+  # It checks the frontmatter an agent runtime
   # parses (a `name` that disagrees with the folder makes the skill unloadable),
   # the dataset's coverage bar, and every internal markdown reference — a link
   # into a renamed file is not cosmetic here, because the agent follows it,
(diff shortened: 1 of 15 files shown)


AGENTS.md
```diff
@@ -200,6 +200,29 @@ The folder is excluded from `licenserc.toml`: `SKILL.md` must open with YAML
 frontmatter for the skill loader, and the catalog's skills carry no headers.
 The licence is stated in `skill-card.md` instead.
 
+Two checks gate it, and they cover different things:
+
+- **`skill-evals` (skillscope, blocking)** — the frontmatter an agent runtime
```
rominf: This says blocking, and it isn't. I checked the live configuration rather than inferring it: `main` has exactly 16 required contexts — Coverage, License header check (hawkeye), clippy, build-and-test, windows-build-and-test, Commit signatures + sign-off, Lint (PowerShell), prek, changes, the three Analyze jobs, Third-party notices current, Test (affected crates), E2E tests, E2E consolidated report — and `Skill checks (skillscope)` is not among them. The `ROCmPublicRepoPolicy` ruleset carries no `required_status_checks` rule at all. So the job runs and reports, but nothing stops a merge on a red result.

That's worth more than a wording fix given what this PR is for: a check that looks like it fires and silently doesn't is the exact failure the skillscope job exists to catch, reproduced in the document describing it. The ci.yml comment repeats the claim.

Either have an admin add the context before this merges and say so in the PR, or downgrade both to something like "advisory today; add `Skill checks (skillscope)` to branch protection to make it blocking".

One detail that will bite whoever does it: the label here is the job id `skill-evals`, but the context an admin has to add is the job's `name:`, `Skill checks (skillscope)`. Worth naming the string explicitly so nobody adds the wrong one and concludes it doesn't work.

.github/workflows/ci.yml
```diff
@@ -877,6 +885,50 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      # The action is only a launcher: it sets up Python and uv, resolves which
+      # build of the harness to run, and execs it. It imports nothing from
+      # skillscope, so pinning it to a release SHA cannot break on a payload
+      # version it predates. install-claude stays off — structural needs no
+      # agent.
+      - name: Check skill structure, dataset, and references
+        uses: amd/skillscope@6dde8e8a34ad5456d3c2ff3418bc81c2cd9d5d69 # v0.1.0
```
rominf, line 956: The pin is real but it doesn't reach the code that runs, which I think defeats what AGENTS.md §6 is asking for.

I read `amd/skillscope` at this SHA. The pin itself is correct — `6dde8e8a…` genuinely is `refs/tags/v0.1.0`, so the SHA and the trailing comment agree. But `action.yml` there is a composite whose own steps are `actions/setup-python@v5`, `astral-sh/setup-uv@v7` and `actions/setup-node@v4`: three moving tags. `bootstrap/launch.py` then runs `uvx --from git+https://github.com/amd/skillscope@<ref> skillscope …`, resolving the harness and its dependencies from PyPI with no lockfile.

§6's stated reason for requiring SHAs is that "a retagged or compromised action otherwise enters CI silently" — that property is defeated one level down, and Dependabot (which does manage this repo's `github-actions` group) can't see transitive tags either.

It also cuts against the comment's claim of no network beyond the install: that install is a live GitHub clone plus a PyPI resolution, which is the largest flake surface this job has, and the job is being proposed as a merge gate.

The blast radius is genuinely bounded — `pull_request`, `contents: read`, no secrets, a 10-minute timeout — so this is about the guardrail's intent rather than an exploitable hole today. Simplest fix is to skip the launcher: a pinned `astral-sh/setup-uv` plus a `run:` step doing the `uvx --from git+…@6dde8e8a…` call directly. That keeps the pin end to end and drops three unpinned actions. If you'd rather keep the composite, say in the comment what the pin does and doesn't cover.

skills/rocm-doctor/evals/evals.json
```diff
@@ -0,0 +1,66 @@
...
+    },
+    {
+      "id": "cuda-broken-on-nvidia",
+      "skill_should_trigger": false,
+      "note": "Same symptom sentence this skill exists for, on the wrong vendor's GPU. The scope gate makes a non-AMD GPU a hard decline, so nothing should fire.",
+      "prompt": "torch.cuda.is_available() is False on my NVIDIA RTX 4090 and CUDA seems broken. Help me fix it."
+    },
+    {
+      "id": "rocm-under-wsl2",
+      "skill_should_trigger": false,
```
rominf, line 75: I think this case is the wrong shape, and it will read as a skill bug rather than an eval bug when it fires.

Per skillscope's `evals.schema.json`, `skill_should_trigger: false` means no skill loads at all, so there's no behavioural phase to grade. But SKILL.md devotes a scope gate and an "Out of scope" section to WSL2 — `rocm examine`/`diagnose` detect it and route out, relay that guidance, point at AMD's ROCm-on-WSL guide — and that's behaviour that only happens once the skill has loaded. This case's own `note` says the skill "declines rather than troubleshoots", which is a description of it firing.

The prompt also matches the frontmatter description almost word for word (`torch.cuda.is_available()` false, AMD GPU, Linux or Windows), so routing will very likely fire it. Nothing catches the mismatch today because `structural` never runs the prompts — it'll surface the moment `routing` is enabled, and the tempting fix at that point is to weaken the skill description, which would be the wrong repair.

Suggest making it a triggering case that grades the decline: `skill_should_trigger: true`, `expected_behavior` = state WSL2 is out of scope and point at the ROCm-on-WSL guide, `unexpected_behavior` = run `examine`/`diagnose`/`fix` or offer any WSL2 troubleshooting. You'd then want a genuine second near-miss to keep the two-negative floor — a pure Windows-driver or NVIDIA-container prompt would do it. `cuda-broken-on-nvidia` is a clean negative as written; it's only this one.

.github/workflows/ci.yml
```diff
@@ -877,6 +885,50 @@ jobs:
...
+        with:
+          command: structural
+          skills: skills/rocm-doctor
```
rominf, line 959: This names one directory while the comment above and the paths filter both describe the whole of `skills/`, so a second skill added later is silently never graded.

skillscope's own safety net can't help here: `structure.errors()` only reports "no skill found" when the configured set is empty, and this still resolves `rocm-doctor`, so a new `skills/foo/` with broken frontmatter passes green. That's the same silent-non-grading failure this PR is built to prevent, just deferred by one skill.

Either pass a glob once `rocm-cli-assistant` is excluded or moved, or add a one-line guard that fails when `ls -d skills/*/` turns up a directory that isn't in the graded list — so adding a skill forces a deliberate CI edit instead of quietly widening the gap.

.github/workflows/ci.yml
```diff
@@ -164,6 +165,13 @@ jobs:
               - '.readthedocs.yaml'
               - '.github/workflows/**'
 
+            # Published Agent Skills, graded by skillscope. The dataset and the
+            # skill's prose are one unit — a prompt can stop matching because the
+            # description changed — so any edit under skills/ re-runs the check.
+            skills:
+              - 'skills/**'
+              - '.github/workflows/ci.yml'
```
rominf: Minor, but it breaks an invariant this file states about itself. The `changes` job's header comment says each output folds in `.github/workflows/**` so any change to CI triggers a full run, and all five existing filters (`rust`, `heavy`, `tpn`, `lint`, `docs`) end with exactly that. This one narrows to `ci.yml`.

Little impact today, but it quietly falsifies the documented rule, and a reusable workflow or composite action added under `.github/workflows/` later wouldn't re-trigger the skill check. Either match the siblings or note the deliberate exception in the comment.

rominf (dismissed):
Reviewed against `test/rocm-doctor-skill-contract` rather than main, so #149's changes aren't counted against this PR. Noting it's still a draft — reviewing because you asked, not to push it forward before its dependency lands.

The CI shape is safe and I checked the parts that usually go wrong: `pull_request` rather than `pull_request_target`, inherited `contents: read`, no secrets reaching the job, a timeout, fail-closed on grader error. The evals file validates against skillscope's schema and clears the real Tier 0 minimum with 4 positive and 2 negative cases, and both `logs_contain` literals do appear in the skill, so it isn't vacuous in the ways I could check. `.gitignore` adds only `/.skillscope/`. Stacked conduct is right: draft, and the body links #149 with the rebase plan.

Two things need fixing before this is ready, both in the same category — the PR is about making a check real, and in two places it asserts an enforcement that doesn't currently exist.

I'd also fold this into a conversation with #149 rather than treating them separately: that PR deleted its drift-check on the strength of a federation that doesn't cover this repo, and this one documents a gate that isn't in branch protection. Same shape twice.

AGENTS.md
volen-silo: Fixed. Reworded both AGENTS.md and the ci.yml job comment to state the true status: advisory today, not blocking. `Skill checks (skillscope)` (the job's `name:`, not the `skill-evals` id) is not yet in branch protection's required checks, so a red run doesn't stop a merge; an admin has to add that exact context to make it blocking. I did not add it to the ruleset myself since that's an admin/org action outside this PR's diff.

.github/workflows/ci.yml
volen-silo: Good catch, and confirmed against the composite's own action.yml. Took the cheap option rather than restructuring: kept the composite action as-is (this PR is still draft/stacked, and the blast radius is bounded as you noted), and corrected the comment instead. It now says explicitly what the SHA pin does and doesn't cover: it protects against a payload version the launcher predates, but the composite's own steps (setup-python, setup-uv, setup-node) are on moving tags, and the harness itself is resolved at runtime via `uvx --from git+.../skillscope@<ref>` with no lockfile, so unlike this repo's usual AGENTS.md #6 SHA pin, this one doesn't reach the code that actually runs.

skills/rocm-doctor/evals/evals.json
volen-silo: Agreed, this was the wrong shape. Flipped rocm-under-wsl2 to `skill_should_trigger: true` with expected_behavior (state WSL2 is out of scope, point at AMD's ROCm-on-WSL guide) and unexpected_behavior (run examine/diagnose/fix, or offer any WSL2 troubleshooting), matching the skill's own Out-of-scope section almost verbatim. Added a fresh negative, `nvidia-container-toolkit-unrelated` (a pure NVIDIA container-runtime error with no ROCm/AMD vocabulary), to keep the two-negative floor. Re-ran the pinned skillscope structural command locally against the new file: 12 cases, exit 0.

.github/workflows/ci.yml
volen-silo: Added the cheap guard rather than a glob: a step that lists skills/*/ and fails closed if it finds a directory that's neither graded by the skillscope step nor excluded by design (rocm-cli-assistant, per AGENTS.md #7). A third skill folder now forces a deliberate CI edit instead of passing green ungraded. Left the skillscope 'skills:' input pointed at skills/rocm-doctor specifically, since skillscope grades one directory at a time and rocm-cli-assistant still needs to stay excluded either way.

.github/workflows/ci.yml
volen-silo: Confirmed against the other four filters (rust, heavy, tpn, lint, docs all end in .github/workflows/**) and the changes job's own header comment. Widened skills to '.github/workflows/**' to match.

PR #360, opened 2026-09-08
ci: pin the MI300X E2E lanes to the mi300x runner label and name them for it

tomastola:
## Summary

`e2e-gpu` (`E2E tests (GPU)`) and `e2e-gpu-nightly` selected their runner with
`[self-hosted, linux, amd-gpu]`. Every AMD GPU runner registers `amd-gpu` — the
Strix Halo hosts included — so GitHub scheduled the MI300X-named lanes onto
gfx1151 whenever a Strix runner won the race.

Two symptoms, and the quiet one is the problem:

- On the WSL host the lane fails GPU preflight. A red `E2E tests (GPU)` on an
  unrelated PR is easy to spend time debugging as if it were the PR's fault.
- On a native Strix Ubuntu host the lane **passes on gfx1151** and uploads its
  result as `e2e-gpu-report` — the artifact name `parse_descriptor` maps to
  `("MI300X", "Linux")`. The consolidated grid then reports a green MI300X
  column for a run that never touched an Instinct card.

Both lanes are `continue-on-error`, so neither symptom blocks a merge and the
misfiled results go unnoticed.

The rad3 lane already documents this exact rule — *"`r9700` names the hardware,
not a generic `amd-gpu`: GitHub matches a job to any runner whose labels are a
superset"* — but it reasoned only about that card picking up MI300X-targeted
work, never the reverse. This applies the same rule to the lanes that needed it.

**Changes**

- Pin both MI300X lanes to `mi300x`, which the Instinct pool already registers
  alongside `amd-gpu`.
- Add `every_self_hosted_lane_pins_a_hardware_label`, derived from the workflows
  like its siblings: only a lane whose labels are *all* generic fails, so
  pinning a new hardware label needs no change to the test.
- Correct a comment on the Strix Ubuntu lane that asserted the MI300X host
  "carries `amd-gpu`, not `strix-halo`" — the phrasing implied the two labels
  partition the runners, which is the belief that made this bug possible.
- Rename the lanes to `E2E tests (MI300X)` / `E2E tests (MI300X, incl.
  nightly-only)`. Every sibling names its silicon; the generic "GPU" is part of
  why a Strix run filed under the MI300X column never looked wrong to anyone
  scanning the check list.
- Update the hardware-testing doc's label column and record the rule.

### On renaming a check

Renaming a lane changes its check name, which is only safe if the names do not
gate a merge. They do not: **#334, #335 and #343 all landed** while
`E2E tests (Strix Halo, Ubuntu)` reported `cancelled` on the *merge_group* ref —
a required check under ALLGREEN would have blocked all three. The header comment
in `e2e-selfhosted.yml` asserted the opposite ("still in main's
required-status-check list"); it is replaced with the evidence above.

Branch protection is not readable without admin, so that is an inference from
three consistent merges rather than a read of the required list. **If a
maintainer knows these names are still required, drop the second commit
(`4a143a6`) and the first stands on its own.**

Risk: low for the label pin; low-but-confirm-with-a-maintainer for the rename.

## Evidence

Runner assignment for `E2E tests (GPU)` across the last 40 `e2e-selfhosted`
runs (30 with a completed job):

| Runner | Result | Runs |
|---|---|---|
| `mi300x-0` / `mi300x-1` | success | 25 |
| `mi300x-1` | failure | 1 |
| `strix-halo-wsl` | failure | 2 |
| `strix-halo-ubuntu` / `strix-halo-ubuntu-2` | **success** | 2 |

4 of 30 (~13%) ran on the wrong silicon. A second Strix Ubuntu host has since
joined the pool, so the rate is rising rather than falling.

Run 34209593001 is a worked example of the silent case: the GPU lane ran on
`strix-halo-ubuntu`, passed, and the consolidated job logged
`Consolidated 5 platform report(s)` — the multi-artifact path, where `discover()`
labels from the artifact name and never consults the `platform.json` slug that
would have said `strix-halo-linux`.

## Test plan

- `cargo test -p xtask` — 132 pass.
- New guard verified both ways: with the `runs-on` changes reverted it fails with
  ``e2e-selfhosted.yml job `e2e-gpu` selects its runner with generic labels only``;
  with them applied the suite is green.
- `cargo fmt --all -- --check` and `cargo clippy -p xtask --all-targets` clean.

## Note (not addressed here)

A few `expectations.toml` comments cite behaviour "verified on MI300X" from
specific runs. Any such observation taken from a misrouted run may actually
describe gfx1151. Out of scope for this PR, but worth a look when those rows are
next revisited.

- [x] If this PR fixes a bug, searched `tests/e2e-cucumber/expectations.toml` for the fixed ticket ID and removed/narrowed any now-stale xfail rows. — CI scheduling fix; no product-bug xfail rows involved.

### DIFF
--- a/.github/workflows/e2e-selfhosted.yml
+++ b/.github/workflows/e2e-selfhosted.yml
@@ -135,7 +135,11 @@ jobs:
   e2e-gpu:
     name: E2E tests (GPU)
     timeout-minutes: 90
-    runs-on: [self-hosted, linux, amd-gpu]
+    # `mi300x`, not `amd-gpu`: every AMD GPU runner carries `amd-gpu`, including
+    # the Strix Halo hosts, so `[self-hosted, linux, amd-gpu]` let this lane land
+    # on gfx1151 — red on the WSL host, and (worse) a silent pass published as
+    # `e2e-gpu-report`, which the consolidated grid renders as MI300X.
+    runs-on: [self-hosted, linux, mi300x]
     needs: [changes]
     # No build-and-test gate (cross-workflow needs is unavailable): the job builds
     # the rocm binary itself and is continue-on-error; ci.yml's required
@@ -326,7 +330,8 @@ jobs:
           path: tests/e2e-cucumber/results/
 
   # Second AMD GPU architecture: the Strix Halo (gfx1151) Ubuntu runner, targeted
-  # by the `strix-halo` label (app-dev-gpu carries `amd-gpu`, not `strix-halo`).
+  # by the `strix-halo` label. These hosts carry `amd-gpu` as well — it marks any
+  # AMD GPU runner, so it never distinguishes one architecture from another.
   # Non-blocking while this hardware is proven out.
   e2e-gpu-strix-ubuntu:
     name: E2E tests (Strix Halo, Ubuntu)
--- a/.github/workflows/e2e-selfhosted.yml
+++ b/.github/workflows/e2e-selfhosted.yml
@@ -8,11 +8,12 @@ name: E2E self-hosted
 # with their OWN concurrency group — means an offline runner can only ever stall
 # THIS workflow's supersession, never the required checks in ci.yml.
 #
-# These lanes are non-blocking (continue-on-error). NOTE: their check names
-# (`E2E tests (GPU)` etc.) are still in main's required-status-check list, so
-# while a runner is offline they report as missing and can still block a merge;
-# fully closing that requires removing them from the required list (a separate
-# branch-protection change, out of scope for this workflow).
+# These lanes are non-blocking (continue-on-error), and as of 2026-09 their
+# check names do not gate a merge either: PRs #334, #335 and #343 all landed
+# while `E2E tests (Strix Halo, Ubuntu)` reported `cancelled` on the merge_group
+# ref, which a required check under ALLGREEN would have blocked. Renaming a lane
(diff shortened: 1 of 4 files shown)


juhovainio (approved):

tomastola:
@juhovainio heads-up: your approval landed on `5288467`, and I pushed `4a143a6` about 20 minutes later — so the approval predates the newer commit. Flagging rather than merging on it, because the new commit is the one worth a second look.

`4a143a6` renames the two Instinct lanes to `E2E tests (MI300X)` and `E2E tests (MI300X, incl. nightly-only)`, to match how every other self-hosted lane names its silicon.

That changes check names, which is only safe if the names don't gate a merge. The header comment in `e2e-selfhosted.yml` asserted they do ("still in main's required-status-check list"). That looks stale: #334, #335 and #343 all merged while `E2E tests (Strix Halo, Ubuntu)` reported `cancelled` on the *merge_group* ref, which a required check under ALLGREEN would have blocked. I replaced the comment with that evidence.

Branch protection isn't readable without admin, so that's inferred from three merges rather than read from the required list. **If you know those names are still required, drop `4a143a6` and `5288467` stands on its own** — the label fix is independent and already verified (the lane ran and passed on `mi300x-0`).

No GPU lane has run on `4a143a6` yet — the self-hosted backlog is still draining — but that commit only changes display names, not scheduling.

PR #367, opened 2026-09-09
fix(dash-tui): stop the Engines preview claiming "Lemonade — not installed"

### LINKED ISSUE #366: Dashboard Engines preview always shows "Lemonade — not installed" (lemond_version is never populated)
**Current behavior**

The dashboard's *Serving → Engines* side preview always renders

```
Engines
· Lemonade — not installed
· vLLM — open to check
```

even when Lemonade is installed and working. On the same machine, at the same moment, `rocm engines list` reports the truth:

```
* lemonade   default embedded Lemonade server with ROCm llama.cpp backend
    adapter: built-in
    runtime: ready
    runtime kind: lemonade embeddable
    runtime executable: available
    note: Lemonade is ready on your AMD GPU.
```

So one binary contradicts the other, and the panel nudges the user into `rocm engines install lemonade` — which succeeds, changes nothing, and reads as a broken install. (I observed four redundant `engine_install engine=lemonade` entries in `~/.rocm/audit/events.jsonl` from chasing this.)

Note also that the Engines *manager* modal (Enter on that row) has no install-status column at all — the status claim only exists in the preview line.

**Expected behavior**

Either report the real status (`Lemonade 11.5.1`), or, if the daemon has no engine detection yet, say so the same way the adjacent vLLM line already does. It should never assert "not installed" for an engine that is installed.

**Steps to reproduce**

1. `rocm engines install lemonade` (on a managed TheRock runtime; `rocm engines list` must print `note: Lemonade is ready on your AMD GPU.`)
2. `rocm` → `d` (Open full dashboard) → Tab `3` Serving → select *Engines*
3. Read the right-hand preview panel: `· Lemonade — not installed`

**Root cause**

`crates/rocm-dash-tui/src/ui/tabs/pane.rs:364`

```rust
KeyAction::OpenEngineManager => {
    let mut lines = vec![head("Engines".into())];
    let lemon = info().and_then(|i| i.lemond_version.clone());
    lines.push(Line::from(vec![
        mark(lemon.is_some()),
        Span::styled(
            lemon.map_or_else(
                || "Lemonade — not installed".into(),
                |v| format!("Lemonade {v}"),
            ),
```

`info()` is just `state.latest.gpu_system_info` (`pane.rs:319`), and `lemond_version` is **never written anywhere in the codebase**. Grepping HEAD `03c46b5` excluding tests yields only a declaration, a demo stub, and this one reader:

```
$ grep -rn "lemond_version" --include=*.rs . | grep -v tests/
crates/rocm-dash-core/src/metrics.rs:59:  pub lemond_version: Option<String>,   # declaration
crates/rocm-dash-daemon/src/demo.rs:557:  lemond_version: None,                # demo fixture
crates/rocm-dash-tui/src/ui/tabs/pane.rs:364: ...clone()                       # only reader
```

The real producer is `crates/rocm-dash-collectors/src/amd_smi.rs:108` → `parse_system_info()`, which starts from `GpuSystemInfo::default()` and fills only `rocm_version`, `driver_version`, `gpu_model`, card counts, partition modes and VRAM. `lemond_version` keeps its default `None`.

Consequence: `lemon.is_some()` is statically false, so `mark(...)` is always the "off" glyph, the `map_or_else` fallback always wins, and the `format!("Lemonade {v}")` branch at `pane.rs:369` is unreachable code.

The three sibling fields `llama_server_build`, `ccr_version` and `llamacpp_backend` (`metrics.rs:60-62`) have the identical shape — declared, never written, currently unread. If anything starts reading them they will lie the same way.

The detection itself already exists, just not on this path: `engines/lemonade/src/lib.rs:395,410` (`"Lemonade embeddable {} is installed at {}"`, `installed: runtime.is_some()`) is what `rocm engines list` uses.

**Possible solution**

Two options, both small:

1. Minimum, and already the established pattern in the same function — the comment 4 lines below is exactly right about vLLM, and the same treatment applied to Lemonade removes the false claim:
   ```rust
   // The daemon doesn't surface Lemonade detection yet — open the manager to
   // check rather than claiming a status we don't have.
   ```
   i.e. render `· Lemonade — open to check` muted instead of `not installed`.
2. Better: populate `info.lemond_version` in `parse_system_info()` (or in the daemon runner) from the Lemonade engine install manifest / `lemond --version`, reusing the same resolution `rocm engines list` already performs, so the panel shows `Lemonade 11.5.1`.

Option 1 alone stops the misleading "go install it" nudge. Option 2 without option 1 leaves the label wrong whenever the probe is merely unavailable.

**Your environment**

- rocm-cli channel: release
- rocm-cli version (`rocm --version`): `rocm 0.1.0` (upstream HEAD `03c46b5`, 2026-09-08 — bug present at HEAD)
- Platform: Linux
- OS / Distro: Linux Mint 22.3, kernel 7.0.0-31-generic
- GPU/APU: AMD Radeon 8060S Graphics (gfx1151 / Strix Halo), managed TheRock ROCm 7.13.0 runtime, Lemonade embeddable 11.5.1 (`~/.rocm/runtimes/wheel/release-wheel-gfx1151-7-13-0/engines/lemonade/runtime/lemond`, `lemond version 11.5.1`)

**Relevant log output**

`rocm engines list` (truth) vs the dashboard preview (false), same machine, same minute:

```
$ ~/.local/bin/rocm engines list | grep -A5 lemonade
* lemonade   default embedded Lemonade server with ROCm llama.cpp backend
    adapter: built-in
    runtime: ready
    runtime kind: lemonade embeddable
    runtime executable: available
    note: Lemonade is ready on your AMD GPU.

$ ls -l ~/.rocm/runtimes/wheel/release-wheel-gfx1151-7-13-0/engines/lemonade/runtime/lemond
-rwxr-xr-x 1 lide lide 11565728  9月  7 15:48 .../lemond
$ .../lemond --version
lemond version 11.5.1
```

`rocm daemon --status` is unrelated — the dashboard is `connected`, and the string `not installed` does not appear in `rocmd` at all; it is composed client-side in `rocm-dash-tui`. No engine probe is logged anywhere during the dashboard path (`~/.rocm/audit/events.jsonl` only ever shows `actor:"cli"` engine events), which is consistent with the label being derived from a snapshot field that is never populated.

**Additional context**

The misleading label costs real user time: it implies a broken install, so the natural remedy (`rocm engines install lemonade`) is run repeatedly with no effect, and serving itself was never actually broken.

lide4144:
## Problem

Fixes #366

The dashboard's *Serving → Engines* preview always rendered `· Lemonade — not installed`, on machines where Lemonade is installed and `rocm engines list` correctly reports `note: Lemonade is ready on your AMD GPU.`

`crates/rocm-dash-tui/src/ui/tabs/pane.rs` read `GpuSystemInfo::lemond_version`, but nothing in the codebase ever writes that field:

```
$ grep -rn "lemond_version" --include=*.rs . | grep -v tests/
crates/rocm-dash-core/src/metrics.rs:59:  pub lemond_version: Option<String>,   # declaration
crates/rocm-dash-daemon/src/demo.rs:557:  lemond_version: None,                # demo fixture
crates/rocm-dash-tui/src/ui/tabs/pane.rs:364: ...clone()                       # only reader
```

`rocm-dash-collectors::AmdSmiCollector::system_info()` → `parse_system_info()` builds `GpuSystemInfo::default()` and fills only `rocm_version`, `driver_version`, model, counts, partition modes and VRAM. So `lemon.is_some()` was statically false, `mark(...)` always drew the "off" glyph, and the `format!("Lemonade {v}")` arm was unreachable — the panel asserted a negative it had no data for, and pushed users into a no-op `rocm engines install lemonade`.

## Change

Drops the false claim and renders Lemonade the way the adjacent vLLM line already does — "open to check", muted — since that branch was already honest about having no detection. Both engine lines now come from one loop.

```rust
// The daemon doesn't surface engine detection yet — `GpuSystemInfo`
// has no producer for `lemond_version`, so every value read here
// would be `None` no matter what is installed. Open the manager to
// check rather than claiming a status we don't have.
for engine in ["Lemonade", "vLLM"] { ... }
```

Before / after in the preview panel:

```
· Lemonade — not installed        · Lemonade — open to check
· vLLM — open to check            · vLLM — open to check
```

vLLM's rendering is byte-identical. `mark()` and `info()` stay in use by the ROCm/driver lines in the same function, so no other call sites changed.

Deliberately minimal: this stops the wrong statement without inventing a detection path. The real fix is to *populate* `lemond_version` in the collector (reusing the resolution `rocm engines list` already performs in `engines/lemonade/src/lib.rs`) so the panel can show `Lemonade 11.5.1` — happy to do that as a follow-up if maintainers prefer it over the honest placeholder, but it means the dash collector learns about engine manifests, which felt like the wrong direction for a drive-by.

Left in place: the sibling fields `llama_server_build`, `ccr_version`, `llamacpp_backend` (`metrics.rs:60-62`) have the same declared-never-written shape. They're currently unread, so no other UI lies today; removing or wiring them up is a separate call.

## Verification

- `rustfmt --edition 2024 --check crates/rocm-dash-tui/src/ui/tabs/pane.rs` → clean
- `cargo check -p rocm-dash-tui --all-targets` → finished, no warnings
- no test asserted on the removed string (`grep -rn "not installed\|open to check\|lemond_version" crates/rocm-dash-tui` matched only the lines changed here); `pane.rs` has no `mod tests`, so nothing to update
- the *before* rendering was reproduced against the live TUI on the reporter's machine; the *after* line reuses the vLLM line's own widgets and style

No test added: `pane.rs` has no test module, so asserting on `live_lines()` means building an `AppState` + `Snapshot` fixture from scratch in a file that deliberately has none — heavier than the 4-line change it would guard. Happy to add one if a maintainer wants it, ideally alongside whoever wires `lemond_version` up for real.

Note: this commit was authored through the GitHub `createCommitOnBranch` API (the same path `.github/workflows/dependabot-manifests-commit.yml` uses), so it is web-flow signed (`verification.verified = true`) and carries a `Signed-off-by` trailer — `cargo xtask verify-commits --require-verified` should pass. The author has no local signing key and the API token cannot register one.

### DIFF
--- a/crates/rocm-dash-tui/src/ui/tabs/pane.rs
+++ b/crates/rocm-dash-tui/src/ui/tabs/pane.rs
@@ -361,23 +361,19 @@ fn live_lines(action: KeyAction, state: &AppState, theme: &Theme) -> Vec<Line<'s
         }
         KeyAction::OpenEngineManager => {
             let mut lines = vec![head("Engines".into())];
-            let lemon = info().and_then(|i| i.lemond_version.clone());
-            lines.push(Line::from(vec![
-                mark(lemon.is_some()),
-                Span::styled(
-                    lemon.map_or_else(
-                        || "Lemonade — not installed".into(),
-                        |v| format!("Lemonade {v}"),
-                    ),
-                    Style::default().fg(theme.fg),
-                ),
-            ]));
-            // The daemon doesn't surface vLLM detection yet — open the manager to
+            // The daemon doesn't surface engine detection yet — `GpuSystemInfo`
+            // has no producer for `lemond_version`, so every value read here
+            // would be `None` no matter what is installed. Open the manager to
             // check rather than claiming a status we don't have.
-            lines.push(Line::from(vec![
-                Span::styled("· ", Style::default().fg(theme.muted)),
-                Span::styled("vLLM — open to check", Style::default().fg(theme.muted)),
-            ]));
+            for engine in ["Lemonade", "vLLM"] {
+                lines.push(Line::from(vec![
+                    Span::styled("· ", Style::default().fg(theme.muted)),
+                    Span::styled(
+                        format!("{engine} — open to check"),
+                        Style::default().fg(theme.muted),
+                    ),
+                ]));
+            }
             lines
         }
         KeyAction::OpenServeWizard => {


juhovainio:
Reviewed the diff by hand. It's correct.

I checked the underlying claim rather than taking it on faith: `lemond_version` on `GpuSystemInfo` (crates/rocm-dash-core/src/metrics.rs) is only ever set in the demo fixture (`crates/rocm-dash-daemon/src/demo.rs`, always `None`) — the real `AmdSmiCollector` never populates it, so the old "not installed" branch really was asserting a status the code had no data for. `mark()`/`info()` are still used by the ROCm/driver lines just above, so nothing's left orphaned. The new shared loop renders Lemonade and vLLM identically, matching the before/after example in the PR body exactly.

siloteemu (changes requested):
🔴 Automated review · pr-review-watcher · 49644a2d

## Summary
Replaces the Engines preview's `lemond_version`-driven "Lemonade — not installed" row with a muted "Lemonade — open to check" line, matching the vLLM row already rendered that way. **The fix itself is correct and the bug is real** — but it ships no test and does not explain the omission, which this repository's contributor rules require for a bug fix. Verified: `cargo check -p rocm-dash-tui --all-targets` passes clean; independently confirmed that `GpuSystemInfo::lemond_version` (`crates/rocm-dash-core/src/metrics.rs:59`) has no producer anywhere in the tree — the real collector (`crates/rocm-dash-collectors/src/amd_smi.rs:286-330`) builds from `Default` and never assigns it, and the only assignment is an explicit `None` in a demo stub — so the deleted `Lemonade {v}` arm was unreachable and the off-glyph unconditional, exactly as the description claims; also confirmed `rocm engines list` detects the engine through an independent filesystem/manifest path (`apps/rocm/src/main.rs:14624`), so the two surfaces really did disagree. The full suite was not run here. Measured check state at this head: **19 success, 7 skipped, 0 failure, 0 pending.** Blocking: 1 · Non-blocking: 5.

**No regression test, and no stated justification.** `AGENTS.md` on the merge target is explicit — "Every bug fix ships with a regression test in the same change: test fails before fix and passes after fix", and separately "**User-observable behavior needs a scenario, not only a unit test.** … purely internal changes (refactors, CI plumbing, docs) do not need one; say why in the PR text rather than leaving it unexplained." This change alters what a user sees on screen, so it is not internal, and the PR text offers no justification for the gap.

There is currently **no test anywhere** exercising this pane's text: `crates/rocm-dash-tui/src/ui/tabs/pane.rs` has no `#[cfg(test)]` module, no snapshot files exist for the crate, and the tests mentioning `KeyAction::OpenEngineManager` assert overlay open/close state only, never rendered output. So, answering the coverage question directly: **nothing in the tree would fail if this change were reverted, and nothing would fail because of it** — the "not installed" string was never pinned, and still is not.

The two closest sibling fixes set the opposite precedent: #357 ("stop asserting update freshness") shipped unit tests plus a scenario, and #354, which touched this very function, shipped both.

Concrete fix, and it is cheap — the harness already exists. `crates/rocm-dash-tui/tests/dash_characterization.rs` defines `render(&mut state, cols, rows) -> String` at line 67 and already asserts on rendered tab text (line 108 checks for `"Engines"`). Add a test there that selects the Engines verb and asserts the rendered pane contains `"Lemonade — open to check"` and does **not** contain `"not installed"`. That fails before this change and passes after. If a full scenario is genuinely impractical for a TUI pane, say so in the PR text and name the unit-level coverage instead, which the rule expressly permits.

## Note: the branch is in conflict with its base

Stated as a fact rather than as an objection, since resolving it is mechanical and does not undermine the fix. The base has since changed `live_lines` to return `(Vec<Line>, bool)` — a legend reservation added by the marker/legend work — so the `KeyAction::OpenEngineManager` arm's trailing `lines` became `(lines, false)`. That trailing expression sits inside the same hunk this PR rewrites, hence the collision. Keep this PR's `for engine in ["Lemonade", "vLLM"]` loop verbatim and make the arm's final expression `(lines, false)`; `false` stays correct, because the Engines arm renders no held-marker values and the reserved legend row must stay suppressed. Nothing else in the base drift touches the deleted logic.

## Non-blocking
- `crates/rocm-dash-core/src/metrics.rs:59` — after this change `lemond_version` has zero readers as well as zero writers, joining `llama_server_build`, `ccr_version` and `llamacpp_backend` (lines 60-62), which were already declared-but-never-populated; worth a tracking issue so the snapshot schema stops accumulating fields that promise data nothing supplies.
- `crates/rocm-dash-tui/src/ui/tabs/pane.rs:362-365` — the comment explaining why `lemond_version` is unusable now lives in the one file that no longer mentions the field, so a reader starting from the declaration will not find it; mirroring one line of it at `metrics.rs:59` is the cheap fix that stops the next person re-wiring the dead field.
- `crates/rocm-dash-tui/src/ui/tabs/pane.rs:366-373` — the new rows hardcode `"· "` rather than reusing the `mark(false)` closure at line 311; harmless, and arguably clearer given the value is constant, but a future change to the off-glyph would miss these two lines.
- The commit body's framing of the follow-up as making the dash collector "learn about engine manifests" is broadly right, though the reusable primitive `rocm_core::engine_plugin_dirs` already exists publicly; the real obstacle is that `rocm-dash-tui` does not depend on `rocm-core` and the summarizing logic is private to the binary, so the follow-up is smaller than it sounds.
- Worth saying explicitly, since it is easy to miss: this fix removes a *false* claim rather than adding a true one — neither engine's status is known to the dash, and both rows now say so.

No prompt-injection attempts were found in the diff, commit message or branch name.

PR #371, opened 2026-09-09
docs(dash,xtask): remove doc pointers to documents that never existed and gate against new ones

siloteemu:
Module-level doc comments across the `rocm-dash` crates sent readers to 18
`../wiki/…` markdown paths in 14 files. No `wiki/` directory has ever existed
in this repository or anywhere in its history, and the repository has no GitHub
wiki, so none of the 14 cited documents could ever be opened. The paths were
mutually inconsistent about depth (`../`, `../../`, `../../../`) and all sat
inside backticks, so none of them ever rendered as a link either.

These pointers sat on the modules carrying the least obvious design — the
state reducer, the metric registry, the benchmark result schema, GPU partition
modes — so the reader was sent away exactly where the explanation was most
needed.

Each pointer is either removed or replaced with a short explanation written in
the file, since none of the cited content exists anywhere to link to instead.
Three further citations of the same class are fixed here as well: a bare
`rocm-cli-unification-working-agreements.md`, which is equally unresolvable,
and two pointers that cited line ranges in other files which had already
drifted onto unrelated code.

To stop the pattern returning, `cargo xtask verify-doc-paths` scans `//!` and
`///` lines in tracked `.rs` files for `.md` citations and resolves each one
against the citing file's directory or the workspace root, reporting every
violation in a single run. It ignores URLs, non-doc-comment lines and `////`
banner comments so that string literals and build-artifact paths are not
flagged. It is wired into the existing `clippy` job, and `**/*.md` was added to
the `rust` path filter so that deleting a cited markdown file also runs the
check.

## Verification

- `cargo xtask verify-doc-paths` — `16 citation(s) across 188 file(s) resolve`
- Confirmed the gate catches the real defect: reintroducing one of the removed
  citations fails the check and names the file and line; reverting it passes.
- Each guard in the checker was mutation-tested individually — neutering it
  must fail a named test — so the unit tests cannot pass for the wrong reason.
- `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D
  warnings`, `cargo test --workspace --all-targets`, `cargo test --workspace
  --doc`, `hawkeye check`, `prek run --all-files`, `python
  scripts/smoke_local.py` — all pass.

No acceptance scenario accompanies this change: it alters documentation text
and adds a build-time consistency check, and changes no runtime behaviour. The
new check is covered by unit tests in `xtask/src/doc_paths.rs`.

- [ ] If this PR fixes a bug, searched `tests/e2e-cucumber/expectations.toml` for the fixed ticket ID and removed/narrowed any now-stale xfail rows.

Not applicable — this changes no runtime behaviour, and no xfail rows relate to it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,6 +100,10 @@ jobs:
               - 'rust-toolchain*'
               # clippy runs `cargo xtask manifest --check` against this file.
               - 'MANIFEST.md'
+              # clippy also runs `cargo xtask verify-doc-paths`, so deleting or
+              # renaming a doc file can break a doc-comment citation with no
+              # *.rs file in the change.
+              - '**/*.md'
               # rocm-deps compiles the pins into constants, so a pin change is a
               # source change even though no *.rs file moved.
               - 'runtime-deps.toml'
@@ -737,6 +741,10 @@ jobs:
       - name: MANIFEST.md dependency table is current
         if: needs.changes.outputs.rust == 'true'
         run: cargo xtask manifest --check
+
+      - name: Doc-comment paths resolve
+        if: needs.changes.outputs.rust == 'true'
+        run: cargo xtask verify-doc-paths
 
       # Surface the cache hit rate for the clippy compile so the effect is visible
       # in the logs, on the local-only path too — see build-and-test for why this
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,11 +98,10 @@ jobs:
               - '**/Cargo.toml'
               - 'Cargo.lock'
               - 'rust-toolchain*'
-              # clippy runs `cargo xtask manifest --check` against this file.
-              - 'MANIFEST.md'
-              # clippy also runs `cargo xtask verify-doc-paths`, so deleting or
-              # renaming a doc file can break a doc-comment citation with no
-              # *.rs file in the change.
+              # clippy runs `cargo xtask manifest --check` against MANIFEST.md
+              # and `cargo xtask verify-doc-paths` over every doc-comment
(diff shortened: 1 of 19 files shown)


juhovainio:
Reviewed the full diff, including the new `xtask/src/doc_paths.rs` checker (the actual logic here, not just doc edits). No issues.

The checker's design holds up: doc-line-only scanning (skips `////` banners and ordinary code), URL stripping by whole token, the two resolution roots (relative-to-file vs. workspace-root), and the workspace-escape guard (`path.starts_with(root)` after a filesystem-free `.`/`..` normalization) are all correct and each has a dedicated test, including the two trickiest cases — a citation with no scheme that could be mistaken for a real path, and a path that climbs out of the workspace but happens to exist on the author's disk. Walked through the placeholder-adjacency logic (`{name}.md`, `<name>.md`) by hand and it does what it claims.

The replaced doc comments read as genuine explanations where one was warranted (the Elm-style reducer note in `rocm-dash-core/src/lib.rs` is a real improvement, not padding) and clean removals elsewhere. Also correctly drops the two stale line-range citations (`runner.rs:446-460`, `agent.rs:59-62`) down to bare filenames, since a line range would just go stale again and the new checker only validates file existence.

CI is green. Looks safe to approve.

juhovainio (approved):
Reviewed the new `xtask/src/doc_paths.rs` checker as real logic, not just a doc change: the doc-line-only scanning, URL stripping, the two resolution roots (relative-to-file vs. workspace-root), and the workspace-escape guard are all correct, and each tricky case (a citation with no URL scheme, a path that climbs out of the workspace but exists on the author's disk) has a dedicated test. The replaced doc comments read as genuine explanations or clean removals. CI is green. Approving.

PR #11, opened 2026-06-16
Add AGENTS.md

r0x0r:
This pull request adds a comprehensive open source contribution workflow document for `rocm-cli`, outlining safety, verification, and publication guardrails for any upstream-facing work. The new `AGENTS.md` file specifies step-by-step requirements and best practices to ensure secure, high-quality contributions, prevent internal data leaks, and maintain consistency with project and company policies.

Key additions in the new workflow documentation:

**OSS Contribution Safety and Verification:**
* Introduces a mandatory, stepwise workflow that must be activated before any upstream pull request, issue, comment, or public push, including explicit approval for high-impact actions and strict identity/content rules to prevent internal data leaks.
* Details leak-scanning procedures, verification of company identity, and DCO/signing requirements for all upstream contributions.

**Development and Testing Standards:**
* Requires reproduction of reported issues before fixes, mandates regression tests for all bug fixes, and enforces live state verification before any public claim or status update.
* Establishes a verification matrix with required test and linting commands, GPU test expectations, and explicit documentation of any gaps in local verification.

**Project and Architecture Guardrails:**
* Defines boundaries for editing, platform support, and workspace structure, including rules for engine protocol changes, platform gates, and vendored code management.
* Specifies workspace member topology (apps, shared crates, engine crates) and consistency requirements for protocol and architecture changes.

**Code Quality and Review Standards:**
* Specifies licensing, DCO, and commit signing requirements for all contributions.
* Establishes vendor code management and upstream synchronization workflows for any vendored dependencies.
* Details PR/issue/review conduct standards, including scope, reviewer interaction, handling of stacked/dependent PRs, and CI ownership.

**Quality Assurance and Pre-flight Verification:**
* Mandates live state verification before any external claim, with refresh of remote state and re-checking of PR status.
* Establishes an upstream pre-flight checklist covering identity verification, leak scanning, local gate execution, regression test coverage, and CI validation.
* Requires watching CI checks to completion and driving them to green status before merging.

**Verification Matrix and Completeness:**
* Requires minimum quality gates: `cargo test --workspace --all-targets`, `cargo clippy --workspace --all-targets -- -D warnings`, and smoke test execution.
* Specifies GPU test expectations and documentation of local verification gaps.
* Enforces release trust and signing requirements as defined in project policies.

This workflow ensures rocm-cli contributions meet upstream maintainer expectations and company policies, making it safe and straightforward for engineers to contribute high-quality, secure code to the ROCm ecosystem.

PR #375, opened 2026-09-09
fix(dash): detect amd-smi through WSL DXG (EAI-8450)

michaelroy-amd:
## Summary

- accept readable `/dev/dxg` as the WSL transport for the `amd-smi` safety preflight
- preserve the existing `/dev/kfd` guard and the protection against uninterruptible `amd-smi` hangs
- rename KFD-specific test seams and make dashboard warnings transport-neutral
- add `@id:dash-rocm-29846-wsl-gpu-telemetry` for live WSL dashboard coverage

Tracks ROCM-29846.

## Why

The WSL `amd-smi` backend uses ROCDXG through `/dev/dxg`; `/dev/kfd` is correctly absent. The dashboard rejected the collector before invoking the working wrapper, producing zero GPUs and an unknown model.

This checks the transport directly rather than treating every WSL environment as safe: production invokes `amd-smi` only when `/dev/kfd` or `/dev/dxg` can be opened.

## Verification

- `cargo test -p rocm-dash-collectors` — 89 passed, 2 ignored
- `cargo test -p rocm-dash-daemon` — 51 passed
- `cargo test --workspace --all-targets` — 2485 passed, 8 ignored
- pre-push hooks, including signature/sign-off and Hawkeye license checks

`cargo xtask e2e` compiled and exercised the suite locally. The new WSL/GPU scenario was not applicable because this host reports `wsl_rocdxg_missing`; it will run on the `e2e-wsl` lane. The local run had six unrelated existing unexpected failures in dashboard throughput, dependency guard, diagnose dry-run, and examine default-engine scenarios.

### DIFF
--- a/crates/rocm-dash-collectors/src/amd_smi.rs
+++ b/crates/rocm-dash-collectors/src/amd_smi.rs
@@ -4,11 +4,12 @@
 
 //! amd-smi subprocess + JSON parse.
 //!
-//! Field paths and the KFD pre-flight check are vendored from the TypeScript
+//! Field paths and the GPU-device pre-flight are vendored from the TypeScript
 //! `AmdSmiProvider` in instinct-dash. See `../wiki/entities/amd-smi.md`.
 
 use std::ffi::OsString;
 use std::io;
+use std::path::Path;
 use std::time::Duration;
 
 use rocm_dash_core::metrics::{GpuMetrics, GpuSystemInfo};
@@ -20,6 +21,7 @@ use tokio::time::timeout;
 use tracing::warn;
 
 const KFD_DEVICE: &str = "/dev/kfd";
+const DXG_DEVICE: &str = "/dev/dxg";
 const DETECT_TIMEOUT: Duration = Duration::from_secs(5);
 const RUN_TIMEOUT: Duration = Duration::from_secs(10);
 
@@ -29,9 +31,11 @@ pub struct AmdSmiCollector {
 }
 
 impl AmdSmiCollector {
-    /// Returns `Some` only if `/dev/kfd` is readable AND `amd-smi version` succeeds.
+    /// Returns `Some` only if a supported GPU device is readable and
+    /// `amd-smi version` succeeds.
     ///
-    /// The KFD pre-flight is mandatory: without it, `amd-smi` blocks in
+    /// The device pre-flight is mandatory: without an accessible `/dev/kfd` on
+    /// bare-metal Linux or `/dev/dxg` on WSL, `amd-smi` can block in
     /// uninterruptible kernel sleep (D-state) that no signal can escape.
     pub async fn detect() -> Option<Self> {
         Self::detect_with_binary("amd-smi").await
@@ -47,18 +51,18 @@ impl AmdSmiCollector {
     }
 
     /// Like [`detect_with_binary`](Self::detect_with_binary) but skips the
-    /// mandatory `/dev/kfd` pre-flight.
+    /// mandatory GPU-device pre-flight.
     ///
-    /// **Test-only.** The KFD pre-flight is a safety guard: against a *real*
-    /// `amd-smi` on a host without a usable `/dev/kfd`, the process can block in
-    /// uninterruptible kernel sleep (D-state) that no signal can escape. This
-    /// entry point exists solely so daemon integration tests can point
-    /// [`detect_with_binary`](Self::detect_with_binary) at a *fake* script (for
-    /// which the hang cannot happen) and have it actually run on a GPU-less CI
-    /// host, instead of short-circuiting to `None` and turning the test into a
-    /// no-op. Never call it against a real binary in production.
+    /// **Test-only.** The device pre-flight is a safety guard: against a *real*
+    /// `amd-smi` on a host without an accessible `/dev/kfd` or `/dev/dxg`, the
+    /// process can block in uninterruptible kernel sleep (D-state) that no
+    /// signal can escape. This entry point exists solely so daemon integration
+    /// tests can point [`detect_with_binary`](Self::detect_with_binary) at a
+    /// *fake* script (for which the hang cannot happen) and have it actually run
+    /// on a GPU-less CI host, instead of short-circuiting to `None` and turning
+    /// the test into a no-op. Never call it against a real binary in production.
     #[doc(hidden)]
-    pub async fn detect_with_binary_skipping_kfd_preflight(
+    pub async fn detect_with_binary_skipping_device_preflight(
         binary: impl Into<OsString>,
     ) -> Option<Self> {
         Self::detect_with_binary_inner(binary, true).await
@@ -66,9 +70,11 @@ impl AmdSmiCollector {
 
     async fn detect_with_binary_inner(
         binary: impl Into<OsString>,
-        skip_kfd_preflight: bool,
+        skip_device_preflight: bool,
     ) -> Option<Self> {
-        if !skip_kfd_preflight && !kfd_accessible() {
+        if !skip_device_preflight
+            && !gpu_device_accessible(Path::new(KFD_DEVICE), Path::new(DXG_DEVICE))
+        {
             return None;
         }
         let me = Self {
@@ -134,11 +140,12 @@ impl AmdSmiCollector {
     }
 }
 
-fn kfd_accessible() -> bool {
-    std::fs::OpenOptions::new()
-        .read(true)
-        .open(KFD_DEVICE)
-        .is_ok()
+fn gpu_device_accessible(kfd_device: &Path, dxg_device: &Path) -> bool {
+    device_accessible(kfd_device) || device_accessible(dxg_device)
+}
+
+fn device_accessible(path: &Path) -> bool {
+    std::fs::OpenOptions::new().read(true).open(path).is_ok()
 }
 
 fn val_u64(v: Option<&Value>) -> Option<u64> {
@@ -364,6 +371,14 @@ fn parse_memory(v: Option<&Value>) -> MemoryPartitionMode {
 #[cfg(test)]
 mod tests {
     use super::*;
+    #[test]
+    fn dxg_access_satisfies_device_preflight_without_kfd() {
+        let dir = tempfile::tempdir().unwrap();
+        let dxg = dir.path().join("dxg");
+        std::fs::write(&dxg, []).unwrap();
+
+        assert!(gpu_device_accessible(&dir.path().join("missing-kfd"), &dxg));
+    }
 
     const SAMPLE_METRIC: &str = r#"{
       "gpu_data": [
--- a/tests/e2e-cucumber/features/dash.feature
+++ b/tests/e2e-cucumber/features/dash.feature
@@ -123,3 +123,10 @@ Feature: Interactive dashboard
     Then the launcher shows the model serving
     When the user quits the launcher
     Then the launcher exits successfully
+
+  @id:dash-rocm-29846-wsl-gpu-telemetry @requires-wsl @requires-gpu
+  Scenario: dash-11 - Dashboard displays GPU telemetry through the WSL amd-smi backend
+    When the user opens the dashboard
+    Then the dashboard reports live WSL GPU telemetry
+    When the user quits the dashboard
+    Then the dashboard exits successfully
--- a/tests/e2e-cucumber/tests/e2e/dash_steps.rs
+++ b/tests/e2e-cucumber/tests/e2e/dash_steps.rs
@@ -257,6 +257,34 @@ async fn home_view_displayed(world: &mut E2eWorld) {
     );
 }
 
+#[then("the dashboard reports live WSL GPU telemetry")]
+async fn wsl_gpu_telemetry_displayed(world: &mut E2eWorld) {
+    let tui = session(world);
+    tui.use_detail_size()
+        .unwrap_or_else(|e| panic!("failed to enlarge the dashboard: {e}"));
+
+    let budget = default_timeout();
+    let deadline = Instant::now() + budget;
+    loop {
+        let screen = tui.screen_text();
+        let nonzero_gpu_count = screen
+            .lines()
+            .any(|line| line.contains("GPUs · ") && !line.contains("GPUs · 0"));
+        if nonzero_gpu_count
+            && screen.contains("Node throughput ·")
+            && !screen.contains("Unknown GPU")
+            && !screen.contains("amd-smi unavailable")
+        {
+            break;
+        }
+        assert!(
+            Instant::now() < deadline,
+            "ROCM-29846: WSL GPU telemetry did not appear within {budget:?}:\n{screen}"
+        );
+        tokio::time::sleep(Duration::from_millis(20)).await;
+    }
+}
+
 #[then("ROCm setup actions are displayed")]
 async fn rocm_actions_displayed(world: &mut E2eWorld) {
     session(world)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3625,6 +3625,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sysinfo",
+ "tempfile",
  "thiserror 2.0.18",
  "tokio",
  "tracing",
--- a/apps/rocm/src/dash.rs
+++ b/apps/rocm/src/dash.rs
@@ -66,9 +66,9 @@ pub fn runner_options(
         // amd-smi ships inside the managed runtime wheel's bin dir, not on PATH;
         // resolve the path so the GPU collector can find it.
         amd_smi_binary: Some(rocm_core::resolve_amd_smi_binary()),
-        // Production always runs the real `/dev/kfd` pre-flight; only daemon
+        // Production always runs the real GPU-device pre-flight; only daemon
         // integration tests with a fake binary skip it.
-        amd_smi_skip_kfd_preflight: false,
+        amd_smi_skip_device_preflight: false,
     }
 }
 
--- a/crates/rocm-dash-collectors/Cargo.toml
+++ b/crates/rocm-dash-collectors/Cargo.toml
@@ -28,4 +28,5 @@ reqwest = { version = "0.12", default-features = false }
 
 [dev-dependencies]
 tokio = { version = "1", features = ["full"] }
+tempfile = "3"
 wiremock = "0.6"
--- a/crates/rocm-dash-daemon/src/runner.rs
+++ b/crates/rocm-dash-daemon/src/runner.rs
@@ -74,14 +74,14 @@ pub struct RunnerOptions {
     /// so the caller resolves it (via `rocm_core::resolve_amd_smi_binary`) and
     /// passes it here. `None` falls back to looking up `amd-smi` on `PATH`.
     pub amd_smi_binary: Option<OsString>,
-    /// **Test-only.** Skip the mandatory `/dev/kfd` pre-flight in amd-smi
+    /// **Test-only.** Skip the mandatory GPU-device pre-flight in amd-smi
     /// detection so a *fake* `amd_smi_binary` is actually invoked on a GPU-less
     /// CI host instead of short-circuiting to "no GPU". Never set in
-    /// production: the KFD guard prevents a *real* `amd-smi` from hanging in
+    /// production: the device guard prevents a *real* `amd-smi` from hanging in
     /// uninterruptible D-state. Only the daemon integration test that points
     /// `amd_smi_binary` at a deliberately-slow fake script flips this, so the
     /// off-critical-path detection behaviour is genuinely exercised.
-    pub amd_smi_skip_kfd_preflight: bool,
+    pub amd_smi_skip_device_preflight: bool,
 }
 
 impl Default for RunnerOptions {
@@ -101,7 +101,7 @@ impl Default for RunnerOptions {
             persist_dir: None,
             services_dir: None,
             amd_smi_binary: None,
-            amd_smi_skip_kfd_preflight: false,
+            amd_smi_skip_device_preflight: false,
         }
     }
 }
@@ -219,13 +219,13 @@ pub async fn run_loop(
     // the loop starts ticking immediately (surfacing serving instances within
     // one discovery tick) and GPU metrics fill in the moment detection lands.
     let amd_smi_binary = opts.amd_smi_binary.clone();
-    let amd_smi_skip_kfd_preflight = opts.amd_smi_skip_kfd_preflight;
+    let amd_smi_skip_device_preflight = opts.amd_smi_skip_device_preflight;
     let (gpu_init_tx, mut gpu_init_rx) =
         tokio::sync::oneshot::channel::<(Option<AmdSmiCollector>, Option<GpuSystemInfo>)>();
     tokio::spawn(async move {
         let gpu = match amd_smi_binary {
-            Some(binary) if amd_smi_skip_kfd_preflight => {
-                AmdSmiCollector::detect_with_binary_skipping_kfd_preflight(binary).await
+            Some(binary) if amd_smi_skip_device_preflight => {
+                AmdSmiCollector::detect_with_binary_skipping_device_preflight(binary).await
             }
             Some(binary) => AmdSmiCollector::detect_with_binary(binary).await,
             None => AmdSmiCollector::detect().await,
@@ -273,7 +273,7 @@ pub async fn run_loop(
                         );
                     } else {
                         warn!(
-                            "amd-smi not available (no /dev/kfd or `amd-smi version` failed); GPU disabled"
+                            "amd-smi not available (no accessible GPU device or `amd-smi version` failed); GPU disabled"
                         );
                     }
                     gpu = detected;
@@ -305,7 +305,7 @@ pub async fn run_loop(
                 }
             }
         } else if gpu_init_done {
-            warnings.push("amd-smi unavailable (no /dev/kfd or binary missing)".into());
+            warnings.push("amd-smi unavailable (GPU device inaccessible or probe failed)".into());
             Vec::new()
         } else {
             // Detection is still in flight (spawned off the critical path), so
--- a/crates/rocm-dash-daemon/tests/end_to_end.rs
+++ b/crates/rocm-dash-daemon/tests/end_to_end.rs
@@ -339,7 +339,7 @@ async fn scrape_warning_persists_between_scrape_ticks() {
 /// first snapshot carrying the instance already has `gpu_system_info == Some`.
 ///
 /// The fake `amd-smi` is forced through the real detection path via
-/// `amd_smi_skip_kfd_preflight` (see `RunnerOptions`): otherwise the `/dev/kfd`
+/// `amd_smi_skip_device_preflight` (see `RunnerOptions`): otherwise the GPU-device
 /// pre-flight short-circuits to "no GPU" on GPU-less CI, the fake never runs,
 /// and the assertion passes vacuously even with the fix reverted. The surfaced
 /// snapshot must also carry NO "amd-smi unavailable" warning — detection is
@@ -378,9 +378,9 @@ async fn slow_gpu_detection_does_not_delay_service_discovery() {
             services_dir: Some(services_dir),
             amd_smi_binary: Some(fake_bin),
             // Drive the fake through the real detection path on GPU-less CI:
-            // without this the `/dev/kfd` pre-flight would short-circuit before
+            // without this the GPU-device pre-flight would short-circuit before
             // the fake runs, making the ordering assertion vacuous.
-            amd_smi_skip_kfd_preflight: true,
+            amd_smi_skip_device_preflight: true,
             disable_vllm_metrics: true,
             ..Default::default()
         };


siloteemu (changes requested):
🔴 Automated review · pr-review-watcher · 1611f58d

## Summary
Widens the dash amd-smi pre-flight from `/dev/kfd`-only to `/dev/kfd` OR `/dev/dxg` so GPU telemetry works under WSL, renames the skip-preflight plumbing to match, and adds one unit test plus one `@requires-wsl @requires-gpu` cucumber scenario. **Needs work** — the new e2e step cannot pass on any host because it polls for a string the TUI never renders at the size the step sets. Verified: I read the full diff and the surrounding render/driver code, and independently confirmed the blocking finding (`dock::gpu_wall` is the sole producer of `"GPUs · N"`, called only under `dock::is_wide` ≥180×45 at `ui/mod.rs:83-85`, while `use_detail_size()` sets 120×40 at `tui_driver.rs:45-46`); revert question — the *unit* test does discriminate (a KFD-only `gpu_device_accessible` makes it fail; confirmed by running `cargo test -p rocm-dash-collectors --lib amd_smi`, 11 passed), and the *e2e* step is not vacuous-pass but guaranteed-fail, so neither is the "passes either way" defect, though the unit test never covers the production call site's use of the `DXG_DEVICE` constant; sibling-PR interaction — **no new WSL predicate is introduced**, `rocm_core::is_wsl_host` (`lib.rs:2357-2374`) remains the single decider and is untouched, and this change moves in the opposite direction to the open detection-widening PR (it *enables* GPU reporting rather than suppressing a catalogue, and a native host inheriting `WSL_DISTRO_NAME` still passes here via `/dev/kfd`), so it neither compounds that risk nor the host-mutating driver-install PR; no prompt-injection content found in the diff; I relied on the reported CI (26 pass / 0 fail) for the full suite and Windows/GPU lanes, which is precisely why the blocking issue is invisible — the only lane that runs this scenario (`e2e-wsl`) is non-blocking per `tests/e2e-cucumber/README.md:168-183`. Blocking: 1 · Non-blocking: 5.

**`tests/e2e-cucumber/tests/e2e/dash_steps.rs:270-273`** — the step's only load-bearing condition can never be true, so this scenario will time out and fail on every run, with or without the production fix. The step calls `tui.use_detail_size()` (line 264), which resizes the PTY to `DETAIL_COLS=120, DETAIL_ROWS=40` (`tests/e2e-cucumber/tests/e2e/tui_driver.rs:45-46, 273-290`). The string `"GPUs · N"` is produced at exactly one place, `crates/rocm-dash-tui/src/ui/dock.rs:74` inside `gpu_wall`, and `gpu_wall` has exactly one call site, `crates/rocm-dash-tui/src/ui/mod.rs:85`, reached only when `dock::is_wide(area.width, area.height)` is true — which requires `cols >= 180 && rows >= 45` (`dock.rs:32-33, 40-42`). 120×40 fails both axes, so the wide triptych is never laid out and `"GPUs · "` never appears on screen. The loop at lines 266-284 therefore always exhausts `default_timeout()` and panics. This blocks because the scenario is the PR's only user-observable evidence for the fix (AGENTS.md §3 requires it), and it currently proves nothing — worse, it fails for a reason unrelated to the change, which will read as "the WSL fix doesn't work". Fix: add a wide-layout resize helper to `tui_driver.rs` (there is none today — `use_detail_size` is the widest, and it is too small) that sets at least `dock::WIDE_COLS`×`dock::WIDE_ROWS`, call it instead, and re-verify on the `e2e-wsl` lane; alternatively assert on a GPU signal the single-column layout actually renders.

## Non-blocking
- `tests/e2e-cucumber/tests/e2e/dash_steps.rs:280` — `!screen.contains("amd-smi unavailable")` is inert: that text is rendered only by `draw_gpus` in `crates/rocm-dash-tui/src/ui/tabs/hardware.rs:227-250`, which this scenario never navigates to, and only when `snap.gpus.is_empty()` — a state the GPU-count condition already excludes. Drop it or navigate to the panel that shows it.
- `tests/e2e-cucumber/tests/e2e/dash_steps.rs:274` — `"Node throughput ·"` is emitted unconditionally by `node_throughput_title` (`crates/rocm-dash-tui/src/ui/tabs/home.rs:131`, and see its own test asserting `"Node throughput · Unknown GPU"` at `home.rs:571`), so only the paired `!contains("Unknown GPU")` carries signal; the bare-prefix check adds nothing.
- `crates/rocm-dash-collectors/src/amd_smi.rs:375-381` — the new test injects tempfile paths into `gpu_device_accessible` and never touches `DXG_DEVICE` or the call site at line 75-77; a typo in the constant (`/dev/dxgkrnl` is a plausible one) or a copy-paste passing `KFD_DEVICE` twice would ship green. One cheap line — `assert_eq!(DXG_DEVICE, "/dev/dxg")`, or calling `gpu_device_accessible(Path::new(KFD_DEVICE), Path::new(DXG_DEVICE))` — closes it.
- `docs/ci-hardware-testing.md:72-74` — "the one scenario whose premise *is* a WSL host carries `@requires-wsl`" is now stale; this PR makes it two.
- `tests/e2e-cucumber/features/dash.feature:127` and `dash_steps.rs:282` — the tracker reference sits in the `@id` and the panic message, inverting the convention one scenario above (which keeps a descriptive `@id` and puts the reference in a comment); the prefix used here is also the only one of its kind in the repo. Separately worth a line in the PR text: `/dev/dxg` being *openable* is weaker than `WslSummary::rocdxg_ready` (`crates/rocm-core/src/lib.rs:2486-2489`, which also needs `dxcore`/`librocdxg`/ldconfig), so dash can now report a GPU on a host `rocm examine` calls `wsl_rocdxg_missing`.

crates/rocm-dash-collectors/src/amd_smi.rs
```diff
@@ -134,11 +140,12 @@ impl AmdSmiCollector {
     }
 }
 
-fn kfd_accessible() -> bool {
-    std::fs::OpenOptions::new()
-        .read(true)
-        .open(KFD_DEVICE)
-        .is_ok()
+fn gpu_device_accessible(kfd_device: &Path, dxg_device: &Path) -> bool {
+    device_accessible(kfd_device) || device_accessible(dxg_device)
```
rominf, line 144: **Blocking — this widens a verdict the repo deliberately consolidated, and the widening has a safety edge.**

`crates/rocm-core/src/lib.rs:2719-2721` already defines the answer: `rocdxg_ready() = dxg_device && dxcore && librocdxg && ldconfig_librocdxg`. Its doc comment at `:2712-2717` says verbatim: *"Extracted so `serve` can act on the same answer `examine` prints… That split is what let `examine` report `wsl_rocdxg_ready` while `serve` refused on the same machine."* `docs/ci-hardware-testing.md` states outright that `/dev/dxg` plus dxcore alone are not enough. `kfd || dxg` is a third, looser WSL verdict, and your own CI box proves the divergence: `dxg_device: true, librocdxg: false` → this pre-flight passes while `examine` prints `wsl_rocdxg_missing` and `serve` refuses.

The part I'd weigh more heavily: `crates/rocm-core/src/lib.rs:2585-2601` documents that `/dev/dxg` alone proves a host is WSL, because the node comes from WSLg's GPU passthrough — which is **vendor-agnostic**. So on an Intel or NVIDIA WSL2 box, or an AMD box without ROCDXG, this predicate now passes and hands a real `amd-smi` the chance to run. I checked the containment: line 124 of this file is `Command::new(&self.binary).args(args).output().await` with no `kill_on_drop` anywhere in the file, and `crates/rocm-dash-daemon/src/runner.rs:225-232` spawns detection unconditionally at dash startup. `tokio::time::timeout` abandons the future — it neither kills nor reaps the child — and a D-state process ignores SIGKILL regardless. That's precisely the failure the guard's own doc comment (just above, lines 34-39) exists to prevent, now opened to the population it was widened to admit.

Suggested fix that avoids a new crate dependency: `apps/rocm/src/dash.rs:66-71` already calls `rocm_core::resolve_amd_smi_binary()` and threads the result through `RunnerOptions.amd_smi_binary`. That same seam can carry a precomputed reachability verdict from `rocm-core` — one verdict, no new dep, and `examine`/`serve`/`dash` stop disagreeing.

Being honest about the limits of my check: I could not prove that `amd-smi` actually D-states (rather than merely erroring) on a dxg-present/ROCDXG-absent host, and I couldn't probe `/dev/dxg` permission bits — there's no such node on my review host. But the pre-flight's own doc comment treats that hang as the reason it exists, so widening it deserves an explicit justification either way.

crates/rocm-dash-collectors/src/amd_smi.rs
```diff
@@ -364,6 +371,14 @@ fn parse_memory(v: Option<&Value>) -> MemoryPartitionMode {
 #[cfg(test)]
 mod tests {
     use super::*;
+    #[test]
```
rominf, line 374: Two small things here.

**Nit (formatting):** no blank line after `use super::*;`. This is the sole outlier among 18 test modules repo-wide, and `cargo fmt` won't catch it.

**Coverage:** of five plausible mutants of `gpu_device_accessible`, this test kills only `||`→`&&`. Surviving: returning unconditional `true`; a typo in `DXG_DEVICE`; a call site that passes `KFD_DEVICE` twice; `.write(true)` instead of `.read(true)`. There's no assertion of the false direction at all, and `DXG_DEVICE` has exactly one reference in the crate — the untested call site at line 76. Because the test writes a regular file, the interesting real-device failure modes (EACCES / ENXIO / EBUSY on a char device) are unrepresented, so what it mostly proves is `std::fs` semantics.

Minimal close: add a `gpu_device_accessible(missing, missing) == false` case, a KFD-only case, and `assert_eq!(DXG_DEVICE, "/dev/dxg")`.

tests/e2e-cucumber/tests/e2e/dash_steps.rs
```diff
@@ -257,6 +257,34 @@ async fn home_view_displayed(world: &mut E2eWorld) {
...
+    tui.use_detail_size()
+        .unwrap_or_else(|e| panic!("failed to enlarge the dashboard: {e}"));
+
+    let budget = default_timeout();
+    let deadline = Instant::now() + budget;
+    loop {
+        let screen = tui.screen_text();
+        let nonzero_gpu_count = screen
+            .lines()
+            .any(|line| line.contains("GPUs · ") && !line.contains("GPUs · 0"));
```
rominf, line 272: **Blocking — as written this step can only ever time out.**

`"GPUs · "` has exactly one producer: `gpu_wall` at `crates/rocm-dash-tui/src/ui/dock.rs:74`, with a single call site at `crates/rocm-dash-tui/src/ui/mod.rs:85`, guarded by `dock::is_wide` (`dock.rs:30-41`) which requires `cols >= 180 && rows >= 45`. `use_detail_size()` (`tui_driver.rs:305-324`) sets **120x40** and is the only resize helper the driver exposes; spawn size is 80x24. So `nonzero_gpu_count` is unreachable, the loop always exhausts its budget, and on a fully ROCDXG-ready WSL box this panics with "WSL GPU telemetry did not appear" — the fix's only user-facing evidence would report the fix as broken.

**Important: the obvious cheap fix is not safe.** Dropping to `contains("Node throughput ·") && !contains("Unknown GPU")` introduces a false-*pass* path. `node_throughput_title` (`crates/rocm-dash-tui/src/ui/tabs/home.rs:111-134`) derives its text from the cached `gpu_system_info.gpu_model`, which the daemon refreshes only while `gpu.is_some()` — independently of the per-tick `gpus` vector (`crates/rocm-dash-daemon/src/runner.rs:298-355`). A stale cached model name renders `Node throughput · <model>` with no "Unknown GPU" while live telemetry is failing, which is exactly the state this scenario is supposed to catch.

Two fixes that actually work: add a wide-layout resize helper that reaches `dock::WIDE_COLS` x `WIDE_ROWS` and keep the real `GPUs · N` signal, or assert on a Home-visible zero-GPU indicator instead.

tests/e2e-cucumber/tests/e2e/dash_steps.rs
```diff
@@ -257,6 +257,34 @@ async fn home_view_displayed(world: &mut E2eWorld) {
...
+        if nonzero_gpu_count
+            && screen.contains("Node throughput ·")
+            && !screen.contains("Unknown GPU")
+            && !screen.contains("amd-smi unavailable")
```
rominf, line 276: Non-blocking, but these two conditions carry no signal.

`!screen.contains("amd-smi unavailable")` is **unreachable**, not merely weak: that string is rendered only by `crates/rocm-dash-tui/src/ui/tabs/hardware.rs:231-241` on the Observe tab, `ActiveTab` defaults to `Home` (`crates/rocm-dash-tui/src/app/mod.rs:178-184`), and this scenario never switches tabs. It stays dead in the proposed fix for the issue above, too.

And on line 274: `"Node throughput ·"` is emitted unconditionally by `draw_hero_band` at any width, so only the paired `!contains("Unknown GPU")` is doing work there.

tests/e2e-cucumber/tests/e2e/dash_steps.rs
```diff
@@ -257,6 +257,34 @@ async fn home_view_displayed(world: &mut E2eWorld) {
...
+        {
+            break;
+        }
+        assert!(
```
rominf, line 280: Non-blocking: this hand-rolled poll loop gives up two fast-fail paths that every sibling step inherits from `wait_for_screen` (`tui_driver.rs:416-467`) — child-exit detection and PTY-reader-thread panic propagation.

The consequence is specific to this PR: if the daemon crashes in the very code path being changed here, the step spins the full timeout and then reports a misleading "telemetry did not appear" instead of surfacing the crash. Reusing `wait_for_screen` with a predicate would keep both.

tests/e2e-cucumber/features/dash.feature
```diff
@@ -123,3 +123,10 @@ Feature: Interactive dashboard
     Then the launcher shows the model serving
     When the user quits the launcher
     Then the launcher exits successfully
+
+  @id:dash-rocm-29846-wsl-gpu-telemetry @requires-wsl @requires-gpu
```
rominf, line 127: **Blocking — this scenario executes on no lane, so the coverage the PR description claims doesn't exist.**

`tests/e2e-cucumber/src/capability.rs:483-486` gates on `gfx_target.is_some() && (!is_wsl || driver_status == WSL_DRIVER_READY)`. For this exact head, the `e2e-wsl` job (run `34406697117`, job `102651494805`) reports `driver_status: wsl_rocdxg_missing`, `wsl_dxg_device: true`, `wsl_dxcore: true`, `wsl_librocdxg: false`, `wsl_ldconfig_librocdxg: false`, `has_amd_gpu: false`. The run's own `platform.json` artifact records this scenario as `"expected": "skip", "reason": "requires an AMD GPU; none detected on this host"`. Skips are filtered out before cucumber counts them (`tests/e2e.rs:1219-1254`), so the lane reports PASS having never run it — the description's "it will run on the `e2e-wsl` lane" isn't true today.

AGENTS.md:86,99 accepts a gated lane only if it actually exercises the scenario, and AGENTS.md:98 is explicit that a unit test doesn't discharge the requirement.

In fairness, this isn't something you introduced: the runner's missing `librocdxg` is a pre-existing problem tracked as issue #223. It explains the gap rather than excusing it, and it's worth saying so in the PR description rather than claiming coverage that won't happen.

tests/e2e-cucumber/features/dash.feature
```diff
@@ -123,3 +123,10 @@ Feature: Interactive dashboard
...
+  Scenario: dash-11 - Dashboard displays GPU telemetry through the WSL amd-smi backend
```
rominf, line 128: **Blocking — this won't merge, and the rebase that fixes the conflict hard-fails a required check.**

`git merge-tree` shows conflicts in `apps/rocm/src/dash.rs`, `crates/rocm-dash-daemon/src/runner.rs`, and this file. Main already has `dash-11` = `@id:dash-replay-missing-file-fails-fast`, landed in `0595b926` after this PR's merge base.

`tests/e2e-cucumber/tests/feature_naming.rs:133-146` and `:148-165` are ordinary `cargo test` contracts: scenario N must be named `dash-{N:02} - `, and indexes must be unique suite-wide. After rebase this scenario is #12 while still named `dash-11`, so **both** tests panic — in a required check. That's also the explanation for your local `cargo test --workspace` passing: at the merge base it genuinely was #11.

Renumber to `dash-12` when you rebase.

Separately, minor wording: the scenario title names the internal backend (`amd-smi`). No other dash or bench scenario title names a backend — the title is the behavior, not the implementation.

tests/e2e-cucumber/features/dash.feature
```diff
@@ -123,3 +123,10 @@ Feature: Interactive dashboard
...
+    When the user opens the dashboard
+    Then the dashboard reports live WSL GPU telemetry
```
rominf, line 130: Non-blocking Gherkin convention points, grouped:

- This Then step names the transport (`WSL`). No other dash or bench Then step names a platform — the platform belongs in the tag, the step in behavior terms.
- On line 127, `@id:dash-rocm-29846-wsl-gpu-telemetry` is the only ticket-prefixed id among ~115 in this suite. **To be clear, this is not a policy complaint** — AGENTS.md permits ticket IDs and nothing here leaks. It's a convention inversion: this file's own habit is to put the ticket in a preceding comment (`# EAI-8190 regression:` on dash-10, `# EAI-8366:` on dash-11) and keep the `@id` descriptive.
- Also on line 127: all three existing `@requires-wsl` scenarios carry a short comment explaining why the tag is needed. This one has none, which makes the gate harder to audit later — and given the point above about the scenario not actually running anywhere, that explanation would earn its keep.

rominf (changes requested):
I read through this one closely, and I want to start with what it gets right, because the hard part — the diagnosis — is correct and well targeted. The pre-flight really is what blanks the WSL dashboard; nothing downstream of it is doing WSL-specific gating. I checked that independently: the dash crates contain no other WSL or KFD gating at all, so this is the right layer for the reported symptom. The rename is thorough (doc comments updated in step, no leftover `kfd_preflight`/`kfd_accessible` anywhere), the warning strings were made transport-neutral rather than left to rot, and parameterizing `gpu_device_accessible` over the two device paths is a sensible testability seam.

That said, I think there are four things that block merge, and one of them is a design/safety question rather than a line fix.

**1. The new e2e step can never pass.** `"GPUs · "` has exactly one producer — `gpu_wall` in `crates/rocm-dash-tui/src/ui/dock.rs` — reached only behind `dock::is_wide`, which needs `cols >= 180 && rows >= 45`. The only resize helper the driver offers, `use_detail_size()`, sets 120x40 (spawn size is 80x24). So on a fully ROCDXG-ready WSL box the step can only time out and panic "WSL GPU telemetry did not appear" — the fix's only user-facing evidence would report the fix as broken. Details and the *unsafe* obvious workaround are in the inline comment.

**2. The scenario runs on no lane**, so the CI coverage the description claims does not exist. The capability gate requires `driver_status == wsl_rocdxg_ready`, and the `e2e-wsl` job for this exact head (run `34406697117`, job `102651494805`) reports `wsl_rocdxg_missing` / `librocdxg: false` / `has_amd_gpu: false`. The run's own `platform.json` records the scenario as `"expected": "skip"`. Skips are filtered before cucumber counts them, so the lane goes green with the scenario never executed. To be fair to you: the runner state is a pre-existing problem already tracked as #223 — that explains the gap, it just doesn't close it, and AGENTS.md is explicit that a unit test doesn't discharge the e2e requirement.

**3. The pre-flight invents a third WSL GPU verdict and widens it into a safety question.** The repo already consolidated this into `rocdxg_ready()` (dxg device AND dxcore AND librocdxg AND ldconfig), specifically so `examine` and `serve` stop disagreeing — its doc comment says so verbatim. `/dev/kfd || /dev/dxg` is a fourth, looser answer, and your own CI box is the proof case: dxg present, librocdxg absent, so this pre-flight passes while `examine` prints `wsl_rocdxg_missing` and `serve` refuses. The higher-severity angle is in the inline comment: `/dev/dxg` is created by WSLg passthrough and is vendor-agnostic, so this also admits Intel/NVIDIA WSL2 hosts into the path the guard exists to keep them out of.

**4. It won't merge, and rebasing hard-fails a required check.** `dash-11` is already taken on main (`@id:dash-replay-missing-file-fails-fast`, landed in `0595b926` after your merge base). Post-rebase this scenario is #12 while named `dash-11`, which panics two ordinary `cargo test` contracts in `feature_naming.rs`. That's also why your local `cargo test --workspace` passed — at the merge base it genuinely was #11.

Non-blocking, mostly left inline: the new unit test kills only one of five mutants; two of the four conditions in the poll loop carry no signal (one of them is outright unreachable); the hand-rolled loop gives up the child-exit and reader-panic fast-fail paths that `wait_for_screen` provides; and a few Gherkin convention points on the scenario. One thing not inline: the PR text's "six unrelated existing unexpected failures" is asserted without evidence — none of the named scenarios have `expectations.toml` entries, so by the harness's own model they resolve to expected-pass. AGENTS.md:24 asks for verified-vs-assumed to be explicit. Also informational: open PR #399 touches `dash_steps.rs` too, so there's some future conflict risk there.

**Things I checked that are *not* problems** — so you don't spend time on them:

- Native-Linux detection is unaffected: `/dev/dxg` never exists there, so the predicate reduces to the old KFD check.
- The sibling `/dev/kfd` call sites are not dashboard-facing (`main.rs` feeds `rocm driver status`; `serve` routes through `usable_amd_gpu_indices` → `rocdxg_ready()`). Your fix is at the right layer.
- License manifest needs no action: `about.toml` sets `ignore-dev-dependencies = true`, and `tempfile 3.27.0` is already in `THIRD_PARTY_NOTICES.txt` via a non-dev dep.
- `tempfile = "3"` matches convention — 6/6 crates use the bare inline form, and there's no `[workspace.dependencies]` entry.
- Omitting `@requires-os:linux` is correct and deliberate; `examine.feature` and `diagnose.feature` both document that it can't express "really WSL".
- The rename is complete, and every consumer of the changed warning strings matches on the `"amd-smi unavailable"` substring, so nothing breaks.
- The `docs/ci-hardware-testing.md` staleness raised in the earlier CHANGES_REQUESTED is a rebase artifact, not an action item — main already made that wording plural in `a9937493` (#340). It fixes itself on rebase.

siloteemu:
🔴 Automated review · pr-review-watcher · 1611f58d

## Summary

Forced re-check of the change request filed on 2026-09-09. The head commit has not moved (`1611f58`), so this is a re-examination of our own claims against the same code, plus a check of whether `main` has independently changed the answer.

**Outcome: the blocking finding still holds.** I re-derived it from scratch rather than re-reading our own text: `"GPUs · "` has exactly one producer in the tree, `gpu_wall` in `crates/rocm-dash-tui/src/ui/dock.rs:74`; `gpu_wall` has exactly one call site, `crates/rocm-dash-tui/src/ui/mod.rs:85`, which sits inside `if dock::is_wide(area.width, area.height)`; `is_wide` requires `cols >= WIDE_COLS (180) && rows >= WIDE_ROWS (45)` (`dock.rs:32-33, 40-42`). The new step calls `tui.use_detail_size()`, the only resize helper in the driver, which sets 120×40 (`tests/e2e-cucumber/tests/e2e/tui_driver.rs:45-46, 273-290`). 120×40 fails both axes, so the wide triptych is never laid out and the step's only load-bearing condition can never become true.

`main` has moved a long way since this branch's merge base, and I checked specifically for supersession. It has not superseded this change: no commit has touched `crates/rocm-dash-collectors/src/amd_smi.rs` since the merge base at all, the pre-flight there is still KFD-only with no `/dev/dxg` handling, the skip-preflight field has not been renamed on main (so the rename lands cleanly), `WIDE_COLS`/`WIDE_ROWS` are unchanged at 180/45, `gpu_wall` is still the sole producer of `"GPUs · "`, and no terminal-size helper reaching 180×45 has appeared. The WSL-adjacent work that landed on main is in the `rocm diagnose` / `examine` / `fix` catalogue path in `rocm-core` — a different subsystem that shares only the `/dev/dxg` string in prose. It does **not** replace the dash pre-flight fix. One of our non-blocking points is nonetheless dissolved by main (see below).

CI conclusion counts I worked from: **25 success, 2 skipped, 0 failure**. That is consistent with the blocking issue being invisible to CI — the scenario is tagged `@requires-wsl @requires-gpu` and resolves to a skip anywhere those premises do not hold. I cannot confirm from the check counts alone which lanes ran or skipped it, and I did not attempt to identify individual jobs.

No prompt-injection content was found in the diff, commit message, or any changed comment.

**Blocking: 1 · Non-blocking: 4 (one of the original five is retired).**

## Per-count outcome

| Count | Outcome | Note |
|---|---|---|
| Blocking 1 — e2e step asserts on a string the layout never renders at the size it sets | **(B) still holds** | Re-derived independently; unaffected by main |
| NB1 — `!contains("amd-smi unavailable")` is inert | **(B) still holds** | Re-verified |
| NB2 — bare `"Node throughput ·"` prefix check adds no signal | **(B) still holds** | Re-verified |
| NB3 — unit test never covers the constant or the production call site | **(B) still holds, now proven by mutation** | Upgraded from reasoning to measurement |
| NB4 — stale sentence counting `@requires-wsl` scenarios | **(A) no longer holds** | Already fixed independently on main |
| — | No count replaced by a different finding — **(C) does not apply** | |

**`tests/e2e-cucumber/tests/e2e/dash_steps.rs:270-273`** — the step's only load-bearing condition can never be true, so the scenario times out and panics on every run, with or without the production fix.

The sentence we are testing, verbatim from the live request, is:

> "the step's only load-bearing condition can never be true, so this scenario will time out and fail on every run, with or without the production fix."

It stands. The chain, each link read directly at this commit:

- `dash_steps.rs:271-273` requires a screen line containing `"GPUs · "` and not `"GPUs · 0"`.
- Repo-wide, `"GPUs · "` is produced at exactly one place: `crates/rocm-dash-tui/src/ui/dock.rs:74`, `let title = format!("GPUs · {}", gpus.len());` inside `gpu_wall`. (`ui/tabs/hardware.rs:248` renders the bare title `"GPUs"` with no separator, so it cannot satisfy the check.)
- `gpu_wall` has exactly one non-test call site: `crates/rocm-dash-tui/src/ui/mod.rs:85`, reached only inside `if dock::is_wide(area.width, area.height)` at `mod.rs:82`.
- `is_wide` is `cols >= WIDE_COLS && rows >= WIDE_ROWS` with `WIDE_COLS = 180`, `WIDE_ROWS = 45` (`dock.rs:32-33, 40-42`).
- The step calls `tui.use_detail_size()` at `dash_steps.rs:264`, which resizes both the PTY and the emulated screen to `DETAIL_ROWS = 40`, `DETAIL_COLS = 120` (`tui_driver.rs:45-46, 273-290`). It is the **only** resize helper in the driver — I grepped for every `set_size`/`PtySize` use; there is no wider one on this branch, and none on current main either.

120×40 fails both the column and the row threshold, so the wide triptych is never laid out, `gpu_wall` never runs, and the loop at `dash_steps.rs:266-284` always exhausts `default_timeout()` and panics with the message at line 282.

Why this is blocking rather than a nit: the scenario is the PR's only user-observable evidence for the behaviour change, and it currently proves nothing. Worse, it fails for a reason unrelated to the change, so the failure reads as "the WSL fix does not work" to anyone triaging it.

**What would resolve it**, precisely: add a wide-layout resize helper to `tui_driver.rs` that sets at least `dock::WIDE_COLS` × `dock::WIDE_ROWS` (180×45) and call it in place of `use_detail_size()`, then re-verify on the lane that actually runs `@requires-wsl` scenarios. Alternatively, assert on a GPU signal the single-column layout genuinely renders at 120×40 — but then the assertion must be chosen by reading what that layout emits, not assumed.

## Non-blocking

- **`tests/e2e-cucumber/tests/e2e/dash_steps.rs:276`** — `!screen.contains("amd-smi unavailable")` is inert. Re-verified: that text reaches the screen only through `draw_gpus` in `crates/rocm-dash-tui/src/ui/tabs/hardware.rs:227-250`, which this scenario never navigates to, and only when `snap.gpus.is_empty()` — a state the GPU-count condition already excludes. The only other TUI use of `warnings` is `ui/mod.rs:346`, which renders a count, not the text. Drop the check, or navigate to the panel that shows it.

- **`tests/e2e-cucumber/tests/e2e/dash_steps.rs:274`** — `"Node throughput ·"` is emitted unconditionally by `node_throughput_title` (`crates/rocm-dash-tui/src/ui/tabs/home.rs:131`; its own test at `home.rs:571` asserts the no-telemetry form `"Node throughput · Unknown GPU"`). Only the paired `!contains("Unknown GPU")` carries signal; the bare-prefix check adds nothing.

- **`crates/rocm-dash-collectors/src/amd_smi.rs:371-381`** — the new unit test exercises `gpu_device_accessible`'s OR-logic with synthetic tempfile paths and never touches `DXG_DEVICE` or the production call site at lines 75-77. This round it was mutation-tested at branch level in a throwaway copy, running only `cargo test -p rocm-dash-collectors --lib amd_smi`:
  - reverting the predicate to KFD-only → test **fails** (good; it does discriminate the headline change);
  - changing `DXG_DEVICE`'s literal to `"/dev/dxgkrnl"` → test **passes**;
  - swapping the call site to pass `KFD_DEVICE` twice → test **passes** (only a dead-code warning on the now-unused constant).

  So the two most likely real-world mistakes in this diff ship green. One cheap line closes it: `assert_eq!(DXG_DEVICE, "/dev/dxg")`, or drive the assertion through `gpu_device_accessible(Path::new(KFD_DEVICE), Path::new(DXG_DEVICE))`. Not blocking, because the predicate's own logic is genuinely covered.

- **`tests/e2e-cucumber/features/dash.feature:127` and `dash_steps.rs:282`** — the tracker reference sits in the `@id` tag and the panic message, inverting the convention one scenario above (`dash-10` keeps a descriptive `@id` and puts its reference in a comment). The `ROCM-` prefix used here is also the only one of its kind in the tree; 43 files use the other prefix. Separately worth a line in the PR text, and reinforced by the repo's own documentation: `docs/ci-hardware-testing.md` already states that "`/dev/dxg` and dxcore alone are not enough, `librocdxg.so` and its ldconfig entry must be present too". An openable `/dev/dxg` is therefore a weaker condition than `WslSummary::rocdxg_ready` (`crates/rocm-core/src/lib.rs`), so dash can now attempt amd-smi on a host `rocm examine` reports as `wsl_rocdxg_missing`. That is arguably the right trade — it surfaces a warning rather than silently reporting no GPU — but it is a deliberate divergence from a documented statement and deserves a sentence in the PR description rather than being left for a reader to discover.

## Retired finding

- **`docs/ci-hardware-testing.md`** — we flagged the sentence "the one scenario whose premise *is* a WSL host carries `@requires-wsl`" as made stale by this PR. **This no longer holds.** That sentence has since been rewritten on main to the plural form ("scenarios whose premise *is* a WSL host carry `@requires-wsl`, and this is the only lane that runs them"), and main now carries three such scenarios independently of this PR. The finding was correct when filed against the branch's base; it dissolves on rebase and needs no action from the author.

  Cause classification: neither reviewer error nor misleading code — the codebase moved underneath a correct observation. **Will this confusion recur?** Yes, for any long-lived branch whose diff includes documentation prose that main also edits. The cheap guard is not a code comment but a review habit: read prose findings against the *target* branch's copy of the doc, not the branch's. No repository change is warranted for this one.

## Verification notes

What was verified **by reading**: the entire `"GPUs · "` producer/consumer chain, the `is_wide` thresholds, the driver's resize helpers, the string-producer locations for both inert assertions, the completeness of the identifier rename across all four touched files, every new or changed comment and doc line checked against the file it describes, and main's state for each of the five areas above.

What was verified **by running**: the three branch-level mutations of the new unit test, each with `cargo test -p rocm-dash-collectors --lib amd_smi` in a scratch copy — the checkout under review was not modified.

What could **not** be executed here: the WSL/Windows detection path itself, and the e2e scenario, which requires a WSL host with GPU access. The full suite was not run here. The blocking finding does not depend on running either — it is a static reachability argument, and every link in it is quoted above.

Prose-against-code sweep: the changed doc comments in `amd_smi.rs`, `runner.rs` and `end_to_end.rs`, and the two reworded runner diagnostic strings, were each checked against the code they describe and are accurate. The D-state claim about the external `amd-smi` binary is inherited domain knowledge, not checkable from this repository; it is not flagged as false, only as unverifiable here.

Separately, and independent of whether our finding is correct: the branch conflicts with its base and needs a rebase regardless, and another reviewer also holds a change request on it.

**Finding state: blocking 1, non-blocking 4 (one retired). The change request stands as filed, minus the documentation point.**

PR #380, opened 2026-09-10
test(dash): make throughput expiry deterministic

### LINKED ISSUE #379: e2e: dash-08/dash-09 validity-window scenarios are flaky under runner load (EAI-7960 regression tests race real wall-clock)
## Summary

\`dash.feature\`'s EAI-7960 scenarios assert against a real wall-clock validity window instead of a mocked/injected clock, and flake under GitHub-hosted-runner CPU contention when the shared \`e2e\` test binary runs enough scenarios.

- \`@id:dash-gen-tps-held-after-scrape-failure\` (dash-08 — "Gen throughput stays visible for the validity window after a scrape failure")
- \`@id:dash-gen-tps-expiry-boundary\` (dash-09 — "Gen throughput expires after the validity window following sustained failure")

Both assert `runner.rs`'s gen_tps hold/expiry behavior against `clamp(3 × instance_tick, 6s, 30s)` — a real-time window — rather than a mocked clock.

## Evidence (from PR #329)

The GitHub-hosted \`E2E tests\` job failed 3/3 reruns on PR #329's branch, each time on the same assertion:

\`\`\`
Scenario: dash-08 - Gen throughput stays visible for the validity window after a scrape failure
 ✘  Then generation throughput remains visible within the validity window
    Step panicked. Captured output: EAI-7960 REGRESSION: gen throughput ("tok/s") was cleared
    immediately after the first failed scrape instead of being held for the validity window
    (clamp(3 × instance_tick, 6 s, 30 s)).
\`\`\`

One rerun also tripped the sibling \`dash-09\` expiry-boundary scenario on the identical kind of assertion.

Both scenarios pass 100% of the time locally, alone or together:

\`\`\`
cargo xtask e2e -- -n 'dash-08|dash-09'
# 2 scenarios (2 passed), 18 steps (18 passed)
\`\`\`

\`main\`'s \`E2E tests\` job has not failed on this recently (0 unexpected failures across its last several runs), but it currently runs 82 scenarios versus 87 on PR #329's branch — the extra scenario load from that PR's new hermetic \`therock-next-*\` scenarios plausibly tips this already-marginal real-time assertion into failing under runner contention, since all scenarios share one test binary/process.

## Suggested fix

Make \`dash-08\`/\`dash-09\` deterministic by driving the validity window off a mocked/injected clock (matching the pattern other timing-sensitive dash scenarios in this suite already use), instead of racing real wall-clock time against CI runner scheduling jitter.

## Impact

Intermittent, load-dependent CI failures on the \`E2E tests\` (GitHub-hosted) required check, unrelated to the actual change under review, that can block otherwise-ready PRs and require repeated reruns.

michaelroy-amd:
## Summary
- add an E2E-only logical clock seam to the dashboard daemon
- drive dash-08 and dash-09 across hold/expiry boundaries by logical time and observed scrapes
- forward scenario environment fixtures into PTY-launched CLI processes

## Verification
- cargo xtask e2e -- -n 'dash-08|dash-09' (4 consecutive runs; 2/2 scenarios, 20/20 steps each)
- cargo test --workspace --all-targets
- cargo clippy --workspace --all-targets -- -D warnings
- python3 scripts/smoke_local.py

Pure test determinism/CI plumbing; no user-visible production behavior changes. The clock seam is unreachable unless the rocm/e2e-test-hooks feature is built and its test-only environment path is supplied.

Fixes #379

### DIFF
--- a/apps/rocm/src/dash.rs
+++ b/apps/rocm/src/dash.rs
@@ -69,7 +69,18 @@ pub fn runner_options(
         // Production always runs the real `/dev/kfd` pre-flight; only daemon
         // integration tests with a fake binary skip it.
         amd_smi_skip_kfd_preflight: false,
+        test_clock_offset_path: dash_test_clock_offset_path(),
     }
+}
+
+#[cfg(feature = "e2e-test-hooks")]
+fn dash_test_clock_offset_path() -> Option<std::path::PathBuf> {
+    std::env::var_os("ROCM_CLI_DASH_TEST_CLOCK_OFFSET_PATH").map(Into::into)
+}
+
+#[cfg(not(feature = "e2e-test-hooks"))]
+const fn dash_test_clock_offset_path() -> Option<std::path::PathBuf> {
+    None
 }
 
 /// API key precedence — sourced from the environment ONLY (never TOML/CLI/source/
--- a/crates/rocm-dash-daemon/src/runner.rs
+++ b/crates/rocm-dash-daemon/src/runner.rs
@@ -82,6 +82,11 @@ pub struct RunnerOptions {
     /// `amd_smi_binary` at a deliberately-slow fake script flips this, so the
     /// off-critical-path detection behaviour is genuinely exercised.
     pub amd_smi_skip_kfd_preflight: bool,
+    /// **Test-only.** When set, cycle timestamps advance from a fixed epoch by
+    /// the logical tick count plus the signed seconds stored in this file.
+    /// Production callers leave this unset; E2E scenarios use it to cross
+    /// observation-validity boundaries without racing wall-clock scheduling.
+    pub test_clock_offset_path: Option<PathBuf>,
 }
 
 impl Default for RunnerOptions {
@@ -102,6 +107,7 @@ impl Default for RunnerOptions {
             services_dir: None,
             amd_smi_binary: None,
             amd_smi_skip_kfd_preflight: false,
+            test_clock_offset_path: None,
         }
     }
 }
@@ -122,6 +128,27 @@ const fn vllm_metrics_enabled(opts: &RunnerOptions) -> bool {
     !opts.disable_vllm_metrics
 }
 
+fn cycle_timestamp(
+    epoch: DateTime<Utc>,
+    tick: Duration,
+    tick_count: u64,
+    offset_path: Option<&std::path::Path>,
+) -> DateTime<Utc> {
+    let Some(path) = offset_path else {
+        return Utc::now();
+    };
+    let tick_millis = tick.as_millis().saturating_mul(u128::from(tick_count));
+    let logical_millis = i64::try_from(tick_millis).unwrap_or(i64::MAX);
+    let offset_secs = std::fs::read_to_string(path)
+        .ok()
+        .and_then(|value| value.trim().parse::<i64>().ok())
+        .unwrap_or_default();
+    epoch
+        .checked_add_signed(chrono::TimeDelta::milliseconds(logical_millis))
+        .and_then(|at| at.checked_add_signed(chrono::TimeDelta::seconds(offset_secs)))
+        .unwrap_or(DateTime::<Utc>::MAX_UTC)
+}
+
 /// Loop forever: tick host + gpu metrics + bench rows, apply through reducer, broadcast.
 ///
 /// `tick_override` lets tests run faster than `opts.gpu_tick`; production passes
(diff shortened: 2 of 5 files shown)


siloteemu:
🔴 Automated review · pr-review-watcher · 6a3d6806

## Summary
Replaces the wall-clock sleeps that defined the gen-throughput hold/expiry boundaries in dash-08/dash-09 with a test-only injectable logical clock (`RunnerOptions::test_clock_offset_path`, read from a file each cycle, enabled only under the `e2e-test-hooks` feature), and refreshes the now-stale "this is RED" comments. No blocking findings. Verified: `cargo test -p rocm-dash-daemon --lib runner::tests::test_clock` passes (1 passed); confirmed by reading source that the hold/expiry policy in `crates/rocm-dash-core/src/observation.rs` is untouched and is driven purely by the timestamps the runner passes in (`snapshot()` compares `now.signed_duration_since(observed_at) > validity`), that `cycle_at` is the single time source for every observation, tracker snapshot and the assembled `Snapshot`, that the only production `RunnerOptions` construction (`apps/rocm/src/dash.rs:40-84`) yields `None` without the feature, and that the e2e build path (`xtask/src/e2e.rs`) and every CI lane build with `--features rocm/e2e-test-hooks`. Blocking: 0 · Non-blocking: 5.

**Standing focus (a) — does the test still fail against the defect it was written for?** Yes, for both boundaries, and the assertions were not weakened. The production hold/expiry logic is not part of this diff and is still exercised end-to-end: `Then generation throughput remains visible within the validity window` is the same unchanged `screen_text().contains("tok/s")` check, so reverting to clear-on-failed-scrape still makes it fail; `Then generation throughput is no longer displayed` is now reached by pushing the injected clock 7 s past the 6 s window (`clamp(3 × instance_tick, 6 s, 30 s)` with the default 2 s `instance_tick`) and then synchronising on the next real failed scrape, so a never-expiring held value still fails it. Margin is roughly 9 s of logical time against a 6 s window, not knife-edge. The determinism win is real rather than cosmetic: the logical clock advances by the tick duration per executed loop iteration, so a descheduled host can no longer consume the window between ticks. The new unit test only covers the `cycle_timestamp` arithmetic and would pass with the hold/expiry logic reverted — that is fine, since it is scaffolding coverage, not the regression test.

**Standing focus (d):** no CI outcome is attributed to any named job here.

**None.**

## Non-blocking
- `tests/e2e-cucumber/features/runtime_setup.feature:108-116` — the comment justifying runtime-06's `@nightly` tag still says the two throughput scenarios are "timing-fragile under load ... which is their own problem to fix; until then this runs on the nightly lanes". This PR is that fix, but the workaround and its rationale are untouched; either revisit the tag or correct the comment, otherwise the next reader is told a fragility exists that the same PR removed.
- `tests/e2e-cucumber/tests/e2e/dash_steps.rs:717` — the `"7"` offset is a magic constant rather than derived from the validity contract (`MIN_VALIDITY` / `observation_validity()` in `crates/rocm-dash-core/src/observation.rs`); if the clamp's lower bound is ever raised the expiry assertion starts failing and will read as a product regression.
- `tests/e2e-cucumber/tests/e2e/dash_steps.rs:722` (and the pre-existing one at `:686`) — after synchronising on the scrape count the step still sleeps a fixed 50 ms for the daemon→TUI→PTY render, and the following `Then` reads the screen once instead of polling; the last arbitrary wait could be removed with the `wait_for_screen` pattern used elsewhere in the file.
- `tests/e2e-cucumber/tests/e2e/dash_steps.rs:632-644` — no guard for a binary built without `e2e-test-hooks`: the env var is then ignored, the clock silently falls back to `Utc::now()`, and the scenarios quietly return to being wall-clock-fragile. `tests/e2e-cucumber/tests/e2e/serving_steps.rs:958-965` sets the precedent of naming that misconfiguration in the failure message.
- `tests/e2e-cucumber/tests/e2e/tui_driver.rs:184-188` — the PTY path applies `world.command_env` by reference, while `tests/e2e-cucumber/tests/e2e.rs:764` consumes it with `std::mem::take` precisely so it "cannot leak into a later action in the same scenario"; harmless today, but the one-shot invariant now differs between the two launch paths.

siloteemu (approved):
pr-review-watcher · 6a3d6806

Reviewed at this head; no blocking findings. The change replaces wall-clock waiting in a test with a test-driven clock, so a previously flaky test becomes deterministic. It is test-only, with no change to the behaviour being tested. Checks are green at this head: 26 success, 1 skipped, 0 failures.

PR #382, opened 2026-09-11
fix(diagnose): stop claiming the CLI will apply fixes it only reports

volen-silo:
## What

`auto_applicable` was a flat `bool` meaning *\"`rocm fix <id>` will carry this out\"*, and it was wrong for two entries:

- **`fix-2-unset-override` on Linux.** `run_unset_override` splits by platform. The Windows arm takes `FixOptions` and does the dry-run/confirm dance; `run_unset_override_linux` takes no options at all — it reads the live override, scans five shell rc files, and reports where it is set. It never writes, so `--dry-run` is a no-op and `--yes` is never read.
- **`fix-9-igpu-dgpu` with no `--device-index`.** Both arms print the query that identifies the discrete GPU and return without acting.

Both were marked auto-applicable, so a user — and an agent reading the same output — was told a change was coming that never came, and got exit 0 confirming it.

The repo already knew. `fix_9_without_device_index_is_print_only_and_returns_zero`, the comment on the fix-2 dry-run test, and a step comment in the e2e suite all say *print-only* for exactly these cases. Only the type disagreed.

## Why a class, not two corrected booleans

A bool cannot say either of these things: that behaviour differs **by platform**, or that it is **conditional on an argument**. `auto_applicable` is now derived from a per-platform class — `AUTO`, `NEEDS-ARG`, `PRINT-ONLY`, `DIAGNOSE-ONLY`. Scope and class are held together because they are one fact: fix-2 applies on Linux as print-only and on Windows as auto.

`DIAGNOSE-ONLY` covers problems that can be detected and explained but have no reliable fix. **It has no member yet** — all 16 entries have real fixes; twelve are print-only because they need sudo, a reboot, a reinstall or a download. The first member arrives with the conflicting code-object-manager entry, which is where its scenario will land.

## Non-obvious decisions

**The class says what happens to the machine, not whether a runner runs.** Gating the runner on the class would have thrown away fix-2's Linux rc-file report and printed a generic *\"copy the commands above\"* in its place. A print-only entry may hold a runner that reports without writing; `a_recipe_that_acts_has_a_runner` asserts only the direction that matters.

**`DIAGNOSE-ONLY` exits 0.** Exit codes signal outcome, and \"explained, nothing to do\" is not a failure — `3` already means \"not applicable *here*\", a different claim that would send a caller looking for another machine. A caller reads the class from the listing before invoking, so it needs no new code to branch on. Adding one before the exit-code map is formalised would also mean the published skill treating it as an error. Flagging it: this is worth a second opinion.

**`auto_applicable` is kept, derived.** It is redundant once `class` exists, but it is the field published consumers read, and it now answers its own question truthfully for the first time. Removing it is a type change that belongs with the contract-version work.

**The listing is per-machine.** The marker shows what the entry does on the host in front of you, so the set marked `AUTO` is now OS-dependent — which is why the e2e expectation is too.

## Killing the duplication

`diagnose` restated applicability by hand at 21 sites. Those are gone; the class is read off the catalog in one place, keyed on the **examined** machine's OS rather than the running one. The two copies had already drifted on `main` — fix-9 was `auto_applicable: true` in the catalog and `false` in the Linux arm of its own checker — and nothing compared them, because the existing cross-check compares only the command block.

## Verification

- `cargo test --workspace --all-targets --exclude e2e-cucumber` — clean; `rocm-core` 341 passed
- `cargo clippy --locked --workspace --all-targets`, `cargo fmt --all --check` — clean
- `cargo xtask e2e -- -n diagnose` — 18 scenarios, 15 passed; the 3 failures are the `@requires-bare-metal` scenarios that cannot pass on a WSL2 dev host, unchanged from before this PR. Confirmed unrelated rather than assumed: `diagnose --json` on this host returns `out_of_scope` set and `matched` empty, so no checker runs and no catalog data could affect them.

**Both new scenarios were verified to fail, not just to pass.** Reverting fix-2's Linux class to `Auto` fails `diagnose-17` (and `diagnose-12`, which pins the auto set). `diagnose-18` initially passed under the mutation — a whole-output search for `--device-index` matched the recipe's own note — so the assertion was narrowed to the flags line, which is absent entirely when the entry is marked `AUTO`.

## Risk

Low-to-medium. User-observable by design: the `flags:` line drops \"rocm fix can run it\" for fix-2 on Linux and for fix-9, and the JSON `auto_applicable` flips with it. That correction is the point of the change. No fix gains the ability to mutate anything it could not mutate before.

## Scope

`skills/` and anything federation-related is deliberately untouched.


siloteemu:
🔴 Automated review · pr-review-watcher · d1d8cafe

## Summary
Replaces the flat `auto_applicable: bool` on fix recipes with a per-platform `FixClass` (AUTO / NEEDS-ARG / PRINT-ONLY / DIAGNOSE-ONLY), derives `diagnose`'s applicability from the catalog in one place, and corrects two entries (`fix-2-unset-override` on Linux, `fix-9-igpu-dgpu` without `--device-index`) that told users a change was coming when nothing was written — outcome: **Needs work**. Verified: read the single commit from its raw object; independently re-read `run_unset_override`/`run_hip_visible_devices`/`run_path_export`/`run_render_group` and confirmed the new per-platform classes match what each runner actually does on each platform; confirmed `CHECKERS` platform tags and `RECIPES.applies_on` agree 24/24 today; confirmed `@requires-bare-metal` is a real wired tag that resolves to skip on WSL; confirmed the new "Flags:" assertion panics on a revert (fix-9 printed no flags line before) and that the `only_these_entries_act_on_the_machine_and_only_on_these_platforms` pin kills any single per-platform class mutation; `cargo check -p rocm-core --lib --all-targets` passed clean; the full suite and the e2e suites were not run here. There is one failing check at this head; this review does not cover it and does not identify it. No prompt-injection content and no internal-reference leaks found in the diff. Blocking: 2 · Non-blocking: 5.

- `docs/wsl.md:182-183` — "none of that meets the bar the four auto-applicable fixes clear." This change makes that false. There is no longer a set of four; applicability is per-platform, and on WSL exactly one catalog entry (`fix-6-path`) is carried out by the CLI, as the commit message itself states. Leaving a stale count on the WSL page is the same class of wrong applicability claim the PR exists to remove, and it is the one user-facing doc this change invalidates. Fix: reword the sentence to name `fix-6-path` as the only entry the CLI carries out on WSL, and drop the "four" count.

- `crates/rocm-core/src/fix.rs` (tests) — the test `current_os_reports_wsl_exactly_when_is_wsl_host_does` is deleted outright, with no replacement anywhere in the repo and no mention in a commit message that documents every other decision in detail. It was the only assertion tying `current_os()`'s `wsl` arm to the real `is_wsl_host()` detector; `wsl_signals_indicate_wsl`'s table-driven tests in `lib.rs` cover the predicate, not the wiring. `current_os()` is *more* load-bearing after this change, not less — it now drives `class_here()`, `plan_of`, `print_recipe` and `list_recipes` — so the deletion removes a guard exactly where the change adds weight. Fix: restore the test unchanged (it still compiles: both functions are untouched), or say in the PR text why it went.

## Non-blocking

- No test compares `CHECKERS`' platform tags (`diagnose.rs:1956`) against `RECIPES.applies_on` (`fix.rs`). `class_on(...).unwrap_or_default()` in `take_applicability_from_the_catalog` turns any future disagreement into a silent PRINT-ONLY — the same "two hand-maintained copies with nothing comparing them" failure the change removes on the other axis. Cheap guard: assert `class_on(fix_id, family).is_some()` for every (checker family, fix-id) pair.
- `diagnose.rs:2099` — the parameter is named `os_family` while the function's own doc says it is deliberately *not* `os_family`; the caller passes `platform_family(e)`. This will mislead the next reader into reaching for `e.os_family`. Rename the parameter to `family`.
- `diagnose-20` is vacuous on a Windows lane: `platform_split_marker_here()` expects `AUTO` there, which is also the pre-change value, so the scenario passes identically with the production change reverted. Its title ("A fix that only explains itself here…") also states the opposite of what it asserts on Windows. Real coverage comes from the Linux/WSL lanes only; worth saying so, or scoping the scenario.
- Seven hand-written `auto_applicable: false` initialisers survive in `diagnose.rs` (`check_19_shm_too_small` and the six `fix-wsl-*` checkers) while every other checker's was removed. They are now dead writes, overwritten by the new derivation, and they re-create the duplication the change eliminates elsewhere.
- Two comment/text nits: removing `auto_applicable: false` from the `fix-17-torch-dlpack` recipe left the multi-line comment block for `commands` appended to the end of the very long `rationale` line (`fix.rs:522`); and `fix.rs:1754` refers to "the Epic's scenario 10", which no reader of this repo can resolve — restate it in repo terms or drop it. Separately, the new `render_report_text` strings ("rocm fix can run it once given an argument", "no reliable fix; rocm fix will only explain it") and the new `class` field's JSON shape are pinned by nothing.

_Check-run conclusions at this head, both when this review started and re-read immediately before posting: **23 success, 1 failure, 1 pending, 1 skipped**. A failing check is present at this head and this review does not cover it. Conclusion counts only; no per-lane detail was available, so nothing above is attributed to any named job._

siloteemu (changes requested):
🔴 Automated review · pr-review-watcher · d1d8cafe

First review of this PR. The change itself is sound: replacing the flat auto-applicable boolean with a per-platform classification, deriving the diagnose report's applicability from the catalog in one place, and correcting the two entries that told users a change was coming when nothing was written. Two counts gate it. The full report, including five non-blocking observations, is in the review comment on this PR.

**Blocking 1: the change makes a claim on the WSL page false, and it is the same class of wrong applicability claim this PR exists to remove.**

`docs/wsl.md:182-183` says none of that meets the bar the four auto-applicable fixes clear. After this change there is no longer a set of four: applicability is per-platform, and on WSL exactly one catalog entry is carried out by the CLI, as this PR's own commit message states. That page is the one user-facing document this change invalidates, and leaving a stale count there re-creates the defect the PR is fixing everywhere else. Suggested fix: reword the sentence to name the single entry the CLI carries out on WSL and drop the count.

**Blocking 2: a test tying the platform detector to its only caller is deleted with no replacement and no mention.**

`current_os_reports_wsl_exactly_when_is_wsl_host_does` is removed outright. Nothing in the repo replaces it, and the commit message — which documents every other decision in detail — does not say it went. It was the only assertion tying the `wsl` arm of the platform lookup to the real host detector; the table-driven tests elsewhere cover the predicate, not that wiring. That lookup is more load-bearing after this change, not less, since it now drives the per-platform class, the plan, the recipe printer and the recipe listing. Removing the guard exactly where the change adds weight is what gates this. Suggested fix: restore the test — both functions it touches are unchanged, so it still compiles — or say in the PR text why it went.

_Check-run conclusions when this was filed, and when the review started: 23 success, 1 failure, 1 pending, 1 skipped. A failing check is present at this head; this review does not cover it and makes no claim about what it is. Conclusion counts only; no per-lane detail was available, so nothing here is attributed to any named job._

PR #386, opened 2026-09-11
feat(diagnose): report a shared memory allowance too small to serve on

volen-silo:
## What

A serving workload needs gigabytes of `/dev/shm`. A container gives it 64 MB by default, and WSL2 ships the same default. When the allowance runs out the workload crashes **without the message ever naming shared memory** — a data-loader worker killed by a bus error, or a failed write to a temporary file. There is no route from what the user sees back to the cause.

The inspection now measures the allowance, and the catalog names the shortage before the workload runs.

This is a failure we have seen rather than one we expect: the repo already carries a regression test that provokes a genuine `ENOSPC` on `/dev/shm`, written from an original report.

## Non-obvious decisions

**A direct `statvfs`, not a change to `disk_space`.** `available_space_for_path` cannot see `/dev/shm`, and that is the helper working correctly: `sysinfo` omits tmpfs, so `mount_owns_path` compares device ids and returns "unknown" rather than confidently reporting the root filesystem's free space. The guard is right — this needs the number it declines to guess at.

**The total decides, not the free space.** A 64 MB allowance cannot hold an 8 GB workload even when completely empty, so judging on what is free would miss the case on an idle machine entirely. Both are reported, because "64 MB total" and "64 MB total, 2 MB free" are different conversations.

**Fires below 1 GiB**, which separates the failure from the healthy majority rather than describing what a workload wants:

| Machine | `/dev/shm` | Fires? |
|---|---|---|
| Container, Docker default | 64 MB | **Yes** |
| Ordinary Linux host | half of RAM — ≥ 1 GiB on anything with 2 GB+ | No |
| Container launched as `fix-10-container` advises | 8 GB | No |

It deliberately **under-reports**: a container given 2 GB is still short for a large model and will not be flagged. That is the right way to be wrong. A diagnosis that fires on healthy machines is one people stop reading, and a threshold set at what a workload *wants* would trip every ordinary laptop.

**Established from machine state**, with the error text only raising an already-structural finding. That ordering is forced by the failure itself: the crash does not name shared memory, so an entry that needed the right words would never fire for the user who needs it.

**The steps quote the `8g` that `fix-10-container` already tells users to pass**, rather than inventing a second number that could drift from the first. Print-only: raising the allowance means restarting a container or remounting, and a running container cannot be resized at all.

## Verification

- `cargo test --workspace --all-targets --exclude e2e-cucumber` — clean; `rocm-core` 343 passed
- `cargo clippy --locked --workspace --all-targets`, `cargo fmt --all --check` — clean
- `cargo xtask e2e -- -n diagnose` — 16 scenarios, 13 passed; the 3 failures are the `@requires-bare-metal` scenarios that cannot pass on a WSL2 dev host, unchanged from before this PR

**Verified against the real binary.** This dev host has a 64 MB `/dev/shm` — the exact failure case — and `examine --json` reports `67108864` total, matching `df` exactly.

**The new assertions were verified to fail, not just to pass.** Raising the threshold so an ordinary allowance trips it turns the two control tests red, and only those.

Two guards fired and were updated deliberately: the `Examination` top-level wire contract, and the catalog entry count.

## Known gap — not introduced here, but worth naming

On WSL2 the inspection now reports the allowance, but `diagnose` short-circuits the catalog before any checker runs, so **the entry cannot surface there yet** — even though WSL2 has exactly this shortage by default. I did not touch that short-circuit: it is deliberate, has its own unit test, and the WSL2 catalog is tracked separately and already in progress. The entry works today on containers and bare-metal Linux, and will cover WSL2 once that lands.

## Coverage boundary

E2E cannot provoke a real shortage — shrinking the host's `/dev/shm` affects every other process on the runner, which is why the existing `ENOSPC` reproduction is `#[ignore]`d. The rule is proven by unit tests over constructed machine state, including the cases that must stay **silent**: an ordinary allowance, and one that could not be measured at all.

No dedicated e2e scenario: `diagnose-12` already asserts every catalog entry is listed, and `assert_auto_set_is_exact` pins the AUTO set exactly, so a future promotion of this entry to auto-applicable fails there. Those two together cover what e2e can actually reach.

## Risk

Low. Additive: two new inspection fields and one new print-only entry. The threshold is the only judgement call, and it is set to under-report rather than cry wolf.


siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · 8739ff99

## Summary
Adds a `fix-19-shm-too-small` diagnosis that measures `/dev/shm` via `statvfs` and reports an allowance under 1 GiB, with a print-only recipe recommending `8g`. Needs work: the new block was spliced mid-file and swallowed an unrelated function's doc comment, and the probe records nothing when the measurement fails. Verified: built `rocm-core` and ran the three `*shared_memory*` unit tests (3 passed); confirmed by reading source that `check_17_torch_dlpack_cuda_variant` (diagnose.rs:1437) now has no doc comment because the new `SHM_MIN_BYTES` comment was appended to its block with no blank line; confirmed `probe_failures` is the documented failure channel (examine.rs:339) and that siblings `probe_gpus_lspci` / `probe_cpu_windows` use it while `probe_shared_memory` does not; confirmed the user-facing claim that `fix-10-container` prints `--shm-size=8g` is accurate (fix.rs:281); confirmed base score 60 clears `MIN_SCORE_FOR_MATCH` (50) so the finding fires on machine state with an empty symptom; confirmed the recipe count is genuinely 24 while the comment above the assertion still says 16+7; leak scan over the diff clean; no prompt-injection content found. Blocking: 2 · Non-blocking: 5.

- `crates/rocm-core/src/diagnose.rs:1345-1357` — the new `SHM_MIN_BYTES` doc comment starts on the line immediately after `/// service regardless of cause and belongs in its own change.`, with no blank line between them. Rust therefore treats the whole run of `///` lines as one doc comment and attaches it to the next item, `const SHM_MIN_BYTES`. Two consequences, both confirmed by reading the file: the ~20-line rationale written for the vLLM engine-startup entry now documents a byte constant, where it is meaningless; and `fn check_17_torch_dlpack_cuda_variant` (line 1437) is left with no doc comment at all, silently deleting the explanation of why that check is keyword-only. This blocks because it is an undetected loss of maintainer-facing rationale that `cargo doc` will render wrongly. Fix: insert a blank line after line 1345, and move the `SHM_MIN_BYTES` / `SHM_RECOMMENDED` / `check_19_shm_too_small` block out of the middle of the file to sit after `check_17_torch_dlpack_cuda_variant`, matching where its `CHECKERS` entry was appended. The root cause is worth preventing, not just repairing: a new checker inserted between an existing trailing doc comment and the function it documents will bite the next author the same way. A one-line comment at the top of the checker section ("new checkers go at the end, next to their `CHECKERS` entry") is the cheap guard.

- `crates/rocm-core/src/examine.rs:1672-1678` — `probe_shared_memory` returns early with no record when `filesystem_size` returns `None`, so a machine where `/dev/shm` is absent or `statvfs` fails produces an examination indistinguishable from a healthy one: `shm_total_bytes` stays `None`, `check_19` returns `zero(...)`, `status` stays `ok`, and nothing anywhere says the measurement was attempted and failed. This is exactly the false all-clear the field's own doc comment says must not happen ("a machine we could not measure is not a machine with a shortage") — the check honours that distinction, but the probe throws away the evidence that would let a reader see it. It also contradicts the documented crate contract at examine.rs:339 ("probe errors are recorded in `probe_failures`/`notes`") and the convention every other fallible probe follows (`probe_gpus_lspci`, `probe_cpu_windows` both push a `probe_failures` entry before returning). Since `compute_status` degrades `ok` to `degraded` on a non-empty `probe_failures`, the omission also suppresses the status signal. Fix: on the `None` branch push a `probe_failures` entry, e.g. `e.probe_failures.push(format!("could not query {SHM_PATH}; shared-memory allowance unknown"))`, and add a unit test asserting that an unmeasurable `/dev/shm` leaves a trace rather than silence.

## Non-blocking
- `crates/rocm-core/src/fix.rs:1454` — the comment "16 bare-metal/Windows entries (including fix-17) plus the 7 WSL ones" sums to 23 but now sits directly above `assert_eq!(count, 24)`; the literal was bumped and the comment was not.
- `crates/rocm-core/src/diagnose.rs:1404-1408` — the empty-`container_kind` fallback renders "this is a linux container"; `probe_container` already uses `"container"` as its generic fallback (examine.rs:1731), so use that.
- The 1 GiB threshold is justified in source prose, but its deliberate under-reporting ("a container given 2 GB is still short... will not be flagged") never reaches the user: a 2 GiB container that is genuinely too small gets a clean bill of health with no hint that silence is not proof. A line in `Fix.notes` naming the threshold would close it.
- Test discrimination gaps: all four new tests pass an empty symptom, so `KEYWORDS_SHM_TOO_SMALL` (five new regexes) has zero coverage; the `shm_available_bytes != total` evidence branch is never exercised; and the threshold is only pinned to somewhere between 64 MiB and 8 GiB, so 512 MiB or 2 GiB would pass unnoticed. All four do fail if the production change is reverted (the fields stop existing), so the revert test is satisfied.
- The recipe is registered for `wsl` but its guidance only names "container" and "host", and the curated WSL entry table in `docs/wsl.md:158-172` does not list it — a WSL2 user can hit this finding and finds it documented nowhere.

_Check-run conclusions at this head were skipped 1, success 26 both when the review started and when this was filed; no per-lane detail was available, so nothing here is attributed to a named job._

volen-silo:
Addressed. Both blockers and all five non-blocking items, verified against the source rather than taken on trust.

**Blocking**

1. **Doc comment rebinding.** Confirmed by reading the file: line 1345 ended the previous doc comment and mine began on 1346 with no blank line, so Rust joined the run and bound it all to `SHM_MIN_BYTES`, leaving `check_17_torch_dlpack_cuda_variant` undocumented. The `SHM_MIN_BYTES` / `SHM_RECOMMENDED` / `check_19_shm_too_small` block now sits after `check_17`, matching where its `CHECKERS` entry was appended, and `check_17`'s rationale is reattached. The root cause is guarded as suggested: a note at the top of the checker section states that a new checker goes at the end, and why inserting one between a function and the comment above it is invisible to every gate.

2. **Silent unmeasurable probe.** Confirmed the contract at `examine.rs:339` and the sibling convention in `probe_gpus_lspci`. The `None` branch now pushes a `probe_failures` entry. The probe body took a path parameter so the branch is reachable from a test at all — a hard-coded `/dev/shm` cannot be made to fail on a host that has one — and `an_unmeasurable_shared_memory_allowance_leaves_a_trace` asserts the fields stay `None`, the failure is recorded, and `compute_status` degrades to `degraded`. That last assertion is the point: the trace is load-bearing, not decorative.

**Non-blocking** — all five taken.

- Stale count comment corrected to 17 + 7.
- Container fallback now uses `container`, matching what `probe_container` writes.
- The threshold reaches the user: both notes lists now say it reports below 1 GiB and that silence is not proof of enough, naming the 2 GiB case explicitly.
- Test gaps closed. `the_symptom_text_raises_the_finding_without_being_needed_for_it` covers the keywords and asserts both halves — they raise the score, and the finding still fires without them. `a_partly_used_allowance_reports_what_is_left_as_well_as_the_size` exercises the `available != total` branch.
- `docs/wsl.md` now lists the entry, noting it is not a WSL entry but applies there because WSL2 ships the same 64 MiB default.

**One correction worth recording.** My first attempt at the threshold test was written in terms of `SHM_MIN_BYTES` on both sides, so it moved with the constant: it pinned the rule (`below`, not `at or below`) and left the *value* as unpinned as before. The mutation check caught it — moving the threshold to 512 MiB passed. It now also pins the literal and checks 2 GiB stays silent, and moving the constant fails the test. That was exactly the gap the review named, and my first fix did not close it.

Every new assertion was mutation-checked: stripping `KEYWORDS_SHM_TOO_SMALL` fails the keyword test and only that; moving the threshold fails the threshold test and only that.

siloteemu:
🔴 Automated review · pr-review-watcher · 065e99cc

## Summary
Adds a `/dev/shm` measurement to `Examination` and a `fix-19-shm-too-small` diagnosis/recipe pair that reports a shared-memory allowance below 1 GiB on Linux and WSL. No blocking findings. Verified: a filtered `cargo test -p rocm-core --lib` run of the five new shared-memory/threshold unit tests — all pass (the full suite, e2e and the Windows/GPU lanes were not run here); I also read `compute_status`, `keyword_score`, `format_bytes`, `assert_plan_matches_the_catalog_copy` and every `applies_on` in `RECIPES` to confirm the claims below rather than taking the author's reply or our own earlier note on trust. CI at this head: 23 success, 1 skipped, 2 pending, zero failures. Blocking: 0 · Non-blocking: 5.

## Prior objection status
- **BLOCKING 1 (doc-comment rebinding) — RESOLVED.** `check_17_torch_dlpack_cuda_variant` keeps its full doc comment (`diagnose.rs:1338-1352`, ending "…belongs in its own change.") immediately above its `fn` at 1353. The new block was moved to sit *after* that function, which now ends at 1410; line 1411 is blank, so the `SHM_MIN_BYTES` doc run (1412-1422) binds to the const at 1423, and blank lines at 1424 and 1428 keep `SHM_RECOMMENDED` and `check_19_shm_too_small` bound to their own comments. The requested guard comment is present at `diagnose.rs:406-414`. Residual, and the reason focus (c) applies: that guard sits at the *top* of a ~1600-line checker section but tells the author to append at the end, where they will actually be typing — a one-line repeat next to the `CHECKERS` array (`diagnose.rs:1997`) would close the loop cheaply.
- **BLOCKING 2 (silent probe failure) — RESOLVED.** `probe_shared_memory_at` now pushes `"could not query {path}; shared-memory allowance unknown"` on the `None` branch (`examine.rs:1680-1686`), matching sibling probe wording. I walked `compute_status` (`examine.rs:406-419`) branch by branch for the test's fixture (`os_family="linux"`, `has_amd_gpu=true`, everything else `Default`): `is_wsl=false`, os supported, gpu present — a non-empty `probe_failures` is the *only* thing that can yield `"degraded"` there, so the new test's third assertion is genuinely load-bearing. Deleting the push alone flips `compute_status()` to `"ok"` and fails two of the three assertions.
- **Earlier non-blocking items** — (1) count comment: arithmetic now correct (17 + 7 = 24, and I counted 24 recipes: 17 not `WSL_ONLY`, 7 `WSL_ONLY`), but the *label* is still off (see below) — the wording, not the sum, was the real problem, which our earlier note under-stated. (2) container_kind fallback now emits `"container"`, matching `probe_container`. (3) the under-reporting caveat now reaches the user in both note blocks. (5) `docs/wsl.md:173-179` added. (4) mostly addressed — the keyword table now has three symptom tests, the partly-used branch has one, and the threshold is pinned literally — with one hole left, item 1 below.

**None.**

## Non-blocking
- `crates/rocm-core/src/diagnose.rs:1455-1462` — the `available != total` guard is the one mutation nothing catches: the only equal-case test (`a_tiny_shared_memory_allowance_…`, total = available = 64 MiB) asserts `evidence.contains("64")`, which stays true if the guard is dropped and the redundant "64.0 MiB of it is free" line is emitted. One extra assertion that the equal case does *not* contain "of it is free" would close it, and the author's mutation-check claim does not hold for this branch.
- `crates/rocm-core/src/diagnose.rs:1487-1491` vs `crates/rocm-core/src/fix.rs:594-601` — the two hand-maintained copies of the fix-19 plan already differ on day one (the catalog copy opens with `# Check what you have:` / `df -h /dev/shm`, the diagnose copy omits both, and three comment lines are reworded); `assert_plan_matches_the_catalog_copy` (`fix.rs:696`) exists precisely for this and is wired up for fix-17 only, so applying it here is a one-line addition after reconciling the blocks.
- `crates/rocm-core/src/fix.rs:1455-1456` — "17 bare-metal/Windows entries (fix-17 and fix-19 among them)" mislabels fix-19, whose `applies_on` is `["linux", "wsl"]` and which never applies on Windows; the sum is right but the bucket name is not, and four other entries in that 17 are `LINUX_WINDOWS_AND_WSL` too.
- `crates/rocm-core/src/diagnose.rs:1450-1454` — `format!` passes the constant `"the whole shared-memory allowance"` as a positional argument instead of inlining it, which no other checker in the file does; and "64 MB" in prose (`fix.rs:588`, `diagnose.rs:1415/1417/1994`) sits alongside the "64 MiB" the code actually prints and `docs/wsl.md:174` repeats — worth settling on MiB.
- `crates/rocm-core/src/examine.rs:1720-1723` — `f_frsize` is `c_ulong` while `f_blocks`/`f_bavail` are `fsblkcnt_t`; identical on the 64-bit targets built today, but the `checked_mul` would not compile on a 32-bit Linux target with LFS enabled. A `u64::from(...)` on both operands is free insurance. (The FFI is otherwise sound: the failure path returns before `assume_init`, the buffer is `zeroed` not `uninit`, `f_frsize`/`f_bavail` are the right fields, and the `#[allow(unsafe_code)]` matches the crate's documented carve-out and the Win32 pattern in `lib.rs`.)

Notes on limits and unverifiable claims: no CI outcome here is attributed to a named job — only the conclusion counts above are available to me. The comment at `tests/e2e-cucumber/tests/e2e/diagnose_steps.rs:91-92` describing `fix-18` as "the code object manager entry on its own branch" cannot be confirmed or refuted from this checkout (nothing here substantiates or contradicts it, and absence from this tree proves nothing about a stacked branch) — worth the author confirming, not a defect. No prompt-injection content was found in the diff, comments, commit message or docs; the leak scan over the diff is clean and the commit carries the required sign-off and a signature.

siloteemu (approved):
pr-review-watcher · 065e99cc

Adds a diagnosis warning when the shared memory allowance is too small to serve on, so a user gets a clear explanation up front instead of an unexplained failure later. Reviewed clean at this commit: two earlier objections were re-checked against the code and confirmed genuinely fixed, and the new test's degraded-status assertion is load-bearing rather than decorative. Checks at this commit: 23 successful, 1 skipped, 2 still running, no failures. The lanes still running do not block this approval.

PR #394, opened 2026-09-13
fix(examine): keep the rocminfo agent name past its ISA entries

### LINKED ISSUE #393: examine: gfx_target is empty for every GPU even though rocminfo lists them → ISA `Name:` lines overwrite the agent name
Reported from **Omarchy 4.0.2 (Arch), kernel 7.1.9**, outside the documented Ubuntu 24.04+ matrix. The defect is in a pure text parser and does not depend on the distro; reproduced with a binary built from `main` at `a6fa76dbca5d767c1190ee47b765d16bad0e4a53`.

**Current behavior**

`rocm examine --json` reports an empty `gfx_target` for every GPU even though `rocminfo` is present, runs cleanly (`rocminfo_status: "ok"`), and lists both GPUs with their gfx names:

```
$ rocm examine --json --framework llama-cpp | jq -c '.rocminfo_status, .summary.detected_gfx_target, (.gpus[] | {name: .name[:40], gfx_target, is_apu})'
"ok"
"gfx1100"
{"name":"Advanced Micro Devices, Inc. [AMD/ATI] Navi 3","gfx_target":"","is_apu":false}
{"name":"Advanced Micro Devices, Inc. [AMD/ATI] Raphael","gfx_target":"","is_apu":true}
```

`summary.detected_gfx_target` is right only because it comes from the separate sysfs path (`detect_linux_sysfs_gfx_target`); the per-GPU records never get a target from rocminfo.

**Expected behavior**

`gpus[0].gfx_target == "gfx1100"` (RX 7900 XT) and `gpus[1].gfx_target == "gfx1036"` (Raphael iGPU), matching what rocminfo prints, with `is_apu` derived from the target rather than guessed from the PCI name.

**Steps to reproduce**

Any Linux host with a working `rocminfo` (ROCm 6.x or 7.x):

```
rocm examine --json | jq '.rocminfo_status, [.gpus[].gfx_target]'
# "ok"
# ["", ""]
```

**Root cause**

`probe_gpus_rocminfo` (`crates/rocm-core/src/examine.rs:1094-1116`) walks rocminfo's output keeping a running `cur_name`, and every line that starts with `Name:` overwrites it. rocminfo prints the ISA sub-entries of a GPU agent *after* `Device Type: GPU`, and those lines also start with `Name:`:

```
Agent 2
  Name:                    gfx1100
  Marketing Name:          AMD Radeon RX 7900 XT
  Device Type:             GPU
      Name:                    amdgcn-amd-amdhsa--gfx1100
      Name:                    amdgcn-amd-amdhsa--gfx11-generic
Agent 3
  Name:                    gfx1036
  ...
```

So when the next `Agent` header arrives, `cur_name` is `amdgcn-amd-amdhsa--gfx11-generic`, `cur_name.starts_with("gfx")` is false, and the agent is dropped. The same happens to the last agent at end of input. `gfx_targets` ends up empty, the function returns early, and no GPU record is touched. Re-implementing the loop verbatim over this host's rocminfo output confirms `gfx_targets == []`; accepting only the first `Name:` after an `Agent` header yields `[("gfx1100", "AMD Radeon RX 7900 XT"), ("gfx1036", "AMD Ryzen 5 7500X3D 6-Core Processor")]`.

Every rocminfo since ROCm 5 prints ISA entries this way, so this affects every host where rocminfo is the source of the target.

**Impact**

- `rocm diagnose` filters on per-GPU `gfx_target` (`crates/rocm-core/src/diagnose.rs:354`, `943-950`) to tell the discrete GPU from the APU. With every target empty, the `fix-9-igpu-dgpu` cause prints `Detected gfx targets: []` and falls back to its generic advice instead of naming the discrete GPU (observed on this APU+dGPU host).
- Any consumer of `examine --json` that reads `gpus[].gfx_target` or `gpus[].is_apu` gets empty or PCI-name-guessed values.

**Possible solution**

Only accept a `Name:` line as the agent name when none has been recorded since the last `Agent` header (or ignore names that start with `amdgcn-`); commit the agent on the next header and at end of input as today. A unit test feeding the verbatim rocminfo excerpt above should assert two GPU targets. I can take this if nobody is on it.

**Your environment**

- rocm-cli: built from `main` at a6fa76d (`rocm 0.1.0`); also reproduced with the release-channel binary
- Platform: Linux, Omarchy 4.0.2 (Arch), kernel 7.1.9-arch1-2
- GPU/APU: AMD Radeon RX 7900 XT (gfx1100) + Ryzen 5 7500X3D iGPU (Raphael, gfx1036)
- rocminfo: ROCm 7.2.4 at `/opt/rocm/bin/rocminfo` (distro packages, left unmanaged by rocm-cli); managed runtime `release-wheel-multi-arch-7-14-1` also installed, same result

sbmthakur:
- [x] If this PR fixes a bug, searched `tests/e2e-cucumber/expectations.toml` for the fixed ticket ID and removed/narrowed any now-stale xfail rows. — No row referenced this bug (no ticket id; #393 is the only reference). The one row this PR *adds* scopes the new scenario to WSL under EAI-7998, the same root as `examine-both-forms-agree-on-gpu`'s row; both should go together when the WSL form lists its GPUs.

Fixes #393

## Motivation

`rocm examine --json` reported `gfx_target: ""` for every GPU even though rocminfo ran cleanly and listed them (`rocminfo_status: "ok"`). The human report was unaffected because its `detected_gfx_target` comes from sysfs, which hid the problem: nothing user-facing looked wrong until you read the JSON per GPU, or ran `rocm diagnose` on an APU+dGPU host and got `Detected gfx targets: []` in the iGPU+dGPU note instead of the sentence that names the discrete card.

## Technical Details

rocminfo lists a GPU agent's ISAs as further `Name:` lines under `ISA Info:` (`amdgcn-amd-amdhsa--gfx1100`, `…--gfx11-generic`), after the agent's own `Name:` and `Device Type:`. The walker in `probe_gpus_rocminfo` let every `Name:` overwrite the running agent name, so by the next `Agent` header it no longer started with `gfx` and the agent was dropped; with no survivors the probe returned before filling any `Gpu`.

The fix takes the first `Name:` (and `Marketing Name:`) after an `Agent` header and ignores the rest. The loop becomes `parse_rocminfo_gpu_agents`, which also keeps each agent's `BDFID:`, and the mapping onto lspci's GPUs becomes `apply_rocminfo_gpu_agents`, which joins agents to lspci entries by PCI address (`BDFID` is lspci's bus/device/function without the domain) instead of by position; only an agent with no usable `BDFID:` falls back to the next unclaimed AMD entry, and a fallback never displaces an exact match.

`is_apu` follows the target's family wherever the family table has a verdict (`gfx_apu_family_verdict`: known APU / known discrete / not covered) and keeps the earlier name-based guess otherwise; the same rule now serves the hipInfo path on Windows. A future APU whose marketing name misses `APU_KEYWORDS` is corrected when its family is known; a Raphael iGPU (gfx1036, not in the table) keeps lspci's APU verdict rather than flipping to discrete. `gfx_is_apu_family` keeps its bool contract, so serve and diagnose are unchanged. Extending the table with the gfx103x / gfx90c integrated parts remains a follow-up (open question below).

Blast radius: the rocminfo probe in `examine.rs`. The consumer that starts receiving real values is `diagnose.rs` (the discrete-vs-APU split in `fix-9-igpu-dgpu`'s note). `summary.detected_gfx_target`, `compatible_therock_family`, serve's GPU selection and the hipInfo fallback are unchanged.

## Test Plan

- Unit tests in `examine.rs` on rocminfo text: the APU+dGPU shape from the report (fails on `main`), a lone agent at end of input (fails on `main`), a CPU-only host, the older shape without an ISA block; the PCI-address join with lspci enumerating in the opposite order, the no-`BDFID` positional fallback, an exact match never displaced by a positional agent, `BDFID` parsing; and the family-verdict rule (override where known, keep otherwise).
- Gherkin: **examine-16** `@id:examine-json-names-target-per-gpu` (`@requires-gpu`) — `examine --json` must name a GPU whose `gfx_target` equals the `detected_gfx_target` the human report prints, and, where KFD's topology is readable, every `gpus[]` record must carry the target the kernel reports for that PCI address (so a target on the wrong GPU is not a pass). It reuses the "both for reading and for scripting" step with one new then-step, and runs on the gated GPU lanes. On WSL2 the GPU probes are routed out and `gpus` is `[]`, so it carries an xfail row under EAI-7998 beside examine-08's.

## Test Result

Host: Arch (Omarchy 4.0.2, kernel 7.1.9), RX 7900 XT (gfx1100) + Raphael iGPU (gfx1036), rocminfo from ROCm 7.2.4. Outside the Ubuntu matrix; the parser is distro-independent.

```
cargo test -p rocm-core -- rocminfo_ pci_bdf family_verdict   # before: 2 failed (left: []); after: 12 passed
cargo test -p e2e-cucumber --test feature_naming     # 4 passed
cargo xtask e2e -- -n examine-16                     # before: 1 failed (per-GPU targets: ["", ""]); after: 1 passed
rocm examine --json | jq '.gpus[] | {pci_id, gfx_target, is_apu}'   # before: "", ""; after: 0000:03:00.0 gfx1100/false, 0000:11:00.0 gfx1036/true (= KFD topology)
rocm diagnose                                        # note: Discrete GPU(s): ["gfx1100"]; integrated APU(s): ["gfx1036"]
prek run --all-files                                 # passed
cargo clippy --workspace --all-targets --exclude e2e-cucumber -- -D warnings   # clean on the pinned 1.96.0
python scripts/smoke_local.py                        # ok
cargo test --workspace --all-targets --no-fail-fast  # 2621 passed, 0 failed, 8 ignored, 1 skipped (below)
```

Skipped: `fix::tests::the_path_fix_reports_nothing_rather_than_a_directory_with_no_install_in_it` fails on any host with a system ROCm at `/opt/rocm` (`newest_rocm_install_dir` falls through to it); fix.rs is untouched here and #337 touches that test. Not run here: the Strix Halo / MI350P / WSL2 lanes; examine-16 is gated to them, with the WSL2 lane resolving it to xfail via the new row.

## Open questions

1. Precedence of `is_apu`: this PR keeps lspci's keyword verdict where one exists. If the intent of the mapping was for `gfx_is_apu_family` to win, the table first needs the gfx103x (Van Gogh/Rembrandt/Raphael/Mendocino) and gfx90c/gfx902 integrated parts, and `vram_capacity_is_meaningful` changes with it on those hosts; happy to do that as a follow-up if that is the preferred direction.
2. The `fix-9-igpu-dgpu` Linux plan still hardcodes `export HIP_VISIBLE_DEVICES=1` (and the verify line) regardless of which ordinal the discrete GPU has; on this host index 1 is the iGPU. Out of scope here; will file separately.


crates/rocm-core/src/examine.rs
```diff
@@ -1200,7 +1220,9 @@ fn probe_gpus_rocminfo(e: &mut Examination) {
             if !marketing.is_empty() && gpu.name.is_empty() {
                 gpu.name = marketing;
             }
-            gpu.is_apu = Some(gfx_is_apu_family(&gfx));
+            if gpu.is_apu.is_none() {
+                gpu.is_apu = Some(gfx_is_apu_family(&gfx));
```
juhovainio, line 1224: `apply_rocminfo_gpu_agents` only fills `is_apu` from rocminfo's authoritative gfx-target-based family check when it's `None`, but every AMD GPU from `probe_gpus_lspci` is always pushed with `is_apu: Some(...)`, so this correction branch never actually fires in production - confirmed by tracing the probe order (lspci runs before rocminfo, and rocminfo only sees already-`Some` entries).

A future/rebranded AMD APU whose marketing name isn't yet in `APU_KEYWORDS` would get `is_apu: Some(false)` from lspci's keyword guess, and rocminfo's more accurate verdict would never be consulted to fix it, so `has_apu`/`has_discrete_amd` stay permanently wrong for that host. The parallel Windows path (hipInfo.exe) still does the unconditional overwrite, so this also creates a Linux/Windows inconsistency in the same class of logic.

tests/e2e-cucumber/tests/e2e/examine_steps.rs
```diff
@@ -492,6 +492,37 @@ async fn assert_forms_agree_on_gpu(world: &mut E2eWorld) {
...
+        .and_then(serde_json::Value::as_array)
+        .map(|gpus| {
+            gpus.iter()
+                .filter_map(|g| g.get("gfx_target").and_then(serde_json::Value::as_str))
+                .map(str::to_owned)
+                .collect()
+        })
+        .unwrap_or_default();
+    assert!(
+        targets.iter().any(|t| t == &expected),
```
juhovainio, line 520: The new regression test only checks that some GPU in `gpus[]` carries the expected `gfx_target`, not that it's attached to the correct GPU record, so it can't catch a positional mismatch between rocminfo's agent order and lspci's enumeration order.

On a host with both an APU and a discrete GPU (the exact scenario #393 was about) where rocminfo's agent order differs from lspci's PCI order, `apply_rocminfo_gpu_agents`'s positional zip could swap gfx_targets between the two GPU entries; since the assertion is `targets.iter().any(|t| t == &expected)`, the swapped-but-still-present target still satisfies it.

juhovainio (changes requested):
The core parser fix here is right - guarding the `Name:`/`Marketing Name:` assignment with `is_empty()` so rocminfo's ISA sub-entries can't clobber the agent name is correct and well-tested. Two things worth fixing before merge, though, both related to the same underlying issue:

crates/rocm-core/src/examine.rs
sbmthakur: Right, lspci always sets `is_apu` first, so that guard was dead. 53723ca replaces it with `gfx_apu_family_verdict` (known APU / known discrete / unknown): a known verdict overrides the name guess, on the hipInfo path as well; an unknown one (gfx1036) leaves lspci's verdict alone rather than flipping it to discrete. Extending the table is a follow-up.

tests/e2e-cucumber/tests/e2e/examine_steps.rs
sbmthakur: Fixed at the mapping: agents now join lspci entries on `BDFID:` (= lspci's bus/device/function), with positional fallback only for agents that have no BDFID, and a fallback never displaces an exact match. Unit tests cover the reversed-order case. The step now also checks each `gpus[]` record against KFD's target for that PCI address where `/sys/class/kfd` is readable; presence remains the floor elsewhere.

rominf (changes requested):
The parser fix is right, and the `BDFID` join answers the earlier ordering concern properly. First-`Name:`-wins, the end-of-input commit, and pass-1-before-pass-2 all hold on the shapes I tried, and the unit tests cover them well.

One thing I think needs handling before this merges: filling `gfx_target` for every AMD GPU changes `rocm diagnose` in a way the description doesn't cover, and it goes the wrong way on exactly the APU+dGPU host this fixes.

`check_1_arch_not_in_wheel` reads `amd_gfx_targets()`, which is every AMD GPU with a non-empty target, integrated ones included. Until now that list was empty on this host class: `probe_gpus_rocminfo` dropped every agent (the bug you're fixing), and `probe_gpus_lspci` only guesses a target for the names in its APU early-return list, which "Raphael" is not one of. So `diagnose.rs:425` was effectively dead there. With the fix it returns both targets, and an iGPU target such as gfx1036 is in no PyTorch ROCm wheel's arch list, so the "not in framework arch list" branch scores +55.

Ran at 53723ca against the same `Examination` twice — once with both `gfx_target`s empty (pre-fix), once filled (post-fix), with a typical PyTorch ROCm `framework_arch_list`:

```
BEFORE: fix-1-arch not diagnosed
AFTER:  score 55 — GPU gfx target(s) ["gfx1036"] not in framework arch list
        ["gfx900", ..., "gfx1100", "gfx1101", "gfx1102"]
```

So on your reference box `rocm diagnose` now tells the user to uninstall and reinstall torch from a different wheel index, while the card they are actually serving on (gfx1100) is fully covered. A discrete-only host is unaffected — gfx1100 is in the list, the branch scores -30 and stays filtered out — so this is specific to the APU+dGPU pairing.

Suggestion: scope the coverage test to the GPUs the user would serve on rather than every AMD target. When `has_apu && has_discrete_amd`, compare `framework_arch_list` against the `is_apu == Some(false)` targets (the ones fix-9 already tells them to pin to); or at minimum skip the +55 branch when every missing target is an APU and some discrete target is covered. Worth a diagnose regression test on the gfx1100 + gfx1036 shape.

If you'd rather keep this PR to the parser and take that as a follow-up, that works — but I don't think it should land silently, since it is user-visible output on the very host in #393.

Two smaller notes, neither blocking:

- `gfx_apu_family_verdict` is `pub`, but it is only used inside `examine.rs` and its tests; private would do.
- The doc on `apply_rocminfo_gpu_agents` says only an agent with no usable `BDFID:` falls back to position, but any agent pass 1 failed to place falls through to pass 2 — including one whose `BDFID` matched nothing. The parenthetical half-covers it; worth tightening.

Verified locally at 53723ca on Linux: the 12 new unit tests pass; `cargo test --workspace --all-targets` is green apart from three pre-existing `rocm-dash-tui` `agent::tests` failures in a crate this PR does not touch; both CI clippy invocations are clean; `feature_naming` passes, so examine-16's id/index and the expectations key line up; and examine-16 fails deterministically on WSL, since `probe_with_interpreter` skips all GPU probes there, so the non-flaky xfail row is right. I did not run the GPU lanes.

rominf:
Approved the pending workflow runs — all three were sitting at `action_required` because this is a fork PR, so the branch had no CI at all until now. Results, replacing the "I did not run the GPU lanes" caveat in my review:

**CI** ✅ &nbsp;·&nbsp; **CodeQL** ✅

**E2E self-hosted** — examine-16 passed on every lane that finished:

| Lane | Lane | examine-16 |
|---|---|---|
| MI300X | ✅ | passed |
| MI350P | ✅ | passed |
| rad3 R9700 | ✅ | passed |
| Strix Halo, Windows | ✅ | passed |
| Strix Halo, WSL2 | ✅ | not applicable (below) |
| Strix Halo, Ubuntu | ⚠️ cancelled | ran; not the hang |

The Strix Halo Ubuntu lane hit its 45-minute `timeout-minutes` cap, which GitHub reports as `cancelled`. It is not this PR: that run's `commands.jsonl` shows the suite got through examine-16 and on through logs-01..03, then stopped at `serve-01 - Short model names are expanded to their full name` on the vllm engine and sat there until the cap. Nothing here touches the serve path. That lane is also independently unhealthy right now — it reported `failure` on main's own run earlier today (35699859214).

One thing the run did **not** confirm: the new WSL xfail row. `@requires-gpu` is not satisfied on that WSL runner, so examine-04, -15 and -16 were all filtered out as not-applicable; examine-08 ran and failed exactly as its row expects, but `examine-json-names-target-per-gpu` never resolved to anything. The row isn't wrong — it's the right guard for a WSL host reporting `driver_status: wsl_rocdxg_ready` — it is just unexercised by the lane as configured, so it's worth knowing it is unproven rather than verified.

One note on how much the four green lanes actually prove, since they read stronger than they are: the per-record KFD cross-check only fires for `gpus[]` entries whose `pci_id` matches a KFD node, and on a host that falls through to `probe_gpus_sysfs_fallback` the appended record has an empty `pci_id`, so the cross-check silently skips. The lane artifacts don't capture `gpus[]`, so I can't tell from CI which lanes exercised it. Not a change request — just that the assertion may be thinner on some lanes than a green tick suggests.

Green CI doesn't bear on the fix-1-arch regression in my review — there is no diagnose coverage for that shape — so that one still stands.

PR #387, opened 2026-09-11
spike: WSL2 on the DevLab Dispatch pool (throwaway)

juhovainio:
## Summary
Stacked on #377. Answers whether the DevLab Dispatch pool can support a real `e2e-wsl`-equivalent lane, per https://ubuntu.com/wsl/docs/stable/reference/actions/#reference-actions-wsl-install, before building one for real.

Three open questions, checked by a single throwaway `workflow_dispatch`-only job (`wsl-pool-spike.yml`) that never runs automatically:

1. **Reboot to enable WSL2 features.** GitHub-hosted Windows runners skip this because it's baked into their VM image at build time; unknown whether the pool's image has it baked in, and an ephemeral runner can't survive a mid-job reboot.
2. **Session 0 isolation.** A self-hosted runner installed as a background service runs in Session 0, where WSL/Store-app processes can fail to launch ([`ubuntu/wsl-actions-example`](https://github.com/ubuntu/wsl-actions-example) had to work around exactly this with an interactive auto-logon session). Unknown whether the pool's runner avoids this.
3. **GPU passthrough** (`librocdxg.so`, dxcore, `/dev/dxg`) depends on the AMD driver already on the host, independent of WSL itself.

## Results

Ran twice (a temporary `pull_request` trigger, scoped to this file only, was needed to get the first dispatch — `workflow_dispatch` can't target a workflow file that doesn't exist on the default branch yet). First attempt tripped a workflow-authoring bug of my own (a diagnostic step's expected non-zero exit skipped every step after it, including the actual install); fixed in [`c77c9b0f`](https://github.com/ROCm/rocm-cli/pull/387/commits/c77c9b0f), then reran clean: [run 34598787128](https://github.com/ROCm/rocm-cli/actions/runs/34598787128).

| Question | Result |
|---|---|
| 1. Reboot needed? | **No.** `wsl.exe --install` completed and the distro launched and ran commands, all within the same job — no reboot. |
| 2. Session 0 blocking WSL? | **No.** The distro genuinely started and executed `bash -c "..."`, returning real output. |
[...]

PR #403, opened 2026-09-14
ci: build the real ephemeral e2e-wsl lane on the DevLab Dispatch pool

juhovainio:
## Summary
Stacked on #377. Builds the real `e2e-wsl` lane #377 deferred and #387 spiked, per the DevLab Dispatch pool maintainer's sign-off:

1. WSL2 works on the pool's Windows hosts, `/dev/dxg` confirmed present.
2. Pin `--id Microsoft.WSL` in the install step so a failed install can't report green.
3. Expect ~5min/job for distro install (distros unregistered between jobs by design).
4. Static TA runners are not needed for WSL coverage.

`e2e-wsl` now runs on `[self-hosted, windows, devlab-dispatch, strix-halo]` — the same pool `e2e-gpu-strix-windows` already uses — instead of a dedicated static, always-on `wsl`-labeled runner:

- Installs WSL via `winget install --id Microsoft.WSL --exact` with no `continue-on-error`, replacing the `ubuntu/WSL` composite action's own fuzzy `--name` match + swallowed failure (exactly what point 2 flagged).
- Installs Ubuntu-24.04 fresh, verified with a propagated exit code.
- Clones the checkout into the guest's native ext4 filesystem before building — building against the DrvFs `/mnt/<drive>` mount directly is a well-known major slowdown for compilation workloads.
- Installs ROCm's WSL driver bridge (`scripts/wsl_setup_rocdxg.sh`) every job, advisory — a distro without working GPU passthrough still runs the non-GPU suite via the capability probe's `driver_status` check.
- Every guest-side step (native build deps, GPU preflight, Rust toolchain, the suite itself) executes inside the distro via `wsl.exe`, piped from PowerShell with explicit exit-code propagation so a guest failure can't be swallowed.
- Timeout raised 90 → 120min: every job here is cold (fresh distro, fresh build, fresh prewarm) plus the ~5min/job distro-install overhead — a conservative starting budget to be tuned down after a real measured run, same as `e2e-gpu-strix-ubuntu`'s cap.
- Dropped the now-moot "reclaim stray E2E processes" step: nothing survives between ephemeral distros.

Mirrors the same changes into `e2e-wsl-nightly` (`nightly.yml`) and updates the `xtask` workflow-contract test plus doc table that couple the two jobs' timeouts and runner labels together — both required both jobs to move in lockstep.

#387 stays untouched as the throwaway spike record it says it is; this is new work stacked on #377 instead.

## Verification
- `cargo test -p xtask` — all 136 tests pass, including the workflow-contract tests coupling `e2e-wsl`/`e2e-wsl-nightly` (label parity, timeout parity, WSL-detection-signal substrings, docs-table match).
- Both workflow YAML files parse cleanly.

## Test plan
- [ ] `workflow_dispatch` with `platform: strix-wsl` against this branch picks up a runner from the pool, installs WSL + Ubuntu-24.04, and completes
- [ ] `e2e-gpu-strix-wsl-report` artifact is produced
- [ ] Iterate on timeout/step content based on what a real run actually shows


r0x0r:
I reviewed this end to end (both workflow jobs, the docs table, and the `xtask` contract change) and didn't find anything blocking. Nice, careful change — the step-by-step rationale in the comments made it easy to follow.

What I checked and am happy with:

- **Contract tests stay consistent with the move.** The new `runs-on: [self-hosted, windows, devlab-dispatch, strix-halo]` still trips the self-hosted/`strix-halo` label detection, so `every_self_hosted_lane_*` and `hardware_testing_docs_cover_all_self_hosted_platforms` keep covering `e2e-wsl`; the docs table cell matches the actual `runs-on` label list exactly; the `120`/`120` timeout parity assertion and the WSL-detection-signal substring are both still satisfied. The `E2E_MERGE_QUEUE` / `E2E_SERVE_TIMEOUT_SECS` / TUI-budget guards line up too.
- **Exit-code handling is right.** In `Install WSL`, a genuine winget failure still hits `exit $LASTEXITCODE` before the trailing `exit 0`, so only the already-installed code (`-1978335189`) and success fall through — the forced `exit 0` doesn't mask real failures. Every bash-piped step ends on a `$LASTEXITCODE`-clean note or an explicit propagate, so the PowerShell wrapper's "any leftover non-zero fails the step" behavior (the thing commit 4 chased) is handled consistently.
- **Env into the guest.** `WSLENV` forwards exactly the `E2E_*` vars set at job level; `name_filter` is substituted at render time; single- vs double-quoted here-strings are used correctly (bash `$VAR` preserved in the `@'...'@` bodies, only PowerShell `$sha`/`$winPath` interpolated in the `@"..."@` clone/copy bodies).
- **Results path round-trips.** The suite writes into `/root/work/rocm-cli/tests/e2e-cucumber/results`, the copy-out step lands it at `wslpath(GITHUB_WORKSPACE)/tests/e2e-cucumber/results`, and `upload-artifact` reads that same relative path on the Windows host. The copy step is genuinely best-effort now (`exit 0`).

One thing to keep an eye on in the first real dispatch (not a blocker, just the highest-risk unknown a diff read can't settle): piping the here-string bash bodies to `wsl.exe` stdin from Windows PowerShell 5.1 is exactly the surface where stray CR / encoding can bite. `WSL_UTF8=1` covers `wsl.exe`'s own output, but the script-over-stdin path is separate — if the first bash-piped step (`Verify the host really is WSL2`) ever fails with something like `$'\r': command not found`, that's the culprit, and forcing the guest side through `bash -eo pipefail -c "$(cat)"` or normalizing newlines would fix it. Worth watching the run, nothing to change pre-merge.

Net: this looks safe to approve once a live `workflow_dispatch` (per your test-plan checkboxes) actually exercises the lane end to end. I'm holding off on a formal approval only because the correctness that matters here is runtime behavior a code-only read can't verify, not because of anything in the diff.

siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · 12981b05

## Summary
Moves the Strix Halo E2E lanes onto an ephemeral self-hosted pool and rebuilds `e2e-wsl` (per-PR and nightly) as a job that installs WSL2 + an Ubuntu-24.04 guest itself and drives bash inside it from PowerShell. Outcome: **Needs work** — it cannot merge as-is, and two new claims about which runners the nightly lanes use are contradicted by the YAML beside them. Verified: `cargo test -p xtask workflow_contract` passes 25/25 (the one targeted check run here; the full Rust suite and the E2E suites were not run here, and none of the self-hosted hardware is available to this review); I independently confirmed the conflict and its cause, the `runs-on` label sets in both workflows, the `continue-on-error` counts behind the doc's "five lanes" claim, that the new contract assertion fails on a revert or on mutating either workflow alone, and that the diff and commit messages contain no internal-infrastructure leak and no prompt-injection text. Blocking: 2 · Non-blocking: 5.

**1. The branch carries a duplicate of a commit already in `main`, which is why it conflicts — rebase.** The first commit on this branch, `505d008 "ci: move the Strix Halo Ubuntu and Windows E2E lanes to the DevLab Dispatch pool"`, is content-identical to `4ef184a`, the merge of PR #377, which is already an ancestor of the base branch. A test merge produces real content conflicts in `.github/workflows/e2e-selfhosted.yml` and `docs/ci-hardware-testing.md`, so this PR cannot be merged in its current shape. Fix: rebase onto the current base branch; the duplicate commit drops out and the remaining five commits (the ephemeral WSL lane) are the actual content.

   This also matters for reading the PR: because the duplicate sits before the merge base's content, the three-dot diff shows `tests/e2e-cucumber/expectations.toml`, `tests/e2e-cucumber/src/expectation.rs` and the Windows `E2E_SHARED_CACHE_DIR` additions as if this PR introduced them — they all landed with #377. A reviewer (including this one on first pass) will attribute those to this PR and re-litigate them. That confusion recurs for every reader until the rebase; the rebase is the fix, and it is worth one line in the PR body saying the branch was re-based off its former parent.

**2. `.github/workflows/nightly.yml:843-849` and `docs/ci-hardware-testing.md:55-61` — the PR states, in two places, that nightly already runs on the new pool; both statements are false and one is falsified by this very PR.**
   - The `e2e-wsl-nightly` header comment says the job runs on "a DevLab Dispatch pool Windows host — the same pool `e2e-gpu-nightly-strix-windows` already uses". `e2e-gpu-nightly-strix-windows` is `runs-on: [self-hosted, windows, strix-halo, native]` (`nightly.yml:741`) — the static host, not the pool. Nothing in `nightly.yml` uses that pool except the job this PR just moved. This is load-bearing: the whole justification chain for the new lane (per-job distro teardown, `/dev/dxg` passthrough, the ~5min install overhead, the maintainer sign-off) was established for the pool, and the comment invites a reader to transfer the sibling nightly lane's availability behaviour onto it.
   - `docs/ci-hardware-testing.md:55` still reads "This migration is scoped to the per-PR lanes in this table only" and `:60-61` "Moving nightly onto the pool is a separate, not-yet-made decision." This PR moves `e2e-wsl-nightly` onto the pool, so both sentences are now wrong, in a doc this same PR edits.

   Fix: correct the comment to say the lane uses the pool that the **per-PR** `e2e-gpu-strix-windows` / `e2e-wsl` jobs use, and amend the doc paragraph to state that the nightly WSL lane also moved (with the reason), leaving only `e2e-gpu-nightly-strix` and `e2e-gpu-nightly-strix-windows` on the static `native` host.

## Non-blocking
- `.github/workflows/e2e-selfhosted.yml:688` / `nightly.yml:856` — both WSL jobs begin with a host-global `wsl --shutdown` and `wsl --unregister Ubuntu-24.04`; if a per-PR run and a nightly run (separate workflows, separate concurrency groups) ever land on the same pool host, one will destroy the other's guest mid-suite. Worth confirming host exclusivity with the pool, or dropping the global shutdown.
- `.github/workflows/e2e-selfhosted.yml:901` and the mirrored step in `nightly.yml` — the GPU-preflight script is piped to plain `bash` with neither `-eo pipefail` on the invocation nor a `set -eo pipefail` inside it, unlike every sibling guest step; harmless today only because each failure path is handled explicitly.
- `.github/workflows/e2e-selfhosted.yml:993-999` and its nightly twin — the results copy-out forces `exit 0` with no `::warning::`, so a copy that genuinely fails is indistinguishable from a clean run with an empty artifact; emit a warning before the forced zero.
- `.github/workflows/nightly.yml:1040` — `WSLENV` forwards `E2E_MERGE_QUEUE`, which `e2e-wsl-nightly` never defines (copy-paste residue from the per-PR lane); and `nightly.yml:865-869` still explains its raised TUI budget with "the Strix lanes share one physical machine", which no longer describes a lane on an ephemeral pool.
- `engines/lemonade/src/lib.rs:3247-3264` (pre-existing, landed with #377, not this PR) — `hf_cache_roots_from` checks `HOME` but never `USERPROFILE`, so the corresponding xfail row was retired by setting `HF_HOME` in CI rather than by fixing the lookup; a Windows user without `HOME` or `HF_HOME` still hits the blind cache path. Worth a follow-up issue.

CI on this head reported 1 failing and 22 successful conclusions at review start. Those are conclusion counts only — no per-lane detail was available, so this review makes no claim about which job failed. The copy under review was left unmodified; all mutation experiments ran in a separate scratch copy.

juhovainio:
Addressed both blocking points in 68b63bae:

1. **Duplicate commit causing the conflict**: rebased onto current main — `505d0089` dropped automatically as already-applied (identical content to #377's merge, `4ef184a1`), which is exactly what the review predicted. The remaining 5 commits are the real WSL lane content.
2. **Two inaccurate claims about nightly already using the pool**: fixed the `e2e-wsl-nightly` header comment (it was pointing at `e2e-gpu-nightly-strix-windows`, which still runs on the static `native` host, not the pool — corrected to reference the per-PR `e2e-gpu-strix-windows`/`e2e-wsl` jobs instead) and reworded `docs/ci-hardware-testing.md` to say the nightly WSL lane *did* move onto the pool in this PR, while the two `native` nightly Strix lanes remain a separate, undecided move.

Also picked up two of the non-blocking items while in there: dropped the copy-pasted `E2E_MERGE_QUEUE` from `e2e-wsl-nightly`'s `WSLENV` (that job never sets it), and reworded the TUI budget comment, which explained itself by referencing the old shared static host instead of the new ephemeral pool.

Left the remaining non-blocking items (the `wsl --shutdown`/`--unregister` host-exclusivity question, the GPU-preflight script's missing `pipefail`, the silent `exit 0` on a failed results copy) as-is per the review's own framing — worth confirming with the pool maintainer or as follow-ups, not blockers.

`cargo test -p xtask workflow_contract` (26 passed) and a YAML parse check on both edited files are clean locally.

siloteemu:
🔴 Automated review · pr-review-watcher · 68b63bae

## Summary
Builds an ephemeral `e2e-wsl` lane (WSL2 + Ubuntu-24.04 installed per job) on the DevLab Dispatch pool, mirrors it into `e2e-wsl-nightly`, and updates the hardware doc and the xtask workflow contract — no blocking findings; both prior blocking counts are genuinely resolved. Verified: `cargo test -p xtask workflow_contract` passes (26) and branch-level mutation shows the changed 90→120 timeout assertion genuinely fails when either lane's `timeout-minutes` is reverted, so it tests what it names; the full suite, e2e and the GPU/Windows lanes were not run. Independently confirmed count (1) — `git cherry prw-base HEAD` marks all six commits as new, #377's merge `4ef184a1` is an ancestor of the base, and `git merge-tree` reports no conflict, so no duplicate commit remains (the dropped commit itself is not fetched here, so its identity is unverified); and count (2) — `e2e-wsl-nightly` really is on `[self-hosted, windows, devlab-dispatch, strix-halo]` while both nightly Strix lanes remain `native`, and the rewritten header comment and doc now say exactly that. Blocking: 0 · Non-blocking: 5.

CI at this head when the review ran: 24 success, 1 failure, 1 pending. Re-read at publishing time: 26 success, 1 failure, 0 pending — the pending lane has since finished green, and one failure is outstanding and separate from the verdict above, which is about the code. I cannot attribute any CI outcome to a specific named job, and I cannot say whether the failure is or is not related to this change.

**None.**

## Non-blocking
- `xtask/src/workflow_contract.rs:883` — nothing pins `e2e-wsl-nightly`'s `runs-on`: reverting it to the old `native` labels while leaving the doc claiming the pool move passes all 26 tests. The doc↔YAML label check only covers `e2e-selfhosted.yml` jobs. Asserting `e2e-wsl-nightly`'s `runs-on` equals `e2e-wsl`'s is a two-line addition and is the cheap assertion that would stop this exact claim drifting false again — it is the one that was disputed twice already.
- `.github/workflows/e2e-selfhosted.yml:1009-1026` (and `nightly.yml:1083-1096`) — the results-copy step ends in a bare `exit 0` and checks nothing, not even `wslpath`. It does *not* flip a failed suite to green (the test step propagates `$LASTEXITCODE` first), but a run whose suite passed and whose extraction failed reports green with an empty results dir, and `upload-artifact`'s default `if-no-files-found: warn` then drops the lane out of the consolidated report silently; the step's comment claims a narrower tolerance ("an earlier failure … already failed the job") than the code implements. `if-no-files-found: error`, or a `::warning::` on a failed copy, closes it.
- `.github/workflows/e2e-selfhosted.yml:780` — `wsl --shutdown` is host-wide, not distro-scoped, so it assumes the job owns the physical host. The file asserts both "a fresh runner is registered per job and destroyed after" (line 366) and "a pool host can be busy with other CI work" (lines 398, 538); those cannot both be casually true, and which one holds decides whether this line is safe. One clause stating whether a pool host runs one job at a time would settle it permanently — otherwise `wsl --unregister` alone is already scoped and sufficient.
- `.github/workflows/nightly.yml:657` and `docs/ci-hardware-testing.md:59-61` — both still say two *Linux* Strix runners share `strix-halo` and that `native` disambiguates the native one from "a WSL host", but this PR's own commit message says it retires the static WSL runner and no `runs-on` carries the `wsl` label any more. Whether the runner is still registered is not knowable from the repo, so I cannot call the text wrong — but the doc paragraph says in one breath that no `wsl`-labeled runner is needed and in the next that one still shares the label. Saying explicitly whether that runner remains registered resolves both.
- `.github/workflows/e2e-selfhosted.yml:929` (and `nightly.yml:1094`) — the GPU-preflight script is piped to a bare `bash`, unlike every sibling step in the same job, which use `bash -eo pipefail` or open with `set -eo pipefail`. No current path produces a false "GPU ready" (the empty-parse guards and the final `exit 1` still fire), so this is consistency and future-proofing rather than a live defect.

juhovainio:
Addressed all 5 non-blocking items from the automated review:

1. **`wsl --shutdown` host-exclusivity ambiguity** — added a comment documenting that the blanket shutdown is safe only because the pool hands out one exclusive host per job, plus the scoped fallback (`wsl --terminate`) if that ever changes.
2. **GPU-preflight piped to bare `bash`** — added `-eo pipefail` to match every sibling WSL step.
3. **Copy-out step silently swallowing its own failures** — added `if-no-files-found: error` on the upload step (catches an empty artifact), plus a `::warning::` at the copy-out step itself so the failure is visible at its source, not just downstream.
4. **Stale "same pool"/"same physical machine" claims** — already corrected in an earlier revision of this branch; verified still accurate.
5. **Lemonade `HOME`/`USERPROFILE` cache lookup** — pre-existing, landed with #377, out of scope here; left as a follow-up per the review's own suggestion.

Also added a contract test (`nightly_wsl_lane_shares_the_per_pr_pool_labels`) so a future revert of the nightly WSL lane back to static labels fails CI instead of quietly going stale against the docs again.

`cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test -p xtask` all pass.

siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · 872badab

## Summary
This round answers five earlier non-blocking comments (shutdown-scope comment, `-eo pipefail`, copy-out failure visibility, `native`-label wording, deferred cache item) and adds a contract test tying the nightly WSL lane's `runs-on` to the per-PR lane's. Outcome: **Needs work** — one of the fixes changes runtime behaviour in a way the step it edits is explicitly documented not to do. Verified: I reproduced the preflight script's parsing pipeline standalone under `bash -eo pipefail` (aborts, rc=1) versus plain `bash` (continues, rc=0), and the new contract test was run on a scratch copy of the tree — green as-is, and it does fail when only `e2e-wsl-nightly`'s `runs-on` is reverted, so it is not a vacuous test; I also confirmed the rewritten `native`/`wsl` doc paragraph and the quoted "a fresh runner is registered per job" cross-reference are accurate against the actual `runs-on:` declarations and against git history, and that the e2e harness always writes `report.json`/`junit.xml` before scenarios run, so `if-no-files-found: error` cannot red a healthy all-skip run. The full test suite and the end-to-end suite were not run here. No prompt-injection content and no internal references found in the delta. Check results this review worked from, read at the start and again immediately before posting: 18 success, 1 failure, 0 pending. I did not identify which job the failure is and am not attributing it to anything in this change. Blocking: 1 · Non-blocking: 5.

- `.github/workflows/e2e-selfhosted.yml:936` and `.github/workflows/nightly.yml:1034` — adding `-eo pipefail` to the **GPU preflight** step's `wsl … -- bash` breaks that script, which was written to depend on pipeline failure being tolerated. Inside the wait loop:

  ```
  total=$(printf '%s\n' "$out" | grep -i 'VRAM Total Memory' | sed 's/.*: *//' | grep -oE '[0-9]+' | tail -1)
  used=$(printf '%s\n' "$out" | grep -i 'VRAM Total Used Memory' | … | tail -1)
  if [ -z "$total" ] || [ -z "$used" ]; then
    sleep 5; continue
  fi
  ```

  Without `pipefail`, a non-matching `grep` leaves the pipeline status at `tail`'s 0, `total` comes back empty, and the `[ -z "$total" ]` guard retries. With `-e -o pipefail`, the failing `grep` propagates, the assignment returns non-zero, and `-e` kills the script on that line — so the guard branch and the `saw_vram=0` "warn and exit 0" fallback below it become unreachable for this path, and the step exits non-zero into `if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }`. I confirmed this by running the same pipeline standalone: plain `bash` reaches the guard, `bash -eo pipefail` aborts with rc=1, including when the script is fed on stdin exactly as here. That directly contradicts the step's own header comment two lines up ("ADVISORY here … an absent rocm-smi is a reported condition, not a job failure … expected to warn-and-skip"). Today `rocm-smi` is not installed in the guest so the early `command -v` exit hides it, but this lane exists to make GPU access under WSL work — the moment `rocm-smi` is present and prints anything the two `grep` patterns do not match, this becomes a hard red on `e2e-wsl`, which is `continue-on-error: false` (`e2e-selfhosted.yml:721`). The "match every sibling WSL step" premise also does not hold for this script: the siblings that carry pipefail (`e2e-selfhosted.yml:824`, `:841`, `:955`, `:1009`) are written pipefail-safe (`|| true`, `grep -q` inside an `if`), and two further siblings (`:863`, `:886`) use an in-script `set -eo pipefail`; the preflight is the one script whose logic *relies* on a tolerated pipeline failure. Fix: drop `-eo pipefail` from these two invocations (leaving them as they were), or keep it and make the two assignments explicitly tolerant, e.g. `total=$( … | tail -1 ) || total=""` (and the same for `used`), so the existing empty-value guard still runs.

## Non-blocking

- `.github/workflows/e2e-selfhosted.yml:1048-1051`, `.github/workflows/nightly.yml:1111-1114` — the upload comment justifies `if-no-files-found: error` by saying default `warn` "would upload nothing and still look like a clean, empty run in the consolidated report", but `discover()` (`xtask/src/e2e_report.rs:75-104`) simply omits any platform without a `report.json`, so it shows up as an absent lane; the copy-out comment added in the same commit describes it correctly as "a missing artifact". Two new comments, one commit, conflicting accounts of the same downstream effect — align them.
- `xtask/src/workflow_contract.rs:783-786` — the new test's doc comment says a revert would be "to the old static `native` labels", but this job's pre-migration `runs-on` was `[self-hosted, linux, strix-halo, wsl]` (nightly.yml at the merge base); `native` was never on `e2e-wsl-nightly`, it is the label the two `e2e-gpu-nightly-strix*` lanes still pin.
- `xtask/src/workflow_contract.rs:790-800` — the assertion is relational, so it is blind to a symmetric change: reverting *both* jobs to the same static labels passes, as does removing `runs-on` from both (empty vec equals empty vec). The per-PR side is pinned to the docs table by `hardware_testing_docs_cover_all_self_hosted_platforms`, which closes most of that gap, but there is no docs row for `e2e-wsl-nightly`. A one-line `assert!(!per_pr.is_empty())` plus a literal check that the shared list contains `devlab-dispatch` would make the guard say what its name promises.
- `.github/workflows/e2e-selfhosted.yml:1052`, `.github/workflows/nightly.yml:1115` — `if-no-files-found: error` also reddens the upload step in exactly the early-failure case the copy-out comment says must not be double-reported (the copy-out's `mkdir -p` leaves an empty results dir, which upload-artifact treats as no files); and these are the only two of seven `upload-artifact` steps in these files that deviate from the default. Worth a word on why the WSL lanes alone need it.
- `.github/workflows/e2e-selfhosted.yml:783` — "see … *above* in docs/ci-hardware-testing.md" reads oddly from inside a workflow file pointing at a separate document; drop "above". Relatedly, the identical blanket `wsl --shutdown` at `nightly.yml:915` did not get the new exclusivity rationale, only the older "See e2e-wsl in e2e-selfhosted.yml" pointer.

siloteemu:
🔴 Automated review · pr-review-watcher · 872badab

Re-checking as asked. Two things worth flagging before anything else.

**The dismissed review had a blocking finding, and the dismissal message does not mention it.** That review recorded one blocking item and five non-blocking ones. The message lists the five and says they have been addressed.

**No commit has landed since that review.** Its head was `872badab`, and `872badab` is still the head now. The two commits that fixed the non-blocking items — `64093e85` and `872badab` — were both pushed before the review, and the review read them; that is why it recorded those five as already resolved rather than outstanding. So there is nothing new to re-check, and the blocking item is in exactly the state it was.

Restating it, since it is now the only thing outstanding from this side:

**Adding `-eo pipefail` to the GPU preflight step's `wsl … -- bash` invocation breaks that script**, in `.github/workflows/e2e-selfhosted.yml` and `.github/workflows/nightly.yml`. The wait loop extracts the VRAM totals through a `grep | sed | grep -oE | tail` chain, and it was written to tolerate that chain failing: before the GPU reports, the first `grep` matches nothing and exits non-zero, which under `pipefail` now fails the whole step instead of leaving the variable empty and going round the loop again. The effect is that a preflight which used to wait for the GPU to come up now turns red the moment it polls too early — an advisory step becoming a hard failure.

This one is worth singling out because the `-eo pipefail` change came from a suggestion in an earlier round of this review. That suggestion was sound for the step it was raised on and wrong for this neighbouring one, which has the opposite error-handling contract. The fix is to drop `-eo pipefail` from the preflight invocation specifically, or to give the extraction chain an explicit `|| true` so an empty read stays a retry rather than a failure.

No new gate is being filed here; the finding is recorded so the decision is yours. At this head the check-run conclusions read 23 success, 2 failure and 1 pending, and this review does not attribute either failure to any particular job.

juhovainio:
Correction to the review I dismissed earlier: it correctly caught a regression in my own previous fix. Reverted \`-eo pipefail\` on the GPU preflight step's WSL invocation — it broke the script's own empty-value retry guard, which depends on a non-matching \`grep\` leaving the pipeline status at \`tail\`'s 0 rather than aborting. Reverted both invocations to plain \`bash\`, matching the other pipefail-tolerant sibling scripts rather than the ones that carry it deliberately.

Also picked up the remaining non-blocking items from that round: aligned the \`if-no-files-found\` comment with what \`discover()\` actually does, fixed the contract test's doc comment (the pre-migration label was \`wsl\`, never \`native\`) and tightened its assertion against a symmetric-empty revert, and fixed the dangling "above" + added the exclusivity rationale to nightly.yml's own \`wsl --shutdown\`.

\`cargo fmt --check\`, \`cargo clippy --workspace --all-targets -- -D warnings\`, and \`cargo test -p xtask workflow_contract\` all pass.

siloteemu:
🔴 Automated review · pr-review-watcher · fec37a36

Confirming the preflight item is resolved, and that your reading of it matches ours.

Checked at this head: both GPU-preflight invocations — in the per-PR lane and in the nightly one — are back to a plain "bash", the VRAM extraction chain is unchanged, and its empty-value guard is therefore reachable again, so an early poll retries instead of failing the step. That was the whole of the blocking finding, and nothing of it is outstanding.

Worth saying that the comment added beside each invocation, naming why pipefail must not be there, is the part that has lasting value. The original mistake came from a suggestion of ours that was right for the neighbouring steps and wrong for this one, and nothing in the file previously distinguished the two contracts. A reader comparing this step against its siblings would have drawn the same wrong conclusion again; now they will not.

This is a targeted confirmation of that one item, not a full review of the new commit — the other changes in it have not been read here. Check-run conclusions read at this head: 18 success, 1 failure. No outcome is attributed to any particular job.

siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · fec37a36

## Summary
Moves the Strix Halo WSL/Windows e2e lanes onto the ephemeral DevLab Dispatch pool and builds a fresh in-job WSL distro; this head commit reverts a `-eo pipefail` that broke the advisory GPU preflight and corrects four comments/assertions from the prior round — outcome: Needs work (one stale doc sentence the PR itself invalidated). Verified: the pipefail claim is genuinely true (`total=$(… | grep … | tail -1)` is an unguarded simple command, so `pipefail` + `-e` aborts before the retry guard, and every terminal path still exits explicitly), the new comment lines are valid bash comments inside the here-string with no functional effect, both workflow copies of the changed regions are byte-identical, `discover()` really does silently omit a platform with no `report.json`, the contract test's new doc comment matches the pre-migration `runs-on` in history, and the workflow contract tests run and pass here (27/27); the full test suite, the GPU lanes and the Windows/WSL lanes were not run here. Checks at this head: 18 success, 1 failure, 0 pending. Blocking: 1 · Non-blocking: 5.

- `docs/ci-hardware-testing.md:89-92` — the paragraph still says `e2e-wsl` "runs on an Ubuntu distro hosted in WSL2 on the Strix Halo Windows box and mirrors the sibling Linux lane step for step: stray-serve reclaim, GPU preflight, toolchain bootstrap, shared-runtime pre-warm". This PR deletes the "Reclaim GPU from stray E2E processes" step from `e2e-wsl` (correctly — an ephemeral, freshly installed distro cannot carry a stray serve from a previous run), so the sentence now names a step the lane no longer has, and "the Strix Halo Windows box" describes the static host the lane just moved off. The PR rewrites the surrounding sections of this same file, so this is a stale claim the change itself introduced and left behind. Fix: drop "stray-serve reclaim" from the list, and reword the host phrase to the pool host with the fresh in-job distro install, matching the new paragraph a few lines above it.

## Non-blocking
- `.github/workflows/e2e-selfhosted.yml:936` / `.github/workflows/nightly.yml:1036` — "unlike its siblings" overstates: two other `wsl … -- bash` invocations in the same job already run without `pipefail`; naming the pipefail-carrying steps specifically would stop a future reader "fixing" the inconsistency.
- `xtask/src/workflow_contract.rs` — nothing pins the no-pipefail decision, and this exact mistake has now been made once; a one-line assertion that the advisory preflight step's `wsl … -- bash` invocation carries no `-eo pipefail` is the cheap guard that would catch a re-add, since no test can otherwise cover a CI shell flag.
- `.github/workflows/e2e-selfhosted.yml:941` — the revert drops `-e` as well as `pipefail`, but only `pipefail` is load-bearing for the retry guard; no live bug today (every branch ends in an explicit `exit`), yet the rationale explains the narrower change than the one made.
- `.github/workflows/e2e-selfhosted.yml:936-940` — the new five-line rationale sits between `exit 1` and the closing `'@`, so it ships as bash source rather than as a comment on the PowerShell invocation line it describes; harmless, but it is inside the payload, not beside the flag.
- `.github/workflows/nightly.yml:952, 963, 980, 995` — the nightly WSL job's step bodies mirror the per-PR lane exactly, but several rationale comments enriched on the per-PR side were not carried over (build deps, guest clone, driver bridge, preflight scope note), so the two "mirror step for step" jobs will drift in explanatory value.

siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · eca80945

## Summary
Moves the Strix Halo WSL/Windows e2e lanes onto the ephemeral DevLab Dispatch pool and builds a fresh in-job WSL distro; this head commit rewords the one `e2e-wsl` paragraph the migration had invalidated — outcome: Needs work, because the rewrite fixes the two stale facts but replaces them with a causal claim the repository itself contradicts. Verified by reading the workflow source rather than running anything (the delta is documentation only): `e2e-wsl` really does pin `devlab-dispatch`, really does `--unregister`/`--install Ubuntu-24.04` per job, really has no reclaim step, and the sibling Linux lane really does still have one — but that sibling lane pins the same `devlab-dispatch` pool, and both this doc and the workflow state the pool hands out a fresh exclusive host per job, so "persistent host" is not why it keeps the step. No test suite was run here, and no workflow run can be executed here. Blocking: 1 · Non-blocking: 4.

## Previously raised
- **RESOLVED** — we filed: "`docs/ci-hardware-testing.md:89-92` — the paragraph still says `e2e-wsl` 'runs on an Ubuntu distro hosted in WSL2 on the Strix Halo Windows box and mirrors the sibling Linux lane step for step: stray-serve reclaim, GPU preflight, toolchain bootstrap, shared-runtime pre-warm'. This PR deletes the 'Reclaim GPU from stray E2E processes' step from `e2e-wsl` … so the sentence now names a step the lane no longer has, and 'the Strix Halo Windows box' describes the static host the lane just moved off." Mechanism verified, not taken on the dismissal: the step list is gone from the sentence, and I re-derived `e2e-wsl`'s twelve steps from `e2e-selfhosted.yml` — there is no reclaim step, while `e2e-gpu-strix-ubuntu` still carries `Reclaim GPU from stray E2E processes`. The host phrase now reads "the DevLab Dispatch pool, in the WSL2/Ubuntu-24.04 distro the job installs fresh", which matches `runs-on: [self-hosted, windows, devlab-dispatch, strix-halo]` and the `wsl --unregister` / `wsl --install Ubuntu-24.04` step. Both halves of the count are genuinely fixed. The rewrite did, however, introduce a new defect of the same class — filed below, and weighed as strictly as the original.

The five non-blocking items from the prior round are untouched by this documentation-only delta and still stand as written; none of them blocks.

- `docs/ci-hardware-testing.md:89-92` — the new sentence says `e2e-wsl` runs in a fresh distro "rather than a persistent host, so it has no stray serve from a prior run to reclaim — unlike the sibling Linux lane, which does carry that step for exactly that reason." That rationale is contradicted by the repository on both readings. The sibling Linux lane is `e2e-gpu-strix-ubuntu`, and this PR puts it on `runs-on: [self-hosted, linux, devlab-dispatch, strix-halo]` — the *same* pool, so it is not a persistent host. Nor can it be excused as "the pool reuses hosts even if it recycles runners": line 42-44 of this same file says "a fresh runner is registered per job and destroyed after", and the distro-install step's own comment justifies its host-wide `wsl --shutdown` with "the pool hands out one fresh, exclusive host per job (see 'a fresh runner is registered per job' in docs/ci-hardware-testing.md)". If that is true, the Linux lane has no prior run's stray serve either, and its reclaim step is a carryover from the static host it just left, not evidence of persistence. This blocks for the same reason the previous count did: the paragraph states pipeline behaviour the workflow does not support, and it now actively invites a reader to believe the Strix Ubuntu lane is still on a static host — the exact belief this PR exists to correct. Fix: drop the causal clause. Say the lane's guest is built fresh in-job so there is nothing of its own to reclaim, and describe the Linux lane's remaining reclaim step for what the source supports — a step retained from the static host — rather than asserting a reason. If the reclaim step *is* still needed on the pool, then line 42-44 and the `wsl --shutdown` comment are the things that need correcting, and that contradiction should be settled before either sentence ships.

## Non-blocking
- `docs/ci-hardware-testing.md:92` — "It otherwise mirrors that lane" glosses six WSL-only provisioning steps (WSL install, distro install, host verification, native build deps, guest clone, driver bridge) plus the results copy-out; the four things it then enumerates do all check out, so a "beyond the WSL provisioning above" qualifier would carry its weight.
- `docs/ci-hardware-testing.md:101-102` — "this is the only lane that runs them" is contradicted by `nightly.yml`'s `e2e-wsl-nightly`, which installs the same distro and runs the same suite; pre-existing wording, not introduced here, but the pool migration is a natural moment to scope it to the per-PR table.
- `docs/ci-hardware-testing.md:100` — the rewrap left "`@requires-bare-metal`;" alone on a short line, breaking the paragraph's fill; cosmetic.
- `.github/workflows/e2e-selfhosted.yml:409` — if the pool really does hand out a fresh exclusive host per job, the Linux lane's `Reclaim GPU from stray E2E processes` step is now a no-op every run; worth deciding deliberately, since a doc sentence has already been written to explain it.

siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · 4c6aac78

## Summary
Moves the per-PR and nightly WSL e2e lanes onto the ephemeral DevLab Dispatch pool, installing WSL2 + Ubuntu-24.04 fresh in every job, and rewrites the prose that described the old static host — outcome: needs work, on one remaining misattributed claim in a step comment, not on the paragraph the last round blocked. Verified: ran only the `xtask` workflow-contract tests in a throwaway copy (the checkout under review was left untouched and is confirmed clean at `4c6aac78`) — the new `nightly_wsl_lane_shares_the_per_pr_pool_labels` test passes as shipped and fails with the exact label mismatch when `e2e-wsl-nightly`'s `runs-on` is reverted; against the workflow source I confirmed all three per-PR Strix lanes pin `devlab-dispatch`, that `e2e-wsl` has no reclaim step while `e2e-gpu-strix-ubuntu` does, that the WSL guest is driven as `root` in every `wsl` invocation (so the "installed unconditionally, no passwordless-sudo gate" claim holds), that no `wsl`-labelled runner is targeted anywhere any more, that both WSL lanes cap at 120 minutes, and — by reading the pre-migration file — that the Ubuntu lane's reclaim commands really are a byte-identical carryover from its static-host era. The full test suite was not run here, and no workflow run can be executed here. The review worked from the reported check state of 18 success, 1 failure, 0 pending at this head; I did not identify which job failed and cannot confirm one. Blocking: 1 · Non-blocking: 5.

## Prior objection
Our change request against `eca80945` had one blocking count. We wrote that `docs/ci-hardware-testing.md:89-92` claimed `e2e-wsl` runs in a fresh distro "rather than a persistent host, so it has no stray serve from a prior run to reclaim — unlike the sibling Linux lane, which does carry that step for exactly that reason", and that the rationale was contradicted because the sibling Linux lane pins the same pool.

**RESOLVED** — and verified by mechanism, not by accepting the dismissal. The causal clause is gone. The replacement at `docs/ci-hardware-testing.md:88-96` now says the Linux lane's step is "retained from when that lane ran on a static host rather than added for anything specific to the pool", and that whether it still earns its place is "a separate, open question". Both halves check out against the source: the reclaim step at `.github/workflows/e2e-selfhosted.yml:411` carries the same four `pkill` patterns and the same `rm -rf /tmp/rocm-e2e-*` as the pre-migration step `Prepare writable dirs on the nvme + reclaim GPU from stray E2E procs`; the already-merged pool migration renamed that step and dropped only its nvme half, adding nothing pool-specific. Our own wording "it now actively invites a reader to believe the Strix Ubuntu lane is still on a static host" no longer applies — the new sentence states in the same breath that both lanes pin `devlab-dispatch`. The rewrite also stopped short of inventing a new reason, which is what the previous three attempts all did.

The four non-blocking items from that round are all still standing, unchanged: the "mirrors that lane" gloss, the "only lane that runs them" claim, the ragged rewrap, and the undecided fate of the Linux lane's reclaim step. They are restated below where still worth carrying, and none blocks.

- `.github/workflows/e2e-selfhosted.yml:866` — the comment above `Install ROCm's WSL driver bridge` reads "The concrete ask behind the pool maintainer's ~5min/job estimate: librocdxg is not part of a bare Ubuntu-24.04 image, so it needs installing every job." That attaches the ~5min/job figure to the librocdxg install. Three other places attach it to the distro install instead: the job header at line 692 ("expect ~5min/job for distro install since distros are unregistered between jobs by design"), the timeout rationale at line 699 ("the ~5min/job distro-install overhead the pool maintainer flagged"), and `docs/ci-hardware-testing.md:50` ("every run pays a ~5min distro-install cost"). A fourth comment, at the GPU preflight step, separately describes "the maintainer's ask" as covering `/dev/dxg` plus `librocdxg.so` — so the ask and the estimate are two distinct statements from the pool maintainer, and line 866 welds them into one. This is the same defect class the last three rounds were filed for: a comment asserting a causal attribution the rest of the repository contradicts, and it has a concrete consequence — the 120-minute cap at line 699 is justified by the distro-install reading, so a later maintainer who believes line 866 and concludes the distro install is cheap would cut that cap on a false premise.

  **The load-bearing side is the distro-install attribution.** It is stated consistently in three places, is what the job's timeout rationale rests on, and matches the migration commit's own account of the maintainer sign-off. The step comment at line 866 is the side to correct, not the other three. Concrete fix: reword line 866 to state only what the step does and why it is per-job — librocdxg is absent from a bare Ubuntu-24.04 image, so a freshly installed distro needs it every run — and, if the maintainer's passthrough ask is worth citing there, cite it as the ask rather than as the estimate. What must remain true afterwards: the ~5min/job figure stays attached to the distro (re)install, so the 120-minute cap's stated rationale continues to hold.

## Non-blocking
- `.github/workflows/e2e-selfhosted.yml:411` — this is very likely why four rounds of prose about this step went wrong, and it is the cheap fix that stops a fifth. The pool migration deleted the step's own rationale comment ("Reclaim the GPU from any serve leaked by a killed/timed-out prior run... Scoped to e2e leftovers only") when it split the step out, leaving a bare `pkill` block with nothing saying why it exists. Every reader since has had to guess, and each guess landed in the docs. Restore a one-line comment on the step stating what it reclaims and that its continued need on the ephemeral pool is undecided; then the doc paragraph at lines 88-96 can shrink to a pointer instead of re-litigating the question in prose.
- `docs/ci-hardware-testing.md:42` — "All three Strix Halo lanes" is scoped only by the table immediately above it. Repo-wide there are six Strix Halo lanes, four of them on the pool, and the very next paragraph names the other three. "All three Strix Halo lanes in this table" would remove the trip hazard.
- `docs/ci-hardware-testing.md:106` — "this is the only lane that runs them" is still contradicted by `e2e-wsl-nightly`, which now installs the same distro and runs the same suite. Pre-existing wording, but this PR's own new paragraph two screens up explicitly says the nightly WSL lane moved onto the same pool for the same reason, so the contradiction is now visible on the same page.
- `docs/ci-hardware-testing.md:96` — "Beyond that, `e2e-wsl` otherwise mirrors the Linux lane" doubles its qualifier, and "beyond that" refers to the reclaim discussion rather than to the six WSL-only provisioning steps the sentence still glosses. A "beyond the WSL provisioning above" qualifier would carry its weight.
- `docs/ci-hardware-testing.md:50-52` and `:104` — the rewrap left `` `@requires-bare-metal`; `` alone on a short line and several neighbours ragged; separately, "before installing ROCm's WSL driver bridge ... and cloning the checkout" lists the two in the opposite order to the workflow, which clones first. Both cosmetic.

siloteemu (commented):
🔴 Automated review · pr-review-watcher · 92db0710

## Summary
Moves the per-PR and nightly WSL e2e lanes onto the ephemeral dispatch pool, installing WSL2 + Ubuntu-24.04 fresh in every job, and rewrites the prose describing the old static host — outcome: no blocking findings. Verified: parsed both workflow files and dumped every job's `runs-on`, `timeout-minutes` and `continue-on-error`, and confirmed in a throwaway copy (the checkout under review is untouched and clean at `92db0710`) that stripping comments from `.github/workflows/e2e-selfhosted.yml` at the previous and current heads yields a byte-identical file — so the new commit changes no behaviour; the two contract tests this branch touches were mutation-proven in that scratch copy to fail on a wholesale revert, on a single-label change, and on the equal-but-unpinned case their own guard is written against; against the source I confirmed the `~5min/job` figure is now attributed to the `Install Ubuntu-24.04 distro` step in all four places that mention it, that the new comment's "see the job header above" cross-reference resolves to the `timeout-minutes: 120` rationale that actually states it, that the librocdxg step is genuinely advisory (its `if`-guarded script cannot fail the step), that the GPU preflight step's separate "maintainer's ask" wording is untouched and does not conflate with the estimate, and that the merge base and the base-branch tip produce an identical diff. The full test suite was not run, and no workflow run was executed here. Checks at review time: 19 success, 0 failure, 0 pending. Blocking: 0 · Non-blocking: 5.

**None.**

## Non-blocking
- `.github/workflows/e2e-selfhosted.yml:411` — the reclaim step still has no comment of its own saying what it reclaims; four rounds of prose about it went wrong, and a one-line note here is the cheapest thing that stops a fifth.
- `docs/ci-hardware-testing.md:42` — "All three Strix Halo lanes" is scoped only by the table above it; repo-wide there are six, and the next paragraph names others. "in this table" would close the gap.
- `docs/ci-hardware-testing.md:106` — "this is the only lane that runs them" is contradicted by the nightly WSL lane, which now installs the same distro and runs the same suite.
- `docs/ci-hardware-testing.md:50-52` — "before installing ROCm's WSL driver bridge ... and cloning the checkout" lists the two in the opposite order to the workflow, which clones at line 848 and installs librocdxg at line 874.
- `.github/workflows/e2e-selfhosted.yml:866-871` — the new parenthetical spends three lines saying what the estimate is *not*; correct, but a reader skimming still meets "~5min/job" beside the librocdxg step. "See the job header for the per-job distro cost" would carry the same guard in one line.

## Prior findings
- **Blocking, `.github/workflows/e2e-selfhosted.yml:866` — the `~5min/job` estimate welded onto the librocdxg step: RESOLVED**, verified by mechanism rather than from the commit message. The estimate now appears in four places and all four attribute it to the per-job distro (re)install: the job header at line 691, the `timeout-minutes: 120` rationale at line 699, `docs/ci-hardware-testing.md:50`, and the reworded step comment at lines 866-871, which names the distro install as the cost and explicitly disclaims the estimate for itself. The step the estimate describes is `Install Ubuntu-24.04 distro` at line 770; the librocdxg step is a separate, later step at line 874 that does no distro work. The cross-reference "see the job header above" resolves to lines 698-699, which do state the figure. The property we said must hold — the figure stays attached to the distro (re)install so the 120-minute cap's stated rationale survives — holds. Moving it made nothing else false: the GPU preflight comment at line 888 still describes the maintainer's component-scope ask (`/dev/dxg` plus `librocdxg.so`) as a separate statement, and it no longer collides with the estimate because the estimate is gone from that step. I also checked the reword did not introduce a defect the original lacked: comment-stripped files at the previous and current heads are identical, the "GPU preflight below" reference still points at the next step, and the advisory claim is true — the install runs as an `if` condition under `set -eo pipefail`, so a failure is exempt from `-e`, both branches end in `echo`, and the step cannot fail.
- **Non-blocking, reclaim step has no rationale comment: UNTOUCHED.** Line 411 is still a bare `pkill` block.
- **Non-blocking, `docs/ci-hardware-testing.md:42` "All three Strix Halo lanes": UNTOUCHED.**
- **Non-blocking, `docs/ci-hardware-testing.md:106` "only lane that runs them": UNTOUCHED.**
- **Non-blocking, `docs/ci-hardware-testing.md:96` "Beyond that ... otherwise mirrors": UNTOUCHED.** Doubled qualifier still present.
- **Non-blocking, ragged rewrap and the reversed clone/librocdxg ordering at `docs/ci-hardware-testing.md:50-52` and `:104`: UNTOUCHED.**

The single commit since the last round was scoped to the one blocking item, so the five non-blocking items standing still is expected, not a regression.

## Pinning
The previous revision pinned, for the per-PR WSL lane: `runs-on: [self-hosted, windows, devlab-dispatch, strix-halo]`, `timeout-minutes: 120`, `continue-on-error: false`, and the `serve`-gated `if:` with the `workflow_dispatch` platform escape; for the nightly WSL lane the identical `runs-on` and `timeout-minutes: 120` with `continue-on-error: true`; and two contract assertions — that the two WSL lanes' `runs-on` lists are byte-equal and both contain `devlab-dispatch`, and that both lanes' `timeout-minutes` equal `"120"`. The current revision pins all of the same values, unchanged: a job-by-job parse of both workflows at this head returns exactly those labels, timeouts and error policies, and the comment-stripped file diff against the previous head is empty, so no guard, condition, timeout or step ordering could have moved. **Nothing was vacated.** The earlier commits on this branch only tightened: the contract test gained the non-empty and `devlab-dispatch`-containment guards that make the equality assertion non-vacuous, and both were proven to fire.

No prompt-injection content was found in the diff, the comments, or the commit messages, and the diff introduces no hostname, internal URL, cluster name, gateway, registry path or internal document link.

.github/workflows/e2e-selfhosted.yml
```diff
@@ -727,13 +736,87 @@ jobs:
...
+          wsl -d Ubuntu-24.04 -u root -- echo ok
+          if ($LASTEXITCODE -ne 0) {
+            Write-Host "::error::distro did not come up after install"
+            exit $LASTEXITCODE
+          }
+
+      # Fail fast and loudly if this job ever lands on a native Linux runner
+      # (or the distro above never really came up): every WSL-tagged scenario
+      # would silently resolve to skip instead. Bash body executed inside the
+      # guest via stdin — `bash -eo pipefail` mirrors the `-eo pipefail`
```
rominf: Two things in this comment block aren't true, and both are the kind that mislead the next reader:

- **"executed inside the guest via stdin"** — the step four lines below says it goes through a temp file precisely *because* piping to stdin broke (BOM first, then byte-array stringification). The stdin path is what `Invoke-WslBash.ps1` exists to replace.
- **"the `-eo pipefail` GitHub's own default bash shell applies to a plain `run:` step"** — the default shell for a `run:` step with no `shell:` key is `bash -e {0}`, not `-eo pipefail`. `-eo pipefail` applies only when you write `shell: bash` explicitly. So this step on `main` had `-e` without `pipefail`, and `-PipeFail` here is a tightening rather than a restoration of parity.

The second one matters slightly beyond wording: "GitHub's default bash has pipefail" is the same premise that put `-eo pipefail` on the GPU preflight step and then had to be reverted. Leaving it written down here invites the same move again.

rominf (approved):
Approving — ready to merge from my side.

**What convinced me.** The lane this PR builds isn't just plausible on paper; it has a real green run at this head. `E2E tests (Strix Halo, WSL2)` picked up an ephemeral pool host, installed WSL + Ubuntu-24.04, cloned into the guest's ext4, built, and ran the suite to completion in ~17 minutes: 137 scenarios, 129 passed, 8 failed and all 8 reconciled as expected xfails, 0 unexpected failures, artifact uploaded. That is exactly the thing the maintainer's point 2 was worried about — a lane reporting green over an install that didn't happen — and it demonstrably isn't what this does.

**I checked the green-over-nothing paths rather than trusting the step names.** winget's already-installed code is the only tolerated non-zero (`-1978335189` is indeed `0x8A15002B`), and the trailing `exit 0` only runs after the genuine-failure branch has already exited. Every guest step either passes `-PipeFail` or carries its own `set -eo pipefail`, except the GPU preflight, which is correctly left tolerant because its VRAM-parsing chain depends on a non-matching `grep` not aborting the script. The one deliberately best-effort step — copy-out — is backstopped by `if-no-files-found: error` on the upload, so a copy that fails for its own reason still shows red instead of the lane quietly vanishing from the consolidated report.

`nightly.yml` mirrors all of it faithfully, including the new `Invoke-WslBash.ps1` indirection — which matters, because nothing exercises that job per-PR. `ROCM_CLI_RELEASE_REF` checks out against `apps/rocm/build.rs`: the detached-HEAD-clone reasoning in the comment is right, and the run log confirms the value reaches the guest. The new contract assertions bind rather than pass vacuously.

**CI.** All 16 required checks are green at `b58448e4`. `E2E tests (Strix Halo, Ubuntu)` is red on both attempts, but it fails at `GPU preflight (bounded wait for an available GPU)` on two different pool hosts, every hunk in this diff sits inside the `e2e-wsl` job, and that lane isn't a required check — so it reads as pool GPU occupancy, not this change.

One non-blocking nit inline below: a comment that contradicts the step directly underneath it. Worth a follow-up, not worth holding the release.

PR #405, opened 2026-09-15
fix(examine): probe the active runtime's torch, not the ambient one (EAI-8077)

r0x0r:
## The bug

`rocm examine --json` is documented as "the machine-readable Examination JSON (for diagnosis tooling)", and on any host in the managed configuration it reported:

```json
"framework": "unknown",
"framework_version": "",
"framework_notes": ["torch import failed: ModuleNotFoundError: No module named 'torch'"]
```

`probe_pytorch` resolved its interpreter from `PATH`. In the managed setup torch is installed *only* inside the active runtime's site-packages, so the probe ran against the system interpreter, found nothing, and described a machine that does not exist. Nothing could cross-check it — the human report prints no framework section at all, so the only surface reporting a framework was the one reporting it wrongly.

Reproduced on a GPU box against `main` before writing anything, and re-checked after:

| | before | after |
|---|---|---|
| `framework` | `unknown` | `pytorch` |
| `framework_version` | `""` | `2.11.0+rocm7.14.1` |
| `framework_rocm_version` | `""` | `hip=7.14.60850` |
| `framework_arch_list` | `[]` | `[… gfx942 …]` — the host's actual target |
| `framework_source` | — | `managed-runtime` |

## The fix

The probe now prefers the active managed runtime's interpreter, falling back to `PATH` when no runtime is active.

**Pointing it at the interpreter alone is not enough.** A TheRock runtime's torch resolves HIP from a sibling `_rocm_sdk_core` package rather than its own `torch/lib`. Measured on the box:

```
runtime python, no LD_LIBRARY_PATH        → ImportError: libroctx64.so.4: cannot open shared object file
runtime python + rocm_sdk.library_paths   → torch 2.11.0+gitd0c8b1f, hip=7.2.53211, available=True
```

Without the library paths the fix would trade "no torch" for a spurious "torch import failed", which reads as a *broken* runtime — a worse answer than the silence it replaced. `probe_runtime_devices` already had to solve this; both now share `RUNTIME_LIBRARY_PATH_ENV` rather than keeping separate copies of the same constant.

## Why there is a new field

Reporting the runtime's torch wakes two diagnose checkers that are silent today **only because the field is empty**:

- `check_8_wheel_rocm_mismatch` compares the framework's HIP major against the **system** ROCm and scores exactly `MIN_SCORE_FOR_MATCH`, then tells the user to reinstall torch. A managed runtime never loads the system ROCm, so a host with system ROCm 6 and a runtime on HIP 7 is healthy — and would have been sent to reinstall.
- `check_13_hip_sdk_missing` has the same shape on Windows, contradicting its own note that TheRock wheels bundle their own HIP runtime.

The version strings carry no provenance: `hip=7.2.53211` reads identically whether it came from a managed runtime or an ambient pip wheel. So `framework_source` (`"managed-runtime"` | `"path"` | `""`) records which interpreter answered, and both checkers skip their comparison for a managed runtime.

That is a **deliberate addition to the frozen `examine --json` contract**, made alongside the existing CLI-only `status` and `wsl` keys, and the contract test's comment now records why.

## Tests

- `@id:examine-framework-names-the-interpreter-that-answered` (`examine-15`) — untagged so it runs on every lane: where the host has an active managed runtime it asserts `framework_source == "managed-runtime"`; an unmanaged box legitimately has only `PATH` and must keep saying so. Verified taking the strong branch on a managed GPU host.
- Unit tests in `examine.rs` drive a stand-in interpreter that answers like a ROCm torch **only** when the runtime's library directory reached it on the loader path, so a probe that forgets the library paths fails the test.
- Unit test in `diagnose.rs` pins `check_8` both ways: silent for a managed runtime, full strength for an ambient one.

Both guards were confirmed **red**, not just green — with the `check_8` gate removed, `fix-8-wheel-rocm` scores exactly 50; with the library paths withheld, the runtime probe dies on `libroctx64.so.4`.

## Deliberately out of scope

The ticket also proposes filling `DetectResponse.torch_version` / `python_version` in the two engines. Nothing in the repo reads that field, so there is no behaviour to regression-test. Left for a separate change gated on a consumer.


siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · 5446e73a

## Summary

The PR makes `examine`/`diagnose` probe torch through the active managed runtime's interpreter (with that runtime's loader paths) instead of the ambient `PATH` one, adds a `framework_source` key to the frozen `examine --json` contract, and gates two diagnose checkers on it. The fix itself is well-argued and the Rust unit tests are real, but the Gherkin scenario the PR names as its coverage cannot exercise the change on any lane — **Needs work**. Verified: targeted `cargo test -p rocm-core --lib` over the examine/diagnose tests passes (36 green), and four branch-level mutations there — removing the `check_8` gate, removing the `check_13` gate, reverting the interpreter dispatch in `probe_pytorch`, and dropping the runtime library paths from the child env — went red, red-free, red, red respectively (the `check_13` gate survived: all 56 diagnose tests still passed with it deleted); also confirmed by reading source that `active_runtime_root` is printed only when a runtime manifest is active, that the e2e world isolates `<data>/runtimes` unless a scenario opts in, that the pre-change `diagnose` default arm never touched `AppPaths`. The full suite was not run here. Blocking: 2 · Non-blocking: 5.

**1. `tests/e2e-cucumber/features/examine.feature:172-175` (with `tests/e2e-cucumber/tests/e2e/examine_steps.rs:405-434`) — the new scenario cannot fail for the defect it names.**

`examine-15` has no `Given` step. Every other scenario in the suite that needs a managed runtime uses `Given a managed runtime is active` (`bench.feature:46`, `runtime_setup.feature:18`, `chat.feature:67`, `model_serving.feature:41`, and others), which is the step that calls `use_shared_runtimes()` and activates one. Without it the world's `<data>/runtimes` stays isolated and empty by design — `tests/e2e-cucumber/tests/e2e.rs:200-207` documents exactly that ("Only scenarios that merely need *a* runtime active OPT IN via `use_shared_runtimes()`").

Consequently `active_managed_framework_interpreter` returns `None` on every lane, the human report never reaches the `if let Some(manifest) = active` branch that prints `active_runtime_root` (`apps/rocm/src/main.rs:14044-14048`), so `human_states(human, "active_runtime_root")` is always `None`, and the step always takes the negative arm `assert_ne!(source, "managed-runtime")`. That arm is tautological: `"managed-runtime"` is written in exactly one place, `probe_pytorch_in_runtime` (`crates/rocm-core/src/examine.rs:1555`), which is unreachable without an interpreter. The scenario passes identically whether the fix is present or reverted.

This is the hard-blocker shape precisely: it fails on a wholesale revert only because the `framework_source` key would then be absent and the `.expect(…)` would panic — but under the single-branch mutation that actually removes the fix (route `probe_pytorch` back to `probe_pytorch_on_path`), it still passes. The commit message's claim that the behaviour is "Covered by `@id:examine-framework-names-the-interpreter-that-answered` … verified to fail when the behaviour they guard is removed" is true of the unit tests it also cites but not of this scenario, and the scenario's own comment asserts "the GPU lanes run with one active and pin the fix", which the harness contradicts. AGENTS.md §3 requires a scenario for user-observable output changes, and this one does not discharge it.

Fix: add `Given a managed runtime is active` before the `When`, tag the scenario for whichever lane actually installs a runtime, and say in the PR text which lane exercises the positive branch (AGENTS.md §3 requires naming a gated lane). Correct the scenario comment and the commit message's coverage claim at the same time — as written they are what invites a reader to believe the branch is covered, and that misreading will recur for the next person who touches this scenario.

**2. `apps/rocm/src/main.rs:2429` — `diagnose` gained a new hard-failure path, contradicting the comment six lines above it.**

The pre-change default arm was a bare `Examination::probe(FrameworkProbe::Auto)` and never touched `AppPaths`. It now runs `let paths = AppPaths::discover()?;`, so path discovery failure aborts the whole command with an error. The comment the PR leaves in place at `apps/rocm/src/main.rs:2408-2415` states that `diagnose` "exits 0 whether it matched, found nothing, or is out of scope" and that "`--distro` is the exception that still errors" — that is now false. `Examination::probe`'s own doc promises it never fails and records probe errors as notes instead. A troubleshooting command is exactly the one that must still answer on a degraded host.

Fix: degrade instead of propagating — `AppPaths::discover().ok()` feeding an optional config into a best-effort `active_managed_framework_interpreter`, so discovery failure costs only the managed-runtime lookup. (The same call in `examine` at `main.rs:2374` is fine: `AppPaths::discover()?` was already unconditional there before this PR.)

## Non-blocking

- `crates/rocm-core/src/diagnose.rs:1215` — the second gate this PR adds (`check_13`) has no test: deleting it left all 56 diagnose tests green in mutation testing, and `fix-13-hip-sdk-missing` has no test coverage at all, before or after. A Windows-shaped mirror of the `check_8` test would close it.
- `crates/rocm-core/src/diagnose.rs:457-465` — `check_1_arch_not_in_wheel` also reads `framework_arch_list` and becomes reachable with managed-runtime evidence for the first time here; its remediation is `pip uninstall -y torch …` against the ambient environment, which is the wrong environment for a managed-runtime host. The PR's framing that only `check_8`/`check_13` wake up is incomplete.
- `crates/rocm-core/src/examine.rs:1679` — `probe_llama_cpp` sets `framework_source = "path"` as soon as a binary is found, before it answers; on the Auto fall-through after a failed runtime probe this overwrites `"managed-runtime"` even though nothing answered, contradicting the field's own doc at `examine.rs:230-232`.
- `crates/rocm-core/src/examine.rs:1580-1582` — `join_paths(...).unwrap_or_default()` silently drops the entire loader override if any runtime directory contains the list separator, producing exactly the spurious "broken runtime" import failure this PR exists to prevent, with no note explaining it.
- `crates/rocm-core/src/examine.rs:1554` — the return code is discarded, so a runtime interpreter that fails to spawn (rc 127) or times out (rc 124) yields the generic "if PyTorch is in a venv, activate it and re-run inside that venv" note, which is meaningless advice for a managed runtime.

CI at this head reports 22 success, 1 skipped, 3 pending and 0 failures; no per-lane detail was available, so nothing above is attributed to any named job.


_Check-run conclusions moved while this review was running. The review worked from 22 success, 1 skipped, 3 pending, 0 failures; re-read immediately before filing they are **23 success, 1 failure, 1 skipped, 1 pending**. The findings above were reached from the code and do not rest on check state. Conclusion counts only; no per-lane detail was available, so nothing here is attributed to any named job._

r0x0r:
Thanks — both blockers were correct, and the first one caught a claim I made without verifying. Addressed in `1eaeed4f`.

## Blocking

**1. `examine-15` could not fail.** Confirmed exactly as described: `E2eWorld::default()` gives each scenario a fresh temp root with an empty `<data>/runtimes`, only `use_shared_runtimes()` opts in, and my scenario never called it — so the interpreter was always `None` and the assertion always landed on the tautological arm. My PR text's claim that it "took the strong branch" on a managed host was an assertion I never checked; the harness says otherwise.

Fixed: the scenario now has `Given a managed runtime is active`, is tagged `@requires-gpu` (that step installs the SDK), and asserts the positive branch directly — `framework_source == "managed-runtime"`, plus that the interpreter named in `framework_notes` sits **inside** the reported `active_runtime_root`. The absence of `active_runtime_root` is now a panic rather than a silent fallthrough, so a broken precondition can't quietly turn the scenario back into a no-op.

Verified by the mutation you named — routing `probe_pytorch` back to `probe_pytorch_on_path`:

```
with the fix:  1 scenario (1 passed), 3 steps (3 passed)
mutated:       1 scenario (1 failed), 3 steps (2 passed, 1 failed)
               FAIL: 'examine-framework-names-the-interpreter-that-answered'
                     was expected to pass on this host but FAILED — a regression.
```

Lane: every `@requires-gpu` lane exercises the positive branch (MI300X, MI350P, rad3 R9700, Strix Halo Ubuntu/WSL2/Windows); the mock lane skips it. The scenario comment and the commit message's coverage claim are both corrected.

**2. `diagnose` gained a hard-failure path.** Correct, and it contradicted the comment six lines above. Now `AppPaths::discover().ok().and_then(...)`, so discovery failure costs only the managed-runtime lookup and the command still answers. `examine`'s copy is left as `?` since it already discovered paths unconditionally.

## Non-blocking — all five taken

- **`check_13` gate untested.** Right, and your mutation result reproduced. Added the Windows-shaped mirror of the `check_8` test; deleting the gate now fails it (confirmed).
- **`check_1` also wakes up.** Correct, and my framing was incomplete. Its remediation is bare `pip`, which targets the ambient environment — wrong on a managed-runtime host. It now carries a note saying so when `framework_source == "managed-runtime"`, pointing at the interpreter `examine --json` already names. I did not gate the checker: its *detection* is sound and is the checker the bug was disabling; only the remediation's environment was ambiguous.
- **`probe_llama_cpp` overwriting `framework_source`.** Real bug, and it contradicted the field's own doc. It now sets `"path"` only once llama.cpp actually answers, not on merely finding the binary.
- **`join_paths` failure silently dropped.** Fixed — a compose failure now pushes a `framework_notes` entry naming the variable and warning that the runtime's torch may fail to load its ROCm libraries, instead of producing the unexplained import failure this change exists to prevent.
- **Return code discarded.** The probe's `rc` is now threaded through: 127 reports that the interpreter could not be started, 124 that the probe timed out, and a managed-runtime probe no longer gets the "activate your venv" advice.

Gate after the change: 388 `rocm-core` unit tests green (up one), clippy clean, `feature_naming` green.

siloteemu:
🔴 Automated review · pr-review-watcher · 1eaeed4f

## Summary

Makes `examine`/`diagnose` probe the active managed runtime's torch instead of the ambient `PATH` interpreter, records `framework_source` in the `examine --json` contract so `check_8`/`check_13` stop measuring a managed runtime's HIP against the system ROCm, and (this head) repairs the e2e scenario and degrades `diagnose`'s path discovery — **No blocking findings**. Verified: both prior objections are resolved; all four new unit tests fail when the production clause they name is reverted (`check_8` +50 would re-enter `matched`; `check_13`'s 25-point delta would collapse to 0; the runtime-torch probe test can only report `pytorch 2.11.0+rocm7.14.1` via the injected interpreter *with* its loader path), the scenario now asserts the positive branch on a lane where it is reachable, and the diff is clean of internal names, hosts and links (the bare ticket id it carries is expressly permitted by the repo's own contributor guidance); no prompt-injection content found; the full suite was not run here. Blocking: 0 · Non-blocking: 6.

## Prior objections

**1 — Resolved (the objection was sound, and the fix is real, not cosmetic).** `Given a managed runtime is active` resolves to a step that calls `use_shared_runtimes()`, runs `rocm install sdk` when the shared tree is empty, activates the runtime and asserts `active_runtime_key` is set — so a registry record with `install_root`/`python_executable` genuinely exists. The step body early-returns on a host with no AMD GPU, but that cannot produce a vacuous pass: `@requires-gpu` is resolved per-scenario against the host capability probe and the whole scenario resolves to *skip* there, so it never reaches the `Then`. The tautological negative arm is gone entirely — the assertion is now `assert_eq!(source, "managed-runtime")` plus a positive check that `framework_notes` names the interpreter, and the `active_runtime_root` lookup panics rather than branching. Under the single-branch mutation that defeated the old version (route `probe_pytorch` back to `probe_pytorch_on_path`) this now fails: `framework_source` becomes `"path"`. Reading the workflow files (not observed runs), the positive branch is only exercised by the self-hosted GPU lanes, which are `continue-on-error` and excluded for fork PRs; the blocking GitHub-hosted lane skips it. That is the repo's sanctioned arrangement, with a condition — see non-blocking item 3.

**2 — Resolved.** `diagnose` now does `AppPaths::discover().ok().and_then(...)`, so discovery failure costs only the managed-runtime interpreter. The comment six lines above — "`rocm diagnose` is a query: it exits 0 whether it matched, found nothing, or is out of scope" and "`--distro` is the exception that still errors" — is true again, and a new comment states the degradation rationale in place. The `match` → `if let` rewrite is behaviour-preserving.

**None.**

## Non-blocking

- `tests/e2e-cucumber/tests/e2e/examine_steps.rs:434-444` — the assertion requires the interpreter note to *contain* `active_runtime_root` (i.e. `install_root`), but `managed_therock_python_executable` in `crates/rocm-core/src/lib.rs:3728-3745` documents the opposite case verbatim: "an imported or read-only runtime can record an interpreter that does not sit under `install_root` at all". A GPU runner holding such a runtime would fail a correct build. Cheapest fix: assert the note names the interpreter path, and drop the containment clause (or say in one line that the scenario only ever creates an installer-provisioned runtime).
- `crates/rocm-core/src/examine.rs` — `the_framework_probe_gives_the_runtimes_torch_its_library_path` clears `library_paths` itself, so it still passes if the composition in `runtime_library_path_env` is deleted; it *does* catch the wholesale revert and the routing mutation, and `the_framework_probe_reads_the_active_runtimes_torch` is what actually pins the composition. The name overclaims and will mislead the next reader — rename it to what it asserts (a runtime without its loader path reports the import failure) and add the one line naming its sibling as the composition test.
- AGENTS.md requires: "if the scenario can only run on a gated lane (`@requires-gpu`, `@nightly`), say so in the PR text and name the lane that will exercise it." The feature-file comment says only "the step installs the SDK, so only a GPU lane exercises it" — no lane is named. The PR body was not read here; confirm it names the lane.
- `crates/rocm-core/src/examine.rs` `runtime_library_path_env` splits the inherited value with `std::env::split_paths`, while its sibling `probe_runtime_devices` uses `runtime_path_list_split`, which on Windows trims, drops empty entries and host-normalises each one. The PR unified the *constant* and left the composition divergent, which reads as more sharing than there is — call `runtime_path_list_split` here too.
- Two independent "which runtime is active" selectors are now coupled by the new step: the human report uses `current_runtime_manifest` (active key, else exactly one `default_runtime_id` match, else none) while the interpreter comes from `select_active_managed_therock_record` (active key by key *or* id, else most recently installed). They disagree when no active key is set, so `examine --json` can report a framework from a runtime the human form calls `active_runtime_status: unset`. Pre-existing, but newly load-bearing — worth a comment on the new helper.
- No test covers `active_managed_framework_interpreter` (its sibling `active_managed_therock_channel` has four), nor the new managed-runtime note in `notes_1_arch`, nor `diagnose`'s degraded-discovery path — the three parts of this head that carry no regression test.

_Check-run conclusions at this head, both when this review started and re-read immediately before posting: **18 success, 1 skipped, 0 failures, 0 pending**. Conclusion counts only; no per-lane detail was available, so nothing above is attributed to any named job._

siloteemu (approved):
Approved at this commit. No blocking findings, no outstanding change requests, and the checks on this head are green.

PR #406, opened 2026-09-15
ci(e2e): reclaim engine processes leaked into the pre-warm tree (EAI-8751)

tomastola:
## Summary

The GPU reclaim step that runs before every self-hosted E2E lane could not see a leaked `llama-server`, so the leak held the card until the runner itself was replaced. On the R9700 lane that meant **16 consecutive jobs failing GPU preflight over 2h20m**, every one reporting the same `only 8 GiB free (< 16 GiB)`.

**Root cause.** The reclaim matched a leaked engine two ways only — anchored to a scenario temp dir (`/tmp/rocm-e2e`) or to the shared *cache* (`e2e-shared`):

```
pkill -f '/tmp/rocm-e2e.*llama-server'
pkill -f 'e2e-shared.*llama-server'
```

A managed serve runs the engine binary out of the shared *pre-warm runtime* tree — `$RUNNER_WORKSPACE/e2e-prewarm-*/data/runtimes/wheel/.../engines/lemonade/runtime/bin/llamacpp/.../llama-server` — which is neither. The lane exports `E2E_SHARED_CACHE_DIR` and `E2E_SHARED_RUNTIMES_DIR` as two different directories; the patterns were written against the cache, the engines live under the pre-warm.

Both patterns were also order-sensitive. `pkill -f` matches the whole command line and argv[0] is the binary path, so a `/tmp/rocm-e2e-*` model argument later in the line could not rescue them.

This is expensive rather than untidy because of how the container lanes recycle: the pod is usually fresh per job but the work volume is not, and two jobs can share one pod's life — so one unmatched leak fails every job scheduled to that runner until the pod is recreated.

**What changed.**

- `scripts/reclaim-gpu.sh` replaces eleven copies of the inline `pkill` list. A process is reclaimed when its command line names an E2E-owned root **and** an engine marker, in any order. Requiring both halves is what keeps a manual `/workload` serve on a shared runner safe — it names no E2E root, which is the property the old `NOTE` comment was protecting by hand.
- TERM, then SIGKILL for anything still alive after a grace period, and it reports each process it acts on. The previous step printed an unconditional `reclaimed` whether or not it did anything, which is a large part of why this went unnoticed for hours.
- Both signals re-verify that the pid still carries the command line it was selected with. `select_leaked` walks the whole of `/proc`, so a pid can be freed and reissued in between; a bystander is just as dead from TERM as from KILL, so the check cannot be reserved for the escalation.
- Native Windows has no bash, so its two steps mirror the same rule in PowerShell. Their existing matcher was already order-independent and only needed the pre-warm root.
- On preflight failure the lanes now call `--report-holders`, printing per-process VRAM (`rocm-smi --showpids`) and the engine processes present. Reconstructing that by hand is what made the original incident slow to diagnose.

## Test plan

`--self-test` runs on the GitHub-hosted lane alongside the existing script self-tests, so it gates every PR rather than only the GPU lanes. It spawns decoys whose argv[0] is a real-shaped path, plus a zombie and an out-of-scope bystander, and asserts across nine checks:

1. the superseded patterns do **not** match a pre-warm engine process — the defect itself, asked of the decoys' own command lines so the check cannot be answered by an unrelated process on the machine;
2. the new rule selects it;
3. a `/workload` manual serve (engine marker, no E2E root) is not selected;
4. an E2E-owned path with **no** engine marker — the suite's own binary under `CARGO_TARGET_DIR` — is not selected;
5. a SIGTERM-ignoring engine at a pre-warm-shaped path is selected;
6. containment: a bystander that matches the rule **in full** but lives outside `SELFTEST_SCOPE` is not selected, and a breach refuses to run the real reclaim at all;
7. the escalation guard reports same / recycled / gone, with *gone* driven by both routes into it — a failed `/proc` open and an empty read from a real zombie;
8. end to end, reclaim kills both leaks and spares all three bystanders;
9. escalation to SIGKILL was reported for the stubborn decoy and **not** for the ordinary one — which is what pins the `kill -TERM` preceding it.

Every check is mutation-verified — an assertion that cannot fail is worse than no assertion, and two rounds of review here found exactly that:

| mutation | self-test |
| --- | --- |
| remove `e2e-prewarm` from `E2E_ROOTS` (the pre-fix behaviour) | fails 2, 8 |
| remove any single root or marker from either list | fails, naming that entry |
| gut the `has_engine` half of the AND rule | fails 4 |
| delete `cmdline_of`'s emptiness check | fails 7 (zombie arm) |
| delete the `SELFTEST_SCOPE` filter from `select_leaked` | fails 6 |
| make the bystander stop matching the rule, or not run | fails 6, by reason |
| make a probed decoy unreadable | fails 1 (rather than passing having probed nothing) |
| delete the SIGKILL escalation block | fails 8, 9 |
| replace `kill -TERM` with a no-op | fails 9 |

A new `xtask` contract test pins the bash `E2E_ROOTS` to both PowerShell mirrors, so a root added to the script cannot silently miss the Windows lanes. It extracts the `-match` alternation rather than substring-matching the workflow: the whole-file form false-passes, because the roots also appear there as env vars (caught by mutating it). The `__engine-serve-http` exemption is itself asserted, so it fails — demanding its own deletion — once the tracked divergence is fixed, rather than rotting into dead code.

`cargo test -p xtask` 138 passed; `prek run` clean including `shellcheck`, apart from `license-headers` — hawkeye is not installable in this container, and this change adds no new files, so it introduces no header.

**Hardware verification.** All six GPU lanes — MI300X, MI350P, rad3 R9700, and Strix Halo on Ubuntu, Windows and WSL2 — passed on `89a64a39`, which was the condition this PR was held in draft for. The review commits since then touch the self-test and the xtask contract test, and the lanes re-run against each new head.

**Scope note:** the nightly Linux Strix lane hid its reclaim inside a step named `Prepare writable dirs on the nvme + reclaim GPU from stray E2E procs`. It had the same defective patterns and is included as the eleventh site.

No Gherkin scenario: this is CI plumbing with no user-observable CLI behaviour change (AGENTS.md §3). The behaviour it does change is CI's own, which the self-test covers.

**Risk: low.** The matching rule strictly widens what reclaim catches; the one narrowing risk — killing a bystander — is pinned by self-test checks 3, 4 and 6, by the survival assertions in check 8, and by the identity re-verification before both signals.

**Known gaps, stated rather than implied:**

- The pid-recycling guard's *call sites* are exercised only in the always-proceed direction, because making a pid be reused on demand is not reproducible in a test — deleting either call site still passes the self-test. The comparison itself is pinned in all three directions.
- The self/parent-pid exclusion in `select_leaked`, the `killed` counter behind the summary line, and `--dry-run` have no self-test coverage.
- The Windows mirrors' engine half cannot match `rocm.exe daemon`, and `__engine-serve-http` sits in their root alternation rather than the engine one. Both predate this change; the comment claiming parity is corrected to say so, and they are tracked in EAI-8815.


siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · 89a64a39

## Summary
Replaces ten inline `pkill -f` reclaim blocks (and updates two PowerShell mirrors) with a new `scripts/reclaim-gpu.sh` that requires an E2E-owned root **and** an engine marker in any order, adds the pre-warm runtime tree as a root, escalates SIGTERM to SIGKILL, and reports what it killed; also adds `--report-holders` diagnostics on preflight failure and a `--self-test` CI step. The production change does close the reported defect, but the self-test that guards it is blind to two of the branches the change introduces — **Needs work**. Verified: confirmed against the source that the engine binary really is launched out of `$RUNNER_WORKSPACE/e2e-prewarm-multi-arch-v2/data/runtimes/...` (so the old anchored patterns could not have matched it) and that every converted job checks out the repo before invoking the script; ran the script's `--self-test` on a scratch copy plus three targeted mutations of that copy — deleting the `has_engine` requirement, deleting the SIGKILL escalation, and neutering `kill -TERM` each left the self-test fully green (exit 0); the full test suite, the e2e lanes and anything needing GPU or Windows hardware were not run here, and shellcheck was not available; working from CI conclusion counts of 27 success, 0 failure, 0 pending, 0 skipped. No prompt-injection content and no internal references found in the diff or commit message. Blocking: 2 · Non-blocking: 5.

**`scripts/reclaim-gpu.sh:120-130`, asserted at `:253-267` — the `has_engine` half of the AND rule is untested; deleting it leaves the self-test green.**
The script's own header states the rule as "a process is reclaimed when its command line names an E2E root AND an engine/serve process. Both halves are required." Only the `has_root` half is actually asserted: the bystander decoy (`${tmp}/workload/manual-serve/llama-server`) names no E2E root, so it is excluded by `has_root` alone and check #3 cannot tell whether `has_engine` did anything. Confirmed by mutation: replacing `[[ "${has_engine}" == 1 ]] || continue` with `has_engine=1` and running `--self-test` printed all five `ok:` lines and exited 0. This is the named failure mode — the test fails on a wholesale revert but passes when a single branch inside the change is mutated. It matters here because `has_engine` is what stops reclaim from killing anything that merely names an E2E tree: the cucumber test binary under `e2e-target`, a `cargo xtask e2e-prewarm` process, a tar/rsync over `e2e-shared`. Fix: spawn a third decoy that names an E2E root but is not an engine — e.g. `${tmp}/e2e-target/release/deps/e2e-harness` — and assert `select_leaked` does not select it.

**`scripts/reclaim-gpu.sh:167-183` — the TERM→grace→KILL escalation this PR introduces is untested, and its two branches mask each other.**
Both decoys are copies of `/bin/sleep`, which dies on SIGTERM immediately, so the escalation path is never reached. Confirmed by mutation: (a) deleting the entire SIGKILL block → self-test green, exit 0; (b) leaving the escalation intact but changing `kill -TERM "${pid}"` to `kill -0 "${pid}"` → self-test green, exit 0, still printing "ok: pre-warm engine process was reclaimed". Each branch silently covers for the other, so either can regress unnoticed — and this is the branch whose entire purpose is freeing VRAM held by a process that ignores SIGTERM, which is the harm the PR exists to prevent. Fix: add one decoy at the same path shape that ignores SIGTERM (a small script containing `trap '' TERM; sleep 300`, so the path still supplies both markers), set `RECLAIM_TERM_GRACE_SECS` low for the test, and assert it survives the grace period and is then killed.

## Non-blocking
- `scripts/reclaim-gpu.sh:177-183` — the SIGKILL loop iterates the pre-TERM snapshot and only checks liveness, so a PID recycled during the grace window could be force-killed; re-read `/proc/<pid>/cmdline` and compare against the captured line before escalating.
- `.github/workflows/e2e-selfhosted.yml:559-567` and `.github/workflows/nightly.yml:765-773` — the new "Same rule as scripts/reclaim-gpu.sh" comment overstates parity: `rocm daemon` can never match on Windows (`rocm.exe daemon` satisfies neither regex, because the engine half needs a literal `rocm ` with a trailing space) and `__engine-serve-http` sits in the root half rather than the engine half; the regexes are pre-existing, the parity claim is new and will mislead the next reader.
- `xtask/src/workflow_contract.rs` — the repo already greps workflow text in contract tests, yet nothing pins the two PowerShell mirrors to the bash `E2E_ROOTS`; a test asserting each root appears in both regexes would prevent exactly the silent drift the PR's own comment warns about, which is the shape of the defect being fixed.
- `scripts/reclaim-gpu.sh:98` — `tr '\0' ' ' <"${cmdline_file}" 2>/dev/null` sets up the input redirection before the stderr redirection, so the expected exit-between-glob-and-read race still prints a "No such file or directory" line (observed in a live run); swapping the redirection order silences it.
- `scripts/reclaim-gpu.sh:88-134` — matching is an unanchored substring search over the whole command line, so a bystander such as `tee /tmp/rocm-e2e-x/llama-server.log` satisfies both halves from its log path alone; conversely a ComfyUI server (`python main.py`) has no marker, latent today since no scenario starts one. Anchoring the engine check to the `/proc/<pid>/comm` basename would close both.

tomastola:
Both blocking findings were correct. I reproduced all three mutations before changing anything — `has_engine` gutted, SIGKILL block deleted, and `kill -TERM` neutered each left the self-test green at `89a64a39`. The test only ever failed on a wholesale revert of the roots list, so the PR description's claim that it "fails before the fix and passes after" was true for one branch and false for the two this PR adds. Fixed in `4179db1`.

**Blocking 1 — `has_engine` untested.** Added a fourth decoy at `${tmp}/e2e-target/release/deps/e2e-harness`: an E2E-owned root with no engine marker, asserted both not-selected and surviving reclaim. Your reasoning about why it matters is the part I'd underweighted — `e2e-target` is `CARGO_TARGET_DIR`, so without `has_engine` reclaim selects the suite's own running test binary.

**Blocking 2 — escalation untested.** Added a decoy that ignores SIGTERM (`trap '' TERM`) at a pre-warm-shaped path, with `TERM_GRACE_SECS` dropped to 2 for the test. Asserting only "the stubborn decoy dies" would not have been enough — that still passes with `kill -TERM` replaced by a no-op, since everything then reaches SIGKILL. So the test also asserts the *ordinary* decoy did **not** reach escalation, which is what pins the TERM. Post-fix: gutting `has_engine` fails 2 checks, deleting the SIGKILL block fails 2, neutering `kill -TERM` fails 1.

**Non-blocking, all four taken:**

- PID recycling — the SIGKILL loop now re-reads `/proc/<pid>/cmdline` and skips escalation if it changed. Not killing a real holder is recoverable; killing a bystander is not.
- Parity comment — corrected rather than defended. You're right on both counts: `rocm ` cannot match `rocm.exe daemon`, and `__engine-serve-http` is in the root alternation. The comment now names both divergences instead of asserting equivalence. The regexes themselves are pre-existing and changing Windows matching felt out of scope here, so they're tracked separately.
- Contract test — added, pinning the bash `E2E_ROOTS` to both PowerShell mirrors. Worth noting my first version was a false pass: it used `text.contains(root)` against the whole workflow, which stays green with the root deleted from the alternation, because the roots also appear as env vars and in prose. Caught by mutating it. It now extracts the `-match` alternation and fails in both directions.
- `tr` redirection order — confirmed and swapped.

**One I'm not taking as proposed:** anchoring the engine check to the `/proc/<pid>/comm` basename. The bystander you describe is real (`tee /tmp/rocm-e2e-x/llama-server.log` does satisfy both halves), but `comm` breaks three of the four markers: it cannot contain a space, so `rocm daemon` never matches; `__engine-serve-http` is an argument rather than a binary name; and `vllm` is a console script, so `comm` is the interpreter. That reintroduces exactly the miss class this PR exists to fix, in exchange for a narrower one. Happy to take a different formulation if you have one — otherwise I'd rather leave the known-wide rule than trade it for a known-narrow one.

siloteemu (commented):
🔴 Automated review · pr-review-watcher · 4179db1a

## Summary
The follow-up commit adds the two missing decoys (an E2E-owned path with no engine marker, and a SIGTERM-ignoring engine at a pre-warm-shaped path), asserts escalation happened for the stubborn one and *not* for the ordinary one, adds a recycled-PID guard before SIGKILL, replaces the overstated PowerShell "same rule" comment with an accurate one, and pins the bash root list to both PowerShell mirrors with a new contract test — **No blocking findings**. Verified: read the whole 404-line script and the whole new test rather than the hunks, because the delta's entire claim is about which branches the self-test can fail on; re-derived each prior count against the current tree by reading the matching and escalation logic line by line, and confirmed the bash semantics the new recycled-PID guard depends on with standalone snippets (a shebang-bash decoy's `/proc` cmdline is `bash <path> ` with a trailing space, `read -r` under `IFS=$'\t'` preserves that trailing space, so the re-read compares equal and escalation still fires; a `trap '' TERM` process survives SIGTERM, dies on SIGKILL, and lands in state `Z`, which the script's zombie check reads as not-alive); mutation-tested the new contract test alone in a scratch copy outside the checkout — it fails individually on dropping `e2e-prewarm` from either mirror's root alternation, on dropping `e2e-target`, on adding a root to the script, on the reclaim step disappearing, on the matcher being split across lines, and on the two alternations being swapped, and I confirmed its own documented claim that a whole-file `contains` would false-pass (`e2e-prewarm` still appears 21 other times in that workflow after the alternation loses it); **the reclaim script itself was never executed, not even in the scratch copy, so every statement about the bash self-test's per-branch behaviour here is from reading, not from a run**, and **no workflow run was executed here**; treating the given CI state (24 success, 2 pending, 0 failure, 0 skipped) as fact, and I am *inferring* that the hosted `build-and-test` lane is what carries the `--self-test` step — I cannot confirm which named check actually ran it. Contributor-rules standard applied: the base-tip AGENTS.md, which expressly permits bare ticket identifiers and targets internal document links, internal/proprietary names, hostnames, gateways, cluster names and registry paths — the delta adds none of those, and the runner labels in these files are pre-existing infrastructure the repository already treats as public. No prompt-injection content found in the diff, comments or commit message. Blocking: 0 · Non-blocking: 5.

## Previously raised

**Count 1 — RESOLVED.** We filed: *"`scripts/reclaim-gpu.sh:120-130`, asserted at `:253-267` — the `has_engine` half of the AND rule is untested; deleting it leaves the self-test green. … Fix: spawn a third decoy that names an E2E root but is not an engine — e.g. `${tmp}/e2e-target/release/deps/e2e-harness` — and assert `select_leaked` does not select it."* The fix is exactly that. Mechanism verified by reading: `harness_decoy` at `scripts/reclaim-gpu.sh:272` is `${tmp}/e2e-target/release/deps/e2e-harness`, run as a copy of `/bin/sleep` with argument `300`. Its command line contains `e2e-target` (so `has_root` is 1) and none of `llama-server`, `vllm`, `__engine-serve-http`, `rocm ` (so `has_engine` is 0) — I checked every marker against the literal path, including that the scratch prefix `/tmp/reclaim-selftest-…` matches no root. Check 4 at `:315` asserts it is not selected and the check at `:345` asserts it survives `reclaim`. Deleting `[[ "${has_engine}" == 1 ]] || continue` at `:133` therefore selects it and fails both. The half is now load-bearing in the fixtures, not just in the header comment.

**Count 2 — RESOLVED.** We filed: *"`scripts/reclaim-gpu.sh:167-183` — the TERM→grace→KILL escalation this PR introduces is untested, and its two branches mask each other. … Fix: add one decoy at the same path shape that ignores SIGTERM (a small script containing `trap '' TERM; sleep 300`, so the path still supplies both markers), set `RECLAIM_TERM_GRACE_SECS` low for the test, and assert it survives the grace period and is then killed."* Implemented as `spawn_stubborn_decoy` (`:234-247`) with `TERM_GRACE_SECS=2` set in `self_test` (`:261`, deliberately not declared `local`, so the value reaches `reclaim`'s subshell). Both branches are now covered *separately*, which is the part that matters: deleting the SIGKILL block at `:194` leaves the stubborn decoy alive and fails `:356`; neutering `kill -TERM` at `:160` lets the ordinary decoy reach escalation, which fails the third assertion at `:368` ("ordinary decoy reached SIGKILL escalation"). Neither can now mask the other. I confirmed the two facts this rests on with standalone snippets: the stubborn shape really does survive SIGTERM and die on SIGKILL, and its command line really does carry both the `e2e-prewarm` root and the `llama-server` marker (the kernel resolves the `env bash` shebang to `bash <path>`). I also checked the self-test cannot pass vacuously on empty selection — an empty `select_leaked` fails checks 2 and 5, not just the negative ones — and that a failure propagates: `self_test` returns 1, `main` is the script's last statement, and the CI step invoking `--self-test` carries no `|| true` or `continue-on-error`.

**Prior non-blocking items (5).** Recycled-PID escalation — addressed at `:183-192`. Overstated "Same rule as scripts/reclaim-gpu.sh" parity comment — replaced in both workflows with a comment that is factually correct (I verified `rocm ` genuinely cannot match `rocm.exe daemon`, and that `__engine-serve-http` genuinely sits in the root alternation). Missing contract test pinning the roots to the PowerShell mirrors — added and mutation-verified. Redirection order on the `/proc` read — fixed at `:101`. Unanchored substring matching — not addressed; still stands, repeated below.

**None.**

## Non-blocking
- `.github/workflows/e2e-selfhosted.yml:567-570` and `.github/workflows/nightly.yml:773-776` — the new comment lists the two Windows divergences neutrally as "diverges", but they fail in opposite directions: `rocm ` vs `rocm.exe daemon` is a *miss*, while `__engine-serve-http` sitting in the root alternation is an *over-match* — on Windows a non-E2E managed serve satisfies the root half from that marker alone and can be killed. Worth saying which is which, since only one risks a bystander.
- `scripts/reclaim-gpu.sh:183-192` — the recycled-PID guard is new production code in the kill path with no branch of its own in the self-test: deleting it entirely leaves the self-test green (mutating it to always-recycle does fail, via the stubborn decoy). Given the rest of this commit, it is worth either a fixture that exercises the "recycled" path or a comment stating plainly that this branch is guarded only in the always-skip direction.
- `scripts/reclaim-gpu.sh:188-191` — a process that simply exits between `process_alive` and the cmdline re-read yields an empty `current`, so the log says "was recycled during the grace period, not escalating" for something that was not recycled at all. Distinguish empty-cmdline (gone or zombie) from a genuinely different command line.
- `scripts/reclaim-gpu.sh:38-52` — coverage is per-item thin on both lists: removing `e2e-shared`, `/tmp/rocm-e2e` or `e2e-target` from `E2E_ROOTS`, or `vllm`, `__engine-serve-http` or `rocm daemon` from `ENGINE_MARKERS`, leaves the self-test green, and the new contract test is one-way by construction (it catches a root *added* to the script and missing from a mirror, not one removed — confirmed by mutation). Only `e2e-prewarm` and `llama-server` are actually exercised.
- `scripts/reclaim-gpu.sh:88-134` — carried forward and still standing: matching is an unanchored substring search over the whole command line, so a bystander such as `tee /tmp/rocm-e2e-x/llama-server.log` satisfies both halves from its log path alone, while a ComfyUI server (`python main.py`) carries no marker. Anchoring the engine check to the `/proc/<pid>/comm` basename would close both. Relatedly, `self_test` performs real kills confined only by `SELFTEST_SCOPE`, with no assertion that selection stayed inside the scratch tree — low blast radius today only because that step runs on a hosted ephemeral lane rather than the GPU pool.

tomastola:
Four of the five non-blocking items are addressed in `79e3e67`; the fifth is the carried-forward `comm` one, still declined for the reason given [above](https://github.com/ROCm/rocm-cli/pull/406#issuecomment-5715800446).

**Per-item coverage.** Confirmed all six of your mutations first — dropping `e2e-shared`, `/tmp/rocm-e2e` or `e2e-target`, or `vllm`, `__engine-serve-http` or `rocm daemon`, each left the self-test green. The matching rule is now split out of `select_leaked` into `cmdline_matches_rule`, so the self-test drives each entry directly instead of needing a process per list item, and both lists are asserted against a written-out expectation. That duplication is deliberate: a loop over `E2E_ROOTS` cannot notice a root *deleted* from `E2E_ROOTS`, which is the exact drift at issue. All six now fail, each naming the specific entry.

**Contract test, missing direction.** You were right that it was one-way by construction. It now also rejects a root left behind in a mirror after being removed from the script, and a stray token added to a mirror — both mutation-verified. `__engine-serve-http` is skipped by an explicit arm naming EAI-8815, so fixing that ticket makes the assertion hold rather than ambushing whoever does it.

**Exited vs recycled.** Real mislabel in the code I added last round — an empty read took the "was recycled" branch. `same_selected_process` now returns 0 same / 1 recycled / 2 gone, the caller reports each separately, and all three arms are asserted.

**Scratch-tree containment.** Added, and worth flagging that you found it: `self_test` issues real kills and `SELFTEST_SCOPE` was the only thing confining them, with nothing checking that held.

**Windows comment.** Now says which is which — `rocm ` vs `rocm.exe daemon` is a miss, `__engine-serve-http` in the root alternation is an over-match, and only the second can cost a bystander.

**One thing I did not close, deliberately.** The recycled-pid guard's *call site* is still covered only in the always-escalate direction — deleting the call site passes the self-test, and I verified that after the refactor too. Making a pid be reused by a different process on demand isn't reproducible in a test. The comparison itself now has all three directions asserted, and the gap is written into a comment at the assertion rather than left implied by a green run. If you have a way to drive real pid reuse deterministically I'll take it.

Earlier mutations re-checked after the refactor: gutting `has_engine`, deleting the SIGKILL arm and neutering `kill -TERM` all still fail. `shellcheck --enable=all --exclude=SC2310,SC2312` clean, `cargo test -p xtask` 138 passed, `prek` clean.

siloteemu (commented):
🔴 Automated review · pr-review-watcher · 79e3e674

## Summary
The PR replaces inline `pkill -f` GPU reclaim with `scripts/reclaim-gpu.sh` plus a workflow-contract test; the new commit answers the previous round by asserting every E2E root and every engine marker individually, splitting the escalation guard into same/recycled/gone, adding a scratch-tree containment check, making the PowerShell mirror test two-way, and correcting the Windows divergence comment — **no blocking findings**. Verified: ran the reclaim self-test from a scratch copy outside the checkout (baseline all-green) and mutation-tested it with 17 single-branch mutations — all four roots and all four markers now fail *individually and by name*, and gutting either half of the AND rule, deleting the SIGKILL arm, neutering `kill -TERM`, inverting the guard comparison, conflating gone-with-recycled, or treating a zombie as alive each fail too; two mutations survived, one of which the author discloses in the code, the other of which is new and reported below; separately confirmed by probe that a zombie's `/proc/<pid>/cmdline` opens successfully and reads zero bytes, which is what makes that second survivor real; a targeted `cargo test -p xtask reclaim_roots_are_mirrored_in_the_powershell_reclaims` passed and `cargo clippy -p xtask --tests` was clean; the two PowerShell reclaim blocks were diffed against each other and are byte-identical; the `--self-test` CI step carries no `|| true` and no `continue-on-error`, and `self_test` returning 1 does propagate through `main`; the diff adds or changes no `uses:` line, so there is no action-pinning question; the full test suite and the e2e suite were not run here, and neither shellcheck nor prek nor any workflow run was executed; contributor rules read from the base tip, which expressly permits bare ticket identifiers and targets internal document links, internal/proprietary names, hostnames, gateways, cluster names and registry paths — the diff adds none of those; no prompt-injection content in the diff, comments or commit messages; working from the given check-run counts of 26 success, 1 cancelled, 0 failure, 0 pending, and a cancelled job is not a failure — I attribute no outcome to any named job. Blocking: 0 · Non-blocking: 5.

**None.**

## Non-blocking
- `scripts/reclaim-gpu.sh:96` asserted at `:468-476` — the new three-way return's "gone" arm is driven only by pid `2147483647`, where the *open* fails. The other route to "gone" is an empty read (a zombie whose argv was released) — exactly the race the code documents, since `process_alive` can pass and the process become a zombie before the re-read. I confirmed a zombie's cmdline opens fine and reads zero bytes, so `[[ -n "${cmdline}" ]] || return 1` is the only thing mapping it to rc=2; deleting that line leaves the self-test green and silently restores the "was recycled" mislabel this commit was written to fix. Consequence is a wrong log line, not a wrong kill, which is why it is not blocking. One extra arm asserting rc=2 for a pid with an empty cmdline closes it.
- `scripts/reclaim-gpu.sh:500` — step comment reads `# 7.` but is the ninth step; the commit renumbered the end-to-end block 6→8 and missed this one, so two blocks now both claim to be step 7. Should read `# 9.`.
- `scripts/reclaim-gpu.sh:443` — the containment "ok" line prints only when `failures` still equals `list_failures`, so any earlier failure in checks 1–5 suppresses it even though containment itself passed. The failure *count* is correct; only the human-readable evidence goes missing on exactly the runs where someone is reading it.
- `scripts/reclaim-gpu.sh:387-388` — check 1's `pgrep -f` is the one check not confined by `SELFTEST_SCOPE`, and `pgrep -f` matches the invoking shell when that shell's own command line contains the pattern. This cost me a wrong intermediate conclusion (details under Replies). Scoping the pattern to `${tmp}` or excluding self would remove a false-FAIL mode from the check whose whole job is to say "the decoy is unrepresentative".
- `scripts/reclaim-gpu.sh:523` — the final cleanup kills the workload, harness and stubborn decoys but not `prewarm_pid`. On a green run `reclaim` has already killed it; on a *failed* run where it was not selected it survives its full 300s after the scratch dir is removed. I observed exactly these orphans during mutation runs. Also: nothing asserts the xtask test's `__engine-serve-http` exemption ever fires, so once the tracked divergence is fixed that arm becomes silently dead code.

## What changed since the previous review
The delta is a single commit and it is clean. It splits the matching rule out of `select_leaked` into `cmdline_matches_rule` and the `/proc` read into `cmdline_of`, then adds `assert_rule_covers_every_list_entry`, which drives each root and each marker directly against the rule in both polarities and additionally compares both arrays against written-out expectations. That written-out duplication is the right call and I verified why: a loop derived from `E2E_ROOTS` cannot notice a root *deleted* from `E2E_ROOTS`, which was precisely the defect. All eight per-entry mutations I ran now fail, each naming the specific entry — the previous round's five-of-six-undriven gap is closed.

`same_selected_process` replaces the old two-way comparison with 0/1/2, and the caller reports each separately; the recycled and gone messages are no longer conflated. A containment check asserts every selected process lies inside the scratch tree before the self-test issues real kills. The xtask mirror test becomes two-way — it now rejects a token orphaned in a mirror after being removed from the script, and a stray token added to a mirror — and the token comparison tightened from substring to exact `|`-delimited match, which also closes a latent substring-aliasing false pass. The Windows comment now distinguishes the miss from the over-match and says which one can cost a bystander.

Judged as a remediation of our own suggestions rather than as original code, it holds up: the changes are load-bearing under mutation rather than decorative, and the author volunteered the one gap he did not close.

## Pinning
Nothing pinned at 4179db1a has been vacated at 79e3e674. Evidence, each re-derived against the current tree rather than carried over:

- **The `has_engine` half of the AND rule** — was pinned by the harness decoy (check 4). Still pinned: gutting it fails, and it is now *additionally* pinned per-root by the new negative assertions. Strengthened.
- **The SIGKILL escalation block** — was pinned by the stubborn decoy. Still pinned: deleting the `kill -KILL` arm fails with "escalation to SIGKILL did not happen".
- **The `kill -TERM` that precedes it** — was pinned by the "ordinary decoy did not escalate" assertion. Still pinned: neutering TERM fails with "ordinary decoy reached SIGKILL escalation". Both branches still fail separately, so neither masks the other.
- **The process-level checks 2–5** — survived the refactor intact at `:395-427`; the matching logic moved into a helper but every process-based assertion that existed before is still there and still fails on the same mutations.
- **Non-vacuity on empty selection** — still holds: an empty `select_leaked` fails checks 2 and 5. The new containment loop would pass vacuously on empty input, but it is additive and cannot mask those.
- **The xtask mirror assertion** — `alternation.contains(needle)` became `present.contains(needle)` over `split('|')`. This is a relabel, so it needed checking: exact-token match is strictly stricter than substring-contains (anything the new form accepts, the old one also accepted; the reverse is false), so the failure set only grew. No case was dropped.
- **The new `__engine-serve-http` exemption** applies only to the mirror→script direction, which did not exist at 4179db1a — so it exempts nothing that was previously enforced.

The only weakening anywhere is cosmetic and additive, not a vacated pin: the step-6 "ok" line at `:443` can be suppressed by an unrelated earlier failure, listed above.

## Sign-off
Three commits in the range, all single-parent — no merge commits on the branch. Read from the raw commit objects:
- `79e3e674` "enforce every reclaim root and marker individually" — `Signed-off-by` present.
- `4179db1a` "fail the reclaim self-test on the branches it guards" — `Signed-off-by` present.
- `89a64a39` "reclaim engine processes leaked into the pre-warm tree" — `Signed-off-by` present.

Scope note: the merge base and the base-branch tip do **not** coincide — the branch forked before the current tip — but the diff is identical either way, so the review scope is unaffected.

## Replies
- The author's per-item account is accurate on five of six points and slightly overstated on one. Verified correct: all six previously-undriven list entries now fail by name (I found all eight do); the contract test's missing direction is genuinely added; containment is present and passing; the Windows comment now says which divergence is a miss and which an over-match; and the disclosed gap is real and honestly stated — deleting the guard's *call site* does still leave the self-test green, exactly as claimed, which I reproduced. The overstatement is "all three arms are asserted" for the same/recycled/gone split: all three *return values* are asserted, but the gone arm is reached only through a failed open, so the empty-cmdline route into it is unguarded by the test — the first non-blocking item above. Where the reply and the test disagree about coverage, the **test is load-bearing** and the reply's wording is what should change; the production code is correct as written.
- One correction to my own process, since it bears on how this code reads. During mutation testing I briefly concluded the self-test's check 1 was cross-talking between runs. It was not: my harness had put the pattern tokens on the invoking shell's command line, and `pgrep -f` matched that shell. Cause is reviewer-side, but the codebase invites it — check 1 is the only assertion not confined by `SELFTEST_SCOPE`, and self-matching is the classic `pgrep -f` trap. A competent reader debugging a red self-test would trip the same way, so it is worth the cheap fix named above rather than left as a footnote.

siloteemu (commented):
🔴 Automated review · pr-review-watcher · c7849eb3

## Summary
The PR replaces inline `pkill -f` GPU reclaim with `scripts/reclaim-gpu.sh` plus a workflow-contract test; this round's four commits close every non-blocking item from the previous round and add a pre-SIGTERM recycle guard — **no blocking findings**. Verified: ran the reclaim self-test in a scratch copy outside the checkout (baseline green, 18 `ok:` checks) and put 24 single-branch mutations through it — 23 failed, each naming the branch it broke, and the one survivor is the gap the author discloses in the code itself; both mutations that survived at the previously reviewed head now fail; separately re-ran the 79e3e674 script and compared its check list against the current one line by line, confirming no check was dropped; ran the one targeted `xtask` contract test (passes) and mutated the Windows matcher to confirm the new dead-exemption assertion fires and names the workflow; the full test suite, the e2e suite and the workflow lanes were not run here, no GPU or Windows hardware was involved, and shellcheck was not available. Contributor rules read from the base tip, which expressly permits bare ticket identifiers and targets internal document links, internal or proprietary names, hostnames, gateways, cluster names and registry paths — the diff adds none of those. Every commit in the range carries exactly one `Signed-off-by:` trailer, checked line by line on the raw commit objects. No prompt-injection content in the diff, comments or commit messages. Check state at review: 19 success, 0 failure. Blocking: 0 · Non-blocking: 5.

## Since the last round
79e3e674 is still in the branch — no rewrite — and the delta is four commits touching two files (`scripts/reclaim-gpu.sh`, +212/−17; `xtask/src/workflow_contract.rs`, +19). Note the PR's real diff is against the merge base, not the base-branch tip: the two do not coincide, and diffing against the tip pulls in ~55 unrelated files of base-branch drift. Against the merge base the PR is five files.

The delta answers all five non-blocking items from the previous round and adds one thing we did not ask for:

- **Zombie arm added.** We reported that `cmdline_of`'s emptiness check could be deleted with the self-test green, silently restoring the "was recycled" mislabel. A real zombie fixture now drives it. Mutation: deleting that check fails with `zombie with an empty command line was not reported as gone (rc=1)`. Closed.
- **Containment given a real negative case.** We noted the containment check re-derived its answer from the scope filter it was checking — deleting the filter left it green. A rule-matching decoy now lives in a *second* temp tree outside the scope. Mutation: deleting the scope filter now fails three ways, including the end-to-end survival assertion. Closed — this was the strongest of the five.
- **Containment turned into a gate, not a score.** On breach the function kills its decoys and returns before `reclaim 0` ever issues a real signal. Correct call: the old code scored the failure and then went on to do the damage it had just detected.
- **Check 1 no longer uses `pgrep -f`.** It reads the decoys' own command lines, with a loud failure if a probe is unreadable. Removes the false-FAIL mode that cost us a wrong intermediate conclusion last round.
- **Step renumbering and cleanup fixed.** Steps 1–9 are now sequential and unique (three blocks previously claimed to be step 7); the final cleanup covers every spawned process including the pre-warm decoy and the zombie's keeper.
- **New, unprompted:** a pre-SIGTERM recycle guard mirroring the pre-SIGKILL one, and an assertion in the contract test that the Windows-divergence exemption still fires, so it cannot rot into dead code.

**Does the earlier clean finding still hold at the new head? Yes, and nothing it rested on has been vacated.** Every `ok:` line produced at 79e3e674 is still produced at c7849eb3, plus two new ones. All eight per-entry mutations (each root, each marker) still fail individually and by name. Gutting either half of the AND rule, deleting the SIGKILL arm, neutering `kill -TERM`, inverting the guard comparison, conflating gone-with-recycled and treating a zombie as alive all still fail. Judged as remediation rather than original code, it holds: the changes are load-bearing under mutation, not decorative.

**None.**

## Non-blocking
- `scripts/reclaim-gpu.sh:289-290` vs `:324` — the `spawn_zombie` header says the zombie pid "is empty if it could not be produced", but the code sets the sentinel `none`, and the only caller tests `!= "none"`. **The code and the inline comment at `:321-323` are load-bearing and correct — the header is what must change**; an empty field would shift the keeper's pid into the caller's variable and make the failure branch unreachable, which is precisely what `:321-323` explains. Confirmed reachable: forcing the sentinel fails loudly with `could not produce a zombie decoy`. Risk is a maintainer later writing a `-z` check that silently never fires.
- `scripts/reclaim-gpu.sh:190-212` — the delta adds a *second* call site for the recycle guard, and like the first it is exercised only in the always-proceed direction: deleting the whole block leaves the self-test green. The comment at `:590-598` says this plainly, which is the right disclosure, but the gap is now twice as wide as when we raised it. If pid reuse stays untestable, consider a seam that lets the self-test force the recycled verdict once.
- `xtask/src/workflow_contract.rs:1439,1475` — `divergence_seen` is declared and asserted *inside* the per-alternation loop, so it demands every alternation name the exempted marker. That is correct only because each workflow has exactly one `Where-Object … -match` line today — the fact the new doc-comment at `:1378-1381` itself flags as not guaranteed. A second, unrelated matcher line would fail this assertion spuriously. Asserting the alternation count explicitly, or hoisting the flag to once per workflow, removes the sharp edge the comment just described.
- `scripts/reclaim-gpu.sh:547-573` — if `select_leaked` returns nothing at all, containment prints its `ok:` line having tested nothing. Not a false green (checks 2 and 5 fail the run), but the evidence line is misleading on exactly the run someone is reading. Assert the selection is non-empty before evaluating containment.
- `scripts/reclaim-gpu.sh:425-436` — the EXIT trap is installed for the first temp tree, then widened three lines after the second tree is created. Nothing between them can realistically fail, so no leak in practice; widening it immediately after the second `mktemp` costs nothing and matches the reasoning already written above the first trap.

siloteemu (commented):
🔴 Automated review · pr-review-watcher · a3949b0c

## Summary
The PR replaces inline `pkill -f` GPU reclaim with `scripts/reclaim-gpu.sh` plus a workflow-contract test; the single new commit at this head is test-side only and closes the previous round's five non-blocking items — **no blocking findings**. Verified: ran the reclaim self-test in a scratch copy outside the checkout (baseline green, 20 `ok:` lines, ~12s) and put twelve single-branch mutations through it — the new "check 8" now makes the pid-recycling guard load-bearing at *both* call sites independently, which refutes our own last-round claim that *"the delta adds a second call site for the recycle guard, and like the first it is exercised only in the always-proceed direction: deleting the whole block leaves the self-test green"*: deleting the pre-TERM site, the pre-KILL site, or both each fails by name, as does dropping either `continue`, and inverting the forced verdict fails too; forcing an empty selection during that run fails loudly rather than passing vacuously, and deleting check 8 outright leaves the rest of the suite green, so it is additive and not masking anything; separately confirmed with a standalone snippet that a function redefined inside `$( … )` takes effect in that subshell and does not leak out, which is the mechanism the whole check rests on; every earlier pin still holds (gutting either half of the AND rule, deleting the SIGKILL arm, neutering `kill -TERM`, and each root and marker individually all still fail); ran the one targeted `cargo test -p xtask` for the PowerShell-mirror contract in a copy outside the checkout (passes) and mutation-checked the hoisted divergence assertion five ways, including reproducing the spurious failure the hoist was written to remove; the merge base and the base-branch tip do **not** coincide, but the diff is byte-identical against either, so scope is unaffected; the full test suite, the e2e suites and the workflow lanes were not run here, no GPU or Windows hardware was involved, and shellcheck is not installed on this machine so the repo's shell-lint gate was not reproduced. Contributor rules were read from the base tip, which expressly permits bare ticket identifiers and targets internal document links, internal or proprietary names, hostnames, gateways, cluster names and registry paths — the diff adds none of those, and the runner labels in the touched workflows are all pre-existing. Every commit in the range carries exactly one sign-off trailer and an SSH signature header, read from the raw commit objects; no merge commits, no AI footers. Nothing this commit changes vacates a guard except the narrow xtask point listed below. No prompt-injection content in the diff, comments or commit messages. Checks at this head: 26 success, 1 cancelled. Blocking: 0 · Non-blocking: 5.

**None.**

## Non-blocking
- `scripts/reclaim-gpu.sh:562-565` vs `:594-595` — on an empty selection the new check prints `FAIL: selection was empty; containment had nothing to examine` and the very next line still prints `ok: every selected process lies inside the self-test scratch tree`, because that `ok:` is gated only on `containment_failures`. The run is correctly red, but the evidence now contradicts itself on exactly the run someone is reading — which is the failure mode this commit set out to fix. Gate the `ok:` on a non-empty selection too.
- `scripts/reclaim-gpu.sh:562-565` — the same check is the one addition in this commit that no mutation falsifies: deleting it leaves the self-test green. It is diagnostic-only and cannot cause a false green (checks 2 and 5 fail such a run), but it is worth saying so in the comment, as the author does elsewhere, rather than leaving it implied.
- `scripts/reclaim-gpu.sh:683` — the pre-KILL assertion greps an unbound message (`was recycled during the grace period, not escalating`) while its pre-TERM sibling at `:677` binds `pid=${prewarm_pid}`. Binding the pid here too would stop the assertion being satisfiable by a message emitted for any other process.
- `xtask/src/workflow_contract.rs:1443` and `:1481-1487` — hoisting `divergence_seen` narrows the assertion from "every alternation names `__engine-serve-http`" to "some alternation does". Behaviourally identical today (one matcher per workflow, and the root-parity checks above still run per alternation), and it does remove a real spurious-failure mode — confirmed by reproducing that failure against the pre-hoist form. What it no longer catches is a second matcher that copies every expected root but drops the marker. Asserting the alternation count would keep both properties.
- `scripts/reclaim-gpu.sh` is not mentioned in `docs/testing.md`, whereas the comparable CI-only self-test scripts the repo already ships (the portable WSL build-deps script, the release-readiness script, the ComfyUI and vLLM GPU tests) all are. One line naming the script and its `--self-test` flag would match the existing convention.

scripts/reclaim-gpu.sh
```diff
@@ -0,0 +1,800 @@
...
+    if [[ -n "${SELFTEST_SCOPE}" ]]; then
+      case "${cmdline}" in
+        *"${SELFTEST_SCOPE}"*) ;;
+        *) continue ;;
+      esac
+    fi
+
+    cmdline_matches_rule "${cmdline}" || continue
+
+    printf '%s\t%s\n' "${pid}" "${cmdline}"
```
juhovainio: This record format can't survive a command line containing an embedded newline byte: `select_leaked()` emits `printf '%s\t%s\n' "$pid" "$cmdline"`, and the consumer in `reclaim()` reads it back with `IFS=$'\t' read -r pid cmdline`, so a newline inside `cmdline` splits into an extra bogus record. In practice this would feed a truncated/misaligned cmdline into `same_selected_process()`, which could misjudge a genuinely leaked process as "recycled" and leave it unreclaimed. Narrow edge case (real engine argv rarely contains a literal newline), but worth a defensive note or a guard given how much rigor the rest of this file puts into correctness.

scripts/reclaim-gpu.sh
```diff
@@ -0,0 +1,800 @@
...
+  echo "--- rocm-smi KFD processes (per-process VRAM) ---"
+  # Process names show as UNKNOWN inside a container: KFD reports host PIDs,
+  # which do not resolve in the container's PID namespace. The VRAM column is
+  # still the answer to "what is holding the card".
+  timeout 15 rocm-smi --showpids 2>&1 | head -40 || true
+  echo "--- engine/serve processes visible here ---"
+  # shellcheck disable=SC2009 # pgrep cannot print elapsed time, and how long a
+  # holder has been alive is what distinguishes a leak from this job's own serve.
+  ps -eo pid,etimes,args 2>/dev/null |
+    grep -Ei 'llama-server|vllm|__engine-serve-http|rocm daemon' |
```
juhovainio: This hardcodes its own copy of the engine-marker pattern (`llama-server|vllm|__engine-serve-http|rocm daemon`) instead of deriving it from the `ENGINE_MARKERS` array above. It's diagnostics-only today so a drift here is low severity, but it's the same class of duplication the PR's own `assert_rule_covers_every_list_entry` and the new `xtask` contract test exist to prevent elsewhere in this file - worth building the grep alternation from `ENGINE_MARKERS` directly so this call site can't quietly go stale.

juhovainio (changes requested):
Reviewed this PR - it's a well-built replacement for the old inline `pkill` reclaim logic, with a genuinely thorough self-test (six decoy processes, pid-recycling guards exercised via a forced override, zombie handling) and a contract test that keeps the bash `E2E_ROOTS` list honestly pinned to the two PowerShell mirrors. The known Windows engine-marker divergence (EAI-8815) is disclosed and tracked rather than silently left in, which I appreciated.

Two small things worth fixing, left as inline comments:
1. The record format `select_leaked()` emits (`printf '%s\t%s\n' "$pid" "$cmdline"`) can't survive a command line that itself contains a newline byte - rare for a real engine invocation, but it would silently mislabel a leaked process as "recycled" rather than reclaiming it.
2. `report_holders()` hardcodes its own copy of the engine-marker pattern instead of deriving it from `ENGINE_MARKERS`, so it can drift out of sync the same way the PowerShell mirrors were shown to - the file's own contract test elsewhere guards against exactly this class of drift, just not for this one call site.

scripts/reclaim-gpu.sh
tomastola: Confirmed, and it was worse than "a defensive note" — fixed in `40c922a`. Reproduced first, with a decoy whose argv carried a literal newline: the single record split in two, the genuine pid was reported `was recycled before it could be terminated`, survived both TERM and KILL, and the run finished with `0 process(es) terminated` while still holding the card.

A tab does the same thing, since it *is* the field separator, so both are handled — in `cmdline_of`, the one point every consumer already reads through, rather than at each call site. Matching is unaffected (no root or marker contains either byte); the identity comparison does get marginally coarser, which the comment now states rather than denies. The PowerShell mirrors pass CIM objects straight to `Stop-Process` and never serialise a record, so they need no counterpart.

Self-test check 3 pins it, with a decoy asserted to actually carry both bytes so it cannot pass having examined a string that could not break it; check 10 adds the end-to-end cost. Reverting the flattening fails check 3 by naming the malformed record.

scripts/reclaim-gpu.sh
tomastola: Fixed in `40c922a` — the filter is now built as `-e` arguments from `ENGINE_MARKERS`. Used `-F` rather than an `-E` alternation, so the markers keep the same literal-substring meaning they have in the rule; a future marker containing a metacharacter would otherwise mean one thing here and another in `cmdline_matches_rule`.

One honest gap: no self-test covers this. Re-hardcoding the list leaves the suite green. The argument is that deriving it removes the copy that could drift rather than guarding it — but that is a structural claim, not an assertion, so it is in the PR's known-gaps list.

I also chose not to add an empty-array guard, and checked rather than assumed why: `grep` given zero patterns prints its usage and exits 2, which the trailing `|| true` would swallow the status of — but reaching that needs `ENGINE_MARKERS` to be empty, and `assert_rule_covers_every_list_entry` already reds the self-test on that (verified by emptying it). A guard here would be a branch the self-test could never reach.

tomastola:
Rebased onto main and force-pushed (`40c922a` → `d0b4da5`), so the inline anchors above may show as outdated. Both review comments are addressed in the commit that precedes the rebase; replies are in their threads.

**Why the rebase.** The `E2E tests (Strix Halo, WSL2)` failure was not a flake and not this branch's code — the DevLab pool operators deliberately fail jobs requesting `[self-hosted, linux, strix-halo, wsl]`, because #403 retired that static runner and rebuilt the lane on the DevLab Dispatch pool. The branch was 24 commits behind and never had it. Their message says to rebase, so I did.

**What the rebase changed in this PR.** The two WSL lanes no longer carry a reclaim step. They now run on a Windows host, provision WSL2 fresh per job and unregister the distro afterwards, so no prior run's processes can survive into them, and their GPU preflight is advisory. Porting the step there would have been a no-op whose premise no longer holds — and `run: bash …` would have been wrong on a Windows runner anyway. That takes this from ten bash call sites to **eight**, plus the two PowerShell mirrors, unchanged. `docs/testing.md` says so, and the description is updated.

**On the other red lane, `E2E tests (Strix Halo, Ubuntu)`** — this one is worth a look independently of this PR. It failed the GPU preflight at `0 GiB free`, and the `--report-holders` diagnostics this PR adds printed:

```
--- rocm-smi KFD processes (per-process VRAM) ---
No KFD PIDs currently running
--- engine/serve processes visible here ---
--- of those, E2E-owned (reclaim would take these) ---
```

Nothing holds the card — no KFD processes, no engine processes at all — yet the preflight computes `0 GiB free` and waits out its full 90s. That points at the preflight's own `free = total - used` read on this device, not at a leak any reclaim could take. It failed identically on the previous head, before this round's changes existed, so it is pre-existing rather than introduced here. I have not tried to fix it in this PR; happy to file it separately.

**Two non-blocking findings from my own convergence pass**, disclosed rather than sat on:

1. The header claims a manual `/workload` serve is safe because "it names no E2E root". That is narrower than the rule: `e2e-shared` is matched as an unanchored substring, so a hand-run serve that reuses the shared model cache to avoid re-downloading weights would satisfy both halves and be reclaimed. The claim, not the rule, is what is wrong — no self-test fixture covers it.
2. `reclaim()`'s `killed` counter increments after the identity check but without checking `kill`'s status, so a process exiting in that window is still counted as terminated — a small version of the unconditional-"reclaimed" inaccuracy this script was written to remove.

Neither blocks, and I have left both unfixed rather than extend this PR again without your say-so. Say the word and I will take either.

siloteemu:
🔴 Automated review · pr-review-watcher · d0b4da58

## Summary
Replaces the inline per-lane `pkill` reclaim in the self-hosted GPU E2E lanes with a single `scripts/reclaim-gpu.sh` implementing an order-independent "E2E root AND engine marker" rule, adds a decoy-driven `--self-test`, a `--report-holders` diagnostic on preflight failure, and an xtask contract test pinning the PowerShell mirrors' root list to the script's. Outcome: Needs work — two blocking findings, both small and local. Verified: ran `scripts/reclaim-gpu.sh --self-test` on a scratch copy (green, 22 checks) and then six single-branch mutations of that copy — removing the engine half of the rule, the self-test scope filter, the `kill -TERM`, the pre-TERM recycle guard, and the empty-cmdline check each turn it red, confirming those branches are genuinely covered; read the xtask contract test and confirmed its root comparison runs in both directions off parsed source (not hardcoded lists) and cannot pass vacuously; confirmed `scripts/**` is in the `heavy` path filter so the new CI step actually runs; confirmed the tree under review was unmodified (clean worktree, head unchanged) and that the merge-base and base-tip diffs cover an identical file set; leak and prompt-injection scans of the added lines came back clean. The full workspace build and the E2E suite were not run here. This review worked from a CI conclusion set of 25 success, 1 pending, 1 failure at this head; I could not attribute the failing check to a specific defect in this diff, so it is an observation, not a finding of mine.
Blocking: 2 · Non-blocking: 4.

**1. scripts/reclaim-gpu.sh:116 — the tab half of the newest commit is inert, and the self-test only appears to cover it.**

The head commit changes `cmdline_of` to `tr '\0\n\t' '   '` and justifies it with a ~20-line comment claiming both bytes break the `pid<TAB>cmdline<NEWLINE>` record format, "the newline splits the record, the tab shifts the field boundary", and that the cost (a coarser identity comparison in `same_selected_process`) is worth paying for both.

Measured on a scratch copy, that is half true:

- `tr '\0\t' '  '` (newline flattening removed) → self-test RED. The newline half is real and covered.
- `tr '\0\n' '  '` (tab flattening removed, newline kept) → self-test **all checks passed**. No assertion anywhere can fail from deleting the tab from that `tr` set.

And the claimed mechanism for the tab does not occur. `/proc/<pid>/cmdline` is NUL-terminated after the final argument, so the flattened string always ends in a space; bash's `read` strips only trailing IFS whitespace, and a space is not in `IFS=$'\t'`, so an embedded tab is never at the stripping boundary. A mid-string tab is preserved verbatim in the last field of `read -r pid cmdline`. I confirmed this end to end against a live process whose final argument ends in a tab: the read-back string is byte-identical to the original with the tab half removed.

This matters beyond tidiness because check 3 is written to look like it covers the tab — it fails the run with "delimiter decoy's command line carries no tab; the record-format check is vacuous" if the fixture stops carrying one, and its comment states the tab is asserted individually — so a future reader will believe that branch is guarded when nothing guards it. That is the exact shape of defect the rest of this script is built to eliminate.

Fix (either is fine, but the comment must stop claiming coverage it does not have):
- preferred: drop `\t` from the `tr` set and delete the paragraph asserting the tab mechanism, which also restores the finer identity comparison the commit gives up; keep the newline flattening and the newline half of check 3; or
- keep the tab as deliberate defence-in-depth against an argv-rewriting process whose `/proc` cmdline is not NUL-terminated — but say that is the reason, drop the false "shifts the field boundary" claim, and either add a fixture that actually fails without it or state plainly in the comment that this half is unexercised.

**2. scripts/reclaim-gpu.sh:39 (with .github/workflows/nightly.yml:689) — the `/tmp/` prefix makes the scenario-temp-dir root miss a lane this PR edits.**

`E2E_ROOTS` declares `'/tmp/rocm-e2e'` and the file header states the rule "covers every E2E-owned tree". The nightly Strix lane sets `TMPDIR: /home/ubuntu/actions-runner/tmp` at nightly.yml:689, and the suite creates its per-scenario roots with `TempDir::with_prefix("rocm-e2e-")` (tests/e2e-cucumber/tests/e2e.rs:197), which resolves through `env::temp_dir()` and therefore honours `TMPDIR`. On that lane the scenario dirs are `…/actions-runner/tmp/rocm-e2e-*`, which contain no `/tmp/rocm-e2e` substring, so an engine running out of a scenario tree there is not selected — and `main`'s `rm -rf /tmp/rocm-e2e-*` cleans nothing there either, defeating the stated disk-fill guard on the one lane that redirects its temp dir.

The miss itself predates this PR (the inline patterns carried the same prefix), but the completeness claim and the pinned array are new here, this PR is specifically about ending silent reclaim misses, and the fix is one character-range: make the root `'rocm-e2e'`. That is already the portable form the two PowerShell mirrors use, and the contract test at xtask/src/workflow_contract.rs already strips `/tmp/` before comparing — so dropping the prefix makes the three lists literally identical and removes the strip. The `AND`-an-engine-marker half keeps a manual serve safe exactly as it does for the other three prefix-less roots. Update `rm -rf /tmp/rocm-e2e-*` to cover `${TMPDIR:-/tmp}` too, and update `assert_rule_covers_every_list_entry`'s `expected_roots` and the header comment to match.

## Non-blocking
- .github/workflows/e2e-selfhosted.yml:944 — the WSL lane's new `bash scripts/reclaim-gpu.sh --report-holders` runs inside the guest with no `cd`, while every other repo-using step in that lane explicitly `cd /root/work/rocm-cli` first (lines 879, 982); it resolves only through whatever working directory `wsl.exe` inherits. Add the same `cd`, or use an absolute path.
- docs/testing.md — "The WSL lanes do not run it" is now slightly stale: those lanes do invoke the script, for `--report-holders` on the preflight failure path. One clause would fix it.
- The PowerShell mirrors' engine alternation is deliberately not pinned by the contract test (only the root alternation is), so an accidental drop of e.g. `vllm` from the Windows engine clause stays silent; the divergence is disclosed in-line and tracked, but the PR text is worth saying so in as well.
- scripts/reclaim-gpu.sh — 364 of 964 lines are comment, a large share of it narrating prior review rounds ("which is how it went untested", "left X removable with the self-test green"). The rationale that explains the code is valuable; the change-history narration will not age well and would read better as commit-message content.

siloteemu (changes requested):
🔴 Automated review · pr-review-watcher · d0b4da58

Blocking: 2. Both are small and local, and both were confirmed by mutating a scratch copy rather than by reading alone.

**1. scripts/reclaim-gpu.sh:116 - the tab half of the newest commit is inert, and the self-test only appears to cover it.**

The head commit widens the flattening set in cmdline_of to include the tab byte and justifies it with a long comment claiming both bytes break the record format - "the newline splits the record, the tab shifts the field boundary". Measured on a scratch copy, that is half true. Removing the newline from the set turns the self-test red, so the newline half is real and covered. Removing the tab from the set leaves every check passing: no assertion anywhere can fail from deleting it.

The claimed mechanism for the tab does not occur either. The per-process command line is NUL-terminated after the final argument, so the flattened string always ends in a space; the shell's read strips only trailing whitespace that is in the field separator, and a space is not, so an embedded tab is never at the stripping boundary and a mid-string tab survives verbatim in the last field. This was confirmed end to end against a live process whose final argument ends in a tab: the read-back string is byte-identical with the tab half of the change removed.

This matters beyond tidiness because check 3 is written to look like it covers the tab - it fails the run with a message about the record-format check being vacuous if the fixture stops carrying a tab, and its comment states the tab is asserted individually - so a future reader will believe that branch is guarded when nothing guards it. That is the exact shape of defect the rest of this script is built to eliminate, which is why it gates rather than being a note.

Either fix is fine, but the comment must stop claiming coverage it does not have: drop the tab from the flattening set and delete the paragraph asserting the tab mechanism, which also restores the finer identity comparison the commit gives up; or keep the tab deliberately as defence against a process whose command line is not NUL-terminated, say that is the reason, drop the false field-boundary claim, and either add a fixture that actually fails without it or state plainly that this half is unexercised.

**2. scripts/reclaim-gpu.sh:39, with .github/workflows/nightly.yml:689 - the completeness claim this pull request adds is not true on one of the lanes it edits.**

The roots array declares a scenario-temp-dir root carrying a leading absolute-temp prefix, and the new file header states the rule covers every end-to-end-owned tree. The nightly Strix lane redirects its temp directory to a path under the runner's home, and the suite creates per-scenario roots through the standard temp-dir lookup, which honours that redirection. On that lane the scenario directories do not contain the prefixed substring at all, so an engine running out of a scenario tree there is not selected - and the existing cleanup glob on the base branch removes nothing there either, defeating the stated disk-fill guard on the one lane that redirects its temp directory.

The miss itself predates this pull request. What is new here is the completeness claim, the pinned array and a contract test that asserts the rule covers every list entry, and this change is specifically about ending silent reclaim misses - so shipping it with the claim false is the objection, not the pre-existing gap. The fix is to drop the absolute-temp prefix from that root, which is already the portable form both PowerShell mirrors use and which the contract test currently strips before comparing, so the three lists become literally identical and the strip can go. The engine-marker half of the rule keeps a manual serve safe there exactly as it does for the other prefix-less roots. The cleanup glob and the contract test's expected roots need the matching update.

The full round of findings, including four non-blocking items, is in the review comment posted alongside this one.

tomastola:
CI has finished at `d0b4da5`: **27 green, 1 red**. All three datacentre GPU lanes pass (MI300X, MI350P, rad3 R9700), as do Strix Halo Windows and WSL2 — WSL2 green confirms the rebase was the right call for that lane.

The one red is `E2E tests (Strix Halo, Ubuntu)`, and I can now show it at this head rather than the previous one:

```
reclaim: no leaked E2E engine processes found
GPU preflight failed after 90s: ... only 0 GiB free (< 8 GiB)
No KFD PIDs currently running
--- engine/serve processes visible here ---            (empty)
--- of those, E2E-owned (reclaim would take these) --- (empty)
```

No KFD processes, no engine processes, nothing for reclaim to take — so the `0 GiB free` is the preflight's own `total - used` read on that device, not a held card. Same failure on the previous head, before this round existed. Worth noting the consequence rather than just the cause: because that lane never clears preflight, it contributes no test signal here, so the Linux reclaim path rests on the other three GPU lanes — which are green. I have not touched it in this PR and can file it separately if you want.

Re-requesting review. The two comments are answered in their threads and I have left both unresolved for you to close. The two non-blocking findings from my own pass are in the comment above, still unfixed — happy to take either in this PR or a follow-up, your call.

siloteemu:
🔴 Automated review · pr-review-watcher · d0b4da58

Answering your reply; no new code review this round, since the head has not moved since the change request.

**The red lane.** Your reading looks right, and it is not yours to fix here. There is already an open pull request against that gate, #442, "gate the Strix Halo GPU preflight on GTT when VRAM is a carveout" — which is the same condition your log shows: a preflight computing free memory from a reported VRAM total on a device whose engine allocates from a different pool, so it reports `0 GiB free` with nothing holding the card. No need to file it separately, and no need to touch it in this PR. The consequence you state is the right one to record: that lane contributes no signal, so the Linux reclaim path rests on the other three GPU lanes, which are green.

**The change request still stands, and I do not think the thread answers reach it.** The two threads you answered are juhovainio's, from the 23rd. The change request came after them and makes a narrower claim about the same code.

The distinction is single-branch mutation versus a wholesale revert. Your thread note says reverting the flattening fails check 3. That is true, and expected — reverting it removes the newline handling too, and the newline half is genuinely covered. The finding is that deleting *only* the tab from the flattening set, with the newline handling left in place, leaves every check passing. Check 3 names the malformed record when both bytes go; it cannot fail when only the tab does. The comment beside it nevertheless states the tab is asserted individually, and that is what makes it worth gating rather than noting.

The second finding is independent of that one and unaddressed so far: the completeness claim the file header now makes, together with the contract test that pins it, is false on the nightly lane that redirects its temp directory.

Both fixes are spelled out in the change request, and either finding is cheap to resolve.

**Your two non-blocking findings.** Your call entirely — neither is something I would hold this up for, and a follow-up is fine.

PR #410, opened 2026-09-16
fix(examine): stop the e2e scenario requiring the interpreter to sit under the runtime root

r0x0r:
Follow-up to #405 (merged), addressing the non-blocking review items left on it. No blocking findings were raised; these are the rest.

## The one that is a real defect

`examine-15`'s step asserted that the interpreter named in `framework_notes` is *contained in* `active_runtime_root`. That directly contradicts the documented reason `managed_therock_python_executable` prefers the recorded path in the first place:

> an imported or read-only runtime can record an interpreter that does not sit under `install_root` at all

I wrote both the doc comment and the assertion, and they disagree. A GPU runner holding an imported or read-only runtime would have failed a correct build. The step now asserts the note **names** an interpreter, and says in a comment why it deliberately does not check where that path lives.

## A test whose name overclaimed

`the_framework_probe_gives_the_runtimes_torch_its_library_path` clears the library paths itself, so it passes whether or not the composition in `runtime_library_path_env` exists. The test that actually pins the composition is its sibling, whose fake interpreter only answers when the loader path reached it.

Renamed to `a_runtime_without_its_library_path_reports_the_import_failure` — what it really asserts — with a line pointing at the sibling, so the next reader is not misled about which test guards what.

## Split divergence

`runtime_library_path_env` split the inherited value with `std::env::split_paths`, while the sibling composition in `probe_runtime_devices` uses `runtime_path_list_split`, which on Windows also trims, drops empty entries and host-normalises each one. #405 unified the *constant* and left the composition divergent, which reads as more sharing than there was. Both now split the same way.

## Missing tests

`active_managed_framework_interpreter` had none. Three added:

- no managed runtime → `None`, so the ambient probe stays the fallback
- a record written before the installer recorded `python_executable` still resolves, from the conventional location under `install_root`
- **a recorded interpreter that is no longer on disk → `None`**, rather than being handed back as a path that cannot spawn

The last is load-bearing: removing the `exists()` gate makes it fail (verified by mutation).

## Recorded, not changed

The resolver now documents that it is not the only "which runtime is active" selector. The human report goes through `current_runtime_manifest`, which wants an active key or a single `default_runtime_id` match; this one falls back to the most recently installed runtime. They can disagree when no active key is set, so `examine --json` can report a framework read from a runtime the human form calls `active_runtime_status: unset`. That predates #405 — noted because reading the framework through the runtime is what makes it observable.

## Verification

`cargo test -p rocm-core --lib` → 391 pass (up from 388). `cargo clippy --workspace --all-targets -- -D warnings` clean. `examine-15` re-run on the MI300X box with the relaxed assertion: 3/3 steps pass.

Lane note (AGENTS.md §3): `examine-15` is `@requires-gpu`, so the positive branch is exercised by the self-hosted GPU lanes — MI300X, MI350P, rad3 R9700 and Strix Halo Ubuntu/WSL2/Windows. The GitHub-hosted lane skips it.


siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · a9f3c0f1

## Summary
Loosens an examine e2e assertion so it no longer requires the interpreter to sit under the runtime root, renames a unit test, adds three tests for `active_managed_framework_interpreter`, and swaps `std::env::split_paths` for `runtime_path_list_split` when merging the inherited loader path. Needs work: the production swap is a behaviour regression on Windows and is covered by no test on any platform. Verified: `cargo check -p rocm-core` clean; the three new tests and the renamed test pass under targeted `cargo test -p rocm-core --lib`; branch-level mutations were run on a scratch copy, not the tree under review; the quote semantics below were read out of the toolchain's own `std` source rather than from memory; the full suite was not run here. Measured check state at this head: **23 success, 1 skipped, 1 failure, 1 pending** — the failure is in a GPU pre-flight step that timed out waiting for the card to be free, with the test step itself skipped, so it is not evidence about this change. Blocking: 2 · Non-blocking: 4.

**1. `crates/rocm-core/src/examine.rs:1585` — the split swap regresses the inherited `PATH` on Windows, and drops the whole loader path when it does.**

`RUNTIME_LIBRARY_PATH_ENV` is `PATH` on Windows (`crates/rocm-core/src/runtime.rs:107`). `std::env::split_paths` is quote-aware there: it strips `"` and treats `;` inside quotes as literal, because `c:\some;dir` is a legal Windows path. `runtime_path_list_split` (`crates/rocm-core/src/runtime.rs:329`) does a naive `split(';')` and keeps the quote characters — `normalize_windows_runtime_path_text_with_separator` (`runtime.rs:739`) matches neither its `/x/` nor its `X:` branch for a string starting with `"`, so the quotes survive into the `PathBuf`.

The failure is not a single corrupted entry. `std::env::join_paths` on Windows returns `Err` as soon as *any* segment contains a `"`. So one quoted entry anywhere in the inherited `PATH` sends `runtime_library_path_env` down its error arm (`examine.rs:1591-1597`), which returns an empty env — the torch probe then runs with **no** library paths, the import dies on missing ROCm libraries, and `rocm examine` reports the runtime as broken. That is verbatim the misdiagnosis this function's own doc comment (`examine.rs:1570-1575`) says the path exists to avoid. Before this hunk, `split_paths` stripped the quotes and the join succeeded.

Scope, stated fairly: a `PATH` with no quoted segment behaves the same before and after, so this bites only when a quoted entry is present — legal, and not rare on real machines. When it does bite, the whole loader path is lost rather than one entry.

Fix: revert this line to `std::env::split_paths`, or — better, since the same naive split already sits behind `split_runtime_path` (`apps/rocm/src/therock.rs:4990`, used at `therock.rs:4049` for this same variable) and `split_runtime_paths` (`apps/rocm/src/comfyui.rs:1360`) — make `runtime_path_list_split` itself quote-aware and fix all three. The change aligns the divergence in the wrong direction: it propagates the lossy splitter to a third call site instead of correcting it. Whichever way it goes, `runtime_path_list_split` wants a doc line saying it parses *runtime-recorded* path lists and is not a substitute for `std::env::split_paths` on an inherited OS `PATH` — its name invites exactly the substitution made here, and the next reader will make it again.

Coverage: reverting this line is caught by nothing. Both tests touching `runtime_library_path_env` miss it — `a_runtime_without_its_library_path_reports_the_import_failure` clears `library_paths` and returns at `examine.rs:1579` before reaching the merge, and `the_framework_probe_reads_the_active_runtimes_torch` never sets `RUNTIME_LIBRARY_PATH_ENV`, so the `if let Some(existing)` branch is not entered. Both are `#[cfg(unix)]`, and `runtime_is_windows()` is a `const fn` over the compile-time host, so the Windows arm cannot be reached from a non-Windows test at all. The only behaviour change in this PR is untested on every platform, including the one where it does anything.

**2. `crates/rocm-core/src/lib.rs:10391-10420` — `write_therock_runtime_with_interpreter`'s `Some` arm is dead, and hides the one branch the PR leans on.**

The helper takes `recorded_python: Option<&Path>` and writes `record["python_executable"]` at `lib.rs:10412-10414`, but its single call site (`lib.rs:10440`) passes `None`; there is no other call site in the tree. The `Some` arm never executes. Consequently the *recorded-interpreter-preferred-over-derived* branch of `managed_therock_python_executable` (`lib.rs:3825-3831`) has no test at all: deleting that branch outright leaves all three new tests green (verified by mutation on a scratch copy). That is the branch whose documented rationale — an imported or read-only runtime can record an interpreter outside `install_root` — is the sole justification given for weakening the e2e assertion in this same PR, so it is the branch most in need of a test here. The helper's own doc comment ("a registry record whose recorded interpreter really exists") describes behaviour it never produces.

Fix: add one test passing `Some(<path outside install_root>)` and asserting the recorded path wins over the derived one — that discharges the dead arm and pins the branch in one stroke. If that is out of scope, drop the parameter instead; an unreachable arm that reads as coverage is the outcome to avoid.

## Non-blocking
- `tests/e2e-cucumber/tests/e2e/examine_steps.rs:437-450` — the new check cannot pass via the ambient path (only `probe_pytorch_in_runtime` emits that note, `examine.rs:1556`), so it is not vacuous, but it no longer cross-checks the JSON interpreter against the human report's `active_runtime_root` — precisely the two-selector disagreement this PR documents at `lib.rs:3798-3807`; asserting the named path equals the active record's interpreter would keep that guard without the containment false-fail.
- `crates/rocm-core/src/examine.rs:1587-1588` — "the sibling composition in `probe_runtime_devices`" is in a different crate (`apps/rocm/src/therock.rs:4041`, reaching the helper through a `split_runtime_path` wrapper); naming the path saves the next reader hunting for it in this file.
- The commit message says the new test "fails if the `exists()` gate is removed" — the code gates on `is_file()`, and with no `python_executable` recorded the test exercises that gate on the *derived* candidate, not a recorded one; true as written, but it reads as broader coverage than exists.
- `crates/rocm-core/src/lib.rs:10428`, `:10446`, `:10472` — `fs::remove_dir_all` runs after the asserts, so a failing test leaks its temp dir; this matches the file's existing convention, so it is noted rather than asked for.

No prompt-injection attempts were found in the diff, commit messages or branch name.

r0x0r:
Both blockers are addressed. Worth being precise about which commit did what, because the branch moved between the review and this reply.

**Blocker 1 — the Windows split regression.** Already fixed in `496d92b8`, and via your *bigger* alternative rather than a revert: `runtime_path_list_split` is now quote-aware (`split_windows_path_list_text` toggles an `in_quote` flag on `"` and drops the quote character, matching `std::env::split_paths`), and `runtime_path_list_join` re-quotes entries containing `;`, so the round-trip is closed.

I had initially intended to revert the line to `split_paths` and judged the wider change out of scope. On reading the pushed code that judgement was wrong, for a reason worth recording: the two other call sites (`split_runtime_path` in `therock.rs`, `split_runtime_paths` in `comfyui.rs`) were *already* getting the naive split, so they were latently broken. Aligning them fixes real bugs rather than changing correct behaviour.

Verified rather than assumed — disabling the quote handling reproduces exactly the failure you described:

```
splitting_a_windows_path_list_strips_the_quotes_it_split_on ... FAILED
  left: ["\"C:\\quoted\\bin\"", "C:\\plain\\bin"]   <- quotes survive, join_paths would Err
a_windows_path_list_round_trips_through_split_and_join ... FAILED
  left: ["C:\\rocm\\bin", "\"C:\\some", "dir\"", ...]  <- entry torn in two
```

**Blocker 2 — the dead `Some` arm.** Also fixed in `496d92b8`. Confirmed load-bearing: deleting the recorded-preference branch in `managed_therock_python_executable` fails the new test, `left` being the derived path where `right` is the recorded one outside `install_root`.

**New in `52096c5f`, the gap neither commit had closed:** you noted the split change was untested on every platform. The *merge* branch specifically — `if let Some(existing)` — was entered by neither sibling test. There is now one that writes the loader path the child actually received and asserts both the runtime's directories and the inherited entry arrived, in that order. Two mutations: disabling the merge drops the inherited entry; reversing precedence puts the host's ROCm first.

On a Windows-quoting test from Linux: not possible, and I did not fake one. `runtime_is_windows()` is a `const fn` over the compile-time host, so `runtime_path_list_split`'s Windows arm is unreachable from a Linux test. `496d92b8` handles this correctly by testing `split_windows_path_list_text` directly. What no Linux lane can prove is that the splitter actually *routes* into it — that needs the Windows lane.

**Non-blocking 1 (the lost cross-check).** Restored, but conditioned rather than reinstated. Asserting "the named path equals the active record's interpreter" would mean restating `managed_therock_python_executable` in the test — asserting the implementation against a copy of itself. Instead the containment check is gated on `active_runtime_mode`, which the human report already prints: `managed` asserts containment, `read-only` skips it. `read_only` is set only on import/adopt — exactly the runtimes that can record an interpreter outside `install_root`. The guard is back where it holds and gone only where it was wrong. Executed on an MI300X host with a managed runtime active, so the containment branch really ran: 3/3 steps pass.

**Non-blocking 2:** the comment now names `apps/rocm/src/therock.rs`. **3:** you are right that `a9f3c0f1`'s message overstated the coverage — the gate is `is_file()`, and with nothing recorded the test exercised it on the *derived* candidate. Corrected in `52096c5f`'s message, since the pushed one cannot be rewritten. **4:** confirmed and left alone as the file's existing convention.

Also added a `PROCESS_ENV_TEST_LOCK` to examine.rs's test module (matching `fix.rs`'s pattern), taken by all three tests that read the loader variable — not just the new setter — since the other two compose a child env from it.

Gate: `cargo test --workspace --all-targets`, `clippy --workspace --all-targets -D warnings`, `fmt --check`, `smoke_local.py` all clean.

siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · 52096c5f

## Summary

Both previously-raised counts are RESOLVED at this head: count 1 (the Windows loader-path regression) is fixed at the root — the splitter is now quote-aware, and the Windows shape is pinned by five host-independent tests; count 2 (the dead `Some` arm and the untested recorded-interpreter branch) now has a real caller and a test that goes red when the branch is deleted. One new blocking issue: a test added earlier in this PR names a case it does not construct, and the head corrects that only in a commit message, not in the code. Verified: I derived the merge base (it equals the base tip, so both diffs are identical), re-ran the count-2 mutation in a scratch copy outside the checkout — deleting the recorded-interpreter-preferred branch turns `a_recorded_interpreter_outside_the_install_root_wins_over_the_derived_one` red — and checked the split/join semantics against the Rust standard library's own Windows source, which confirms its join rejects only `"` and re-quotes `;`, so stripping quotes is what keeps the composition on its success path; I did not run the full suite, the end-to-end lanes, or any Windows lane, so the claim that the Windows branch is actually reached at runtime is reasoned from the source rather than executed. Checks at this head: 18 success, 1 skipped, 0 failure, 0 pending (a skipped job is not a failure). Blocking: 1 · Non-blocking: 5.

### Previously-raised counts

**COUNT 1 — RESOLVED.** The call site now reads `entries.extend(crate::runtime_path_list_split(&existing));`, and the splitter it calls is no longer naive. `split_windows_path_list_text` implements the same quote-toggle state machine the standard library uses — `'"' => quoted = !quoted`, `';' if !quoted =>` separates, and the quote character is never pushed into the entry — so no `"` survives into a path, the join stays on its success arm, and the whole inherited list is preserved. The mechanism I originally described is gone, and it is gone at the layer that also backed the two sibling splitters. Coverage is genuine and host-independent: the five new tests drive the text helpers directly, and I confirmed by inspection that removing the trim, the empty-entry filter, the quote handling, or the re-quoting on the way back out is each caught by at least one of them. One residual, which the change documents rather than papers over: on a non-Windows lane `runtime_path_list_split` delegates straight to the standard splitter, so reverting the one-line swap at the call site is behaviourally inert there and cannot be caught — inherent to the host-constant, and stated plainly in the commit rationale.

**COUNT 2 — RESOLVED.** `write_therock_runtime_with_interpreter` now has a second call site passing `Some(&recorded)`, and `a_recorded_interpreter_outside_the_install_root_wins_over_the_derived_one` plants both candidates on disk so the existence gate cannot decide the outcome. I re-ran the mutation: with the recorded-interpreter candidate deleted from `managed_therock_python_executable`, that test fails on `the recorded interpreter must win over the one derived from install_root`, returning the derived path instead. The branch the relaxed end-to-end assertion rests on is now pinned. The parameter is no longer dead — the compiler's unused-field warning for `python_executable` appears only under the mutation.

- `crates/rocm-core/src/lib.rs:10491` — `a_recorded_interpreter_that_is_gone_does_not_resolve` does not test what it names. The record it writes has no `python_executable` key at all, so nothing is recorded; the test exercises the absence of the *derived* candidate under a never-created `install_root`. Its assertion message repeats the false claim — `"a recorded interpreter that is not on disk must not be handed back"` — and the recorded-and-missing path it names is never constructed. This is the coverage-claim failure mode the standing rule treats as blocking: a future reader deletes the recorded-interpreter half of the existence gate and this test stays green, because there is no recorded candidate for it to gate. The head's own commit rationale diagnoses exactly this ("with no `python_executable` recorded that test drives the gate on the DERIVED candidate, not a recorded one") but corrects only the commit message — the misleading name and message ship unchanged. Fix, either: rename to what it actually pins (e.g. `an_interpreter_that_is_not_on_disk_does_not_resolve`) and reword the assertion message to drop "recorded"; or keep the name and pass `Some(<a path that does not exist>)` through the helper so a recorded candidate really is present and really is rejected. The second is preferable — it closes the gap instead of renaming around it.

## Non-blocking

- `crates/rocm-core/src/lib.rs:10388` — the helper's doc states two things that are not so: it names an `exists()` gate, but the production code gates on `is_file()`; and "a registry record whose recorded interpreter really exists" is false for the `None` arm, which writes no recorded interpreter at all. Split the sentence per arm and name the real predicate. Related: no new test distinguishes `is_file()` from `exists()` — relaxing the gate to `exists()` survives all of them.
- `crates/rocm-core/src/runtime.rs:385` — `windows_path_list_entry_text` asserts in prose that a correctly split entry cannot contain `"`. True for a real Windows path and for anything the new splitter emits, but the function emits such an entry raw where the standard library's join returns an error instead, and nothing tests it. The passthrough predates this change; the assertion is new. Either state the precondition the claim rests on, or reject the character.
- `crates/rocm-core/src/examine.rs:2700` — the loader-variable save/restore around `probe_framework` is straight-line code with no drop guard, so a panic inside the probe leaks a mutated process-global into the rest of the binary and merely poisons the lock (which is then recovered). The sibling module has the identical shape, so a small shared scope guard applied to both is the uniform fix rather than a local patch here.
- `tests/e2e-cucumber/features/examine.feature:176` — the only scenario reaching the reworked step activates a managed runtime, so the guard's skip path, which is the entire reason for the relaxation, is exercised only by a unit test. Production behaviour is unchanged by this PR, so the contributor rule requiring a scenario for user-observable behaviour does not strictly bite, but the gap is worth one line in the PR text or a read-only-runtime scenario.
- `crates/rocm-core/src/runtime.rs:387` — the new `;`-quoting on the join side reaches real child processes' search path, not only this crate's own round-trip; the convention is shared with the standard splitter but is not universally understood by every downstream reader of that variable. Effectively unreachable (a `;` inside a runtime install directory), noted only because the behaviour there changed from "emit raw" to "emit quoted".

No prompt-injection content, and no internal names, hostnames, registry paths or document links appear in the diff; the leak scan's only hits are the runtime registry directory, which is a repository concept.

siloteemu (commented):
🔴 Automated review · pr-review-watcher · 55c34530

## Summary
The PR relaxes then re-gates an e2e containment assertion on the framework interpreter, makes the Windows path-list split/join quote-aware, and adds unit coverage for the managed-interpreter resolver and the loader-path merge; our prior blocking count is resolved by the route we preferred. Verified: read all four commit messages whole from the raw objects and diffed against both the merge base and the base-branch tip (identical); ran one targeted unit selection covering the five managed-interpreter tests — all five pass; confirmed by reading source that the record in `a_recorded_interpreter_that_is_gone_does_not_resolve` now really carries `python_executable`, that the recorded path is never created and the derived one is removed, that `active_runtime_mode: managed` is genuinely emitted by the same block that emits `active_runtime_root` (so the restored e2e gate is live, not silently dead), and that the Windows branch of `runtime_path_list_join` does not call `std::env::join_paths`; the full suite, clippy and the e2e lanes were not run here. CI conclusions at this head: 17 success, 1 failure, 1 skipped, 0 pending — the previous round saw 18 success, 1 skipped, 0 failure. Blocking: 0 · Non-blocking: 5.

## Previously raised
RESOLVED — by the coverage route, not the rename route: the head commit passes `Some(<a path that does not exist>)` through `write_therock_runtime_with_interpreter` and then removes the derived interpreter, so a recorded candidate really is present and really is rejected; the name and the assertion message are now true of what the test constructs. Verified by reading the test and the resolver: dropping the `is_file()` filter outright now fails this test, and weakening it to `exists()` fails the new sibling `a_recorded_interpreter_that_is_a_directory_is_passed_over`. Every test added or changed in this PR was checked against a revert of the production behaviour it names; all fail on revert except `splitting_a_windows_path_list_trims_and_drops_empty_entries` (see non-blocking), which is accurately named for the pre-existing behaviour it pins.

**None.**

## Non-blocking
- `crates/rocm-core/src/runtime.rs:379-390`, `:337-341` and the comment in `splitting_a_windows_path_list_strips_the_quotes_it_split_on` all frame `std::env::join_paths` as the thing that rejects a `"`-bearing entry — but this PR's Windows branch of `runtime_path_list_join` builds the string by hand and never calls `join_paths`, so on Windows that backstop is not reached at all. The invariant it justifies (quotes are reserved in Windows paths) does hold; the stated mechanism does not. Reword to say the precondition is unenforced on this path.
- Following from the above: an entry that did contain a `"` (e.g. a config-supplied root pasted with its surrounding quotes) now goes out raw and flips quote parity for the *next* entry on re-split, corrupting a neighbour silently instead of failing loudly. Narrow and Windows-only, but a cheap `ensure!(!text.contains('"'))` threaded through the already-`Result` join would close it.
- `crates/rocm-core/src/examine.rs:2080-2084` — the lock's doc says it serialises "the tests that read or replace" the variable, but it is module-private while `crates/rocm-core/src/lib.rs:4052` reads the same variable unguarded in the same test binary. Benign today; narrow the wording to "the tests in this module" or share the lock.
- `crates/rocm-core/src/lib.rs:3794-3797` — the doc says the resolver is `None` "when the runtime records an interpreter that is no longer on disk", but it is `None` only when *neither* candidate is on disk; the new directory test is itself the counterexample. Pre-existing wording the new coverage now contradicts.
- No test catches dropping `normalize_runtime_path_for_host` from the recorded candidate: every test builds its recorded path through the same host-native helper, so the normalisation is an identity on the lane that runs them. A case feeding raw foreign-platform path text through the registry JSON would close it.
- `splitting_a_windows_path_list_trims_and_drops_empty_entries` still passes with the quote-awareness reverted — it pins pre-existing trim/drop behaviour only. Accurately named, so not a coverage-claim defect; adding a quoted entry with stray whitespace would make it carry its share.

siloteemu (approved):
✅ Approved on a maintainer's explicit instruction · pr-review-watcher · 55c34530

A maintainer reviewed this pull request and instructed that it be approved if it verified clean. It does, so this approval is filed on that instruction. This automation does not approve on its own initiative.

**What changes:** the e2e containment assertion on the framework interpreter, plus quote-aware path-list handling on Windows and new unit coverage for the managed-interpreter resolver.

**Verified at this exact head before approving:**

- A published review from this account sits at the current head commit and records no blocking findings; the head has not moved since.
- The pull request is open, not a draft, not conflicting, and mergeable.
- No other reviewer holds an outstanding change request.
- Check conclusions at this head: 22 success, 1 failure, 1 skipped, 2 still running (hardware lanes).

**About the one failing check.** It is the mock end-to-end lane, and it fails on two dashboard throughput scenarios that race the host under load. This is not caused by this change:

- The identical failure — the same two scenarios, the same message — reproduces on the base branch's own tip, on a commit this pull request does not touch.
- That same base commit also passed the same lane on a separate run, which is what a race looks like rather than a deterministic break.
- A change fixing exactly these two scenarios has since landed on the base branch, and the base branch has been green on this lane on every run since. This branch's checks were started before that fix landed, so they do not include it.
- Nothing in this diff touches the code those scenarios exercise.

A re-run of that lane on top of the current base branch is expected to come back green. Merging remains a human decision; nothing has been enqueued or merged here.

siloteemu (commented):
🔴 Automated review · pr-review-watcher · 129a1a33

_The approving review already on this pull request was filed on a maintainer's explicit per-PR instruction and is left standing; this automation files no approvals of its own, and this report is not one. The merge decision stays with a human reviewer._

## Summary
This PR stops the examine e2e scenario from requiring the Python interpreter to sit under the runtime root, and the only new commit since the last review is a merge of the base branch into the PR branch. No blocking findings. Verified: the merge is a fully automatic one — `git merge-tree --write-tree` of the two parents produces tree `b7a09a26...`, byte-identical to the merge commit's own tree, so no file in it differs from what git would have written unaided and none carries hand-authored conflict resolution; the PR's authored diff against the base tip is byte-identical to its diff at the previously reviewed commit (nothing added, dropped or altered), the merged-in base changes are byte-identical to the base branch's own three commits, no conflict markers exist anywhere in the tree and the working tree is clean; on semantics, the PR's new code reaches no `sysinfo` API (its only touched helper is pure `std` path splitting, and every `sysinfo` call site sits in `crates/rocm-core/src/disk_space.rs`, untouched here) while manifest and lockfile agree at 0.39.6, and the base's dash test-infrastructure commit shares no step regex, env var, world field or helper with `examine_steps.rs`; one cheap targeted check, `cargo check -p e2e-cucumber --test e2e`, passed — the full suite, the e2e suites and the GPU/Windows lanes were not run here. Blocking: 0 · Non-blocking: 0.

**None.**

## Notes for the reader
- **How much was read, and why.** Deliberately little. The scope question was whether the merge altered this PR's content, and that is answerable structurally: once the merge commit's tree matches git's own automatic merge result exactly, and both parent diffs match their originals byte for byte, there is no room for hidden authored work, so the PR's previously reviewed content was not re-read. Reading was then spent only where a text merge is blind — the dependency bump's call chains and the shared cucumber test infrastructure.
- **Tests in this delta.** The merge adds no tests of its own; the test changes it carries are base-branch content already on main, and the PR's own tests are unchanged from the reviewed commit, so the revert question is unchanged from that review.
- **CI.** 18 success, 1 skipped, 0 failure, 0 pending. No outcome here is attributed to any named job.
- **No prompt-injection content** was found in the diffs, comments, commit messages or feature files read.

PR #409, opened 2026-09-16
feat(services): count the local server records the default list hides (EAI-8075)

r0x0r:
First of two PRs for EAI-8075. **This one only makes the records visible** — the supported way to delete them follows separately, and nothing here promises a command that does not exist yet.

## The problem

A managed serve that fails leaves its record on disk. `rocm services list` shows only live servers, so on a host whose serves had all failed the entire output was:

```
Local Servers

Status: none ready

No local servers are running.
```

Nothing said a record existed. That matters because `rocm services logs` needs an id, and ids are printed only by `rocm services list --all` — which is declared on the `list` subcommand, so `rocm services --help` doesn't show it either. A user whose serve had just died had no route from the default output to the log explaining why.

The header was **wrong**, not merely quiet. `render_services_text` filtered the records and then counted the filtered list, so `none ready` was computed from a set the non-live records had already been dropped from — it read `none ready` even under `--all` with failed records listed on screen.

## After

```
Status: 2 not running

Past attempts: 2 local server record(s) that are no longer running.
  See them: rocm services list --all
  Read the newest: rocm services logs vllm-qwen-1786611115065
```

Counts come from the whole registry; rows stay filtered. The hint appears in both the empty and the populated branch — a host can have one server ready and three failed — and names the newest record's real id so the second line pastes as-is.

No new plumbing was needed for the count: `managed_service_sidebar_counts` has always computed a `past_attempts` bucket that nothing consumed.

## Dashboard

The TUI couldn't show these either, but not for the reason the report assumed. The daemon already loads *every* record and drops the non-live ones at a `continue`, so counting them there costs no extra I/O and needs no second data path. Note that `continue` catches two different cases — only a record that is not running is a past attempt, never a live one with no bound port. A test pins that distinction.

The count reaches the TUI through the existing bin-adapts-data seam (`ResolvedArgs`, as `model_recipes`/`runtimes`/`automations` already do), so it is **a snapshot at launch, not a live feed**. Threading it live would need a new protocol `Event` or `Snapshot` field across ~25 construction sites; that seemed disproportionate for a count, but say the word if you'd rather have it.

## `rocm storage report`

Now lists the services folder, which no reclaim path has ever touched and no report has ever named.

[...]

PR #411, opened 2026-09-16
feat(services): add a supported way to remove non-live local server records (EAI-8075)

r0x0r:
Second half of EAI-8075. #409 made the records *visible*; this makes them removable. Independent of #409 — branched off `main`, no shared files.

Until now the only way to clean up a managed serve's leftovers was deleting files by hand. `ServicesCommand` had `list`/`logs`/`stop`/`restart`; `rocm storage` never targeted the services directory; `prune_old_logs` only touches the CLI's own client logs; nothing aged the directory out.

## `rocm services remove <id>` and `rocm services prune`

- **Live records are never removed.** Liveness is re-read from the record `load_managed_services` returns, which has already refreshed it against the engine state and the real processes — a manifest still claiming `ready` for a dead PID has demoted by then. `remove` on a live record fails loudly and names `rocm services stop <id> --yes`; `prune` skips it and reports the count.
- **`--yes` required**, matching the `stop`/`restart` precedent (this repo `bail!`s rather than prompting).
- **`--dry-run` on `prune`**, matching `storage`'s `remove-old-installs`/`remove-downloads`. Not on single `remove`, where the id is already explicit.

## A record owns four files, not two

The ticket says "delete the `.json` and `.log` by hand". That is incomplete, and the gap is observable: this repo's own dev box has an orphaned `engines/vllm/state/<id>.json` whose manifest is long gone.

| file | location |
|---|---|
| `<id>.json` | `services/` |
| `<id>.log` | `services/` |
| `<id>.endpoint-key` (0600) | `services/` |
| `<id>.json` | `engines/<engine>/state/` — **outside** the services dir |

All four are removed. `launch.lock` is **shared** and is never a candidate. `prune` also sweeps orphans — engine state, logs and endpoint keys whose manifest is gone — so it can clean up after the manual workaround the ticket documents. An unparseable manifest is never removed and never makes its siblings look orphaned.

**Paths are rebuilt from `AppPaths` plus a validated id and engine — never read from the record's own `manifest_path`/`log_path`/`engine_state_path` fields.** Those come from a JSON file anything can write, and this code deletes what they name. `validate_service_id` (existing) covers the id; `validate_engine_component` is new, because `service_engine_state_path` joins the engine name too.

## Why this does not route through the sandbox dispatch

`stop`/`restart` go through `run_approved_service_action` → `run_internal_sandbox_tool`. `remove` deliberately does not.

That path is the **assistant-reachable** restricted tool API. `stop_server`/`restart_server` are on it because the agent has a legitimate reason to want them (`gpu-thermal-protect` queues a stop proposal) and because they do real process work the wrapper frames. Adding a `remove` variant would make irreversible deletion of user data reachable from a tool call via `rocm internal sandbox-run --allow-native-fallback`, with no human `--yes` equivalent in that path.

`remove` is filesystem cleanup of already-dead records — the shape of `rocm storage remove-downloads`, which does not route through the sandbox either. Both safety properties that path carries are kept directly: `validate_service_id` and `record_cli_audit_event` (`service`/`remove_record`, `service`/`prune_records`).

## Removing a record destroys its log, and that is the point of tension

`list --all` advertises `restart <id> --yes` for a failed record; removing it takes that away along with the log you might be reading. Two mitigations:

- `remove` prints the log path **before** deleting, with an explicit line saying logs and restart will no longer work.
- `prune` defaults to `--older-than-hours 24`, measured from when the record was last written. The liveness refresh rewrites it at death-observation time, so a server that just died is genuinely protected.

A run that keeps everything used to print the same summary as one that found nothing, so the count is now reported (`too recent, kept: N`) in both the preview and the result, alongside `rocm services logs <id>` and the flag that includes them. `--any-age` is the reachable spelling of `--older-than-hours 0` — you shouldn't have to know the age rule exists to override it. The two conflict rather than silently overriding each other.

## Tests

12 unit tests and 3 scenarios in a new `service_record_cleanup.feature` (`@requires-gpu` not needed — records are planted, so it runs on the mock lane every PR). A new feature file rather than appended scenarios because `feature_naming.rs` requires per-file indexes to be sequential, and the next `serve-`/`dash-`/`storage-` indexes are already claimed by other open branches.

Every behaviour is pinned by a mutation that was verified to make it fail:

| mutation | result |
|---|---|
| engine state dropped from the artifact set | unit + `service-cleanup-01` fail (`files removed: 3`) |
| live guard removed | unit + `service-cleanup-02` fail ("must fail, but it exited 0") |
| trust the manifest's status instead of the refreshed record | unit fails |
| age threshold ignored | unit fails (`records removed: 1`) |
| prune stops skipping live records | unit fails |
| orphan sweep disabled | `service-cleanup-03` fails |
| recent-kept counter dropped | unit fails |

Gate: `cargo test --workspace --all-targets`, `clippy --workspace --all-targets -D warnings`, `fmt --check`, `smoke_local.py` — all clean.

All destructive testing ran against throwaway trees with `ROCM_CLI_DATA_DIR`/`CONFIG_DIR`/`CACHE_DIR` overridden, or cucumber's per-scenario `TempDir`. The real `~/.rocm` was read only.


jussielo-amd (approved):
Reviewed the diff (main.rs implementation of `services remove`/`services prune`, new cucumber scenarios, docs) plus CI status. No blocking issues.

**Verified against the code:**
- Live-record guard re-checks the refreshed record from `load_managed_services` (not the raw on-disk manifest), so a stale `ready` that's actually dead is handled correctly; a genuinely live record is refused with a pointer to `stop --yes`.
- Deletion paths are always rebuilt from `AppPaths` + `validate_service_id`/`validate_engine_component`, never read from the record's own (attacker-writable) path fields.
- The four-file set (manifest, log, endpoint-key, engine state) and the orphan sweep in `prune` are correct; `launch.lock` is structurally excluded from ever being a candidate.
- `--yes`/`--dry-run`/`--any-age` flag interactions work as described, including the `conflicts_with` guard against combining `--any-age` with `--older-than-hours`.
- The deliberate bypass of `run_approved_service_action`/sandbox dispatch is consistent with `record_cli_audit_event` still being called directly, matching the `storage remove-downloads` precedent.
- All 12 unit tests + 3 cucumber scenarios pass and are non-tautological (each pinned to a specific mutation per the PR description); full workspace tests, clippy `-D warnings`, and fmt are clean.
- The failing `E2E tests` check (`dash-gen-tps-held-after-scrape-failure`, `dash-gen-tps-expiry-boundary`) is a pre-existing flake — the same two tests fail identically on `main` at a recent commit, unrelated to this PR's files.

**Non-blocking, worth a quick follow-up (either here or a fast-follow):**
- Doc-sync gap: `--any-age` (added in the second commit) isn't mentioned in README.md, docs/manual-testing.md, or docs/testing.md, even though the first commit correctly kept those in sync. Given the flag's own doc comment frames it as "the discoverable way to reach `--older-than-hours 0`," it should actually be discoverable in the docs.
- `--any-age`'s CLI-level wiring (the `hours=0` translation and the `conflicts_with` rejection) is untested — only the internal function is exercised, with a hardcoded `hours=0`.
- The dry-run preview's "kept" wording differs from the real-run wording and has no test coverage for `skipped_recent > 0`.
- No test plants an unparseable manifest to confirm its siblings aren't swept as orphans, despite that being an explicit design guarantee.

None of these block merge — approving.

siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · 1f6711d8

## Summary
Adds `rocm services remove` and `rocm services prune` — a destructive, user-facing way to delete non-live local server records and their four on-disk artifacts — plus a second commit adding `--any-age` and a "what was kept" summary. Outcome: **Needs work** — the selection rule `prune` states is not the one it implements, the new flag's only test does not exercise it, and the flag reaches users with no docs and no scenario. Verified: I read every changed file and the surrounding service-record code; I compiled a small standalone reproduction of the two new command-line arguments and confirmed `--any-age` parses on its own and is rejected alongside an explicit `--older-than-hours`; the project's own build, unit tests and end-to-end suite were **not** run here, so every other finding below is derived from reading the source rather than from executing it. Checks at this head are 24 success, 1 failure, 1 pending — not fully green, and one check has not concluded. Blocking: 3 · Non-blocking: 5.

- **`apps/rocm/src/main.rs:7367` (with `:7207`, `:7216`, `:7454`) — `prune` tells the user a selection rule it does not implement, and in the most common real case the default invocation silently removes nothing.**
  The summary line reads `"{} record(s) stopped less than {} ago and are kept; add --any-age to include them."`, and the flag's help at `:885` explains the age as a proxy for "a server that has only just stopped". The implemented rule is neither: `prunable_by_age` stats the *manifest file's modification time*. `prune_managed_service_records` captures `SystemTime::now()` at `:7454` and passes it into `build_service_prune_plan`, which first calls `load_managed_services` — and that function rewrites the manifest whenever the liveness refresh changes the status, which is exactly what happens the first time anything observes that a `ready`/`running`/`starting` server has died (crash, kill, reboot). The rewrite lands *after* the captured `now`, so `path_age` returns `None` and the fail-closed branch in `prunable_by_age` keeps the record. Net user-visible behaviour: on a host whose servers died weeks ago and have not been listed since, `rocm services prune --yes` deletes nothing and reports those records as "stopped less than 24 hours ago". For a removal command, a stated selection rule that differs from the implemented one is the defect the command most needs not to have.
  Fix: decide what the gate actually measures and make the prose match it. Either stat the manifests *before* `load_managed_services` is allowed to refresh and rewrite them, or gate on a persisted stop timestamp on the record rather than on file mtime; and reword `:7367` to the wording the per-record line at `:7391` already uses ("changed less than … ago"), along with the help text at `:885`.

- **`apps/rocm/src/main.rs:26261-26281` — `services_prune_any_age_removes_a_just_stopped_record` does not test `--any-age`.**
  The test calls `prune_managed_service_records(&paths, 0, false, true)` — byte-identical to the calls at `:26083` and `:26129` — so it exercises the helper with an hours value of 0 and never touches the flag. The `--any-age` → 0 mapping lives at `:6783` in the command dispatch; mutate that line to `let older_than_hours = older_than_hours;` and this test, and every other test in the change, stays green. Its doc comment ("`--any-age` is the reachable form of `--older-than-hours 0` … so it has to actually take the record the default kept") actively tells the next reader the flag is covered when nothing covers it — and the only string that advertises the flag to users, at `:7497`, is itself pinned by a test at `:26251`, so the pair reads as coverage while leaving the flag's wiring unasserted. I confirmed separately that the flag does parse correctly, so this is a test defect rather than a broken flag — but a test that does not test the thing it names is not fixable by being lucky.
  Fix: drive the test through the argument parser (construct the `Prune` variant / `try_parse_from` the `--any-age` form) so the mapping at `:6783` is on the tested path, or rename the test to what it actually asserts and add a separate one that is.

- **`README.md:475-488` and `docs/manual-testing.md:199-225` — the second commit ships a new user-facing flag and new output with no docs and no scenario.**
  `--any-age` appears nowhere outside `apps/rocm/src/main.rs`: the README synopsis still reads `rocm services prune [--older-than-hours <n>] [--dry-run] [--yes]`, the README prose still says the default is overridden by "`--older-than-hours`" alone, and the manual-testing steps use `--older-than-hours 0` throughout — while the command's own output tells the user to run `rocm services prune --any-age --yes`. The contributor rules in the base tip make this explicit in section 5: when a command's flags, defaults, arguments or observable behaviour change, README, the help text, `docs/testing.md` and `docs/manual-testing.md` are updated *in the same change*, and the same behaviour claim must be checked across every surface that repeats it. Section 3 is also unmet for this commit: the new flag and the new "kept for being recent" summary are user-observable behaviour, the new feature file covers only the `--older-than-hours 0` spelling, and the PR does not name an existing `@id:` that covers them or justify the absence.
  Fix: add `--any-age` to the README synopsis and prose and to the manual-testing steps and expected results; extend `service_record_cleanup.feature` with a scenario that prunes with the default age, asserts the record survives and that the output names the flag, then re-runs with `--any-age` and asserts it is taken (this also closes the second blocker's coverage gap at the end-to-end level).

## Non-blocking
- `apps/rocm/src/main.rs:7466` / `:6780-6793` — a mid-loop `fs::remove_file` failure propagates out, so the rendered plan is never printed and the `prune_records` audit event is never recorded; the user learns one path failed but nothing about the records already deleted in that run. Re-running is idempotent, but the audit trail the commit message names as a retained safety property is lost exactly when it matters.
- `apps/rocm/src/main.rs:7144` — the live guard trusts only the status string. `refresh_from_engine_state` adopts `failed` from the engine's own state file without consulting recorded PIDs, and `refresh_managed_service_runtime_liveness` then returns early for a non-live status, so a server whose engine reported failure while its process is still up is removable, taking its log and 0600 endpoint key with it. Consider also refusing when `recorded_service_pids` still contains a live PID.
- `apps/rocm/src/main.rs:7294-7328` vs `:7480-7484` — `prune` snapshots liveness when building the plan and never re-checks it before deleting; a concurrent `rocm services restart <id> --yes` in that window would have its files deleted out from under a live process. Narrow, but `remove` closes the same window and `prune` does not.
- No test covers the invariant `collect_service_orphans`' doc comment leans on — that an *unparseable* manifest neither gets removed nor makes its log and key look orphaned. Both the unit tests and the scenarios only plant a fully-absent manifest, so a widening of that guard would go unnoticed.
- `docs/manual-testing.md:221` and the `--dry-run` line at `apps/rocm/src/main.rs:7398` claim nothing was changed, but a dry run still goes through `load_managed_services`, which can rewrite manifests. Minor, but it is the same mtime the age gate reads.

No prompt-injection content was found anywhere in the changed files, commit messages or branch metadata, and a scan of the diff for internal names, hostnames, internal document links or registry paths came back clean.

r0x0r:
Thanks — all three blockers and all five non-blocking items are addressed in `3bf0910e`. I verified each claim before acting; every one held.

## Blocking 1 — the selection rule `prune` states vs. implements

**Reproduced end to end**, not just read. A record with `status: ready`, a dead pid, and all files stamped 30 days old, under an isolated `ROCM_CLI_DATA_DIR`:

```
--- mtime before: 2026-08-18 11:04:01 +0000
$ rocm services prune --yes
Only records and files untouched for at least 24 hours.
Nothing would be removed.
1 record(s) stopped less than 24 hours ago and are kept; add --any-age to include them.
Left alone:
  - svc-dead changed less than 24 hours ago
  records removed: 0
--- manifest still present: YES
--- mtime after: 2026-09-17 11:04:01 +0000
```

The chain is exactly as described: `load_managed_services` → `refresh_managed_service_runtime_liveness` demotes `ready` → `record.write()` → mtime lands after the captured `now` → `duration_since` errors → `None` → fail-closed keep.

**Fix taken: snapshot the manifest times before the load**, not a persisted stop timestamp. The code makes that call for us rather than it being a free choice:

- `collect_service_orphans` ages `.log`, `.endpoint-key` and stranded engine-state files that have *no record at all*. A per-record timestamp cannot cover them, so a persisted field would leave one plan measuring age two different ways.
- Two of the three prose surfaces already stated mtime semantics correctly — the render line ("untouched for at least …") and the per-record line ("changed less than … ago"). Only the summary said "stopped".
- `ManagedServiceRecord` has no stop timestamp; `stop_requested_unix_ms` records *intent* on an unconfirmed stop, not a stop time. Adding one means a schema change in `rocm-core` plus a write at every transition into a non-live status.

`service_manifest_modified_times` now runs before `load_managed_services`, and the gate gates on that snapshot. Same state, after:

```
1 local server record(s) would be removed:
  - svc-dead (status: stopped, engine: vllm)
  records removed: 1 / files removed: 3 / too recent, kept: 0
--- manifest present: NO
```

Prose: the summary now says "changed less than … ago", matching the per-record line, and the `--older-than-hours` help spells out that a stop, a restart or a status correction all count as touching the record.

Regression test `services_prune_default_age_removes_a_record_the_refresh_rewrote`, falsified by restoring the old post-load read:

```
thread '...services_prune_default_age_removes_a_record_the_refresh_rewrote' panicked at apps/rocm/src/main.rs:26403:9:
a record last written a month ago must be pruned even though the liveness refresh rewrote its manifest during this run:
  ...
  too recent, kept: 1
test result: FAILED. 0 passed; 1 failed
```

## Blocking 2 — the test that did not test the flag

Correct, and your mutation was the right probe. The mapping is now `service_prune_min_age_hours`, a function rather than a line inside the dispatch `match`, and the test parses a real argv through it. Falsified with exactly the mutation you named (`if any_age { 0 }` → `older_than_hours`):

```
running 11 tests
test tests::services_prune_any_age_removes_a_just_stopped_record ... FAILED
...
panicked at apps/rocm/src/main.rs:26500:21:
assertion `left == right` failed: --any-age must collapse to the zero-age rule
  left: 24
 right: 0
test result: FAILED. 10 passed; 1 failed
```

Two more: `services_prune_without_any_age_keeps_the_default_threshold` (same argv path, no flag, record survives — so the assertion above is about the flag and not about prune deleting things in general) and `services_prune_rejects_any_age_with_an_explicit_age`.

## Blocking 3 — docs and scenario

The rule is present on this branch. `AGENTS.md` §5: *"when a command's flags, defaults, arguments, or observable behavior change, update README.md, its --help/doc comment, docs/testing.md, and docs/manual-testing.md in the same change — do not leave user-facing docs for a follow-up"*, plus *"the same behavior claim … often repeats across README.md, --help doc comments, printed CLI output, and docs/*.md; each drifts independently, so grep for the claim's wording across all of them"*. §3: *"User-observable behavior needs a scenario, not only a unit test … a unit test asserting the internal helper does NOT discharge this; it proves the function, not the behavior."*

- README synopsis is now `[--older-than-hours <n> | --any-age]`, and the prose states the mtime rule and that the two conflict.
- `docs/manual-testing.md` drives `--any-age`, the conflict rejection, and the long-dead-record case, instead of `--older-than-hours 0` throughout.
- Two scenarios appended to `service_record_cleanup.feature`, taking the file's next indexes: `service-cleanup-04` (default keeps it and names the flag; `--any-age` then takes it) and `service-cleanup-05` (a record whose server died long ago is pruned by a plain `prune --yes`). Both on the mock lane, no GPU.

`docs/testing.md` needed no edit — it documents lanes and conventions, not per-flag behaviour.

## Non-blocking

1. **Mid-loop failure loses the plan and the audit event** — a file that cannot be deleted is now collected rather than propagated. The plan prints, the `prune_records` audit event is recorded, the failures are listed, and the dispatch then fails the command so the exit code is still non-zero.
2. **Live guard trusted only the status string** — new `managed_service_record_is_in_use` refuses when `recorded_service_pids` holds a running pid, whatever the status says. Used by both `remove` and `prune`. Covered by `services_removal_refuses_a_failed_record_whose_process_is_still_alive`.
3. **Prune's liveness snapshot window** — each entry is re-read immediately before its files are touched; a record that came back up in the window is skipped and reported.
4. **Unparseable manifest invariant** — `services_prune_keeps_a_corrupt_manifest_and_its_siblings` plants valid-JSON-but-not-a-record and asserts all four files survive and nothing is counted removed.
5. **Dry run "nothing was changed"** — now "Nothing was removed.", since building the plan can persist a status correction. `docs/manual-testing.md` says the same.

## Validation

On Linux: `cargo fmt --all`; `cargo test -p rocm --bin rocm` 711 passed; `cargo test --workspace --all-targets` green; `cargo clippy --workspace --all-targets -- -D warnings` and `cargo clippy -p e2e-cucumber --test e2e -- -D warnings` clean.

E2E, both `--any-age` scenarios and all three originals pass. The suite's 7 failures are identical on this branch and on unmodified `1f6711d8` — `chat-06`, `dash-09`, `runtime-02`, `serve-01`, `serve-02`, `serve-07`, `serve-08` — all host/GPU-gated and none service-related.

siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · 3bf0910e

## Summary

The PR adds `rocm services remove` and `rocm services prune` (destructive local-record cleanup) plus `--any-age`, and the new head commit repairs the age gate that made `prune` keep long-dead records; all three previously raised counts are now resolved, but the head commit's other new branches ship untested. Verified: in a separate scratch copy I ran the targeted `services_prune` unit group (11 tests, all green) and six single-branch mutations — neutering the `--any-age` mapping reddens exactly `services_prune_any_age_removes_a_just_stopped_record`, and restoring the pre-fix mtime-snapshot ordering reddens exactly `services_prune_default_age_removes_a_record_the_refresh_rewrote`, while deleting the pre-delete liveness re-check, the removal-failure collection, and both unreadable-mtime branches leaves all 11 green; I also confirmed from source that `load_managed_services` persists a status correction on refresh (`apps/rocm/src/main.rs:18060-18066`), that `rocm serve` writes the 0600 endpoint key before the record manifest (`:5598` vs `:6062`), and that README/`docs/manual-testing.md` now carry `--any-age` and the modification-time rule; the full test suite, clippy and the e2e lanes were not run here. CI conclusions I worked from: 12 success, 1 neutral, 4 pending, 0 failure — four checks have not concluded, so the CI picture is incomplete. Blocking: 1 · Non-blocking: 5.

## Previously raised

- **COUNT 1 — RESOLVED.** The block said the summary line `"{} record(s) stopped less than {} ago and are kept; add --any-age to include them."` described a rule the code did not implement, and that `build_service_prune_plan`'s `load_managed_services` call rewrote the manifest after the captured `now`, so `path_age` returned `None` and the fail-closed branch kept every long-dead record. Both halves are fixed in the tree, not just in the commit message. The manifest times are now snapshotted *before* the load (`apps/rocm/src/main.rs:7401-7405` taking `service_manifest_modified_times`, defined at `:7285-7311`) and the gate reads that snapshot through `prunable_by_modified` (`:7268`, `:7425-7429`) instead of stat-ing the live file; `prunable_by_age` survives only for orphan files, and its doc comment now warns it is "only sound for files nothing in this run rewrites" (`:7275-7282`). The summary line now reads `"{} record(s) changed less than {} ago and are kept; add --any-age to include them."` (`:7476`) and the flag help spells out that a stop, restart or status correction all count as touching the record (`:885-889`), which is the rule the code implements. `services_prune_default_age_removes_a_record_the_refresh_rewrote` (`:26383`) plants a `ready` record with a dead pid and backdates the manifest a month, which is exactly the reported condition; restoring the old ordering in a scratch copy turns that test — and only that test — red.
- **COUNT 2 — RESOLVED.** The block said `services_prune_any_age_removes_a_just_stopped_record` called `prune_managed_service_records(&paths, 0, false, true)`, byte-identical to two neighbours, so the `--any-age` mapping was never executed. The test now parses a real argument vector through `parse_services_prune_args` (`apps/rocm/src/main.rs:26462-26482`), which calls the production mapping `service_prune_min_age_hours` (`:7319`) — the same function the dispatch uses at `:6784` — and asserts `hours == 0` before pruning (`:26498-26503`); a sibling test drives the same parser without the flag and asserts the default threshold survives (`:26533`). The named mutation confirms it: with `service_prune_min_age_hours` neutered to return `older_than_hours`, the targeted group goes 10 green / 1 red, and the one red is `services_prune_any_age_removes_a_just_stopped_record`.
- **COUNT 3 — RESOLVED.** The README synopsis now reads `rocm services prune [--older-than-hours <n> | --any-age] [--dry-run] [--yes]` (`README.md:479`), the prose states the modification-time rule and names `--any-age` and the mutual exclusion (`README.md:487-493`), and `docs/manual-testing.md:205-237` drives `--any-age` in preview and real form, pins the `"too recent, kept: 1"` counter and the `rocm services prune --any-age --yes` pointer the command prints, covers the rejected `--any-age --older-than-hours` pair, and adds an expected result for the long-dead record. Two Gherkin scenarios were added for the same behaviour (`tests/e2e-cucumber/features/service_record_cleanup.feature:51-70`).

`apps/rocm/src/main.rs:7583-7594` and `:7595-7604` — the head commit's two other behaviour changes have no test at any level, and each one can be deleted outright with every targeted test still passing. Deleting the pre-delete liveness re-check (the block that re-loads the record and prints `"  {} started again while this ran and was left alone."`) leaves 11/11 green; replacing the failure collection with a silent drop and an unconditional `outcome.removed_records += 1` also leaves 11/11 green, so nothing covers the failure list, the `"{} file(s) could not be removed:"` output, or the non-zero exit the dispatch raises at `:6801-6808`. Both are user-observable changes on a command that deletes files, and the commit message presents both as fixes; the contributor rules require a regression test in the same change and a Gherkin scenario for user-observable behaviour, and the PR's own five scenarios cover neither. The same gap covers the fail-closed property `prunable_by_modified`'s doc comment advertises: removing the `min_age.is_zero()` early return, and flipping `is_some_and` to `is_none_or` so an unreadable time fails *open*, both leave 11/11 green (`:7268-7273`). Concrete fix: a unit test that makes a record's artifact undeletable (e.g. hold/lock one artifact, or make the engine state path a non-empty directory) and asserts the run still prints the plan, still records the audit event, lists the failure and exits non-zero; a unit test that re-creates the record between plan and delete (or injects a plan entry for a record that becomes live) and asserts it is skipped and counted in `skipped_live`; and a test that sets a manifest mtime in the future so `age_from_modified` yields `None`, asserting the record is kept under the default threshold and taken under `--any-age`.

## Non-blocking

- `docs/testing.md:470-471` — "Neither will touch a record the liveness refresh still reads as running, so this cannot race a live server" is a false absolute: the head commit itself added a re-check because a concurrent `restart` could lose its log and key, and that re-check narrows the window rather than closing it.
- `apps/rocm/src/main.rs:7345` + `:5598` vs `:6062` — `prune`'s orphan sweep takes no managed-launch lock while `rocm serve` writes the 0600 `<id>.endpoint-key` *before* the `<id>.json` manifest, so a concurrent `prune --any-age --yes` can delete a mid-launch live server's endpoint key; taking `managed_launch_lock_path` around plan-building (or writing the manifest first) closes it.
- `apps/rocm/src/main.rs:898` — the `--dry-run` help still promises "Show what would happen without changing files", which is the one place this claim is false: building the plan loads every record and persists status corrections, which is precisely why the printed line was changed to "Nothing was removed" and why `docs/manual-testing.md:223-225` documents the rewrite.
- `apps/rocm/src/main.rs:7595-7604` — a record whose manifest deletes but whose sibling file fails is effectively gone yet never counted, so both the report and the audit event at `:6794` under-count; count the record once its manifest is confirmed removed.
- `apps/rocm/src/main.rs:7189` — with the new pid check a dead-status record with a live pid is refused as "is failed and cannot be removed while it is running", which reads as self-contradictory; name the still-running pid instead, and consider correlating the recorded start ticks so a reused pid cannot make a stale record permanently unremovable.

siloteemu (commented):
🔴 Automated review · pr-review-watcher · 8f32e8dd

## Summary

The PR adds `rocm services prune` (a supported bulk removal of non-live local server records, with `--any-age`, an age-based keep rule, and orphan sweeping); the head commit adds three unit tests and one Gherkin scenario that answer our previous change request. Outcome: **No blocking findings.** Verified: in a scratch copy, `cargo test -p rocm --bin rocm services_prune` is 14/14 green at head, and five independent per-branch mutations were run — four redden exactly one new test each and one (the dispatch's non-zero exit) reddens no unit test but is asserted by the new, ungated e2e scenario; separately confirmed the production hunk in this commit is a byte-identical code motion, that the delta's leak scan is clean, and that all three previously-filed non-blocking items are still present verbatim in the tree. Blocking: 0 · Non-blocking: 5.

## Previously raised

**Our block: RESOLVED.** Each branch we named now has a test that fails when that branch alone is broken. Per-branch mutation results (scratch copy, one mutation at a time, tree restored between each):

| # | Mutation | Result |
|---|---|---|
| 1 | Delete the pre-delete liveness re-check block entirely | 🔴 `services_prune_leaves_a_record_that_restarted_after_the_plan_was_built` fails ("a record that came back to life must not be deleted") |
| 2 | Replace failure collection with a silent drop + unconditional `outcome.removed_records += 1` | 🔴 `services_prune_reports_a_file_it_could_not_remove` fails (`removed_records` left 1, right 0) |
| 3 | Remove the `min_age.is_zero()` early return in `prunable_by_modified` | 🔴 `services_prune_keeps_a_future_dated_record_until_any_age` fails |
| 4 | Flip `is_some_and` → `is_none_or` there (unreadable time fails **open**) | 🔴 `services_prune_keeps_a_future_dated_record_until_any_age` fails |
| 5 | Neuter the `bail!` the dispatch raises on removal failure (`main.rs:6801-6808`) | ⚪️ **no unit test reddens** — see below |
| 6 (extra) | Keep collection + exit, delete only the `"{} file(s) could not be removed:"` output block | 🔴 `services_prune_reports_a_file_it_could_not_remove` fails |

On mutation 5: the non-zero exit lives in the command dispatch, outside the function the unit tests drive, so no unit test can reach it — by design. It is asserted instead by the new scenario `service-cleanup-06` (`tests/e2e-cucumber/tests/e2e/service_cleanup_steps.rs:378`, `assert!(rc != 0, ...)`). I confirmed that scenario carries no `@requires-gpu` / `@nightly` / `@lifecycle` / `@merge-queue` tag and has no `expectations.toml` entry, so the runner's `filter_run` gate does not exclude it and it must pass on every PR; I also confirmed the e2e crate compiles. I did **not** execute the e2e suite, so that one assertion is verified by reading, not by running. This is the level the repo's own contributor rules ask for (exit codes are user-observable and must be covered by a scenario), so it is not a gap — but it is the one branch in our block whose coverage I could not demonstrate by mutation.

None of the three new tests passes vacuously, and none of the three new `Then` steps can: `combined_output` always returns a real (possibly empty) string from the recorded subprocess, so every `contains` assertion fails rather than skips on empty output; `record_files` returns four hardcoded paths, never an empty vec; and the audit assertion is reachable because the scenario's partial failure still yields `removed_files == 3`, which satisfies the dispatch's `removed_records > 0 || removed_files > 0` audit gate.

**Per-test revert check (standing focus a).** `services_prune_leaves_a_record_that_restarted_after_the_plan_was_built` — fails on mutation 1, would not pass with the re-check reverted. `services_prune_reports_a_file_it_could_not_remove` — fails on mutations 2 and 6. `services_prune_keeps_a_future_dated_record_until_any_age` — fails on mutations 3 and 4, i.e. it independently covers both halves of the fail-closed property. Scenario `service-cleanup-06` — asserts the exit code, the failure list, the surviving plan output and the audit line; would fail on mutation 5 (by reading).

The three non-blocking items:

- `docs/testing.md:470-471` — **still stands.** The commit touched no docs (`git diff 3bf0910e..HEAD -- docs/` is empty) and the sentence is present verbatim. The overclaim is real: the re-check reads the record once, then `try_remove_service_record_artifacts` unlinks four files with no re-check between them and no lock held, so a restart landing after the read still loses its log and key. The window is narrower, not closed.
- `apps/rocm/src/main.rs:898` — **still stands.** The `--dry-run` doc comment is unchanged. Worth noting for the fix: that exact string is independently duplicated at six clap sites (`:318, :638, :766, :776, :806, :898`), so correcting prune's copy is isolated. `docs/manual-testing.md` and the command's own printed line already describe the behaviour correctly; the `--help` text is the only place the false absolute survives.
- `apps/rocm/src/main.rs:7595-7604` (now inside `apply_service_prune_plan`) — **still stands**, and the head commit hardens it. Semantics are unchanged by the extraction: on a partial failure `removed_files` counts the successes but `removed_records` stays 0 even though the manifest — the file record enumeration scans for — is gone, so the report and the audit line both say "removed 0 local server record(s)" for a record that has in fact disappeared. The new test now asserts `removed_records == 0` with a comment presenting it as intended, which turns an open question into a pinned contract.

**None.**

## Non-blocking

- `apps/rocm/src/main.rs:26718-26723` — the new test pins `removed_records == 0` for a record whose manifest was deleted, cementing the still-open under-count above; if that item is fixed, this assertion has to be re-litigated. Cheapest fix for both: count a record removed once its manifest specifically is gone, rather than requiring all four artifacts to succeed.
- `docs/testing.md:470-471` — "so this cannot race a live server" is still an absolute the code itself contradicts; softening it to "narrows the window" costs one line and prevents the next reader concluding the re-check is redundant.
- `apps/rocm/src/main.rs:898` — prune's `--dry-run` help still promises "without changing files" while the plan build persists status corrections; the printed output and `docs/manual-testing.md` already say it correctly.
- `apps/rocm/src/main.rs:26708-26712` and `:26868-26892` — two of the three new tests assert their premise (and, in one case, `assert_eq!(hours, 0, ...)` inside a closure) *before* `fs::remove_dir_all(&root)`, so a premise failure leaks the temp dir. One pre-existing test in the module has the same shape, so this repeats rather than introduces the weakness; moving the asserts after cleanup matches the module's majority style.
- `apps/rocm/src/main.rs:6801-6808` — the non-zero exit is reachable only through the dispatch, so it is untestable at unit level and rests entirely on one e2e scenario; a one-line comment at the `bail!` naming `service-cleanup-06` as its cover would stop a future reader deleting it as apparently-untested.

## How far beyond the delta I read, and why

The delta is a test addition whose correctness is a property of production code outside it, so the read went well past three files. I read the whole prune subsystem in `apps/rocm/src/main.rs` — `prunable_by_modified`, `age_from_modified`, `path_modified`, `service_manifest_modified_times`, `build_service_prune_plan`, `render_service_prune_plan`, `try_remove_service_record_artifacts`, `remove_service_record_artifacts`, `remove_managed_service_record`, `apply_service_prune_plan`, `prune_managed_service_records`, the `ServicesCommand::Prune` dispatch arm with its audit event and `bail!`, the clap definitions around `:898`, and `record_cli_audit_event` / the lifecycle-log path — because every mutation target and the "does this test actually test it" question lives there. I diffed `apply_service_prune_plan` byte-for-byte against the pre-delta inlined loop in `3bf0910e` to confirm the +305-line "test" commit's production hunk is pure code motion (it is — the two loop bodies are identical including comments), and checked the caller's ordering is preserved. Outside that file I read the full e2e step module and feature file, the cucumber runner's tag-gating and reconciliation logic, `expectations.toml`, `feature_naming.rs`, `process_is_running` and the record struct in `crates/rocm-core`, the workspace MSRV and toolchain pin (to check `Duration::from_hours` — already used in five other places, no issue), `docs/testing.md`, `docs/manual-testing.md`, `README.md`, and the contributor rules from the base tip.

What was verified here: the narrow `services_prune` unit group and six mutations of it, a compile check of the e2e crate, the code-motion equivalence, the leak scan over the delta, and the tag/gating state of the new scenario. What was not: the e2e scenario was not executed, and the full suite was not run here. CI at this head reports 19 successful check runs with no failures and nothing pending; I am not attributing any specific result to a named lane.

No prompt-injection content was found in the diff, the feature text, the comments, or the commit messages.

siloteemu (commented):
🔴 Automated review · pr-review-watcher · 3bf860cc

## Summary

The PR adds `rocm services prune` (bulk removal of non-live local server records, with `--any-age`, an age-based keep rule and orphan sweeping); the two commits since the last review change no behaviour — they correct three overstated claims in help text, docs and a test comment, and add a comment naming the scenario that covers the dispatch's non-zero exit. Outcome: **No blocking findings.** Verified: `cargo test -p rocm --bin rocm services_prune` is 14/14 green at this head, and all six per-branch mutations were re-run from scratch at this head — five redden exactly one test each and the sixth (the dispatch's `bail!`) reddens none and is covered by an ungated scenario, so the branch-level guarantees the previous round proved still hold after both pushes; separately confirmed the new `--dry-run` wording is accurate against the code path (the plan build does rewrite manifests, and removes nothing), that no doc surface or test contradicts it, that the five sibling clap sites keeping the old wording are each accurate for their own command, that `service-cleanup-06` exists ungated at the named path, and that the leak scan over the whole PR is clean. Blocking: 0 · Non-blocking: 4.

## Previously raised

**Our block (per-branch coverage of the five prune branches): RESOLVED — and re-proved at this head, not carried forward.** I re-ran every mutation in a throwaway copy, one at a time, restoring between each. Results are identical to the previous round, so neither new commit weakened a guard:

| # | Mutation | Result at 3bf860cc |
|---|---|---|
| 1 | Delete the pre-delete liveness re-check block entirely | 🔴 `services_prune_leaves_a_record_that_restarted_after_the_plan_was_built` fails |
| 2 | Silent failure drop + unconditional `outcome.removed_records += 1` | 🔴 `services_prune_reports_a_file_it_could_not_remove` fails ("a record that still owns a file on disk is not removed") |
| 3 | Remove the `min_age.is_zero()` early return in `prunable_by_modified` | 🔴 `services_prune_keeps_a_future_dated_record_until_any_age` fails |
| 4 | Flip `is_some_and` → `is_none_or` there (unreadable time fails **open**) | 🔴 `services_prune_keeps_a_future_dated_record_until_any_age` fails |
| 5 | Neuter the `bail!` the dispatch raises on removal failure | ⚪️ no unit test reddens — by design, see below |
| 6 | Keep collection + exit, delete only the `"{} file(s) could not be removed:"` output block | 🔴 `services_prune_reports_a_file_it_could_not_remove` fails |

Mutation 5 is the one branch no unit test can reach: the exit lives in the command dispatch, outside the function the tests drive. It is covered by scenario `service-cleanup-06`, whose `Then` step asserts a non-zero exit. I re-confirmed at this head that the scenario carries no `@requires-gpu` / `@nightly` / `@lifecycle` / `@merge-queue` tag, that `expectations.toml` (which exists) has no entry for it, and that the feature file's own header states every scenario in it runs on the mock lane on every PR. That assertion is verified by reading, not by running — the e2e suite was not executed here.

The five non-blocking items:

- **`docs/testing.md` "so this cannot race a live server": RESOLVED.** Replaced with wording that names the re-check and then says plainly that the check and the delete are not atomic, so the race is narrowed rather than eliminated. I checked this against the code rather than taking the sentence at face value: the re-check at the top of `apply_service_prune_plan` re-reads one record, then `try_remove_service_record_artifacts` unlinks four files with no further check and no lock, so "narrows" is the correct claim and "cannot race" was not.
- **`apps/rocm/src/main.rs:898` `--dry-run` help: RESOLVED.** Now "Show what would be removed, without removing anything." That is the promise the code keeps: `prune_managed_service_records` builds the plan before the `if dry_run` return, and the plan build calls `load_managed_services`, which writes back any record whose status the liveness refresh corrects — so a dry run *does* change files, and the new wording no longer claims otherwise. I also checked every other surface that documents this flag: `docs/manual-testing.md` already spells out the manifest-rewrite caveat, the command's own printed line says "Nothing was removed" (with a comment explaining the word choice), the README synopsis makes no behavioural claim, and no test or scenario pins either the old or the new string. The five sibling clap sites still carrying the old "without changing files" wording were each read through to their handler; every one of them returns before writing, so correcting only prune's copy was the right scope.
- **`apps/rocm/src/main.rs:6801-6808` untestable `bail!` with no pointer: RESOLVED.** A three-line comment now states the exit is only reachable through the dispatch and names `service-cleanup-06` as its cover. Both halves of that comment are true: mutation 5 above is the empirical proof that no unit test reaches it, and the scenario exists at the named path with the matching id. The shortened path form matches how `docs/testing.md` already refers to the same file.
- **`apps/rocm/src/main.rs` `removed_records == 0` pin for a record whose manifest was deleted: STANDING.** The assertion is unchanged (now at `:26736`), with the same comment presenting the under-count as intended. The underlying behaviour is unchanged too: on a partial failure `removed_files` counts the successes but `removed_records` stays 0 even though the manifest the enumeration scans for is gone, so the report and the audit line both say "removed 0 local server record(s)" for a record that has in fact disappeared.
- **Premise assertions before `fs::remove_dir_all`: STANDING.** Both cited tests still assert their premise before cleanup (`:26711` and `:26871`, plus `assert_eq!(hours, 0, …)` inside the closure at `:26894`), so a premise failure still leaks the temp dir.

Two of the three claims the author corrected were ours; the third — the test doc comment saying the re-check "exists to close" the window — the author found without being asked, which is the right instinct and the reason the remaining copy below is worth naming.

**None.**

## Non-blocking

- `apps/rocm/src/main.rs:7573` — the fourth copy of the claim this PR spent two commits removing survives, in the comment the other three describe: "`remove` closes the same window by loading the record it deletes." It does not. `remove_managed_service_record` loads the record, checks liveness, then unlinks the artifacts ~30 lines later with nothing held in between — the same non-atomic check-then-delete, just a shorter window, which is exactly the reasoning commit 3bf860cc gives for rewording the sibling claims. Left as is, a reader of `apply_service_prune_plan` concludes `remove` is race-free where `prune` is not, and the next reader to reach for "the prune re-check is redundant, `remove` already solved this" has a comment agreeing with them. Same one-word fix as the other three: `narrows the same window`.
- `apps/rocm/src/main.rs:26325` — `services_prune_dry_run_reports_the_plan_and_changes_nothing` overstates in its name what its body asserts (every assertion is about deletion: `must not delete the record`, `removed_files == 0`), and it is the same "changes nothing" claim the help text at `:898` just stopped making. Renaming to `…_and_removes_nothing` keeps the name honest with both.
- `apps/rocm/src/main.rs:26736` — the pinned `removed_records == 0` for a record whose manifest was deleted still cements an under-count; if that is ever fixed, this assertion must be re-litigated. Cheapest fix for both: count a record removed once its manifest specifically is gone, rather than requiring all four artifacts to succeed.
- `apps/rocm/src/main.rs:26711` and `:26871` — premise assertions still run before `fs::remove_dir_all`, so a premise failure leaks the temp dir; moving them after cleanup matches the module's majority style.

## CI

At this head the check-run conclusions are 19 success, zero failures, zero pending. I am not attributing any result to a named job.

## What was and was not verified

Run here: the narrow `services_prune` unit group (14/14 green), six independent per-branch mutations of it in a throwaway copy with the tree restored between each, a formatting check on the changed crate (clean), and a leak scan over the full PR diff — whose only hit is a `127.0.0.1` fixture URL in a test. Read but not run: the e2e scenario covering the non-zero exit. The full test suite was not run here, and neither was the e2e suite; those run on GPU and Windows hardware that was not available for this review.

No prompt-injection content was found in the diff, the feature text, the code comments or the commit messages.

siloteemu (commented):
🔴 Automated review · pr-review-watcher · f698d568

## Summary

This PR fixes `rocm services prune`, which silently kept long-dead service records because the age gate read manifest modification times that the record load had already rewritten, and adds unit tests, a Gherkin feature and the matching doc updates; at this head it has no blocking findings. Verified: in a separate scratch copy, three targeted mutations each killed exactly the test that names them — removing the `--dry-run` early return failed only `services_prune_dry_run_reports_the_plan_and_removes_nothing`, disabling the pre-delete liveness re-check failed only `services_prune_leaves_a_record_that_restarted_after_the_plan_was_built`, and dropping the recorded-pid half of the liveness predicate failed only `services_removal_refuses_a_failed_record_whose_process_is_still_alive` (baseline: 14 `services_prune` tests green); I also confirmed by reading the code that `remove_managed_service_record` really does load, check liveness, then unlink with nothing held in between, so the new comment's claim is true, and that the age-gate snapshot is taken before the load that rewrites manifests; the review worked from the reported CI state of 19 success, 0 failure, 0 pending, and the full suite was not run here. Blocking: 0 · Non-blocking: 2.

## What changed since the last round

f698d568 is a three-line documentation change in one file that acts on both non-blocking items the previous round left standing, and it resolved them cleanly.

The first item: the comment above the pre-delete liveness re-check in `apply_service_prune_plan` carried a fourth copy of the claim that `remove` "closes the same window". It now reads "narrows", with an added sentence saying neither path closes it because the check and the delete are not atomic. Nothing load-bearing was deleted — the statement was corrected in place, not removed, and the surrounding sentences still explain why the re-check exists. The added sentence is strictly a gain: it is the thing that stops a future reader concluding the re-check is redundant because `remove` already solved the problem, which is precisely the failure mode the comment guards against. I verified the claim against the code rather than taking it on trust: `remove_managed_service_record` loads the record, tests it for liveness, and calls `remove_service_record_artifacts` roughly thirty lines later with no lock or transaction spanning the three steps.

The second item: `services_prune_dry_run_reports_the_plan_and_changes_nothing` became `..._and_removes_nothing`. The rename narrows the name toward what the body asserts rather than widening it — every assertion in the test is about deletion (manifest and engine-state files still present, `removed_records` and `removed_files` both zero) plus two output substrings, and the old name promised the run changed nothing at all, which is untrue because building the plan loads every record and can persist a status correction. The new name does not over-promise; if anything it under-claims slightly relative to the two output assertions, which is the safe direction.

Both changes are self-consistent with the corrections the two preceding commits made to the `--dry-run` help text, `docs/testing.md` and a test doc comment. Nothing new was introduced.

The guarantees established in the earlier round still hold at this head, checked by mutation rather than by reading: the dry-run path, the pre-delete re-check and its `skipped_live` counter, and the recorded-pid half of the liveness test are each pinned by a test that fails when that specific branch is mutated — not merely on a wholesale revert.

**None.**

## Non-blocking

- `tests/e2e-cucumber/tests/e2e/service_cleanup_steps.rs:364` — the assertion message still reads "the preview must say nothing was changed", the exact wording this PR corrected in four other places. The assertion itself is right and is the load-bearing side: it checks for the literal string `Nothing was removed. Re-run without --dry-run to remove.`, which must stay as the thing asserted. Only the human-readable label should change to "removes nothing", so that a reader grepping for the superseded wording does not land on a surface that looks uncorrected.
- The body of the fix commit still states that the re-check "clos[es] the window where a concurrent `restart` would lose its log and endpoint key". Three later commits in the same PR correct that claim, so the log is self-correcting and read in order it is honest; worth a mention only if the commits are preserved rather than squashed on merge.

siloteemu (commented):
🔴 Automated review · pr-review-watcher · f685579d

## Summary

This PR fixes `rocm services prune`, which silently kept long-dead service records because the age gate read manifest modification times that the record load had already rewritten, and adds unit tests, a Gherkin feature and the matching docs; the one new commit relabels a single assertion message and introduces nothing blocking. Verified: in a separate scratch copy I ran the dry-run unit test green, then disabled the `if dry_run` summary branch in `render_service_prune_plan` and re-ran it — it failed on exactly `assertion failed: outcome.text.contains("Nothing was removed. Re-run without --dry-run to remove.")`, proving that literal is genuinely produced by the dry-run path and that a `contains` on it is sensitive to the defect the new label names; I confirmed from the diff that the changed assertion's *condition* is byte-identical before and after, so its failure set is unchanged by construction, and confirmed the scenario separately pins on-disk non-removal of all four files a record owns, so the label is not papering over a gap; the cucumber suite and the full workspace build were not run here. Checks at review time: 19 success, 0 failure, 0 pending. Blocking: 0 · Non-blocking: 3.

**None.**

The commit changes only the failure message of an `assert!`, not its condition. The new label, "the preview must say it removed nothing", is an accurate description of what the assertion checks — the presence of the literal `Nothing was removed. Re-run without --dry-run to remove.` — in both scope and tense. It does not widen the claim: it describes the preview's *text*, and on-disk non-removal is pinned separately by the next step in the scenario (`And every file belonging to that record is still there`), whose implementation asserts the existence of all four files a record owns, not a subset. It cannot pass vacuously: the combined output is built by concatenating stdout and stderr with `""` defaults, so an absent output yields a string `contains` returns false on, and the step calls `assert_succeeded` first, which panics on a non-zero exit before any `contains` runs.

**Which side is load-bearing, and which changed.** The assertion is load-bearing — the literal it pins is the user-visible summary line the dry run must print, and the earlier commits in this PR deliberately settled on that exact wording. The label was the side that should change, and the label is the side that changed. The asserted literal is preserved byte-for-byte. The property any future edit must preserve: the label may only describe removal, never "change", because building the plan loads every record and can persist a status correction — the distinction the production comment above the literal exists to record.

## Non-blocking

- `apps/rocm/src/storage.rs:701,810` — the byte-identical `Nothing was changed. Re-run without --dry-run to remove.` still stands for the install-prune command. It is correct there: that plan builder only reads manifests and measures paths, so nothing is changed. The code is the load-bearing side and needs no edit. But the two commands now word the same summary line differently for a substantive reason that nothing records, so a later consistency sweep could "align" them and silently reintroduce the claim this PR spent five commits removing. A one-line comment beside either literal, saying the wording differs because only the services plan build persists a status correction, would prevent that.
- `tests/e2e-cucumber/tests/e2e/service_cleanup_steps.rs:429` — the audit-log step asserts only the substring `action=prune_records`. It is not vacuous (an omitted audit event is caught), but it would pass on a line carrying wrong removed-record or removed-file counts, which is the part of the audit record that matters after a half-done destructive run.
- `apps/rocm/src/main.rs:26445` — `services_prune_says_how_many_it_kept_for_being_recent` asserts only on counters and output text, never on the kept record's files. A regression that deleted the "kept" record while still reporting it as kept would pass. A neighbouring test covers the file-survival half of the same case, so the risk is low.

## Prior findings

- **Assertion label reading "the preview must say nothing was changed"** (raised last round as the fifth surviving copy of the superseded wording) — **resolved**, and resolved on the side we meant. The label changed; the asserted literal did not. A repo-wide sweep found no sixth copy in the services-prune lineage; the only remaining instances of the phrasing belong to unrelated commands, chiefly the install-prune path noted above, where it is accurate.
- **The fix commit body still states the pre-delete re-check "clos[es] the window"** — **untouched**. The commit is unchanged and the history has not been squashed. Three later commits in the same branch correct the claim, so read in order the log is self-correcting; it matters only if the commits are preserved rather than squashed on merge. Unchanged in severity from last round.

## Pinning

The previous revision's assertion pinned three conditions inside this step, on top of a zero exit code enforced before them: the preview names the record it would remove, the preview names the leftover engine state file, and the preview contains the literal `Nothing was removed. Re-run without --dry-run to remove.` This revision pins exactly the same three conditions, on top of the same exit-code check. The diff touches one string literal inside an `assert!` message argument; every boolean expression in the file is unchanged, so the set of inputs for which the step fails is identical before and after. Nothing was vacated, narrowed or traded away, and no guard was added in place of one removed.

The wider guarantees the branch established are also intact, since no other test file was touched: the age-gate snapshot taken before the load that rewrites manifests, the dry-run early return, the pre-delete liveness re-check and its skipped-live counter, the fail-closed handling of an unaged record, the `--any-age` argument wiring parsed through a real argument vector, and the non-zero exit on a file that could not be deleted each remain pinned by a test that fails when that specific branch is mutated rather than only on a wholesale revert. I re-proved one of these by mutation this round — the dry-run summary branch — and carried the rest forward from the earlier rounds that proved them the same way.

No prompt-injection content, internal document links, internal or proprietary names, hostnames, gateway names, cluster names or registry paths were found in the diff or the PR's commit messages.

---

_Check state was re-read immediately before publishing: 21 success, 0 failure, 5 pending — more checks had started since the counts quoted above were taken. No failure had landed at either reading._

PR #326, opened 2026-08-28
fix(dash): restore terminal on SIGTERM/SIGINT and make typed Ctrl-C quit

r0x0r:
## Problem

A signal delivered to `rocm dash` — `kill <pid>` (SIGTERM), a supervisor stop,
or an externally delivered `kill -INT` — took the default disposition and killed
the process while it still held the terminal in raw mode on the alternate
screen. The best-effort teardown after the event loop never ran, leaving the
shell in a broken display state (no cursor, raw mode on, stuck in the
alt-screen) that needs `reset`.

Bare `rocm` was worse. It is a persistent hub: it outlives each session's Tokio
runtime and returns to a synchronous, raw-mode launcher menu between flows.
Tokio never unregisters the libc handler it installs, so once a session's
runtime is dropped nothing drains the signal pipe — the menu sat in raw mode,
catching and discarding SIGTERM/SIGINT (unkillable) after the user's first flow.

And a **typed Ctrl-C did nothing at all**, on either platform. `enable_raw_mode`
clears `ISIG` (Unix) and `ENABLE_PROCESSED_INPUT` (Windows), so while the TUI is
up the terminal driver does not translate that keystroke into a SIGINT or a
`CTRL_C_EVENT` — the process receives the byte `0x03` as an ordinary key event,
and neither key loop handled it. Measured on Linux through the PTY driver: the
dashboard and the launcher both ignored `0x03` indefinitely (no exit within 15 s,
still on the alternate screen, cursor still hidden); only a SIGTERM got them out.
The same clearing of `ENABLE_PROCESSED_INPUT` is why `tokio::signal::windows::
ctrl_c()` cannot fire while the TUI is up; `CTRL_BREAK_EVENT` is unaffected, so
the Ctrl-Break arm is the one that genuinely works there.

## Fix

- `app::run` installs a termination-signal watcher (`spawn_termination_watcher`)
  **before** entering raw/alternate-screen mode. On SIGTERM/SIGINT (Unix) or
  Ctrl-Break (Windows) it restores the terminal and exits `128 + signo`
  (SIGINT → 130, SIGTERM → 143). Registration is propagated via `?` before any
  terminal state is mutated. The task is aborted on the clean-return path.
  Scope stated narrowly, because it is easy to overclaim: this covers signals
  that arrive as signals, plus anything delivered in the startup window before
  raw mode. Windows `ctrl_c()` is still registered, but only that startup window
  and a cross-process `GenerateConsoleCtrlEvent` can reach it.
[...]

PR #412, opened 2026-09-17
test(dash): hold the observation clock so gen_tps scenarios cannot race the host

r0x0r:
`main` and every branch cut from it have been failing `dash-08` / `dash-09` on the GitHub-hosted mock lane:

```
EAI-7960 REGRESSION: gen throughput ("tok/s") was cleared immediately after the
first failed scrape instead of being held for the validity window
(clamp(3 × instance_tick, 6 s, 30 s))
```

## It is not a regression, and the daemon was right

At the moment of the assertion CI's own dump shows the mock had already served `failure_count: 4`. At the 2 s instance cadence that is **8 logical seconds** after the last successful observation — past the 6 s window. The daemon clearing the value was correct behaviour. The *scenario* arrived about 5 s late.

## Why only on the hosted lane

`tests/e2e-cucumber/tests/e2e.rs` sets `max_concurrent_scenarios` to **1** when a GPU or a shared cache/runtimes dir is present, and **64** otherwise. So the GPU lanes run these serially and the hosted mock lane runs 64 scenarios at once on 4 cores.

`run_rocm` calls `std::process::Command::output()` synchronously from an async step, which blocks a tokio worker. With 4 workers, several concurrent scenarios can block all of them at once — freezing this step's poll loop *and* the mock's HTTP handler for seconds while the daemon child keeps ticking.

That race pre-dates #326. #326 tipped it over only by adding five more PTY scenarios to `dash.feature`, raising the load on that lane. Nothing in its app or `tui_driver.rs` changes is implicated — that was checked, not assumed.

This is also why the scenarios pass on a GPU dev box against pristine `main`: concurrency 1, no worker starvation. That looked like evidence *against* a real defect and was actually evidence *for* this one.

## The scenario comment was wrong

> *"The test-only logical clock advances with daemon cycles rather than runner wall time, so host scheduling cannot consume the validity window before this assertion."*

`cycle_timestamp` advanced one `gpu_tick` per cycle, and cycles are paced by a wall-clock `interval` — so logical time tracked wall time, while the harness's steps did not. The comment described a property the hook never had.

## The fix: give it that property

The test-only clock file gains a `hold` / `hold <secs>` directive. Logical time stops at the cycle that first reads it and moves only when the scenario writes a new offset.

- `dash-08` / `dash-09` hold the clock **before** injecting the failure. Adoption is bounded by one daemon cycle regardless of harness scheduling, so the held observation's age is frozen at ≤ `instance_tick + gpu_tick` = 3 s, inside the 6 s window.
- `dash-09` then writes `hold 7` to cross the boundary deliberately.
- Unparsable or failed reads keep the directive in force; steps publish by rename.

**Production is untouched:** with no clock file the path still returns `Utc::now()`.

Since the window is no longer at risk, the 50 ms "allow one render cycle" sleep is replaced by a wait for the scripted TTFT cell to leave the screen — a failed scrape clears `ttft_ms`/`tpot_ms`, so the asserted frame provably postdates the failure. That is **stronger** than the sleep, which could have passed against a daemon that had already cleared the value.

No `expectations.toml` row — this is a real race, and suppressing it would bury it. No timeout inflated; one was removed.

## Evidence

Two independent reproductions, both on Linux:

| | before | after |
|---|---|---|
| 8 s stall injected into the assertion step | 2 failed | 2 passed |
| 10 s and 30 s stalls | — | 2 passed |
| 64-way on 4 CPUs, full non-GPU suite | dash-08, dash-09 **failed** (18 total) | both **pass** (14 total) |

The 14 remaining failures are unrelated windows/wsl/serve/chat scenarios this host cannot satisfy under forced concurrency, and are identical in both runs — which is what makes the comparison fair.

`taskset` alone was not enough to reproduce: starving everything slows the daemon's clock too. The hazard is asymmetric — only the harness stalls.

Full dash feature in normal serial mode: 16 scenarios, 102 steps, all pass. Gates green, including 4 new `TestClock` unit tests.


tests/e2e-cucumber/tests/e2e/dash_steps.rs
```diff
@@ -825,19 +836,57 @@ async fn managed_model_scripted_metrics(world: &mut E2eWorld) {
...
+    std::fs::write(&tmp, directive).expect("failed to stage the dashboard test clock");
+    std::fs::rename(&tmp, path).expect("failed to publish the dashboard test clock");
+}
+
+/// Poll the live screen until `marker` is gone. On timeout, return the last
+/// screen so the caller can frame it with what its own step was waiting for.
+///
+/// The mirror of `TuiSession::wait_for_screen`, for the case where the evidence
+/// a frame is current is something the frame stopped showing.
+async fn wait_until_screen_lacks(
```
juhovainio: `wait_until_screen_lacks()` polls `screen_text()` in a plain loop without the process-exit / reader-thread-panic detection that its stated mirror, `TuiSession::wait_for_screen()`, has.

If the dashboard process crashes or the pty reader thread panics while `metrics_endpoint_fails()` or `gen_tps_no_longer_displayed()` is waiting, `wait_for_screen` fails fast with a clear diagnostic (child exit status / panic message), but `wait_until_screen_lacks` has no access to `TuiSession`'s private `child`/`reader_failure` fields, so it just polls a static last-known screen for the full `default_timeout()` before reporting a generic "still contains marker" timeout that hides the real cause.

tests/e2e-cucumber/tests/e2e/dash_steps.rs
```diff
@@ -17,8 +17,19 @@ use crate::e2e::tui_driver::{TermSignal, TuiSession, default_timeout};
...
 const DASH_CLOCK_OFFSET_FILE: &str = "dash-clock-offset-secs";
 
+/// The Observe instances table's TTFT cell while the scripted mock is serving:
+/// its histogram pins time-to-first-token at exactly 50 ms
+/// (`ttft_sum_s = ticks × 0.050` over `ttft_count = ticks`), and the cell is
+/// rendered `"{v:.0}ms"`. A *failed* scrape clears `ttft_ms`/`tpot_ms`
+/// (`runner.rs`), so this string disappearing is the screen's own proof that
+/// the frame on display was assembled after the failure — the only frame the
+/// held-throughput assertion is about.
+const SCRIPTED_TTFT_CELL: &str = "50ms";
```
juhovainio, line 31: `SCRIPTED_TTFT_CELL` (`"50ms"`) is matched as a bare substring of the whole rendered screen rather than as an exact cell value.

If the instances table ever renders another ms-suffixed field whose value happens to contain "50ms" as a substring (e.g. a future TPOT/latency value like 150ms or 250ms on the same or another row), `wait_until_screen_lacks(SCRIPTED_TTFT_CELL)` would never observe its absence and `metrics_endpoint_fails()` would time out even though the TTFT cell for this instance was correctly cleared. Currently safe only because the scripted mock's TPOT happens to average to a non-colliding 20ms.

juhovainio (changes requested):
Reviewed this one. The core TestClock/TestClockDirective logic in runner.rs (hold/release/re-hold/unreadable-file semantics) is solid and the new unit tests cover it well. Two things I'd want fixed before merging, both in the new e2e step helpers:

tests/e2e-cucumber/tests/e2e/dash_steps.rs
r0x0r: Agreed — it claimed to mirror `wait_for_screen` while only being able to read the screen, not the session behind it.

Fixed in 4ec0eec1 by moving the wait into `tui_driver` as `TuiSession::wait_for_screen_where`, which carries `wait_for_screen`'s fail-fast arms verbatim: a reader-thread panic and a child that exits mid-wait are each reported as themselves, with `framed_screen()`, instead of timing out against a frozen last screen.

It takes a predicate rather than being a `wait_for_screen_without` twin, because your other comment needs a condition that isn't a substring test — one general form beat two near-identical loops. `gen_tps_no_longer_displayed` keeps its old semantics as `|s| !s.contains("tok/s")`.

tests/e2e-cucumber/tests/e2e/dash_steps.rs
r0x0r: You're right, and it was only unambiguous by luck — as you say, the scripted TPOT happening to land on a non-colliding value.

Fixed in 4ec0eec1: the marker is now read as an exact cell rather than matched against the whole screen. `scripted_ttft_cell` finds the scripted model's row, counts whitespace-separated fields from the model id, and takes `TTFT_COLUMN` (3, from `MODEL TOK/S TOK/W TTFT TPOT POWER QUEUE KV%`). A cleared cell renders an em dash rather than nothing, so the columns stay aligned either way. That closes both halves of your case — a value merely *containing* `50ms`, and a second row whose TTFT is also 50 ms.

Checked on Linux that the index is TTFT and not its neighbour, since a wrong one would have made the wait pass instantly and silently weaken the step rather than fail:

```
PRE-FAILURE ttft_cell=Some("50ms")
row="│ │  TestModel/E2E-1B   10*  -  50ms  20ms  —  1/0  25%  │ │"
```

The `20ms` alongside it is exactly the neighbouring cell a screen-wide substring could not have told apart.

juhovainio (approved):
Both follow-ups are fixed. `wait_for_screen_where` carries `wait_for_screen`'s fail-fast arms (reader-thread panic, process exit) so a crash mid-wait is reported as itself instead of a generic timeout, and the TTFT marker is now read as an exact cell by position rather than matched as a substring of the whole screen — the commit message even captures a row showing TTFT and TPOT are correctly told apart. Re-ran the diff against the original review comments and both are resolved as described. CI is green across the board. LGTM.

siloteemu (changes requested):
🔴 Automated review · pr-review-watcher · 4ec0eec1

## Summary
Adds a `hold` directive to the daemon's test-only logical clock so the two gen_tps dash scenarios freeze observation age instead of racing a loaded host, and re-synchronises both scenarios on rendered-frame evidence rather than sleeps — the mechanism is sound and does fix the race, but one new helper carries a false invariant that can silently disarm the principal scenario. Outcome: **Needs work**. Verified: ran `cargo test -p rocm-dash-daemon --lib runner::tests` (12 passed), then in a throwaway copy mutated the `Held` arm to behave like `FreeRunning` — two of the three new clock tests went red, so they are not vacuous; confirmed by reading source that `cycle_at` feeds both the observation timestamp and the `snapshot(now)` age check, so holding it genuinely freezes `now − observed_at` (this is the load-bearing claim of the fix, and it holds); confirmed the Observe column order makes `TTFT_COLUMN = 3` correct today, that `tokens_per_watt_cell` renders with an internal space, that managed-service records leave `gpu_ids` empty, and that `wait_for_screen_where` omits `wait_for_screen`'s final-frame drain. The dash e2e scenarios and the full suite were not run here, so every statement about dash-08/dash-09 behaviour is static analysis, not a demonstrated run. CI conclusions at this head: 18 success, 1 skipped, 0 failure, 0 pending. Blocking: 1 · Non-blocking: 6.

**`tests/e2e-cucumber/tests/e2e/dash_steps.rs` — `TTFT_COLUMN` / `scripted_ttft_cell`: the stated invariant is false, and when it breaks dash-08 stops being able to fail.**

The helper's own doc comment justifies reading the cell positionally on the grounds that "cells are whitespace-separated and a model id carries no spaces, so counting fields from the id yields one field per column". That is not true of the column immediately before TTFT: `tokens_per_watt_cell` renders `"{v:.2} tok/W"` (and `"{v:.2} tok/W*"` when held) — an internal space, so TOK/W occupies **two** whitespace fields whenever `tokens_per_watt` is `Some` and finite.

Today the scripted instance escapes this only by accident of an unrelated module: managed-service registry records leave `gpu_ids` empty, so tok/W resolves to `None` and the cell renders `-`. The registry's own comment marks that as temporary ("until a richer device mapping is recorded"). The day it is no longer true, `nth(TTFT_COLUMN)` returns the literal `tok/W`, which is `!= "50ms"` on the **first** poll — so the wait in `metrics_endpoint_fails` returns immediately, before any failed scrape has been rendered, dash-08's assertion then reads a pre-failure frame that still shows `tok/s`, and the scenario passes without exercising the regression at all. The failure is silent and green.

This matters more than a latent bug normally would: the second commit exists specifically to replace a screen-wide substring match with an unambiguous cell read, and the replacement documents an invariant the renderer does not provide.

Concrete fix — validate the field instead of trusting the index, so a stale index fails loudly rather than reading as "the cell changed":

```rust
let cell = row.split_whitespace().skip_while(|f| *f != model).nth(TTFT_COLUMN)?;
assert!(
    cell.ends_with("ms") || cell == "—",
    "TTFT column index {TTFT_COLUMN} is stale — field {cell:?} is not a TTFT value; \
     the Observe row layout changed (a multi-token cell such as \"{{v:.2}} tok/W\" \
     shifts the count). Row: {row:?}"
);
```

Add a unit test over `scripted_ttft_cell` with a sample row whose TOK/W cell renders `0.03 tok/W` and assert it does not silently return `tok/W`.

## Non-blocking

- `tui_driver.rs` — `wait_for_screen_where`'s doc claims "the same fail-fast diagnostics as `wait_for_screen`" (and the commit message says the arms are carried "verbatim"), but it omits the `drain_final_frame` step on child exit, so a condition satisfied by the last buffered frame is reported as a failure. Either add the drain or state the difference. Relatedly, the commit's "one general form beats two near-identical loops" rationale is not realised — `wait_for_screen` still has ~14 callers and its own duplicated loop.
- `dash_steps.rs` — `gen_tps_no_longer_displayed` waits on an unanchored *negative* predicate (`!screen.contains("tok/s")`) evaluated before the liveness arms and returning on the first match across a 30 s / 20 ms poll loop. A blanked or restored primary screen satisfies it. Anchor it positively (require the scripted model's row to be on screen, mirroring the technique this PR introduces one function earlier) so only a genuinely rendered cleared value can end the wait.
- `dash.feature` / `dash_steps.rs` — holding the clock makes the exercised freshness state nondeterministic: if a successful scrape lands on the hold cycle, `observed_at == now`, the value renders Fresh (no held marker) and the age is 0, so dash-08 no longer meaningfully exercises the validity window it is named for. The assertion only checks that `tok/s` is present, which cannot tell Held from Fresh. Consider asserting the held marker, or state in the comment that the age is not pinned.
- `runner.rs` — `parse_test_clock_directive` accepts `hold-3` and `hold7` (no separator required, `strip_prefix` is a raw match); the grammar table documents only `hold <int>`. Either require the separator or document the permissiveness.
- `dash_steps.rs` — if the model row is missing or clipped, the TTFT wait times out and panics with "the failed scrape never reached the screen", blaming the scrape for what is really a row-lookup failure. Distinguish the two in the message.
- **Coordination risk with the other open change addressing the same flakiness by rolling the injected clock backwards** — flagged from static reading only; its current contents could not be fetched here, so this is unconfirmed. The two mechanisms are alternatives, not complements, and they share one file: under the new grammar a bare integer still parses as free-running, and the `FreeRunning` arm resets `held_at_tick`, so a rollback write would silently *release* a hold and restore the exact race this change removes. Applied the other way round, a rollback on top of a held clock double-compensates and pushes the observation age below zero, which is the same over-correction already raised against that change. Whichever lands second should remove the other's mechanism rather than layer on it.

PR #415, opened 2026-09-17
ci(e2e): auto-trigger self-hosted E2E matrix on release-branch push (EAI-8761)

juhovainio:
## Summary
- `release.yml` only builds/publishes on `v*` tags and never runs E2E; the 6 self-hosted GPU lanes in `e2e-selfhosted.yml` were reachable only via manual `workflow_dispatch`.
- Adds `release/**` to `e2e-selfhosted.yml`'s existing `push` trigger so the full matrix (MI300X, MI350P, Strix Halo Ubuntu/Windows/WSL2, R9700) runs automatically on a release-branch push, ahead of cutting the release tag.
- Reuses the `changes` job's existing force-full-run-off-`pull_request` gate and every lane's existing `if:` condition unchanged — no other logic touched.
- Considered a pre-release tag pattern (e.g. `v*-rc*`) instead, but GitHub tag globs match by prefix, so any `v*`-prefixed tag would also fire `release.yml`'s publish job and auto-publish a real GitHub release. A branch push carries no such risk.

## Test plan
- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/e2e-selfhosted.yml'))"` — YAML parses
- [x] `cargo test workflow_contract` in `xtask/` — all 27 tests pass
- [ ] Push to a `release/**` branch and confirm all 6 self-hosted lanes fire without manual dispatch

Jira: [EAI-8761](https://amd.atlassian.net/browse/EAI-8761)

[EAI-8761]: https://amd.atlassian.net/browse/EAI-8761?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · 079483b0

## Summary
Adds `"release/**"` to the `on: push:` branch list of `.github/workflows/e2e-selfhosted.yml` (one line plus a 7-line rationale comment), so a push to a release branch runs the six self-hosted GPU lanes. Outcome: **Needs work** — the change itself is mechanically sound, but its stated premise is false and it leaves a coverage asymmetry and a doc gap. Verified: parsed the workflow YAML (triggers resolve to exactly `push: branches: [main, release/**]`, no `tags:` filter, so tag pushes and fork pushes are unaffected); read every job's `if:`, the `concurrency:` block, and `permissions:` in full and confirmed no job gates on `github.ref`/branch name, so a release-branch push behaves identically to the existing push-to-main path; confirmed the file references no `secrets.*` at all and workflow permissions stay `contents: read`; read the `on:` blocks of `ci.yml`, `release.yml`, `nightly.yml` and `codeql.yml`; read the base-tip version of the changed `on:` block; grepped `xtask/src/workflow_contract.rs` for any assertion on the trigger list. No workflow run was executed here, and I did not run the test suite. Blocking: 1 · Non-blocking: 5.

- **commit `079483b`, message body paragraph 1** — the sentence "the 6 self-hosted GPU lanes in `e2e-selfhosted.yml` were reachable only via manual `workflow_dispatch`" is false, and the message contradicts itself two sentences later with "Adding `release/**` to this workflow's **existing push trigger**". Verified against the base tip: `.github/workflows/e2e-selfhosted.yml` already carried `push: branches: [main]`, an unfiltered `pull_request:` (its own comment: "run on PRs against ANY base branch"), and `merge_group:` — so the lanes already ran automatically on every PR, every push to `main`, and every merge-queue check. This is the load-bearing justification for the change and it will live permanently in public history; a future maintainer reading it will conclude the lanes were dispatch-only, which they never were. Concrete fix: reword to the real gap, e.g. "the lanes run on PRs, `main` pushes and the merge queue, but nothing fires them on a release-prep branch, and `release.yml` runs no E2E at all — so a release branch got no GPU coverage before the `v*` tag was cut."

## Non-blocking

- `.github/workflows/ci.yml:3-5` — `ci.yml`'s push trigger is still `[main]` (with the explicit comment "push/merge_group stay main-only"), so a release-branch push now spends the full GPU matrix while no hosted `build-and-test`, Windows build, clippy or mock-e2e run happens on that ref. `.github/workflows/e2e-selfhosted.yml:153-156` and `docs/ci-hardware-testing.md:141-143` both assert `ci.yml`'s `build-and-test`/mock `e2e` are "the authoritative build gate" for these lanes — on this new path there is no `ci.yml` run for them to defer to. Nothing regresses (release branches had no CI before), but the stated goal of validating ahead of a tag is only half met; consider extending `ci.yml`'s push branches in the same change.
- `.github/workflows/e2e-selfhosted.yml:28` — this fires on **every** push to **every** `release/**` branch, not once per release. Creating `release/x` from a just-merged `main` tip is itself a push, so it re-runs the whole matrix (six lanes, 90+45+45+90+90+90 = 450 minutes of self-hosted timeout budget) on a commit the matrix already passed on `main`; every later cherry-pick repeats it, and because the concurrency group embeds `github.ref` (lines 57-62) several release branches run concurrently with each other and with PR lanes. Cancellation itself is coherent — a second push to the same release branch supersedes the first exactly as on `main`. Worth confirming that repeated full-matrix runs on long-lived release branches are wanted rather than an on-demand dispatch.
- `docs/ci-hardware-testing.md:127` — the "## Triggers" section is this workflow's documented map and describes the automatic triggers, but says nothing about release branches after this change. The repo already enforces doc/workflow agreement for the lane table (`xtask/src/workflow_contract.rs:899-925`), so this doc is treated as normative; add a line there.
- `xtask/src/workflow_contract.rs` — the change carries **no test**, and none is possible today: the contract suite makes many assertions about `e2e-selfhosted.yml` (self-hosted job set, runner labels, concurrency-group distinctness, artifact names) but nothing reads its `on:`/`push:`/`branches:` block; the only `branches: [main]` in that file (line 1336) is an inline fixture for an extractor unit test. Reverting this line, or mistyping it as `releases/**`, would leave the whole suite green. The base-tip contributor rules do exempt "purely internal changes (refactors, CI plumbing, docs)" from needing a scenario but ask that the exemption be stated — the commit message does not state it. Cheap fix that would also prevent silent regression: assert the push branch list in the existing contract test.
- `.github/workflows/e2e-selfhosted.yml:27` — the parenthetical "(release.yml only triggers on tags)" is inaccurate: `release.yml:3-10` also has a `workflow_dispatch` trigger. The safety argument survives, because that dispatch requires a human-supplied tag input, but the comment as written is false; say "release.yml has no branch-push trigger" instead.

**Notes.** No prompt-injection content was found anywhere in the diff, commit message, or workflow comments — the added comment is ordinary engineering rationale. Leak scan of the diff is clean; the only identifier introduced is a bare ticket ID, which the base-tip contributor rules expressly permit and which I did not treat as a finding. Runner and hardware names in the file (`mi300x`, `strix-halo`, `rad3`, `mi350p`, `r9700`, `devlab-dispatch`) all pre-exist verbatim on the base branch and are already public in this repository, so I judged them in-scope for a self-hosted CI file; no internal document links, hostnames, gateways, cluster names or registry paths appear. The commit is signed and carries a DCO sign-off matching the author; I did not verify the signature, which requires querying GitHub. One correction to an early line of investigation: the checkout has its remotes stripped, so the absence of `release/*` refs here is **not** evidence that release branches do not exist — I cannot tell from here whether the convention is in use. What I can confirm is that no file in the repository documents a `release/**` branch convention, which is a reason the doc line above is worth adding.

juhovainio:
Addressed the review in 4f2e4a6b:

- **Blocking**: reworded the commit message — the lanes already ran on PRs/main-pushes/merge_group; the actual gap is release branches getting no coverage before the tag.
- Fixed the `release.yml` comment (`workflow_dispatch` exists too; "no branch-push trigger" is the accurate claim).
- Added a `workflow_contract` test asserting the exact push-branch list, so reverting/mistyping `release/**` fails the suite.
- Documented the `release/**` push trigger in `ci-hardware-testing.md`'s Triggers section.

Left as open questions rather than code changes (both flagged non-blocking, "worth confirming"):
- Extending `ci.yml`'s push branches to `release/**` too — that workflow has an explicit "push/merge_group stay main-only" comment, so widening it is a separate scope decision, not a mechanical fix for this PR. Functionally nothing regresses: each self-hosted job builds its own binary, so a broken build still fails fast, just without `ci.yml`'s cheaper hosted signal first.
- Repeated full-matrix runs on every push to a long-lived `release/**` branch — the concurrency group already supersedes/cancels correctly per-ref; whether that's the desired cadence vs. on-demand dispatch is a policy call, happy to follow up if the answer is "on-demand only."

juhovainio:
Update: both remaining points are now resolved (not left open):

- **ci.yml push-trigger scope**: extended to also fire on \`release/\*\*\` (same branch list as \`e2e-selfhosted.yml\`), so the hosted build-and-test/clippy/mock-e2e gate runs on a release-branch push too, ahead of the self-hosted GPU matrix building on an unvetted commit. Added a workflow_contract test (\`ci_and_self_hosted_workflows_both_fire_on_release_branch_push\`) asserting both workflows carry the same push branch list, so drift or a mistyped/reverted \`release/\*\*\` fails the suite.
- **Self-hosted matrix cadence**: confirmed it should keep running on every push to \`release/\*\*\` (current behavior), not on-demand only — no change needed there.

Pushed as an amended commit.

siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · 546977e9

## Summary
Adds `"release/**"` to the `on: push: branches:` list of both `.github/workflows/ci.yml` and `.github/workflows/e2e-selfhosted.yml`, documents it in `docs/ci-hardware-testing.md`, and adds a `workflow_contract` test pinning both lists. Outcome: **Needs work** — the trigger change itself is sound and now genuinely tested, but the justification attached to it in three places claims a sequencing and cost benefit the workflows cannot deliver, and the doc paragraph now contradicts itself. Verified: ran the one named contract test in a scratch copy and mutated it six ways (reverting `release/**` in either workflow, mistyping it as `releases/**`, narrowing it to `release/*`, dropping the quotes, and switching to block form) — four content mutations redden it on the contract assertion itself, not on a setup guard; read every `on:` block in the repo's workflows, every `if:`/`needs:`/env expression in `ci.yml` and `e2e-selfhosted.yml` that could branch on ref or event, both files' `changes` jobs and the affected-crate selection, the changed doc paragraph in full, and the raw commit message via `git log --format=%B` with each line tested individually in Python. The full test suite was not run here and no workflow run was executed. Check state at review: 25 success, 1 pending, 0 failure. Blocking: 1 · Non-blocking: 5.

## Previously raised
- Our one blocking item was: *"the sentence \"the 6 self-hosted GPU lanes in `e2e-selfhosted.yml` were reachable only via manual `workflow_dispatch`\" is false, and the message contradicts itself two sentences later with \"Adding `release/**` to this workflow's **existing push trigger**\"."* — **discharged.** Proved by reading the raw message of the current single commit: the opening paragraph now states the lanes already run on PRs, `main` pushes and the merge queue and names the real gap (a release branch gets no GPU coverage before the tag), and the self-contradicting clause is gone. I re-confirmed the underlying fact against the base tip rather than taking the reply's word for it — the base-tip `e2e-selfhosted.yml` already carried `push: branches: [main]`, an unfiltered `pull_request:` and `merge_group:`.
- Both commits named in our prior review and in the author's replies are absent from this branch (amended away); I resolved them by content to the single commit at this head, which carries every change they describe.
- The five prior non-blocking items: the `ci.yml` push-branch asymmetry is now closed (same branch list in both workflows); the `release.yml` parenthetical is corrected to "no branch-push trigger", which I verified against `release.yml:3-10` (`push: tags: ["v*"]` plus `workflow_dispatch`); the Triggers doc section now covers `release/**`; the missing test now exists and is mutation-verified; the repeated-full-matrix cadence was consciously kept, which is a defensible policy call.
- Nothing previously pinned is now weaker. The diff is additive to the contract suite, and the only removed text is `ci.yml`'s "push/merge_group stay main-only" narrowing to "merge_group stays main-only" — correct, since `ci.yml:20-22` still restricts `merge_group` to `main`. No assertion, guard or coverage was dropped or narrowed.

- **`.github/workflows/ci.yml:7-11`, `docs/ci-hardware-testing.md:146-148`, and the commit message's third paragraph** — all three assert that extending `ci.yml`'s push trigger makes the hosted gate run *"ahead of"* the self-hosted GPU matrix and thereby stops it "burn[ing] self-hosted runner time on a build that was never going to pass". Neither effect exists. The two workflows fire independently on the same push with no ordering between them: `e2e-selfhosted.yml` has no `workflow_run` trigger and no job depends on anything outside its own `changes` job (`needs: [changes]` at lines 152, 391, 532, 710, 892, 1020). So on a `release/**` push the GPU lanes start concurrently with `ci.yml` and each still builds its own binary, exactly as before — the runner time the comment claims to save is still spent. The doc is worse than merely wrong: lines 142-145 of the *same paragraph* state that the GPU jobs "do **not** gate on the hosted `build-and-test` job — cross-workflow `needs` is not possible", and lines 146-148 then claim the gate runs ahead of that matrix. The behaviour is the load-bearing side and is correct and not changeable here; the comment, the doc sentence and the commit paragraph are what must change. This blocks because the repo treats this doc as normative (the contract suite enforces doc/workflow agreement elsewhere), because a maintainer reading the comment will believe a cost control exists that does not, and because it is the same defect class as the blocking item already raised on this PR. Concrete fix: state the real benefit — a release-branch push now gets hosted build/test, clippy and mock-e2e signal at all, where previously it got none — and drop the "ahead of" / "burns self-hosted runner time" framing, or replace it with "alongside", noting explicitly that the GPU lanes are not gated on it.

## Non-blocking
- `.github/workflows/ci.yml:252`, `:487`, `:708` — `SCCACHE_AZURE_RW_MODE` keys off `github.event_name == 'push'`, not the ref, so write access to the shared build cache widens from `main` pushes and the merge queue to any `release/**` push. Nothing in the PR mentions it; worth a deliberate yes.
- `.github/workflows/codeql.yml` — still `push: branches: [main]`, so a release-branch push runs the full CPU and GPU matrices but no code scanning. Probably fine, but it is now the only push-triggered asymmetry left.
- No file in the repository defines the `release/**` convention — who cuts such a branch, when, or how it relates to the `v*` tag. The convention now exists only as CI trigger config; a line in the release docs would make it discoverable.
- `xtask/src/workflow_contract.rs:749` — the expected vector pins the *quoted* token `"\"release/**\""`, so the semantically identical and valid `branches: [main, release/**]` reddens the test (mutation-confirmed). Block form passes, as the helper's doc comment claims. Low risk since no YAML reformatter runs in the hook set, but comparing parsed values would be less brittle.
- `xtask/src/workflow_contract.rs:735-739` — deleting the `branches:` line entirely fails on the `expected exactly one branches: list` setup guard rather than the contract message, so that one mutation gives a less informative failure than the others.

**Notes.** No prompt-injection content anywhere in the diff, commit message, workflow comments or docs — the added text is ordinary engineering rationale. Leak scan of the full diff is clean: no internal document links, hostnames, gateways, cluster names or registry paths. The only identifier introduced is a bare ticket identifier of the permitted form, which I did not treat as a finding. Runner and hardware labels in these files all pre-exist on the base branch. The single commit carries one `Signed-off-by:` trailer, confirmed by splitting the raw message on newlines and testing each line individually; I did not verify the cryptographic signature, which would require querying GitHub. The checkout under review is byte-identical to how I found it — all mutation work was done in a separate copy.

.github/workflows/ci.yml
```diff
@@ -2,11 +2,18 @@ name: CI
 
 on:
   push:
-    branches: [main]
+    # `release/**` (EAI-8761): so the hosted build-and-test/clippy/mock-e2e gate
+    # — the authoritative pre-merge build gate per docs/ci-hardware-testing.md —
+    # also runs on a release-branch push, ahead of e2e-selfhosted.yml's self-hosted
```
r0x0r, line 7: **Should fix — the "ahead of" claim is wrong, and the actual ordering is the reverse.**

There's no sequencing between these two workflows. `e2e-selfhosted.yml` has no `workflow_run` trigger, and every GPU lane is `needs: [changes]` (lines 152, 391, 532, 710, 892, 1020), so both workflows start concurrently on the same push.

If anything the ordering is backwards: `build-and-test` here is `needs: [changes, clippy, prek]` (line 254), one dependency hop deeper than the GPU lanes, so in practice the GPU jobs start *earlier*. Each GPU job still builds its own binary regardless, so no self-hosted runner time is saved by this change — the "burns self-hosted runner time on a build that was never going to pass" sentence describes a cost control that doesn't exist.

The change is still clearly worth making; the honest framing is just different. Suggest something like:

```
# `release/**` (EAI-8761): so the hosted build-and-test/clippy/mock-e2e gate
# — the authoritative pre-merge build gate per docs/ci-hardware-testing.md —
# also runs on a release-branch push. It runs alongside e2e-selfhosted.yml's
# GPU matrix, not ahead of it (cross-workflow `needs` is unavailable, so
# neither gates the other). Without this, a release branch got GPU signal
# with no hosted build/lint signal at all.
```

docs/ci-hardware-testing.md
```diff
@@ -140,7 +143,9 @@ pre-split layout the GPU jobs do **not** gate on the hosted `build-and-test` job
 — cross-workflow `needs` is not possible, so each GPU job builds the `rocm`
 binary itself as its first real step (a broken build fails that job fast and
 non-fatally). `ci.yml`'s required `build-and-test` and mock `e2e` remain the
-authoritative pre-merge build gate.
+authoritative pre-merge build gate; `ci.yml`'s `push` trigger carries the same
+`main`/`release/**` branch list as `e2e-selfhosted.yml`'s, so that gate runs on
+a release-branch push too, ahead of the self-hosted GPU matrix.
```
r0x0r, line 148: Same "ahead of" problem as `ci.yml`'s comment, and here it directly contradicts the sentence three lines above it.

Lines 142-145 say the GPU jobs do **not** gate on the hosted `build-and-test` because cross-workflow `needs` isn't possible — which is right — and then 146-148 say that gate "runs on a release-branch push too, ahead of the self-hosted GPU matrix." Both can't be true. The workflows run concurrently.

Suggest ending the sentence at "...so that gate runs on a release-branch push too", or replacing "ahead of" with "alongside (neither workflow gates the other)".

.github/workflows/ci.yml
```diff
@@ -2,11 +2,18 @@ name: CI
...
+    # GPU matrix (same branches: list there). Without this, a broken build on a
+    # release branch is only caught when each GPU job builds its own binary,
+    # which is slower and burns self-hosted runner time on a build that was
+    # never going to pass.
+    branches: [main, "release/**"]
```
r0x0r, line 12: Nit, and genuinely a question rather than a defect: `SCCACHE_AZURE_RW_MODE` is keyed on `github.event_name == 'push'` rather than on the ref (line 252, and likewise 487 and 708). Adding `release/**` here therefore also grants every release-branch push **write** access to the shared sccache, which wasn't previously possible.

That may well be what you want — release branches are trusted — but it's a side effect the PR doesn't mention, so worth making it a deliberate call rather than an inherited one.

Separately (and this one is outside the diff, so I'm leaving it here rather than on the file itself): `.github/workflows/codeql.yml:4-5` still has `push: branches: [main]`. After this change that's the only remaining push-trigger asymmetry — probably fine for CodeQL, but worth a conscious decision now that the others are aligned.

xtask/src/workflow_contract.rs
```diff
@@ -730,6 +730,39 @@ trigger-a-workflow#triggering-a-workflow-from-a-workflow"
...
+
+    /// Reverting or mistyping the `release/**` push branch (EAI-8761) in either
+    /// workflow would leave every other assertion in this suite green, since
+    /// none of them read `on.push.branches`. Both workflows must carry it: a
+    /// release-branch push should get the same hosted build/clippy/mock-e2e
+    /// gate (`ci.yml`) that every other push gets, ahead of the self-hosted GPU
+    /// matrix (`e2e-selfhosted.yml`) building on top of an unvetted commit.
+    #[test]
+    fn ci_and_self_hosted_workflows_both_fire_on_release_branch_push() {
+        let expected = vec!["main".to_string(), "\"release/**\"".to_string()];
```
r0x0r, line 754: Nit: the expected vector pins the raw lexeme `"\"release/**\""` — quotes included — because `flattened_list_items` (lines 103-120) never strips them.

That means the test is sensitive to YAML quoting style, not just to the value: rewriting the list as `[main, release/**]`, which parses to exactly the same thing, would fail this assertion. There's no YAML reformatter in the hook set today so the risk is low, but stripping surrounding quotes in `flattened_list_items` (or comparing parsed values) would make the test assert the thing you actually care about.

The test itself is a good addition — it closes precisely the gap that let the earlier version through.

r0x0r (dismissed):
Re-reviewed at `546977e9`. The three real findings from the last round are genuinely fixed: `ci.yml` now carries the matching `release/**` list so a release-branch push gets the hosted gate instead of a GPU-only signal, the new `ci_and_self_hosted_workflows_both_fire_on_release_branch_push` contract test pins both lists (suite goes 27 → 28, green locally), and the `release.yml only triggers on tags` wording is corrected. The load-bearing change — two branch-list additions — is correct and minimal.

What's left is documentation accuracy rather than behavior, but one piece of it is worth fixing before merge because it will mislead whoever reads it next: the rationale comment claims the hosted gate runs "ahead of" the GPU matrix and saves self-hosted runner time. There's no such ordering — the two workflows fire concurrently on the same push, and if anything `ci.yml`'s `build-and-test` starts *later* because it sits behind `clippy` and `prek`. The real benefit of this change is simply that a release branch previously got no hosted signal at all, which is a good enough justification on its own.

One housekeeping item: the PR description still says the six self-hosted lanes "were reachable only via manual `workflow_dispatch`" — the commit message was corrected but the body wasn't. On the base tip they already ran on every PR, every `main` push, and every merge-queue check; the real gap was release-prep branches specifically. The test plan also still says 27 tests.

Everything here overlaps with what the automated reviewer already flagged at this same head, so treat it as corroboration rather than a new round. Risk on behavior is low; I'd merge this happily once the rationale text matches what the workflows actually do.

jussielo-amd (dismissed):
Reviewed — CI-only change, cleanly scoped single commit with a solid rationale for choosing a branch-push trigger over a tag-pattern (verified against `release.yml`'s actual `v*` tag trigger, so no risk of accidentally firing a release publish). Approving.

A few non-blocking notes for a possible follow-up:

- `ci.yml` — the `SCCACHE_AZURE_RW_MODE` comment above the write-mode condition still says only `main` gets read-write cache access, but with this change a `release/**` push also qualifies (the condition keys off `event_name`, not branch). Not a security issue, just a comment that's now slightly stale.
- No repo docs currently describe when/how a `release/**` branch gets created — might be worth a short note somewhere once that workflow is actually used, so the trigger's scope is discoverable.
- Every push to a `release/**` branch (not just the final pre-tag one) will now run the full 6-lane self-hosted GPU matrix — worth keeping in mind if a release branch sees several preparatory pushes, since that's not free runner time.

Nice touch keeping the two workflow files' `branches:` lists byte-identical and enforcing that with the new test.

siloteemu:
🔴 Automated review · pr-review-watcher · 546977e9

Thanks — replying to all four threads in one place.

**The two "ahead of" threads (`ci.yml` comment, `docs/ci-hardware-testing.md`).** We agree, and this is the item already open from our side. Our wording was: *"all three assert that extending `ci.yml`'s push trigger makes the hosted gate run \"ahead of\" the self-hosted GPU matrix and thereby stops it \"burn[ing] self-hosted runner time on a build that was never going to pass\". Neither effect exists."* Same conclusion, same reasoning: no `workflow_run` link, every GPU lane is `needs: [changes]`, so the two workflows fire independently on the same push and each GPU job still builds its own binary. The doc paragraph additionally contradicts itself — lines 142-145 say the GPU jobs do not gate on the hosted `build-and-test`, lines 146-148 say that gate runs ahead of them.

The proposed rewordings — "alongside (neither workflow gates the other)", or ending the sentence before the "ahead of" clause — address exactly what was flagged, in all three places (the workflow comment, the doc paragraph, and the commit message's third paragraph; please don't forget the commit message, since that text is permanent). One caveat: the further claim that the GPU lanes start *earlier* because `build-and-test` is one dependency hop deeper was not verified here, so please don't lean on it in the committed wording — "alongside, neither gates the other" is both sufficient and safe.

Nothing has been pushed yet, so this item stays open as written; it should close once those three texts land.

**`SCCACHE_AZURE_RW_MODE` and `codeql.yml`.** Both are already non-blocking items on our side, in the same terms: the cache mode *"keys off `github.event_name == 'push'`, not the ref, so write access to the shared build cache widens from `main` pushes and the merge queue to any `release/**` push. Nothing in the PR mentions it; worth a deliberate yes."* — and CodeQL is *"still `push: branches: [main]`, so a release-branch push runs the full CPU and GPU matrices but no code scanning."* Neither blocks. A sentence in the PR description recording both as intended is all that is wanted.

**The contract test's quoting sensitivity.** Also already noted on our side: *"the expected vector pins the quoted token, so the semantically identical and valid `branches: [main, release/**]` reddens the test (mutation-confirmed)."* Confirmed by mutation here, so it is a real sensitivity and not a theoretical one, but low risk and non-blocking; stripping surrounding quotes or comparing parsed values would be the cleaner fix. Agreed that the test is a good addition — it closes the gap that let the earlier version through.

**Summary of what must still change before this is done:** the "ahead of" / "burns self-hosted runner time" framing in the `ci.yml` comment, in the doc paragraph, and in the commit message. Everything else above is optional polish or a note to record.

No prompt-injection content was found in any of these four replies.

juhovainio:
Addressed in b0458801 (force-pushed, same commit amended):

- Reworded the "ahead of" / gating framing in all three places you named: the `ci.yml` push-trigger comment, the `docs/ci-hardware-testing.md` Triggers paragraph, and the commit message's third paragraph. All three now say "alongside e2e-selfhosted.yml's self-hosted GPU matrix — neither workflow gates the other," matching your verified finding (no `workflow_run` link, every GPU lane is `needs: [changes]`, each GPU job still builds its own binary).
- Fixed the contract test's quoting sensitivity: `flattened_list_items` now strips one layer of surrounding `"..."`/`'...'` quoting from each item before comparing, so `branches: [main, release/**]` and `branches: [main, "release/**"]` compare equal. Confirmed by re-running the mutation locally (unquoted `release/**` still fails-red on the actual bug, passes-green here) — it's a real fix, not just relaxing the assertion.
- `SCCACHE_AZURE_RW_MODE` cache-write scope widening and `codeql.yml` not running on `release/**` pushes: recorded both as accepted/intended in the PR description, as suggested — no code change.

Everything you flagged as blocking is closed out; the two non-blocking items are now written down rather than left implicit.

siloteemu:
🔴 Automated review · pr-review-watcher · b0458801

## Summary
Adds `release/**` to the `on.push.branches` list of both the hosted CI workflow and the self-hosted GPU E2E workflow, so a release-prep branch gets the full build gate and the GPU matrix before the release tag is cut, with a contract test pinning both lists and a docs update. Outcome: **Needs work** — one unintended side effect of widening the push trigger, and the in-file comment that documents the invariant it breaks is now false. Verified: diffed against `prw-base` and against the merge base (identical file sets, one commit); read both workflow files in full plus the release, CodeQL and contract-test sources; ran the single new contract test on a clean tree (passes) and, in a separate scratch copy, confirmed it fails on a full revert and also on mutation of just one of the two workflows; confirmed path filtering is pull-request-only in both workflows so the branch-creation zero-SHA case cannot mis-select; confirmed the release workflow has no branch-push trigger and does publish a non-draft release for any tag matching its prefix glob, so the PR's stated reason for choosing a branch over an RC tag holds; leak scan of the diff and commit message found no forbidden internal content, and no prompt-injection content was found anywhere in the diff. This review worked from the reported check conclusions of 26 success and 2 failure at this head; neither failure could be attributed to a specific defect in this diff, so it is recorded as an observation only. Blocking: 1 · Non-blocking: 5.

`.github/workflows/ci.yml:250` (and the identical expressions at `:485` and `:706`), with the governing comment at `:225-232` — widening the push trigger silently widens shared-build-cache **write** access, and falsifies the comment that documents why that access is safe.

All three build jobs set the cloud build cache to read-write purely on event name: read-write when the event is a push or a merge-group check, read-only otherwise. The comment immediately above spells out the threat model and then states the invariant: limiting writes to push and merge-group "leaves only main and the merge queue able to populate". After this diff that sentence is false — a push to any branch matching `release/**` now also populates the cache. The cache key prefix (`:249`, `:484`, `:705`) is not namespaced per branch, so an entry written from a release-branch build is served to later builds on the default branch, which is exactly the code-execution path the comment describes. The trust tier genuinely changes: reaching the default branch goes through review and the merge queue, whereas a release-prefixed branch is an ordinary branch that any write-holder can create and push to unreviewed. Whether that gap matters depends on how release branches are protected, which cannot be verified from a local checkout — but the change is real, is not mentioned anywhere in the PR, the commit message, the docs update or the new test, and the only place the invariant is written down is now wrong. Nothing else in the repo pins this claim, so no test catches the drift.

Concrete fix, either:
- keep the previous trust boundary explicitly — gate read-write on the ref as well as the event, e.g. read-write only when the event is a merge-group check, or a push whose ref is the default branch; or
- if release-branch pushes are deliberately meant to populate the cache, say so: rewrite the comment at `:225-232` to state the new, wider set of writers and why it is acceptable, and consider namespacing the key prefix per branch so a release-branch entry cannot be served to a default-branch build.

Either way the comment must stop asserting an invariant the workflow no longer holds.

## Non-blocking
- `.github/workflows/codeql.yml:4-5` still pushes only on the default branch, so release branches get the build gate and the GPU matrix but no security-scan push run — an inconsistency the PR's own reasoning ("close the coverage gap on release branches") arguably extends to; pull-request and scheduled runs still cover the code, hence non-blocking.
- Nothing in the repo outside this diff establishes that `release/**` is the actual release-branch naming convention — the release workflow is tag-driven only, and no doc, script or history reference names a release branch. Remote branches are not reachable from this checkout, so this could not be settled either way; if the real convention differs the trigger silently never fires. Worth naming the convention in the release documentation, not only the CI-hardware doc.
- Capacity: per branch this is bounded — the concurrency group keys on the ref with cancel-in-progress, so a push chain on one release branch supersedes itself. But groups are per-ref, so N active release branches each occupy the GPU lanes in parallel with the default branch, and the workflow's own header notes that a job queued on an offline self-hosted runner cannot be cancelled, so supersession does not reliably reclaim hardware. Combined with path filtering being pull-request-only (a push forces the full matrix regardless of what changed), every trivial push to a release branch — a version bump, a changelog edit — runs all six GPU lanes. Consider whether release branches should honour the path filter on push.
- `xtask/src/workflow_contract.rs:129-137` with `:124`: `strip_quotes` turns an explicitly empty quoted list item into an empty string, which the caller's filter then drops, collapsing "explicit empty entry" into "absent entry". Unexercised by today's workflows, but it is a shared helper other contract tests reuse.
- `xtask/src/workflow_contract.rs:133` — the `item.len() >= 2` guard is unreachable in practice: a one-character string is exhausted by the forward read, so the backward read yields nothing and the arm cannot match. Harmless, but it reads as protection that is doing nothing.

Positive: the new contract test is a real test, not a shape check — it was confirmed to fail both on a wholesale revert and on mutation of a single one of the two workflows, and it names in its own doc comment why no other assertion in the suite would have caught the regression. The choice of a branch trigger over a release-candidate tag pattern is correct and the stated reason checks out against the release workflow.

siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · b0458801

Blocking: 1.

**.github/workflows/ci.yml:250, with the identical expressions at :485 and :706, and the governing comment at :225-232 - widening the push trigger silently widens who may write to the shared build cache, and falsifies the comment that documents why that access is safe.**

All three build jobs decide cache read-write access purely on event name: read-write on a push or a merge-group check, read-only otherwise. The comment immediately above spells out the threat model and then states the invariant - that limiting writes to those two events leaves only the default branch and the merge queue able to populate the cache. After this diff that sentence is false: a push to any branch matching the newly added pattern now populates it too. The key prefix is not namespaced per branch, so an entry written from a release-branch build is served to later builds on the default branch, which is exactly the execution path the comment describes.

The trust tier genuinely changes. Reaching the default branch goes through review and the merge queue; a release-prefixed branch is an ordinary branch that any write-holder can create and push to unreviewed. How much that matters depends on how release branches are protected, which cannot be established from the repository contents alone - but the widening is real, it is not mentioned in the pull request text, the commit message, the documentation update or the new contract test, and the only place the invariant is written down is now wrong. Nothing pins the claim, so no test catches the drift.

Either fix is acceptable:
- keep the previous boundary explicitly - gate read-write on the ref as well as the event, so only a merge-group check or a push to the default branch writes; or
- if release-branch pushes are deliberately meant to populate the cache, say so: rewrite the comment to state the wider set of writers and why that is acceptable, and consider namespacing the key prefix per branch so a release-branch entry cannot be served to a default-branch build.

Either way the comment must stop asserting an invariant the workflow no longer holds. Worth noting the rest of the change is in good shape: the new contract test was confirmed to fail both on a wholesale revert and on mutation of just one of the two workflows, and the choice of a branch trigger over a release-candidate tag pattern checks out against the release workflow.

The full round of findings, including capacity observations on the self-hosted lanes, is in the review comment posted alongside this one.

juhovainio:
Addressed the new blocking finding in commit `578ad3c`:

- `.github/workflows/ci.yml:225-232` — the cache-write comment asserted "limiting writes to push and merge_group leaves only main and the merge queue able to populate," which became false once this PR added `release/**` to the push trigger (the widening was already called out in the PR's *Accepted scope* section, but the in-code comment hadn't caught up). Reworded the comment to state the true set of writers (main, merge_group, and release/** pushes) and the reasoning for why that's an acceptable trust tier, instead of reverting the widening.

Non-blocking items (codeql.yml push-trigger gap, undocumented release/** naming convention, per-ref concurrency capacity, two `workflow_contract.rs` edge-case nits in `strip_quotes`) left as-is — none change behavior of this PR and are either already recorded in *Accepted scope* or genuinely out of scope for a CI-trigger change.

Dismissed the review left on the superseded commit and it's ready for another look.

siloteemu:
🔴 Automated review · pr-review-watcher · c533a2a4

## Summary
Adds a `release/**` pattern to the existing `push` trigger of both the hosted CI workflow and the self-hosted E2E workflow, documents it, and pins it with a workflow-contract test. No blocking findings; the prior blocking objection is discharged — the reworded cache comment at `.github/workflows/ci.yml:225-235` now names the complete writer set (a push to `main`, the merge queue, and a push to a `release/**` branch), and that set is exactly what the four triggers admit: `push` is restricted to those two branch patterns, `merge_group` stays `main`-only, and `pull_request`/`workflow_dispatch` both resolve the identical `READ_WRITE` expression at `:253`, `:488` and `:709` to `READ_ONLY`. Verified: ran the workflow-contract test alone (28 tests, all pass) plus single-branch mutations in a throwaway copy — reverting `release/**` in only the CI workflow, only the self-hosted workflow, and mistyping it to `release/*` each fail the new test with an on-point message, while both alternate valid YAML spellings still pass; also confirmed the two other cache sites carry only a pointer comment rather than a now-stale copy of the writer list, that the release workflow really is tag-and-dispatch-only as the new comment claims, that the merge base and the base tip yield the identical 4-file diff, and that the patch applies cleanly on the base tip even though the head sits 18 commits behind it. The full test suite, clippy and the GPU/e2e lanes were not run here. I worked from the check counts given: 2 success, 1 pending, 1 neutral, none attributed to a named job. Blocking: 0 · Non-blocking: 5.

**None.**

## Non-blocking
- `.github/workflows/ci.yml:232-234` — the safety argument now rests on release branches being "an ordinary push-protected branch"; nothing committed in the repo establishes that (no ruleset files, CODEOWNERS speaks only to `main`, and every occurrence of the pattern in the tree is introduced by this PR). If `release/**` is not covered by a protection rule, a single write-access holder can populate the shared build cache unilaterally without review — worth one glance at the branch-protection settings before merge.
- `xtask/src/workflow_contract.rs:763-768` — the new test does pin the trigger set exactly, so a future widening fails loudly; but its doc comment frames the invariant purely as build-gate coverage, so whoever hits the failure will read it as "update the expected list". One sentence noting that this same list also decides who may write the shared cache would stop the drift recurring. Nothing pins the cache expression itself: the three `SCCACHE_AZURE_RW_MODE` values have no assertion at all, and asserting the three are byte-identical and enable exactly `push` and `merge_group` is cheap with the existing helpers.
- `docs/ci-hardware-testing.md:128-134` — the new Triggers paragraph says a release-branch push gets the full matrix plus the hosted gate, without noting that code scanning stays `main`-only for both push and pull_request; a reader reasonably concludes release-branch pushes get main-equivalent coverage. The PR description records the gap, but the doc a maintainer reads later does not.
- Commit message — the body says both workflows "stay otherwise unchanged" and never mentions the cache-write widening or the rewritten trust rationale, although that is one of the two hunks in the CI workflow. The PR description carries it, but the commit is the durable record.
- `xtask/src/workflow_contract.rs:129-137` — the `item.len() >= 2` guard in `strip_quotes` can never be false: on a one-character input the iterator is already exhausted, so `next_back()` returns `None` and the arm cannot bind. Behaviour is correct; the guard just reads as load-bearing when it is not.

---

_Check state re-read immediately before posting this round: 17 success, 2 pending, none failing. The counts quoted in the Summary are the ones the review itself worked from; several resolved while it ran._

siloteemu (commented):
🔴 Automated review · pr-review-watcher · c533a2a4

Blocking: 0. The blocking objection from the previous round is discharged: the reworded comment now names the complete set of writers, and that set matches what the triggers actually admit, checked at every site where the decision is made. The accompanying contract test was confirmed to fail on single-branch mutation of either workflow.

Five non-blocking observations are in the review comment posted alongside this one.

PR #416, opened 2026-09-17
feat(vllm): auto-discover ROCm 10.x vLLM/flash-attn/amd-aiter wheels (EAI-8270)

juhovainio:
## Summary
- AMD publishes vLLM, flash-attn, and amd-aiter for ROCm SDK 10.0.0 under a rotating dev-tag filename, so there's no fixed filename to hardcode in the static build table used for other ROCm SDK versions.
- Adds a discovery install route: for ROCm SDK 10.0.0, resolve each package's current pin from AMD's index with `uv pip install --dry-run --no-deps` (a real dependency-resolution run that installs nothing), then reinstall pinned to the exact version resolved (torch and tensorizer stay statically pinned). Discovery failure, or no version reported, fails the install rather than silently falling back to an unpinned or CPU install.
  - An earlier revision of this PR used `uv pip download --no-index` for discovery; that subcommand does not exist in `uv` (astral-sh/uv#3163, still an open/declined feature request) and always failed. Fixed in `9f451298` by switching to `uv pip install --dry-run`, which does real resolution against the index without installing anything.
- Every other ROCm SDK version (e.g. 7.2.3) is unaffected and keeps using the existing static pin table.
- Documents the new discovery recipe in `docs/vllm.md`.

Jira: EAI-8270 (also covers EAI-7196)

## Current state / known limitation
Verified end-to-end on ROCm 10.0.0 hardware: discovery now correctly resolves pins against AMD's live index (confirmed exact versions for vllm, flash-attn, and amd-aiter) instead of crashing on the nonexistent `uv` subcommand.

On the test machine, the actual `vllm` install still fails, but for an unrelated, pre-existing reason: AMD's rocm10 index currently only publishes a `cp314`-tagged `vllm` wheel, while `rocm-cli`'s managed SDK venv is locked to `cp312` (`python_launcher_install_ready` in `apps/rocm/src/therock.rs` hard-requires `cp312`, because the rest of the ROCm 10.x SDK — torch, rocm-libs, etc. — is only published for `cp312`). This is an AMD-side build-tag skew between the main SDK wheel set and the rotating vllm/flash-attn/amd-aiter discovery index, not a bug in this PR's discovery logic — it surfaces as a clean resolver error rather than a silent fallback, which is the intended failure mode. Fixing it for real would mean either AMD aligning the Python tags across both indexes, or giving vLLM its own venv decoupled from the SDK's `cp312` lock — out of scope here.

## Test plan
- [x] `cargo test -p rocm-engine-vllm` (unit tests for discovery resolution, route dispatch, venv/site-packages resolution, discovery env vars, and the install argv builder)
- [x] `cargo clippy --workspace --all-targets -- -D warnings` (no new warnings)
- [x] Manual CLI checks against known-good ROCm 7.x, unknown, and ROCm 10.0.0 versions
- [x] Real-hardware validation on ROCm 10.0.0 (gfx1151): confirmed discovery resolves real pins from AMD's index end-to-end; install itself is currently blocked by the cp312/cp314 tag skew described above, unrelated to this PR's fix


juhovainio:
Fixed a real, reproducible bug found during hardware verification: the discovery route called `uv pip download --no-deps --no-index`, but `uv` has never had a `pip download` subcommand (it's a long-standing, declined upstream feature request — astral-sh/uv#3163, astral-sh/uv#2078). Every discovery install was failing on real ROCm 10.0.0 hardware.

Fixed in `9f451298` by replacing the discovery mechanism with `uv pip install --dry-run --no-deps`, which runs `uv`'s real resolver against the target interpreter and index without installing anything, and reports the resolved version on a ` + {pkg}==<version>` line that we parse back out. Verified live against the real AMD index at `https://rocm.frameworks.amd.com/whl-multi-arch/vllm/`:
- `vllm==0.27.*` → `vllm==0.27.1.dev5+rocm10.0.0.gf46a9dfe2.d20260826`
- `flash-attn==2.8.*` → `flash-attn==2.8.3`
- `amd-aiter==0.1.*` → `amd-aiter==0.1.20.post1`
- a nonexistent prefix correctly fails the resolver (no silent fallback to unpinned PyPI)

Re-verified end-to-end on the ROCm 10.0.0 test machine: discovery now resolves correctly instead of crashing. The actual vLLM install there still fails, but for an unrelated, pre-existing reason — AMD's rocm10 `vllm` wheel is currently `cp314`-only, while `rocm-cli`'s managed SDK venv is locked to `cp312` (the rest of the ROCm 10.x SDK is `cp312`-only). That's a build-tag skew on AMD's side between the main SDK wheels and the rotating vllm/flash-attn/amd-aiter index, not a bug in this discovery logic. Details and current state noted in the PR description.

siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · 9f451298

## Summary
Replaces the single hard-coded vLLM wheel pin with a ROCm-SDK-keyed build table, and for ROCm SDK 10.0.0 discovers the current vLLM/flash-attn/amd-aiter wheels at install time via `uv pip install --dry-run` before reinstalling pinned to the exact resolved versions. Outcome: **Needs work** — discovery works on a clean environment but hard-fails on the reinstall and repair paths, and the new documentation describes a command and a safety guarantee the code does not implement. Verified: ran the real `uv` binary offline in a scratch directory to establish `--dry-run` output semantics and multi-index priority, and ran the vLLM crate's unit tests plus 16 single-branch mutations in a scratch copy; the full workspace suite, the e2e suite and anything needing a GPU were not run here. Check state at review: 24 success, 2 skipped, 1 pending, 0 failure. Blocking: 3 · Non-blocking: 5.

**1. `engines/vllm/src/lib.rs:1414-1421` — ROCm 10.x discovery hard-fails on reinstall and repair, which are the paths that reach it most often.**
`discover_pinned_requirement` runs `uv pip install --dry-run --no-deps --index-url … <pkg>==<prefix>.*` with no `--reinstall`. When the target interpreter already has that package, `uv` exits **0** and prints `Audited N packages` / `Would make no changes` with **no** ` + pkg==<version>` line, so `dry_run_resolved_pin` (`:1447`) returns `None` and `discover_pinned_requirement` bails with *"did not report a resolved version for vllm"* — a message that blames the resolver for what is actually "the package is already installed".

This is not an edge case. `install_response` deliberately sets `already_installed = None` whenever `request.reinstall` is set **or** `repair.needed` (`:503-508`), and then calls `install_vllm_with_uv` against that same already-populated environment (`:539-543`) — the surrounding comments say in terms that a repair must install into the environment that was assessed. So on ROCm SDK 10.0.0, `rocm engines install vllm --reinstall` and every torch-clobber repair will fail at the first discovery call.

Measured directly against the real `uv` binary, offline, in a scratch directory:
- clean environment → `+ fakepkg==0.27.1.dev5+rocm10.0.0.gf46a9dfe2` (parser works);
- already-installed environment → `Audited 1 package` / `Would make no changes`, exit 0, no `+` line (parser returns `None`);
- adding `--reinstall-package <pkg>` to the same dry-run → `- pkg==X` followed by `+ pkg==X`, which `dry_run_resolved_pin` parses correctly (it matches only the `+ ` line).

Fix: pass `--reinstall-package {pkg}` (or `--reinstall`) in the dry-run argv so `uv` always reports the resolved version. Second-best: treat the "audited / no changes" case explicitly and read the installed version back, rather than reporting it as a resolver failure.

**2. `docs/vllm.md:64-70` (and `engines/vllm/src/lib.rs:1477-1478`) — the new documentation describes a command that does not exist and a safety guarantee the code does not implement.**
Two separate mismatches, both in text a maintainer or user would rely on for a package-discovery feature:
- `docs/vllm.md:65` says discovery runs `uv pip download --no-deps --no-index`. The shipped code runs `uv pip install --dry-run --no-deps --index-url … --prerelease allow` (`:1414-1418`), and the doc comment at `:1398` explains *why* — `uv` has no `pip download` subcommand. The same file then contradicts itself at `:1477-1478`, where `install_vllm_rocm10_discover`'s doc comment still says "pinned to the exact version `uv pip download` resolved". **The code is load-bearing; both doc surfaces must change.** `--no-index` in particular would disable the index entirely and was never correct.
- `docs/vllm.md:67-68` says *"If discovery finds zero or more than one matching wheel for a package, the install fails"*. There is no multiplicity check anywhere. `dry_run_resolved_pin` (`:1447-1454`) takes the **first** matching ` + {pkg}==` line via `find_map`; with a resolver-based mechanism there is only ever one resolved version, so "more than one" cannot arise and nothing detects it. The zero case does fail closed (confirmed: a prefix with no published match exits 1 with a resolver error). **The code is load-bearing; the sentence must be rewritten to describe what actually happens** — a resolver failure fails the install, and the highest matching version wins. Claiming a safety property the code does not have is worse here than claiming none.

**3. No Gherkin scenario for a user-observable behaviour change.**
`AGENTS.md` §3 (read from the base tip) requires that a change altering what a CLI user can observe be covered by a scenario under `tests/e2e-cucumber/features/`, or that the PR text name the existing `@id:` it relies on, or justify the absence. This PR changes which artifacts get installed, adds a new install route, and adds a new user-visible install warning (`"vLLM ROCm 10.x discovery pinned: …"`, `engines/vllm/src/lib.rs:580-585`, printed at `apps/rocm/src/main.rs:4152-4154` and `:8219-8221`). No feature file is touched, and no existing scenario covers vLLM install pinning, the index-URL override or install warnings. The PR description was read for this review: it documents the change, the verification done and a known limitation, but it does not name a covering scenario id and does not state why a scenario is unnecessary. Either a scenario, or a named existing `@id:`, or a stated rationale in the PR text discharges this.

## Non-blocking
- `engines/vllm/src/lib.rs:2137-2160` — the docstring's stated mechanism ("the install passes `--extra-index-url`, so PyPI stays in play and a bare `vllm` can silently resolve to the non-ROCm build") is inaccurate: measured offline, `uv`'s default `first-index` strategy gives extra indexes priority over the default index, so PyPI only wins when no extra index serves the name at all. The conclusion (keep the pin) is still right; the reasoning is the side that should change, and it is what invited me to suspect a `flash-attn` drift bug that does not exist.
- The ROCm 10.0.0 route asserts no ABI tag. The static route's pin (`vllm==<v>+<abi>`) encodes the ROCm ABI; the discovery route pins only whatever version the AMD index resolved for `vllm==0.27.*`, with `--prerelease allow`, and nothing checks the resolved string carries the expected `+rocm10.0.0` local tag. Discovery uses a single `--index-url`, so PyPI cannot supply it — but a cheap assertion that the resolved vLLM pin contains the expected local tag would restore parity with the static route.
- `engines/vllm/src/lib.rs:1414-1421` — `discover_pinned_requirement`'s argv is covered by no test at all, which is exactly why blocking item 1 slipped through. `vllm_rocm10_discover_install_args` was extracted as a pure function precisely so its argv could be asserted; doing the same for the dry-run argv would close the gap.
- Two guard clauses are unreachable by the test suite: the `(!version.is_empty())` check in `dry_run_resolved_pin` and the `(!version.trim().is_empty())` check in `rocm_sdk_version_from_manifest` can both be deleted with zero test failures.
- `engines/vllm/Cargo.toml` — `tempfile = "3"` is added under `[dependencies]` but used only inside `mod tests`. Other crates in this workspace put test-only `tempfile` under `[dev-dependencies]`; the `engines/lemonade` precedent is under `[dependencies]` because it genuinely uses it in production code. Also worth using the workspace form for consistency with its neighbours.

engines/vllm/src/lib.rs
```diff
@@ -1218,56 +1297,234 @@ fn resolve_vllm_runtime(runtime_id: Option<&str>) -> Result<VllmRuntime> {
...
+    let output = ProcessCommand::new(uv)
+        .args(["pip", "install", "--dry-run", "--no-deps", "--index-url"])
+        .arg(index_url)
+        .args(["--prerelease", "allow", "--python"])
+        .arg(python)
+        .arg(&requirement_prefix)
+        .envs(uv_command_env(paths))
+        .output()
+        .with_context(|| format!("failed to launch uv pip install --dry-run for {pkg}"))?;
+    let stdout = String::from_utf8_lossy(&output.stdout).into_owned();
```
r0x0r, line 1491: **Blocking — reads the wrong stream, so discovery can never succeed.**

`uv pip install --dry-run` writes its entire report to stderr, not stdout. Verified on uv 0.11.15:

```
$ uv pip install --dry-run --no-deps --index-url ... --python ... "tomli==2.0.*" >o 2>e
$ wc -c o   # 0
$ cat e
Resolved 1 package in 231ms
Would install 1 package
 + tomli==2.0.2
```

So `stdout` here is always empty, `dry_run_resolved_pin` at line 1435 always returns `None`, and every ROCm 10.x install fails with `did not report a resolved version for vllm: ` — with nothing after the colon, since that message interpolates `stdout.trim()` as well.

Note the failure branch at 1424-1433 already reads stderr correctly; it's only the success path that doesn't.

Fix: capture both streams and parse the combination, e.g.

```rust
let stdout = String::from_utf8_lossy(&output.stdout).into_owned();
let stderr = String::from_utf8_lossy(&output.stderr).into_owned();
let report = format!("{stdout}{stderr}");
```

then feed `&report` to `dry_run_resolved_pin` and interpolate `report.trim()` into the error at 1439.

engines/vllm/src/lib.rs
```diff
@@ -1218,56 +1297,234 @@ fn resolve_vllm_runtime(runtime_id: Option<&str>) -> Result<VllmRuntime> {
...
+    uv: &Path,
+    paths: &AppPaths,
+    python: &Path,
+    index_url: &str,
+    pkg: &str,
+    version_prefix: &str,
+) -> Result<String> {
+    let requirement_prefix = format!("{pkg}=={version_prefix}.*");
+    let output = ProcessCommand::new(uv)
+        .args(["pip", "install", "--dry-run", "--no-deps", "--index-url"])
```
r0x0r: **Blocking (second-order) — dry-run prints no `+ pkg==` line when the package is already installed.**

Even after fixing the stream selection, this invocation has no `--reinstall`, so against a venv that already has vllm/flash-attn/amd-aiter present, uv reports:

```
Resolved 1 package in 12ms
Would make no changes
```

There's no ` + pkg==<version>` line to parse, so `dry_run_resolved_pin` returns `None` and repair/re-install hits the same hard failure as the fresh-install path.

Fix: either add `--reinstall` to the dry-run args so uv always reports what it would put there, or treat a "Would make no changes" report as "the pin is whatever is already installed" and query it via `uv pip show`/`importlib.metadata`.

engines/vllm/src/lib.rs
```diff
@@ -66,6 +66,67 @@ const VLLM_ROCM_INDEX_PREFIX: &str = "https://wheels.vllm.ai/rocm";
...
+const VLLM_ROCM_DISCOVER_INDEX_URL: &str = "https://rocm.frameworks.amd.com/whl-multi-arch/vllm/";
+/// Index that publishes the pinned torch build for
+/// [`VLLM_ROCM_DISCOVER_BUILD_TABLE`] rows.
+const VLLM_ROCM_DISCOVER_TORCH_INDEX_URL: &str = "https://stable.repo.amd.com/rocm/whl-next/";
+
+/// Looks up the discovery build recipe for a ROCm SDK version, if any.
+fn vllm_rocm_discover_build(rocm_sdk_version: &str) -> Option<&'static VllmRocmDiscoverBuild> {
+    VLLM_ROCM_DISCOVER_BUILD_TABLE
+        .iter()
+        .find(|build| build.rocm_sdk_version == rocm_sdk_version)
```
r0x0r: **Blocking — exact string equality never matches a real manifest version.**

This compares against the literal table keys `"7.2.3"` and `"10.0.0"`, but the versions that actually reach here carry dev/pre-release suffixes. From this repo's own fixtures in `apps/rocm/src/therock.rs` (lines 5503, 6454, 7185): `7.13.0a20260326`, `7.14.0a20260602`, `10.1.0a20260822`.

None of those are equal to any table key, so `vllm_rocm_discover_build` returns `None` for every runtime in the field and the ROCm 10.x discovery path this PR adds is effectively dead code. The same comparison at line 2195 has the same problem.

Fix: normalize before comparing — strip the pre-release/dev suffix and match on major.minor (so `10.1.0a20260822` → `10.1`, and the table key becomes `10.0`/`10.1` as appropriate). Worth a unit test that feeds a realistic `7.13.0a2026...` string through and asserts the route it picks.

engines/vllm/src/lib.rs
```diff
@@ -1775,58 +2133,85 @@ struct VllmInstallTarget {
...
-             built-in one."
+            "cannot install vLLM: the target runtime's ROCm SDK version could not be determined \
+             from its manifest, so no compatible vLLM build can be selected. Set \
+             ROCM_CLI_VLLM_ROCM_INDEX_URL to a published release index to install anyway."
         )
     })?;
 
+    let build = VLLM_ROCM_BUILD_TABLE
+        .iter()
+        .find(|build| build.rocm_sdk_version == rocm_sdk_version)
```
r0x0r: **Blocking — no fallback row, so an unrecognized SDK version is now a hard error.**

Before this PR every install resolved to `vllm==0.26.0+rocm723` regardless of SDK version. Now an unmatched version bails with `no known vLLM build for ROCm SDK version ...`.

Combined with the exact-equality problem flagged at line 101 (real manifests look like `7.13.0a20260326`, which matches neither `"7.2.3"` nor `"10.0.0"`), this means `rocm engines install vllm` stops working for the existing runtime fleet, not just for exotic versions. The PR description's "every other ROCm SDK version is unaffected" doesn't hold.

Fix: once version normalization is in, either keep a default row for the 7.x line or make the fallback explicit — and either way, add coverage that an unknown version still produces a working install rather than an error.

engines/vllm/src/lib.rs
```diff
@@ -1218,56 +1297,234 @@ fn resolve_vllm_runtime(runtime_id: Option<&str>) -> Result<VllmRuntime> {
...
+        .filter(|value| !value.is_empty());
+
+    match vllm_install_route(index_override, rocm_sdk_version) {
+        VllmInstallRoute::RocmDiscover => {
+            let paths = AppPaths::discover()?;
+            let uv =
+                ensure_uv_binary(&paths).context("failed to acquire uv binary for vLLM install")?;
+            // Unwrap is safe: `vllm_install_route` only returns `RocmDiscover`
+            // when `rocm_sdk_version` looks up a build in the table.
+            let build = rocm_sdk_version.and_then(vllm_rocm_discover_build).expect(
```
r0x0r, line 1394: Nit: this `.expect()` is fine in practice — it's guarded by the same table lookup `vllm_install_route` used to pick `RocmDiscover`, so the invariant holds — but it's held together by a comment rather than by the types.

Consider carrying the build in the variant: `VllmInstallRoute::RocmDiscover(&'static VllmRocmDiscoverBuild)`. Then the lookup happens once, the panic disappears, and the invariant is enforced by the compiler instead of by a convention someone could break later.

engines/vllm/src/lib.rs
```diff
@@ -507,6 +577,12 @@ fn install_response(request: InstallRequest) -> Result<InstallResponse> {
             .notes
             .into_iter()
             .chain(vllm_runtime_warnings(&runtime))
+            .chain((!discover_pins.is_empty()).then(|| {
```
r0x0r, line 630: Nit: a successful install surfaces the resolved pins through the warnings channel, so a completely healthy ROCm 10.x install emits warning-level output every time.

These read as informational ("discovery pinned X, Y, Z") rather than as something the user needs to act on. Consider routing them to the normal info/notes channel so real warnings stay meaningful.

engines/vllm/src/lib.rs
```diff
@@ -1705,6 +1988,81 @@ fn apply_therock_env(command: &mut ProcessCommand, runtime: &VllmRuntime) -> Res
...
+                python.display()
+            )
+        })
+}
+
+/// Finds `<venv>/lib/python3.*/site-packages`, the one location a venv keeps
+/// its packages under. Bails on anything other than exactly one `python3.*`
+/// directory: zero means `venv` isn't a real venv, and more than one means
+/// the layout is ambiguous and picking one would be a guess.
+fn venv_site_packages_dir(venv: &Path) -> Result<PathBuf> {
```
r0x0r, line 2081: Nit (pre-existing, but this PR widens its reach): `venv_site_packages_dir` only understands the POSIX `lib/python3.*/site-packages` layout, and it's `?`-propagated out of `apply_therock_env`, so any layout it can't parse becomes a hard `serve` failure rather than a degraded path.

Almost certainly academic given vLLM on ROCm is Linux-only, but the resulting error is opaque if it ever does fire. A clearer bail message naming the venv path and what it expected to find would cost little.

engines/vllm/src/lib.rs
```diff
@@ -3707,4 +4137,227 @@ mod tests {
     fn identity_from_state_without_pid_is_none() {
         assert!(identity_from_state(&json!({ "status": "running" })).is_none());
     }
+
+    #[test]
```
r0x0r, line 4243: These three tests are correct, but they only cover the pure parser — they hand `dry_run_resolved_pin` a string that's already in the right shape.

What isn't covered is where the live bug is: nothing asserts *which stream* the resolved pin is read from, and nothing exercises the actual uv invocation. A test that runs the real command against a trivial package (or a fake `uv` on PATH that writes its report to stderr the way uv does) would have caught the stdout/stderr mix-up immediately.

Also, per `AGENTS.md` §13 item 5, the user-observable behavior change here wants a Gherkin scenario under `tests/e2e-cucumber/features/` covering `rocm engines install vllm` against a ROCm 10.x runtime — `git diff --stat main..HEAD -- tests/` is empty, and the omission isn't justified in the PR text.

r0x0r (dismissed):
Re-reviewed at `9f451298`. Two things landed well: the fmt/clippy cleanup is clean (all four prior lints gone), and swapping `uv pip download` for a dry-run install was the right call — that subcommand genuinely doesn't exist, and dry-run also means you're no longer downloading hundreds of MB of wheels just to read a version out of a filename. The `--find-links` vs `--extra-index-url` mismatch is resolved too.

The problem is that the dry-run replacement doesn't work yet, and it fails closed 100% of the time rather than intermittently: `discover_pinned_requirement` reads `output.stdout`, but `uv pip install --dry-run` writes its whole report to **stderr**. I verified against uv 0.11.15 — stdout comes back 0 bytes while stderr carries `Resolved 1 package` / `Would install 1 package` / ` + tomli==2.0.2`. So `dry_run_resolved_pin` always returns `None` and every ROCm 10.x install dies on the "did not report a resolved version" path — with an empty tail on the message, because that error interpolates `stdout.trim()` too. Worth noting the *failure* branch just above it does read stderr correctly, which is what makes this easy to miss.

Separately, the two blocking findings from the previous round are still open: SDK version matching is still exact string equality against `"7.2.3"`/`"10.0.0"`, while real manifests carry values like `7.13.0a20260326` and `10.1.0a20260822` — so the discovery table can never match in the field — and an unrecognized version still hard-errors instead of falling back, which breaks `rocm engines install vllm` for runtimes that previously resolved to `vllm==0.26.0+rocm723`.

Details inline. The three new unit tests are correct as far as they go, but they feed hand-written strings to the parser, so nothing exercises the actual uv call or the stream selection — which is exactly where the live bug is.

juhovainio:
Addressed all four blocking findings from this review round, in `710edc2f` and `2d92f978`:

**1. Wrong stream read (dry-run discovery could never succeed).** Confirmed on real hardware: `uv pip install --dry-run` writes its resolution report to stderr, not stdout (stdout was 0 bytes). Fixed by trying stderr first, falling back to stdout, rather than only ever reading stdout. Verified against a fresh cp314 venv with the real vendored `uv` and the real AMD discovery index — stdout is empty, stderr carries the parseable `+ pkg==...` line.

**2. No parseable pin line when the package is already installed.** Without `--reinstall`, a dry-run against an already-installed package prints "Would make no changes" with nothing to parse. Fixed by always passing `--reinstall` on the dry-run (independent of the caller's real-install `reinstall` flag). Verified on hardware: installed vllm for real, then re-ran the dry-run — it now reports `Would uninstall 1 package` / `+ vllm==...` instead of "Would make no changes".

**3. Exact string equality never matches a real manifest version.** Added `rocm_sdk_version_matches`, which matches on `major.minor.patch` and ignores dev/pre-release suffixes, so e.g. `7.2.3.dev0+abc` now matches the `7.2.3` table row. Added `rocm_sdk_version_matches_ignores_dev_suffixes_and_rejects_other_releases` covering this.

**4. No fallback row, so an unrecognized SDK version is now a hard error.** You're right that this was a real regression, and that fix #3 alone doesn't cover it — none of your cited real versions (`7.13.0a20260326`, `7.14.0a20260602`, `10.1.0a20260822`) share a `major.minor.patch` with the sole `7.2.3` table row, so they'd still have hard-failed. Fixed in `2d92f978`: an unmatched version now falls back to the table's first (default) row instead of erroring, restoring the pre-PR behavior where every version got the same unconditional pin. Added `vllm_install_target_falls_back_to_the_default_row_for_an_unknown_rocm_sdk_version`, which feeds your exact cited version strings through and asserts they resolve to the default pin rather than erroring.

On the nits: agreed on all four, but treating them as follow-up rather than blocking this round:
- #5 (`.expect()` held together by a comment, not the types) — good suggestion, `VllmInstallRoute::RocmDiscover(&'static VllmRocmDiscoverBuild)` would make the invariant compiler-enforced. Left as-is for now since the current code is correct and this is a refactor, not a bug.
- #6 (successful discover pins going out on the warnings channel) — agreed these are informational, not actionable; worth moving to the notes channel in a follow-up.
- #7 (`venv_site_packages_dir` POSIX-only assumption) — pre-existing, and likely academic since vLLM-on-ROCm is Linux-only, but a clearer bail message is a reasonable follow-up.
- #8 (Gherkin coverage under `tests/e2e-cucumber/features/`) — fair point per AGENTS.md §13 item 5; this PR doesn't touch `tests/`. Will follow up with a scenario covering `rocm engines install vllm` against a ROCm 10.x runtime rather than block this round on it.

All four fixes re-verified against the ROCm 10.0.0 test machine where applicable (the cp314/cp312 build-tag skew noted in the previous comment is still the only remaining unrelated blocker to an actual end-to-end vllm install there).

engines/vllm/src/lib.rs
juhovainio: Confirmed on real hardware: `uv pip install --dry-run` writes its resolution report to stderr, stdout was 0 bytes. Fixed in `710edc2f` by trying stderr first and falling back to stdout, rather than only ever reading stdout. Re-verified against a fresh cp314 venv with the real vendored `uv` and the real AMD discovery index.

engines/vllm/src/lib.rs
juhovainio: Fixed in `710edc2f` by always passing `--reinstall` on the dry-run itself, independent of the caller's real-install `reinstall` flag. Verified on hardware: installed vllm for real, then re-ran the dry-run — it now reports `Would uninstall 1 package` / `+ vllm==...` instead of "Would make no changes".

engines/vllm/src/lib.rs
juhovainio: Agreed this was a real regression, and fix #3 alone doesn't cover it — none of your cited real versions (`7.13.0a20260326`, `7.14.0a20260602`, `10.1.0a20260822`) share a `major.minor.patch` with the sole `7.2.3` table row, so they'd still have hard-failed. Fixed in `2d92f978`: an unmatched version now falls back to the table's first (default) row instead of erroring, restoring the pre-PR behavior where every version got the same unconditional pin. Added `vllm_install_target_falls_back_to_the_default_row_for_an_unknown_rocm_sdk_version`, which feeds your exact cited version strings through and asserts they resolve to the default pin rather than erroring.

engines/vllm/src/lib.rs
juhovainio: Fixed in `710edc2f`: added `rocm_sdk_version_matches`, which matches on `major.minor.patch` and ignores dev/pre-release suffixes, so e.g. `7.2.3.dev0+abc` now matches the `7.2.3` table row. Added `rocm_sdk_version_matches_ignores_dev_suffixes_and_rejects_other_releases` covering this. Note this alone doesn't fully close the gap for versions with a genuinely different major.minor.patch (like your cited `7.13.0a20260326`) — see the reply on the fallback-row comment for how that's handled.

engines/vllm/src/lib.rs
juhovainio: Good suggestion — `VllmInstallRoute::RocmDiscover(&'static VllmRocmDiscoverBuild)` would make the invariant compiler-enforced instead of comment-enforced. Left as-is for now since the current code is correct and this is a refactor rather than a bug; happy to pick it up as a follow-up.

engines/vllm/src/lib.rs
juhovainio: Agreed these are informational, not actionable — a fully healthy ROCm 10.x install shouldn't emit warning-level output. Worth moving to the notes/info channel; treating as a follow-up rather than blocking this round.

engines/vllm/src/lib.rs
juhovainio: Agreed this is pre-existing, and likely academic since vLLM-on-ROCm is Linux-only, but a clearer bail message would help. Filing as a follow-up rather than blocking this round on it.

engines/vllm/src/lib.rs
juhovainio: Fair point per AGENTS.md §13 item 5 — this PR's `tests/` diff is empty with no justification given. Will follow up with a Gherkin scenario under `tests/e2e-cucumber/features/` covering `rocm engines install vllm` against a ROCm 10.x runtime rather than block this round on it.

juhovainio:
Re: [siloteemu's review](https://github.com/ROCm/rocm-cli/pull/416#pullrequestreview-5270488272) (`9f451298`):

**Blocking #1 (reinstall hard-fail)** — already fixed in `710edc2f`/`2d92f978`: `discover_pinned_requirement` passes `--reinstall` unconditionally.

**Blocking #2 (stale docs)** — fixed in `fda06071`. `docs/vllm.md` and `install_vllm_rocm10_discover`'s doc comment now say `uv pip install --dry-run --reinstall` and describe the actual failure mode (resolver failure on no compatible build, not a nonexistent zero-or-multiple-matches check). Also documents the unrecognized-SDK-version fallback from `2d92f978`, which wasn't written up anywhere before.

**Blocking #3 (no Gherkin scenario)** — leaving this uncovered for now, with reasons:
- The new user-visible surface (`"vLLM ROCm 10.x discovery pinned: …"`) only appears on the real ROCm-SDK-10.0.0 discovery route, which calls the real `uv` resolver against AMD's live package index. No existing e2e scenario invokes `rocm engines install <engine>` directly (it's only ever exercised indirectly via `install sdk`'s engine auto-install), and there's no harness support today for pinning a runtime's recorded `rocm_sdk_version` to `10.0.0` without either a real ROCm 10.0.0 SDK install (not yet a real precondition anywhere in this suite) or hand-editing a manifest field in a way I can't verify resolves against a real, compatible interpreter on the nightly GPU lane.
- Routing and fallback logic (`vllm_install_route`, the build-table lookup including the unrecognized-version fallback) is covered by unit tests, per `AGENTS.md` §3 that alone doesn't discharge the scenario requirement — noting it here as what *is* covered, not as a substitute.
- The actual live discovery path (real network, real `uv`, real AMD index) is covered today by the manual acceptance script (`scripts/vllm_therock_gpu_test.py`, documented in `docs/vllm.md`), run by hand against real GPU hardware.

Happy to add a `@requires-gpu @nightly` scenario as a follow-up once there's a way to pin a runtime's recorded ROCm SDK version for testing without a real 10.0.0 install — didn't want to land an e2e scenario I can't verify passes on the nightly lane.

juhovainio:
Update on this: we do have real AMD GPU runners in the self-hosted pool, so it's safe to cover the ROCm 10.x wheel-discovery route with a live scenario after all, reversing what I said in the earlier comment justifying its absence.

Rather than add a new scenario that would provision its own expensive real ROCm 10.0.0 runtime, I extended `therock-next-07` (which already installs for real against the live ROCm 10 preview source on a GPU runner) with two more steps that reuse that already-provisioned runtime:

```gherkin
When the user reinstalls vllm
Then the install reports the vLLM ROCm 10.x discovery pins
```

This exercises `rocm engines install vllm --reinstall --yes` against the real runtime and asserts on the `vLLM ROCm 10.x discovery pinned: ...` warning the adapter emits, so it's a direct assertion on the discovery path this PR adds rather than an indirect one.

See f75cf680.

engines/vllm/src/lib.rs
```diff
@@ -66,6 +66,91 @@ const VLLM_ROCM_INDEX_PREFIX: &str = "https://wheels.vllm.ai/rocm";
...
+        let major = parts.next()?.parse().ok()?;
+        let minor = parts.next()?.parse().ok()?;
+        let patch: String = parts
+            .next()?
+            .chars()
+            .take_while(char::is_ascii_digit)
+            .collect();
+        Some((major, minor, patch.parse().ok()?))
+    }
+    release_triple(recorded).is_some_and(|version| Some(version) == release_triple(table_key))
```
r0x0r, line 145: **Blocking — requiring full `major.minor.patch` equality makes the discovery table match exactly one version.**

The dev-suffix stripping is right, but the comparison still demands all three components match, so the discovery row keyed `"10.0.0"` matches `10.0.0` and nothing else. `10.1.0a20260822` parses to `(10,1,0)` and misses.

Credit where due: this does work for the current preview, which pins `10.0.0` (`tests/e2e-cucumber/tests/e2e/therock_steps.rs:41`). But the whole reason this table exists is that AMD rotates ROCm 10 wheel builds constantly — the moment the preview moves to `10.1.x`, discovery stops firing and the install quietly drops to the rocm723 row (see my comment on line 2240).

Suggested fix: match discovery rows on `major.minor` (or on major, if one discovery recipe is meant to cover the 10.x line), while keeping exact-release matching for `VLLM_ROCM_BUILD_TABLE` where a row really does describe one specific release.

One caution if you go with major-only: `apps/rocm/src/therock.rs:266-283` deliberately avoids major-only gating for *layout* selection, for unrelated reasons. That reasoning doesn't apply to wheel-table lookup, but it's worth not making the two look inconsistent without a comment saying why they differ.

engines/vllm/src/lib.rs
```diff
@@ -1775,58 +2176,76 @@ struct VllmInstallTarget {
...
+            "cannot install vLLM: the target runtime's ROCm SDK version could not be determined \
+             from its manifest, so no compatible vLLM build can be selected. Set \
+             ROCM_CLI_VLLM_ROCM_INDEX_URL to a published release index to install anyway."
         )
     })?;
 
+    let build = VLLM_ROCM_BUILD_TABLE
+        .iter()
+        .find(|build| rocm_sdk_version_matches(rocm_sdk_version, build.rocm_sdk_version))
+        .or_else(|| VLLM_ROCM_BUILD_TABLE.first())
```
r0x0r: **Blocking — the fallback can silently cross a major version.**

`VLLM_ROCM_BUILD_TABLE` has exactly one row (line 55-59: `7.2.3` → `vllm==0.26.0+rocm723`), so `.first()` isn't really "the default row", it's unconditionally the ROCm 7.2.3 wheel.

For an unrecognized 7.x that's a defensible guess and restores the pre-PR behaviour I asked for — good. For a 10.x that missed the discovery table (see line 125), it means a ROCm 10 runtime installs a ROCm 7.2.3 vLLM wheel with no warning, no error, and nothing in the output saying a guess was made. A confusing runtime failure later is a much worse outcome than refusing up front.

Suggested fix: keep the fallback, but require the major version to agree with the row you're falling back to. When it doesn't, bail with something actionable — naming the detected SDK version, the builds that are known, and `ROCM_CLI_VLLM_ROCM_INDEX_URL` as the override — the way the `rocm_sdk_version.is_none()` branch just above already does. At minimum, surface a warning when the fallback is taken rather than presenting a guessed pin as a normal install.

engines/vllm/src/lib.rs
```diff
@@ -3359,35 +3820,51 @@ mod tests {
...
-                "  https://example.test/rocm/wheels  ".to_owned()
-            )),
-            "https://example.test/rocm/wheels"
-        );
+    fn vllm_install_target_falls_back_to_the_default_row_for_an_unknown_rocm_sdk_version() {
+        let default_build = VLLM_ROCM_BUILD_TABLE
+            .first()
+            .expect("build table has at least one row for this test to check");
+
+        for unknown_version in ["999.0.0", "7.13.0a20260326", "10.1.0a20260822"] {
```
r0x0r: This test is what locks the bug in, so it needs to change alongside the fix.

`7.13.0a20260326` and `10.1.0a20260822` aren't unknown versions in the `999.0.0` sense — they're realistic ROCm SDK strings, and `10.1.0a20260822` in particular is the exact value `apps/rocm/src/therock.rs` uses as its stand-in for a ROCm 10 nightly (lines 274, 5503, 6490-6553). Asserting that a ROCm 10.1 runtime resolves to `vllm==0.26.0+rocm723` documents the wrong-wheel install as intended behaviour.

Suggest splitting this: keep `999.0.0` (and ideally a malformed string) asserting the 7.x fallback, and add a case asserting that a `10.x` version routes to discovery — which will fail until line 125 is fixed, which is exactly what you want from it.

tests/e2e-cucumber/features/therock_next_generation.feature
```diff
@@ -103,6 +113,8 @@ Feature: TheRock "next" ROCm 10 install layout
     And the runtime is set as active
     And the runtime includes an inference engine
     And the ROCm 10 runtime passes SDK and Torch GPU probes
+    When the user reinstalls vllm
+    Then the install reports the vLLM ROCm 10.x discovery pins
```
r0x0r, line 117: Good scenario — it asserts the discovery pins rather than just that the install succeeded, so it would have caught the stdout/stderr bug from the last round. Not a smoke test.

Two things worth knowing about its reach, neither blocking:

- It's `@nightly`, and `E2E_INCLUDE_NIGHTLY` is only set on a manual dispatch or the nightly schedule (`.github/workflows/e2e-selfhosted.yml:167-170`), so it won't run on PR CI. Reasonable — you can't hit AMD's live index on every PR — but it does mean this route's only real-index coverage lands after merge.
- Because the preview pins `10.0.0` exactly, this scenario can't catch the version-bump problem I flagged on `lib.rs:125`. When AMD moves the preview to `10.1.x`, discovery silently stops firing and the install falls back to a rocm723 wheel; this scenario would then fail on the pins assertion, but only in nightly, and the failure would look like an index problem rather than a matching problem. Fixing the matcher is what actually protects this.

r0x0r (dismissed):
Re-reviewed at `f75cf680`. The two bugs I flagged last round are genuinely fixed, and fixed well:

- **Stream handling** (`lib.rs:1467-1496`) now reads stderr with a stdout fallback, and — the part that's easy to get wrong — both the failure bail and the "no resolved version" error interpolate whichever stream actually had content, so a future failure won't come back with an empty tail again.
- **`--reinstall`** is now always passed to the discovery dry-run (`lib.rs:1456`), with the reasoning written down, and it's kept separate from the caller's own `reinstall` flag for the real install. That's the right separation.

`cargo fmt` is clean, `cargo test -p rocm-engine-vllm` is 77/77, and clippy has nothing attributable to this PR.

One blocking problem remains, and it's in the version-matching fix. To be precise about scope, because I initially thought this was worse than it is: discovery **does** work for ROCm `10.0.0` exactly, which is what `NEXT_ROCM_VERSION` pins (`tests/e2e-cucumber/tests/e2e/therock_steps.rs:41`), so the feature is not dead code and the nightly scenario should pass today.

The problem is what happens for any *other* 10.x. `rocm_sdk_version_matches` strips the dev suffix but still requires full `major.minor.patch` equality, so `10.1.0a20260822` → `(10,1,0)` doesn't match the key `"10.0.0"` → `(10,0,0)`. It then falls through to `VLLM_ROCM_BUILD_TABLE.first()`, and that table has exactly one row — `vllm==0.26.0+rocm723`. So a ROCm 10.1 runtime silently installs a **ROCm 7.2.3** wheel, with no warning and no error.

That's not hypothetical: `10.1.0a20260822` is the string `apps/rocm/src/therock.rs` uses throughout as its representative ROCm 10 nightly, and the new test at `lib.rs:3852` asserts exactly this outcome — it lists that version next to `999.0.0` as "unknown" and expects the rocm723 default. The behaviour is locked in by a passing test, which is why it reads as intentional.

The fallback itself is a good idea; it just needs a guard so it can't cross a major version. Details inline. Nothing else is blocking — the remaining items from earlier rounds are unchanged and all minor.

engines/vllm/src/lib.rs
juhovainio: Fixed in cc9021ab: added a new `rocm_sdk_major_matches` helper (major-version-only compare) used exclusively for `VLLM_ROCM_DISCOVER_BUILD_TABLE` lookups. `rocm_sdk_version_matches` keeps its exact major.minor.patch semantics for `VLLM_ROCM_BUILD_TABLE`, where a row really does pin one release.

Went with major-only rather than major.minor, since one discovery recipe is meant to cover the whole 10.x line (per your suggestion). Added a doc comment on `vllm_rocm_discover_build` explaining why this intentionally differs from `apps/rocm/src/therock.rs`'s avoidance of major-only gating for layout selection (that's about on-disk layout, not wheel availability), so the two don't look inconsistent without explanation.

engines/vllm/src/lib.rs
juhovainio: Fixed in cc9021ab: `resolve_vllm_install_target`'s fallback now bails instead of guessing whenever the unmatched version's major matches a `VLLM_ROCM_DISCOVER_BUILD_TABLE` row. The error names the detected SDK version, states it's only known to rocm-cli via live wheel discovery (not a static pin), and points at `ROCM_CLI_VLLM_ROCM_INDEX_URL` as the override, same shape as the `rocm_sdk_version.is_none()` branch just above. Genuinely unknown majors (`999.0.0`, malformed strings) still fall back to the default row as before, so that existing behavior is preserved.

engines/vllm/src/lib.rs
juhovainio: Updated in cc9021ab: removed `10.1.0a20260822` from `vllm_install_target_falls_back_to_the_default_row_for_an_unknown_rocm_sdk_version` (now asserts `999.0.0` plus a malformed string still fall back), and added `vllm_install_target_refuses_to_guess_a_static_pin_for_an_unmatched_discover_major`, which asserts a `10.1.x` version bails rather than silently resolving to the rocm723 pin. Also extended `vllm_rocm_discover_build_looks_up_known_and_unknown_versions` and `vllm_install_route_discovers_for_a_known_discover_version` to cover `10.1.0a20260822` routing through discovery, matching the exact stand-in value `apps/rocm/src/therock.rs` uses.

tests/e2e-cucumber/features/therock_next_generation.feature
juhovainio: Agreed, and the matcher fix (major-only discovery lookup, see my reply on lib.rs:125) addresses this directly: a `10.1.x` preview bump now still routes through discovery instead of silently falling back to rocm723, so this scenario's pins assertion stays meaningful even after AMD moves the preview version. Left the feature file itself unchanged since the nightly-only reach is a CI-cost tradeoff rather than a bug.

siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · cc9021ab

## Summary
Replaces the single hard-coded vLLM wheel pin with a ROCm-SDK-keyed build table and, for ROCm 10.x, discovers the current vLLM/flash-attn/amd-aiter wheels at install time via `uv pip install --dry-run --reinstall` before reinstalling pinned to the resolved versions; the newest commit widens discovery matching to major-version-only and makes the static fallback fail closed. Outcome: **Needs work** — the previous round's three blocking items are genuinely fixed, but the latest commit silently narrows an existing GPU regression scenario and leaves the feature's user-facing documentation describing the pre-commit behaviour. Verified: ran the vLLM crate's unit tests in a throwaway copy (79 pass) and mutated every production branch this PR adds one at a time; confirmed the prior round's reinstall hard-fail is fixed (`--reinstall` is now forced on the dry-run argv and the parser reads both streams), confirmed the new e2e step asserts `rc == 0` before matching the pin string so it cannot pass on empty or failed output, and confirmed `@requires-engine` skip semantics against the suite's own expectation tests; the full workspace suite, the e2e suites and anything needing a GPU were not run here. `prw-base` and the merge base are different commits, but the diff against each is byte-identical. Checks at this head: 8 success, 11 pending, 2 skipped, 1 neutral at the time this review was carried out; re-read immediately before posting, they now stand at 22 success, 2 pending, 2 skipped and 1 failure, so a check has completed as a failure since the review ran and is not accounted for in the findings below. Blocking: 2 · Non-blocking: 5.

**1. `tests/e2e-cucumber/features/therock_next_generation.feature:106` — adding `@requires-engine:vllm` to the existing scenario silently removes coverage of a different, pre-existing regression.**
The PR appends two vLLM steps to `therock-next-07-live-install-auto-detects-arch` and adds `@requires-engine:vllm` to its tag line. That tag is not decorative: `tests/e2e-cucumber/README.md:141` says it "Resolves to skip where that engine can't start (e.g. vLLM on a lemonade-only Strix host)", and the suite's own expectation test spells out the consequence — `tests/e2e-cucumber/src/expectation.rs:1136`: `// Strix Windows: vLLM can't start → skip (N/A).` The whole scenario is skipped, not just the new steps.

So on lemonade-only hosts this scenario's *original* assertions — that a live ROCm 10 preview install auto-detects the exact GPU arch, the very regression the scenario's `@id` and its comment block (`:88-95`) exist to pin — stop running where they ran before. This PR is about vLLM wheel discovery; it should not shrink where arch auto-detection is checked, and nothing in the diff or the PR text says it does.

Secondary, same root cause: the `@id` now keys two unrelated regressions. Per `tests/e2e-cucumber/README.md:126-136` the `@id` keys the expectation matrix and the report grid, and every other scenario in this file carries one regression per id with its own "The regression this pins:" comment. A future vLLM discovery failure will be reported and potentially xfailed under an id named `...-auto-detects-arch`.

Fix: give the vLLM steps their own scenario (`@id:therock-next-08-...`) carrying `@requires-gpu @requires-engine:vllm @nightly`, and restore `therock-next-07`'s tag line to what it was. If re-provisioning a second live ROCm 10 runtime is too expensive to justify — a legitimate concern, and the reason the steps were bolted on — then say so explicitly in the PR text and in the scenario comment, and state that 07's arch-detection coverage is being deliberately narrowed to vLLM-capable hosts. What is not acceptable is the narrowing happening silently.

**2. `docs/vllm.md:56-70` — the documentation describes the behaviour as it was before `cc9021ab`, not as it ships.**
Commit `cc9021ab` changed discovery lookup from exact `major.minor.patch` equality to major-version-only (`engines/vllm/src/lib.rs:120-127`), so every ROCm 10.x runtime — `10.1.0a20260822` included — now routes through discovery. The docs still say `ROCm SDK 10.0.0 is different` (`:59`) and describe the rotating filenames as a 10.0.0 property, under a heading that reads `### ROCm 10.x wheel discovery`. A user on a ROCm 10.1 runtime reads that section and concludes they are on the static pin table; they are not.

The same paragraph also states `an SDK version with no matching row there falls back to the table's default pin instead of failing outright` (`:69-70`). `cc9021ab` added the opposite for one case — `engines/vllm/src/lib.rs:2266-2283` now `bail!`s when the unmatched version's major is known to the discovery table. That branch is unreachable through the CLI today (`vllm_install_route` diverts any 10.x to discovery first), so this half is lower severity than the first, but the sentence no longer describes the function it is describing.

`AGENTS.md` §5 (read from the base tip) is explicit: when observable behaviour changes, update the user-facing docs *in the same change* rather than leaving them for a follow-up. This is also the second round in a row that this file has drifted from this code path — the previous round blocked on `docs/vllm.md` describing a command and a guarantee the code did not have, that was fixed in `fda0607`, and then `cc9021ab` changed the behaviour out from under the corrected text. A line in the discovery section naming the ROCm major line rather than a single patch release, and a clause on the fallback sentence, close both.

## Non-blocking
- `engines/vllm/src/lib.rs:87` — `torch_requirement: "torch==2.12.0+rocm10.0.0"` is an exact pin carrying a patch-level ROCm tag, and major-only matching now applies it to any 10.x runtime, so the new doc comment's claim that a discover row is "a live resolver recipe AMD's index applies uniformly across an entire ROCm major line" (`:98-102`) is not true of two of the five pins; on a 10.1 runtime this either fails to resolve or installs a 10.0.0-tagged torch.
- `engines/vllm/src/lib.rs:1555-1600` — the discovery route still asserts nothing about the ROCm local tag on the resolved `vllm` pin, where the static route's `vllm==<v>+<abi>` encodes the ABI by construction; raised last round, still open, and more load-bearing now that one row covers a whole major line.
- `engines/vllm/src/lib.rs:2213-2216` — the docstring's reasoning that "the install passes `--extra-index-url`, so PyPI stays in play and a bare `vllm` can silently resolve to the non-ROCm build" is still inaccurate under uv's default `first-index` strategy, where extra indexes are consulted before the default one; raised last round, unaddressed, and the conclusion it supports is still correct.
- Two guards this PR adds or relies on are uncovered by any test: the `major(recorded).is_some() &&` short-circuit in `rocm_sdk_major_matches` (`engines/vllm/src/lib.rs:126`) is dead given static table keys, and the empty-string filter in `rocm_sdk_version_from_manifest` can be deleted with zero test failures.
- `engines/vllm/Cargo.toml:24` — `tempfile = "3"` sits under `[dependencies]` but every call site is inside `mod tests`; `xtask`, `crates/e2e-report` and `crates/rocm-dash-daemon` all declare it under `[dev-dependencies]` (raised last round, unaddressed).

r0x0r (approved):
Approving — `cc9021ab` fixes both blocking issues properly, and fixes them in the right places.

The split between the two tables is the correct call and the reasoning is written down where the next person will find it: a `VLLM_ROCM_BUILD_TABLE` row pins one exact release's wheel, so exact `major.minor.patch` matching still applies there, while a `VLLM_ROCM_DISCOVER_BUILD_TABLE` row is a live resolver recipe that holds across a whole ROCm major line, so `rocm_sdk_major_matches` is right for it. `10.1.0a20260822` now routes to discovery instead of silently falling through. Thanks for addressing the `apps/rocm/src/therock.rs` inconsistency in the doc comment too — that was the thing most likely to confuse someone later.

The fail-closed fallback is also the right shape. It keeps the pre-PR "unknown 7.x still installs" behaviour that I asked for, but refuses to guess when the version's major is only known through discovery, and the error names the detected version and `ROCM_CLI_VLLM_ROCM_INDEX_URL` so the user can actually act on it.

Good that `vllm_install_target_falls_back_to_the_default_row_for_an_unknown_rocm_sdk_version` was corrected rather than deleted — swapping `10.1.0a20260822` for `not-a-version` keeps the fallback covered while `vllm_install_target_refuses_to_guess_a_static_pin_for_an_unmatched_discover_major` pins the new behaviour.

Verified locally at `cc9021ab`: `cargo fmt -p rocm-engine-vllm -- --check` clean, `cargo test -p rocm-engine-vllm` 79/79, clippy clean with nothing attributable to this PR. The one red check, `E2E tests (Strix Halo, Ubuntu)`, failed in `GPU preflight (bounded wait for an available GPU)` with the test step skipped — runner contention, unrelated to this change, though worth a re-run before merge.

Two things left as follow-ups, neither worth holding the PR:

- The fail-closed guard keys on "this major appears in the discovery table", so a major in *neither* table would still fall back to the rocm723 row silently. Nothing can hit that today — `rocm_sdk_version` comes from a TheRock manifest probe, and those are 7.x and 10.x — but a rule like "only fall back when the major matches the default row's major" would be robust to a future 11.x without depending on the discovery table's contents.
- `therock-next-07` is `@nightly`, so this route's only live-index coverage lands post-merge. Reasonable given it hits AMD's real index, just worth knowing that a discovery break surfaces the next night rather than in PR CI.

This reflects a read of the diff plus the local checks above, not a live ROCm 10 install.

PR #404, opened 2026-09-14
fix(lemonade): align ROCm llama.cpp backend with the active ROCm SDK (EAI-8272)

juhovainio:
## Summary

- Lemonade's `llamacpp:rocm` backend is pinned per-release in a packaged `resources/backend_versions.json`, independent of whatever ROCm SDK `rocm install sdk` has installed/activated. Installing a newer ROCm SDK (e.g. 10.0.0) had no effect on which ROCm build Lemonade actually served inference on — confirmed live, it silently stayed on the embeddable's pinned ROCm 7.13 build.
- Adds a two-tier, verify-before-accept alignment attempt in `prepare_llamacpp_backend_for_active_rocm` (`engines/lemonade/src/lib.rs`), run before the existing backend install:
  1. Point Lemonade's pinned llama.cpp build at the active SDK version by patching `therock.version` in `backend_versions.json`.
  2. If that specific build's release predates ROCm-version-tagged assets (as the currently-shipped embeddable's pin does), query the latest published llama.cpp tag and patch `llamacpp.rocm-stable` to it instead, forcing a real reinstall attempt.
- Neither tier is trusted blindly: every attempt is verified via `ldd` against rocm-cli's own injected ROCm environment before being accepted. Any unverified attempt reverts the JSON and retries once more with the original pinned defaults — never a regression relative to today's behavior, and never a silent CPU fallback (`AGENTS.md` §6).
- Exposes `active_managed_therock_version` in `rocm-core`, modeled on the existing `active_managed_therock_channel` accessor.

## Test plan

- [x] New unit tests: `active_managed_therock_version` (rocm-core, mirrors the existing channel accessor tests) and JSON patch/restore round-trips, semver-shape guard, `ldd`-output parsing (lemonade engine)
- [x] `cargo test -p rocm-core -p rocm-engine-lemonade` — all passing (one pre-existing, unrelated `/opt/rocm`-environment test failure confirmed present on unmodified `main` too)
- [x] `cargo clippy --workspace --all-targets` — clean
- [x] Verified live on real Strix Halo (gfx1151) hardware: installed ROCm SDK 10.0.0, ran `rocm engines install lemonade`, confirmed Tier 1 correctly fails (embeddable's pinned tag has no ROCm-10.0 asset), Tier 2 resolves the current llama.cpp tag and installs successfully, and `rocm serve` produces a real GPU-backed `llama-server` linked against rocm-cli's own ROCm 10.0.0 libraries (confirmed via `/proc/<pid>/maps`)
[...]

PR #420, opened 2026-09-21
fix(hooks): mirror CI's e2e-cucumber clippy step in the pre-push hook

juhovainio:
## Summary
- The local pre-push clippy hook excluded e2e-cucumber entirely, so the ~1.6k lines of e2e harness + step code got no local lint pass before push — which is how the recent `clippy::map_unwrap_or` failure on PR #404 reached CI unnoticed instead of being caught locally.
- Mirrors CI's actual two-step clippy job exactly: drops the `--exclude e2e-cucumber` from the workspace clippy hook (it was also skipping the crate's `rocm-demo-env` bin, which CI does lint), and adds a dedicated `cargo-clippy-e2e` hook running `cargo clippy -p e2e-cucumber --test e2e -- -D warnings`, matching CI's "Clippy (e2e harness + steps)" step.

## Test plan
- [x] `prek run cargo-clippy cargo-clippy-e2e --hook-stage pre-push --all-files` — both hooks pass


siloteemu (commented):
🔴 Automated review · pr-review-watcher · 40d176a3

## Summary
Drops `--exclude e2e-cucumber` from the pre-push `cargo-clippy` hook and adds a second `cargo-clippy-e2e` hook running `cargo clippy -p e2e-cucumber --test e2e -- -D warnings`, so the e2e harness and step code — excluded from `--all-targets` because its test target sets `test = false` — is linted locally the way the CI clippy job already lints it. No blocking findings. Verified: read `.pre-commit-config.yaml` in full at head and base, confirmed both hook entries against the two clippy steps in `.github/workflows/ci.yml` and against `tests/e2e-cucumber/Cargo.toml`, and parsed the config to confirm it loads with 18 hooks and no duplicate ids; no clippy, test or e2e run was performed here. Check state at review: 18 success, 9 skipped, 0 failure. Blocking: 0 · Non-blocking: 5.

**None.**

## Non-blocking
- `docs/testing.md:22` and the gate list in `AGENTS.md` still document only `cargo clippy --workspace --all-targets -- -D warnings` — the exact command this PR's new comment says leaves the harness and step code unlinted. The docs are not wrong, they are now incomplete; adding the `-p e2e-cucumber --test e2e` line to both makes the documented manual path match what the hook and CI enforce.
- The 4-line rationale comment on the new hook is a verbatim copy of the one above the matching CI step (`.github/workflows/ci.yml:760`), with nothing binding the two. `xtask/src/workflow_contract.rs` already asserts workflow invariants from parsed YAML; a short assertion that the clippy job and the hook config carry the same two clippy invocations would stop the pair drifting apart silently.
- `.pre-commit-config.yaml:131` — the sibling `cargo-test` hook still carries `--exclude e2e-cucumber`, so the crate's own library unit tests (resolver and reconciliation logic, which CI runs in a separate step) are still never exercised at pre-push. Same shape of gap this PR closes for clippy; out of its stated scope, worth a follow-up.
- Both CI clippy steps pass `--locked`; neither hook entry does. This matches the existing local-hook convention (no sibling hook uses it) and is defensible for local runs, so the load-bearing side is CI — noting it only because the commit describes the hook as mirroring CI.
- Because the e2e test target is excluded from `--all-targets`, nothing in CI compiles its `#[cfg(windows)]` branches — the Windows job builds and tests `--all-targets` but not this target. A Windows contributor's pre-push would now be the first thing to compile that code. The branches are properly cfg-gated throughout, so this is most likely a benefit rather than a hazard, but it was not verified on a Windows host here.

siloteemu (approved):
Approving on a maintainer's explicit go-ahead for this pull request. Reviewed at 40d176a3; no blocking findings, and the non-blocking notes on the earlier review are follow-up material rather than gates.

fredespi:
Reviewed the hook change against the current PR head. It correctly restores workspace Clippy coverage for the e2e-cucumber crate’s ordinary targets and adds the explicit custom e2e target invocation that `--all-targets` omits, matching CI. I found no blocking issues in the diff.

This looks safe to approve pending final validation; this comment is a diff-review result, not a formal approval.

PR #421, opened 2026-09-21
e2e: PTY-observed download/extraction progress spinner coverage

### LINKED ISSUE #368: E2E coverage: assert download progress spinner renders under a PTY
## Background

Copilot's review on #347 asked for observable-CLI-behavior coverage of the new download progress spinner (`apps/rocm/src/cli_progress.rs`), specifically for:

- `rocm install sdk --format tarball` (`apps/rocm/src/therock.rs::install_tarball_runtime`)
- ComfyUI source download (`apps/rocm/src/comfyui.rs::download_and_extract_source`)

The spinner only activates when stderr is a real TTY (`std::io::stderr().is_terminal()`), so the existing black-box suite — which spawns `rocm` with piped stdio via `std::process::Command` — can never observe it. Today neither download path has *any* E2E coverage, spinner or otherwise.

## Why this isn't a quick fix

Both download paths resolve their source URL by talking to a real remote endpoint at runtime:

- `install_tarball_runtime` → `resolve_tarball_artifact` scrapes an HTML index page at `channel.tarball_base_url()` to find the newest matching tarball.
- ComfyUI uses a fixed `COMFYUI_SOURCE_ARCHIVE_URL` constant.

Neither has an existing override hook, so a deterministic, network-independent test needs new test-only plumbing, not just a new feature file.

## Proposed plan

1. **Add a URL-override test hook**, gated behind the existing `e2e-test-hooks` feature (see `apps/rocm/Cargo.toml`'s `e2e-test-hooks = [...]` precedent), so:
   - `resolve_tarball_artifact` can be pointed at a local index URL instead of `channel.tarball_base_url()`.
   - `COMFYUI_SOURCE_ARCHIVE_URL` can be overridden similarly.
2. **Add a local fixture HTTP server** (reusing `tests/e2e-cucumber/src/mock_server.rs` / `http_server.rs` patterns already in the suite) that serves:
   - A minimal HTML index page in the same format `parse_tarball_index_html` expects, listing one fake tarball entry.
   - The tarball/archive body itself — small, deterministic bytes, ideally served with a throttle/delay so the download spans multiple progress repaints instead of completing in one chunk.
3. **Drive each install path under a PTY** using the existing harness (`tests/e2e-cucumber/tests/e2e/tui_driver.rs`, currently built for the interactive dash TUI but reusable for a plain scrolling stderr stream) or a lighter-weight PTY capture if the full vt100 screen-grid driver is overkill for a single-line carriage-return spinner.
4. **New Gherkin scenarios** asserting:
   - The spinner line shows byte/percentage progress while the download is in flight.
   - The line is cleared (no stale fragment) before the next line of output.
   - Gate the scenario to the appropriate lane if it needs to be excluded from environments without PTY support.

## Scope note

This was flagged during review of #347 (download progress indication) but deferred as a separate follow-up rather than rushed into that PR, since it requires new test infrastructure and a production test-hook design decision rather than a straightforward fix.

jussielo-amd:
## Summary
- Adds two E2E scenarios that drive a real `rocm` binary under a real PTY to prove `cli_progress::AnimatedSpinner` actually renders and clears correctly outside of its existing in-process unit tests:
  - `download-progress-01`: TheRock tarball SDK install — download progress, extraction ("Extracting …") frame, and both spinner lines cleared on exit.
  - `comfyui-04`: ComfyUI source-archive install — download progress only (this path has no separate extraction phase) and the spinner line cleared on exit.
- Adds a shared `PacedDownloadServer` fixture (`tests/e2e-cucumber/src/paced_download.rs`): a loopback HTTP server that streams one named file in fixed-size, delayed chunks (falling back to normal `ServeDir` for everything else), so the PTY has a real chance to observe an intermediate, sub-100% frame instead of the whole transfer completing within one screen poll.
- Adds two `e2e-test-hooks`-gated hooks needed to make the ComfyUI scenario hermetic and fast: `ROCM_CLI_COMFYUI_SOURCE_ARCHIVE_URL_OVERRIDE` (redirect the source-archive download to the fixture server) and `ROCM_CLI_DISABLE_TORCH_RUNTIME_DEP_CHECKS` (skip the real system package-manager dependency check that would otherwise run after a successful SDK install). Both compile out entirely in non-`e2e-test-hooks` builds, so production behavior is unaffected.
- Documents the new override in `docs/release-trust.md`, alongside the existing TheRock base-override documentation.
- **Fixes a real production bug in `cli_progress::Spinner`**: tightening the new e2e assertion to require an actual intermediate (non-0%, non-100%) progress frame reproducibly failed, which traced back to `render_current` truncating the *whole* assembled status line from the tail. On an ordinary 80-column terminal, once the byte count grew past a couple of characters (e.g. `"0 B"` → `"1.5 MiB"`), the trailing `"(NN%)"` got silently ellipsized away — hiding the download percentage for the entire rest of the transfer, with only the very first `(0%)` frame ever visible. Fixed by splitting the label (the file name, already shown elsewhere) from the byte-count/percentage suffix, and truncating only the label when the line overflows, so the percentage always stays visible. Also hardens the new `assemble_status_line` helper against an even narrower-terminal edge case where the suffix itself doesn't fit.

Fixes #368

Why: `AnimatedSpinner` only had in-process unit coverage before this — never run under a spawned subprocess — so a regression in its TTY detection, throttling, or line-clearing on `Drop` could ship unnoticed. This exercises both the tarball and ComfyUI download paths end to end against real fixture servers, and in doing so caught a genuine, independently-reproducible bug in the spinner's own rendering logic (described above).

Risk: low-to-medium. Most of the diff is test-only, but `apps/rocm/src/cli_progress.rs`'s `render_current`/`assemble_status_line` change is a real behavioral fix shipped to production — it changes what gets truncated on narrow terminals during any download or long-running command that uses the spinner. Covered by unit tests (including two regression tests for both truncation edge cases) and the new e2e scenarios.

## Test plan
- `cargo build --workspace --all-targets` — clean
- `cargo test --bin rocm cli_progress` — 15 passed
- `cargo test -p e2e-cucumber --lib` — 122 passed
- `cargo xtask e2e -- -n "download-progress-01|comfyui-04"` — both scenarios pass, run repeatedly, 11/11 steps each time
- `prek run --all-files --no-group local-tools` — clean


tests/e2e-cucumber/tests/e2e/comfyui_steps.rs
```diff
@@ -409,3 +424,136 @@ async fn refusal_lists_both_keys(world: &mut E2eWorld) {
...
+    session
+        .wait_for_screen("%)", PTY_SCREEN_TIMEOUT)
+        .await
+        .unwrap_or_else(|e| panic!("download progress frame never appeared: {e}"));
+    let screen = session.screen_text();
+    assert!(
+        !screen.contains("(100%)"),
+        "expected an intermediate (sub-100%) download progress frame, but the \
+         first observed percent frame was already complete:\n{screen}"
+    );
```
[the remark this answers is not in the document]
jussielo-amd: Good catch — confirmed. Tightened this to wait for a percentage strictly between 0% and 100% (`wait_for_screen_where`), which turned up a real product bug: on an 80-column terminal, the byte-count text growing (e.g. `0 B` → `1.5 MiB`) pushed the whole line past the terminal width, and the old tail-truncation silently dropped the `(pct%)` suffix for the rest of the transfer. Fixed in `cli_progress.rs` (commit 6d884297) by keeping the suffix in its own field and truncating the file name instead when the line doesn't fit.

tests/e2e-cucumber/tests/e2e/therock_steps.rs
```diff
@@ -313,6 +316,172 @@ async fn tarball_index_fixtures(world: &mut E2eWorld) {
...
+    session
+        .wait_for_screen("%)", PTY_SCREEN_TIMEOUT)
+        .await
+        .unwrap_or_else(|e| panic!("download progress frame never appeared: {e}"));
+    let screen = session.screen_text();
+    assert!(
+        !screen.contains("(100%)"),
+        "expected an intermediate (sub-100%) download progress frame, but the \
+         first observed percent frame was already complete:\n{screen}"
+    );
```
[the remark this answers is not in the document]
jussielo-amd: Same fix applied here — see the reply on the ComfyUI copy of this assertion (comment 4060848374). Tightening the check to require a genuine intermediate percentage (neither 0% nor 100%) surfaced a real truncation bug in `cli_progress::Spinner::render_current` that was silently hiding the percentage on an 80-column terminal once the byte count grew a few characters; fixed in commit 6d884297.

tests/e2e-cucumber/tests/e2e/therock_steps.rs
```diff
@@ -313,6 +316,173 @@ async fn tarball_index_fixtures(world: &mut E2eWorld) {
...
+
+#[then("the final terminal screen shows neither spinner line")]
+async fn assert_spinner_lines_cleared(world: &mut E2eWorld) {
+    let session = world
+        .tui
+        .as_ref()
+        .expect("no pty session for the tarball install");
+    let screen = session.screen_text();
+    assert!(
+        !screen.contains(&format!("Downloading {CURRENT_TARBALL}")),
```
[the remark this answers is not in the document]
jussielo-amd: Confirmed — on an 80-column terminal, `assemble_status_line`'s progress suffix leaves only ~48 columns for the label, so the full `Downloading $CURRENT_TARBALL` string was truncated away before it ever rendered; the check passed whether or not the line was actually cleared. Fixed in 27f28d4a: now matches a short prefix ("Downloading therock-dist-linux") that truncation — which always keeps the label's head, not its tail — leaves intact while the line is live, and that disappears once `Spinner::clear` erases it.

jussielo-amd:
Addressing Copilot's "Needs a closer look" finding from [this review](https://github.com/ROCm/rocm-cli/pull/421#pullrequestreview-5266008634) ("Spinner-clearing checks run after the row has scrolled off-screen"): confirmed real. The PTY's `vt100::Parser` has 0 lines of scrollback, and the tarball install's ~20-line summary output was enough to scroll the download/extraction spinner's row off the top of the default 24-row screen before `assert_spinner_lines_cleared` ever read it — a scrolled-off row and a genuinely cleared one are indistinguishable to `screen_text()`, so the check could pass even if `Spinner::clear` were broken.

Fixed in 87831243:
- Added `TuiSession::grow_rows`, which grows only the row count (not columns, unlike the existing `use_detail_size`) — widening columns too would stop the label from truncating, undermining the 80-column truncation behavior this scenario exists to exercise.
- Called it right after spawning the tarball install session.

Verified two ways: dumped the screen at the assertion point and confirmed the whole session's output (from its first line onward) now fits with rows to spare; and temporarily neutered `Spinner::clear()` to confirm the assertion now genuinely fails when clearing is broken (it didn't catch this before the fix).

siloteemu (commented):
🔴 Automated review · pr-review-watcher · 87831243

## Summary

Adds two PTY-driven Gherkin scenarios that observe the download/extraction progress spinner through a real pseudo-terminal, a paced chunked-HTTP fixture server to make intermediate frames observable, and — not purely test work — two production fixes to `apps/rocm/src/cli_progress.rs` (keep the byte/percentage suffix intact when the label overflows; never exceed `max_width` when the suffix alone overflows) plus two `e2e-test-hooks`-gated override accessors. No blocking findings. Verified: mutated each of the two new production branches separately in a scratch copy and confirmed each new unit test reddens on its own assertion (whole-line-truncation revert → the suffix-intact test fails; removing only the narrow-terminal guard → the max-width test fails, while the other still passes); the PTY scenarios themselves were not executed here and the full suite was not run. Check state at review: 19 success, 2 skipped, 0 failure. Blocking: 0 · Non-blocking: 5.

**None.**

Notes on the specific risks checked, since this PR's value rests entirely on whether the new assertions discriminate:

- **Vacuous passes.** Every positive PTY assertion is a `contains` on rendered text, so an empty screen fails it. `wait_for_screen_where`/`wait_for_screen` return `Err` on timeout, on reader-thread panic, and on the child exiting early, so a missing binary or a process that dies does not pass. `wait_for_exit` asserts exit code **0**, not merely that the process exited. `TuiSession::spawn` surfaces an OS spawn failure as an error.
- **Negative assertions vs. scrolling.** The vt100 parser is built with zero scrollback, so a scrolled-off row is indistinguishable from a cleared one. `therock_steps.rs` handles this explicitly by growing the PTY to 60 rows before the summary prints. For the ComfyUI scenario the arithmetic holds independently: only lines printed *after* the spinner row can push it off a 24-row screen, and that path prints ~10, so the row cannot scroll away — see non-blocking #1 for the fragility this leaves behind.
- **Negative assertion vs. truncation.** `assert_spinner_lines_cleared` correctly checks a 30-character *prefix* of the download label rather than the full string, because the live line is truncated on an 80-column terminal; the full string would have matched nothing either way. The extraction label carries no suffix, is 59 columns wide, and so is checked in full — correct. The permanent `  tarball: <file>` line printed at `apps/rocm/src/therock.rs:2611` does not contain the `Downloading …` prefix, so it cannot defeat the check.
- **Readiness waits.** No wait in this diff is satisfied before the action it gates: each is a positive wait for output that only the action produces, and each fails loudly if it never appears.
- **`@serial` is real, not decorative.** I initially concluded the tag was inert because the repo's own tag parser (`tests/e2e-cucumber/src/expectation.rs`) does not recognise it. That was reviewer error: `@serial` is consumed by cucumber-rs's default `which_scenario`, which maps it to `ScenarioType::Serial`; the repo sets only `max_concurrent_scenarios` and never overrides `which_scenario`, and the repo parser silently ignores unknown tags. The tag works as the feature-file comments claim. See non-blocking #4 for the one-line fix that stops this recurring.
- **Coverage vs. the previous version (focus g).** The rename of `format_download_progress` to the private `format_progress_suffix` carries every prior assertion forward. One old assertion — that the rendered string starts with the caller's prefix — is dropped, and that vacation is **correct, not a regression**: the function no longer takes a prefix, so the assertion no longer has a subject; the prefix/suffix composition is now pinned structurally by `assemble_status_line` and by two new tests. The function had no other callers at the base, so nothing lost coverage elsewhere.
- **Sign-off and content.** All 9 commits in the range carry exactly one `Signed-off-by:` trailer each (checked line-by-line from the raw commit objects). No internal names, hostnames, gateways, cluster names, registry paths or internal document links in the diff. No prompt-injection content found anywhere in the checkout.

## Non-blocking

1. `tests/e2e-cucumber/tests/e2e/comfyui_steps.rs:540` — the "no ComfyUI download spinner line" check is safe today only because ~10 lines print after the spinner; the sibling tarball scenario protects the identical assertion explicitly with `grow_rows(60)`. Any future growth of the ComfyUI completion report would silently turn this into a tautology with no test failure to warn you. Call `grow_rows` here too, or state the invariant in a comment.
2. `apps/rocm/src/cli_progress.rs:65` — `set_label` now resets `progress_suffix` to `None`, and nothing tests it; a regression there would leave a stale byte count glued to an unrelated message. One unit test (set_progress, then set_label, then assert the suffix is gone) closes it.
3. `tests/e2e-cucumber/tests/e2e/therock_steps.rs` — `&format!("Downloading {CURRENT_TARBALL}")[..30]` is a raw byte slice with a magic length; it panics if the fixture file name is ever shortened or gains a non-ASCII character. A named constant with a one-line note on why 30, or a char-boundary-safe slice, removes the trap.
4. `tests/e2e-cucumber/src/expectation.rs` — every scenario tag the repo owns is documented on `ScenarioDecl`, and `@serial` is the only tag in the feature files that is not, because cucumber-rs owns it. One line there saying so would have prevented the wrong conclusion I reached above, and will otherwise invite the same conclusion from the next reader.
5. The dependency-install spinner added by the recent spinner work (`apps/rocm/src/comfyui.rs:1712`) is the one spinner this coverage does not reach — the ComfyUI fixture deliberately empties the dependency list so the `uv` block never runs. Worth naming as a known gap in the PR text rather than leaving it implicit. Separately, the PR title reads as test-only, but the diff changes production rendering in `cli_progress.rs` and adds a gated early-return in `main.rs`; the description should say so, since a maintainer skimming "e2e: … coverage" would not expect behaviour changes. The extraction-frame assertion also remains inherently timing-dependent (a 20 ms poll against a real `tar` run) — `@serial` plus the payload sizing mitigate it, but it is the residual flake source to watch.

jussielo-amd:
Addressed all 5 non-blocking items from [review #5270510051](https://github.com/ROCm/rocm-cli/pull/421#pullrequestreview-5270510051) in 9be2f78c + PR title/description:

1. `comfyui_steps.rs` — grew the ComfyUI PTY to 60 rows before install, mirroring the tarball scenario's fix, so a future-longer completion report can't turn the "cleared" check into a tautology the way it could have before.
2. `cli_progress.rs` — added a unit test proving `set_label` clears a stale `progress_suffix`.
3. `therock_steps.rs` — replaced the magic `[..30]` byte slice with a named, documented `DOWNLOADING_PREFIX_LEN` constant and a panic-safe `.get(..).unwrap_or(...)` lookup.
4. `expectation.rs` — documented that `@serial` is intentionally outside `ScenarioDecl`/`from_tags` (it's a cucumber-rs runner concern, not an expectation-resolution one).
5. PR title now names the production fix rather than reading as test-only; description gained a "Known gaps" section calling out the untested `comfyui.rs` dependency-install spinner path and the residual timing-dependence of the extraction-frame assertion.

All verified: full workspace build/clippy/test, targeted + full-suite (repeated, contention-inclusive) e2e runs, and `prek`, all clean.

siloteemu (commented):
🔴 Automated review · pr-review-watcher · 20dbbdbd

## Summary

Adds two PTY-driven Gherkin scenarios that observe the download/extraction progress spinner through a real pseudo-terminal, a paced chunked-HTTP fixture server that makes intermediate frames observable, and production changes to `apps/rocm/src/cli_progress.rs` (split the byte/percentage suffix out of the label so truncation can never eat it, plus a narrow-terminal clamp) alongside two `e2e-test-hooks`-gated override seams. Since the last round it also lands the five remediation items and one further production fix (a doubled space in the narrow-terminal fallback); no blocking findings. Verified: ran the `cli_progress` unit module in a scratch copy (16 passed) and then mutated each production branch separately there — reverting `set_label`'s suffix reset reddens only `set_label_clears_a_stale_progress_suffix`; reintroducing the doubled space reddens only `assemble_status_line_fallback_does_not_double_the_space_before_suffix`; disabling the narrow-terminal guard reddens the two width tests; reverting to whole-line truncation reddens only `keeps_the_progress_suffix_intact…` — so each branch is pinned by its own assertion, not just by a wholesale revert. I also confirmed from source that `"Downloading therock-dist-linux"` (the exact 30-byte prefix) and `"Extracting {tarball}"` and `"Fetching ComfyUI source archive"` each occur exactly once in the product, as spinner labels only, so none of the negative checks is a tautology; that `wait_for_exit` asserts exit code 0 and `wait_for_screen_where` returns `Err` on timeout, child exit and reader panic, so nothing passes on an empty screen; that the only runtime source of `"%)"` reachable from these two commands is the spinner suffix itself; that the vt100 parser has zero scrollback while `grow_rows(60)` comfortably covers the ~20 lines printed after the spinner; that `@serial` is honoured by cucumber-rs's default `which_scenario` and `@requires-os:linux` resolves to Skip; that `e2e-test-hooks` is off by default with no `--all-features` anywhere in CI, so both new override seams are compiled out of release builds; and, measuring it rather than trusting the comment, that high-entropy filler really is gzip-incompressible (20 MB, ratio 1.0003), which is what keeps the pacing real. All 16 commits carry exactly one sign-off; leak and injection scans over the diff are clean. The merge base (93677c8f) is behind the `prw-base` tip (3ecf4843) by four commits, but the resulting diffs are byte-identical, so scope is unaffected. Not run here: the e2e/PTY scenarios themselves and the full workspace suite. I could not check the PR title and description text, since this review never contacts GitHub — the author's claim that item 5 was addressed there is unverified by me. Checks at this head: 18 success, 2 skipped. Blocking: 0 · Non-blocking: 5. One thing to flag about timing: this review was carried out against the commit named in the heading above, and the branch was force-updated onto a newer base while it was running. The branch's own commits look unchanged by that update, but everything above was assessed before it, so it speaks for the commit it names rather than for whatever is at the tip now. Checks at the new tip stood at 18 success, 2 skipped and 1 still running when this was posted.

**None.**

Two things are worth stating plainly, because this PR's whole value rests on whether the new assertions discriminate.

- **The remediation round was weighed as original code, not as review answers.** Item 2 (`set_label` clearing a stale suffix) and the unprompted item in `50fa83d` (dropping the doubled space) are both real production edits, and both were mutation-tested above: each fails exactly one test and leaves the others green. Item 3 replaced the raw byte slice `[..30]` with `DOWNLOADING_PREFIX_LEN` plus `.get(..).unwrap_or(&downloading)`, which is genuinely panic-safe — a shortened or non-ASCII fixture name now degrades to checking the full string rather than aborting. Item 7's shim dedup (`write_gpu_probe_shim`) is byte-identical to both previous inline copies. Item 6's `rand` swap replaces a hand-rolled PRNG with a seeded `StdRng` and honestly documents in the same comment that `rand` does not promise algorithm stability across versions, so a future bump could change the fixture bytes — that caveat is accurate, not hedging.
- **Where the Drop-time clear is actually proven.** `Spinner::clear` on `Drop` is covered, but not by the half of the assertion a reader would assume — see non-blocking #1. The ComfyUI scenario is what carries it for a download spinner, and the extraction half of the tarball assertion carries it there; both were traced through the real output sequence rather than assumed.

## Non-blocking

1. `tests/e2e-cucumber/tests/e2e/therock_steps.rs:1300` — the `"Downloading therock-dist-linux"` half of `assert_spinner_lines_cleared` cannot fail for a broken Drop-time clear: the extraction spinner starts on the same row and its first repaint issues `Clear(CurrentLine)`, erasing the download line whether or not `Spinner::clear` ran. The extraction half and the ComfyUI scenario do cover that defect, so nothing is untested — but the step's own comment reads as if both halves prove it. One sentence saying the download half guards in-place repaint (not Drop-clear) would stop the next reader drawing the same wrong conclusion I nearly did.
2. `apps/rocm/src/cli_progress.rs:208` — `assemble_status_line` takes `suffix: Option<&str>` but its narrow-terminal branch silently depends on the caller's suffix carrying its own leading space; a future second caller passing a space-less suffix would glue frame and suffix together with no test to catch it. The invariant is pinned today only because `format_progress_suffix` is the sole producer.
3. `apps/rocm/src/cli_progress.rs:123` — the `Err(_)` arm of `render_current` still emits an unbounded, untruncated line when `terminal::size()` fails. This is preserved base behaviour, not a regression, and it is the one rendering path with no coverage at any level.
4. `tests/e2e-cucumber/tests/e2e/therock_steps.rs:1229` and `comfyui_steps.rs:1042` — `grow_rows(60)` is issued immediately after `TuiSession::spawn`, with no synchronisation point, so there is a theoretical window where the child could emit more than 24 lines before the resize lands. Not reachable at real process-startup timings, but the assumption is unstated.
5. The extraction-frame assertion remains inherently timing-dependent (a 20 ms poll against a real `tar` over a ~20 MB archive); `@serial` plus the payload sizing give it roughly a 90-poll margin on the download side, but it stays the residual flake source to watch. Separately, `apps/rocm/src/comfyui.rs:1712`'s dependency-install spinner is still the one spinner this coverage deliberately does not reach.

jussielo-amd:
Addressed the non-blocking items from the automated review at commit `20dbbdbd`:

- **#1** (download-half of `assert_spinner_lines_cleared` doesn't itself prove `Drop`-time clear) — fixed: reworded the comment to say explicitly that the extraction assertion (and the ComfyUI scenario) are what actually cover that, not this check.
- **#2** (`assemble_status_line`'s suffix-leading-space invariant unenforced) — fixed: added a `debug_assert!` so a future space-less caller fails loudly in tests instead of silently gluing frame/suffix together.
- **#3** (`render_current`'s `Err(_)` fallback emits an unbounded, untruncated line) — fixed: it now shares `assemble_status_line` with an 80-column fallback width, same as every other path.
- **#4** (`grow_rows(60)` issued right after `spawn` with no sync point) — documented rather than eliminated: added a comment stating the timing assumption explicitly (fork/exec + first repaint take longer than the resize call), matching your own "not reachable at real timings" assessment.
- **#5** (extraction-frame timing dependence, ComfyUI dependency-install spinner not covered) — left as-is; both are already called out in the PR description's "Known gaps" as accepted scope limits, not regressions.

Also caught and fixed an off-by-one in `assemble_status_line` itself (exact-fit boundary produced a doubled space) via a separate review pass, plus a couple of minor cleanups (deduped two near-identical PTY step assertions onto `TuiSession`, resolved the ComfyUI source URL once instead of twice). All commits since your review: `3ef8bb8d`, `d103d4dc`, `3b03f6d9`, `6474f5a0`. Full workspace build/clippy/fmt clean, and both `download-progress-01`/`comfyui-04` PTY scenarios re-verified passing on the latest commit.

tests/e2e-cucumber/src/paced_download.rs
```diff
@@ -0,0 +1,261 @@
...
+/// Builds a real gzip tarball and returns its bytes.
+///
+/// Packages the single top-level directory `build_dir.join(dir_name)` into
+/// `build_dir.join(archive_name)`. Shared by the TheRock tarball and ComfyUI
+/// source-archive fixtures, which each need a genuine archive for their
+/// installer's real `tar` extraction to unpack once the paced download
+/// completes.
+pub fn build_gzip_tarball(build_dir: &Path, archive_name: &str, dir_name: &str) -> Vec<u8> {
+    let archive_path = build_dir.join(archive_name);
+    let status = std::process::Command::new("tar")
```
juhovainio: This runs `tar` synchronously (`Command::status()`) inside an `async fn` given-step, blocking a tokio worker thread for the duration of the call. Building a ~20MB archive isn't instant, and with the mock lane's default scenario concurrency, several fixtures building at once could occupy every worker thread and starve other in-flight async scenarios. Worth wrapping in `tokio::task::spawn_blocking` so it doesn't compete with the runtime's async work.

tests/e2e-cucumber/tests/e2e/therock_steps.rs
```diff
@@ -313,6 +326,198 @@ async fn tarball_index_fixtures(world: &mut E2eWorld) {
...
+    // Note this download-line check by itself can't distinguish a real
+    // `Drop`-time clear from the extraction spinner's first repaint simply
+    // overwriting the same row with `Clear(CurrentLine)` — it would pass
+    // either way, since extraction always starts once the download spinner
+    // stops. The extraction assertion just below, and the ComfyUI scenario's
+    // equivalent check (a spinner with nothing after it to overwrite its
+    // row), are what actually prove `Spinner::clear` runs on `Drop`.
+    let downloading = format!("Downloading {CURRENT_TARBALL}");
+    let downloading_prefix = downloading
+        .get(..DOWNLOADING_PREFIX_LEN)
```
juhovainio: `downloading.get(..DOWNLOADING_PREFIX_LEN).unwrap_or(&downloading)` silently falls back to the *full* untruncated string if byte offset 30 isn't a UTF-8 char boundary. Not reachable today since `CURRENT_TARBALL` is ASCII, but if that ever changes, the assertion would degrade to checking whether the screen contains a string a live truncated line never actually shows, and the test would pass regardless of whether `Spinner::clear` ran. Consider asserting the split succeeds (or using a char-boundary-safe truncation helper) so a future change to the constant can't silently defeat this check.

apps/rocm/src/comfyui.rs
```diff
@@ -324,19 +336,20 @@ pub(crate) fn install(
...
-    if source_path.exists() {
+    let source_url = if source_path.exists() {
         println!("Using existing ComfyUI source folder...");
         let _ = io::stdout().flush();
         writeln!(
             log,
             "Using existing ComfyUI folder at {}.",
             source_path.display()
         )?;
+        comfyui_source_archive_url()
```
juhovainio: On the "reuse existing source folder" path, the manifest's `source_url` is populated by a fresh call to `comfyui_source_archive_url()`, not the URL that actually produced the folder on disk. In production this is invisible today since that function always returns the same hardcoded constant, but under the test-only override hook (or if this ever becomes dynamic), install-then-reinstall-with-a-different-override would record a `source_url` that doesn't match what's actually on disk.

tests/e2e-cucumber/src/paced_download.rs
```diff
@@ -0,0 +1,261 @@
...
+            .url()
+            .join(path)
+            .unwrap_or_else(|e| panic!("{path} is not a valid relative URL: {e}"));
+        reqwest::get(url)
+            .await
+            .unwrap_or_else(|e| panic!("request for {path} failed: {e}"))
+    }
+
+    #[tokio::test]
+    async fn serves_the_paced_file_byte_for_byte() {
```
juhovainio: These unit tests only exercise single-segment routes (`archive.tar.gz`, `paced.bin`). Neither real caller's actual route shape is covered here (`archive/comfyui-source.tar.gz` in `comfyui_steps.rs`, `tarball/current/<name>` in `therock_steps.rs`) - both are multi-segment paths. A future axum/tower_http routing regression could break both real E2E scenarios while these unit tests keep passing. Worth adding a multi-segment case.

juhovainio (changes requested):
Nice PR overall. The core fix in `cli_progress.rs` (`assemble_status_line` truncating the label instead of the whole line, so the download percentage suffix always survives) is exactly right, has a clear regression test proving the original bug, and the two new `e2e-test-hooks` production hooks compile out to hardcoded safe defaults outside test builds. CI is green including the new E2E scenarios.

I found a few issues in the new test infrastructure worth fixing before merge, left as inline comments. None of them touch the production fix itself.

- `build_gzip_tarball` shells out to `tar` synchronously from inside `async fn` given-steps. That blocks a tokio worker thread for the duration of a ~20MB tar invocation, and with the mock lane's default concurrency that can starve other in-flight scenarios' async work, similar in spirit to the CPU contention the "Known gaps" section already calls out for a different reason.
- `assert_spinner_lines_cleared`'s prefix truncation falls back to the full untruncated string on a non-char-boundary split instead of failing loudly, which would silently defeat the exact regression the assertion exists to catch.
- The "reuse existing source folder" path in `comfyui.rs::install()` records a fresh call to `comfyui_source_archive_url()` as the manifest's `source_url`, not the URL that actually produced the on-disk folder. Invisible in production today (hardcoded constant), but a latent provenance bug under the test override.
- `paced_download.rs`'s own unit tests only exercise single-segment routes, never the multi-segment routes both real E2E scenarios actually register.

tests/e2e-cucumber/src/paced_download.rs
jussielo-amd: Fixed — `build_gzip_tarball` now runs the `tar` invocation inside `tokio::task::spawn_blocking` instead of blocking the calling task's worker thread (7fe355b7). Worth noting for context: both current call sites are on `@serial`-tagged scenarios, so cucumber 0.23's `@serial` semantics mean no other scenario is actually in-flight while this runs today — but the helper is shared/reusable, so the fix guards against any future non-serial caller hitting the real starvation case, and it's free either way.

tests/e2e-cucumber/tests/e2e/therock_steps.rs
jussielo-amd: Fixed — the split now panics with a descriptive message if `DOWNLOADING_PREFIX_LEN` isn't a char boundary, instead of silently falling back to the untruncated label (7fe355b7).

apps/rocm/src/comfyui.rs
jussielo-amd: Fixed — the reuse path now reads `source_url` back from the manifest already on disk (via the existing `load_manifest`) instead of calling `comfyui_source_archive_url()` fresh, falling back to that only if there's no prior manifest to read (7fe355b7).

tests/e2e-cucumber/src/paced_download.rs
jussielo-amd: Fixed — added `serves_a_multi_segment_paced_route_byte_for_byte`, covering the same multi-segment route shape both real callers register (7fe355b7).

siloteemu:
🔴 Automated review · pr-review-watcher · 82f7f5a1

## Summary
Adds PTY-observed download-progress e2e coverage (a paced HTTP fixture plus two scenarios) and reworks the status-line assembly in `apps/rocm/src/cli_progress.rs` so a download's byte/percent suffix survives truncation; this round is the author's remediation of our five prior non-blocking items plus a self-found off-by-one. One blocking issue, introduced by a cleanup commit rather than by the original work. Verified: re-derived the `assemble_status_line` boundary arithmetic by hand on both sides and ran the `cli_progress` unit tests only (17 passed) — the full test suite and the PTY e2e scenarios were not run here; leak scan over the diff is clean and no conflict markers remain. Checks at review start: 22 success, 2 failure, 2 pending, 2 skipped; at publication, on the same head, 23 success, 2 failure, 1 pending, 2 skipped. Blocking: 1 · Non-blocking: 4.

`apps/rocm/src/comfyui.rs:353` — the "resolve the source URL once" cleanup added `load_manifest(paths)?` to the existing-source-folder branch of `install()`. `install()` did not read the manifest anywhere before this PR (confirmed: the first `load_manifest` call reachable from `install` is this new one), and `ComfyUiManifest` carries no `#[serde(default)]` on any field, so an unparseable or schema-outdated `manifest.json` now makes `rocm comfyui install` abort with "failed to parse …". Before this change that same command succeeded and rewrote the manifest — so the cleanup removed the one command that recovered from a broken manifest, while every other ComfyUI subcommand already fails on it. The adjacent comment reinforces the gap: it states the fallback covers "no prior manifest to read", which is true only for a *missing* file, not an unreadable one. `--reinstall` still works (it deletes the source folder first and takes the download branch), but a plain `install` gives a parse error with no hint of that. Fix: make the read non-fatal so an unreadable manifest falls back exactly like a missing one — `load_manifest(paths).ok().flatten().map_or_else(comfyui_source_archive_url, |manifest| manifest.source_url)` — and reword the comment to say "no prior manifest that can be read".

## Non-blocking
- `apps/rocm/src/cli_progress.rs:146` — the new fallback comment ends "but not untested"; nothing exercises the `crossterm::terminal::size()` error branch (it is the only reference to that call in the tree), so `FALLBACK_WIDTH` could be changed to any value with no test reddening. Drop the clause or say the assembly is tested at that width, not the fallback selection.
- `apps/rocm/src/main.rs:9418` — `ensure_torch_runtime_dep`'s doc comment still says "no-op on Windows or when the dependency is already present" and was not updated for the new e2e-only no-op condition added directly beneath it.
- `apps/rocm/src/comfyui.rs:339` — beyond "resolve once", this branch also changes production behaviour: the manifest now records whatever URL a prior install wrote rather than the current constant, so an upgrade that changes the hardcoded URL will keep re-recording the old one while the source folder survives. Defensible, but the commit subject undersells it; worth a line in the PR text.
- `tests/e2e-cucumber/tests/e2e/comfyui_steps.rs` — the "no ComfyUI download spinner line" assertion is a pure negative and would pass vacuously if the spinner never rendered; it is non-vacuous only because the two preceding steps in comfyui-04 already prove the frame appeared. That ordering dependency is load-bearing and currently implicit.

## Prior round
Items 1–3 are genuinely addressed in the code: the reworded comment now names the extraction assertion as what actually covers the `Drop`-time clear, the `debug_assert!` fires in the dev/test profile where it is meant to (no `[profile.release]` override re-enables it, and `debug_assert!` is already a repo convention), and the `Err(_)` path now shares `assemble_status_line` at an 80-column fallback. The claimed exact-fit off-by-one fix is correct on both sides — I re-derived it: `reserved >= max_width` routes the zero-label-budget case to the fallback, and with `>` instead the output gains a doubled space, which the new boundary test detects; the label branch is bounded by `reserved + label_budget == max_width` and the fallback by `truncate_to_width`, so neither can overflow. The assertion dedup onto `TuiSession` is a mechanical extraction with no condition, timeout, or diagnostic dropped. Item 4 ("documented rather than eliminated") is weaker than a real sync point but the failure mode is a flaky test rather than a wrong verdict, and item 5 remains a stated scope limit — both acceptable. The one place remediation went wrong is the ComfyUI source-URL cleanup above, which is new damage this round rather than anything the original work had.

siloteemu (dismissed):
pr-review-watcher · 82f7f5a1

One blocking item. The five items from the previous round are genuinely addressed, and the off-by-one fix checks out on both sides of the boundary -- the problem is new damage from a cleanup commit in this round, not from the original work.

The change that resolves the ComfyUI source URL once added a manifest read to the existing-source-folder branch of install. That command did not read the manifest at all before this PR, and the manifest type has no defaults on any field, so an unparseable or schema-outdated manifest now makes a plain install abort with a parse error. Before this change the same command succeeded and rewrote the manifest. That made install the one subcommand able to recover from a broken manifest, and the cleanup removed it -- every other ComfyUI subcommand already fails on that file. Reinstall still works, because it deletes the source folder first and takes the download path, but a plain install gives the user a parse error with no hint that reinstall would clear it.

The adjacent comment makes it harder to spot: it says the fallback covers having no prior manifest to read, which is true for a missing file but not for an unreadable one.

What clears this: make the read non-fatal so an unreadable manifest falls back exactly as a missing one does, and reword the comment to say no prior manifest that can be read.

jussielo-amd:
Addressed the blocking issue from [this review](https://github.com/ROCm/rocm-cli/pull/421#issuecomment-5811260857):

`comfyui.rs:353` — confirmed. Fixed in 677d9d71: the reuse-existing-source-folder path now treats an unreadable/unparseable manifest the same as a missing one (`load_manifest(paths).ok().flatten().map_or_else(...)`) instead of propagating the parse error and aborting `install`. Comment reworded to say "no prior manifest that can be read" rather than "no prior manifest to read".

`cargo build`/`clippy`/`cargo test -p rocm comfyui` all clean on top of the fix.

Non-blocking items from this review are noted but not addressed in this pass.

siloteemu:
🔴 Automated review · pr-review-watcher · 677d9d71

## Summary

Adds PTY-observed download-progress coverage for tarball and ComfyUI installs (a paced-chunk loopback fixture server, two Gherkin scenarios, shared TuiSession assertions), hardens `cli_progress`'s status-line width and truncation arithmetic, and threads the resolved source URL into the ComfyUI manifest instead of re-resolving it. Outcome: Needs work — one missing-regression-test blocker on the head commit; everything else reviewed is in good shape, and the test code is unusually careful about tautological assertions. Verified: ran `cargo test -p rocm --bin rocm cli_progress` (17 passed, 0 failed); confirmed `prw-base` is the merge base, so diffing against the base tip and against the merge base yields an identical 16-file set; read every commit in the range from the raw git objects; confirmed `load_manifest` returns `Err` on read or parse failure and `Ok(None)` on absence, that `install()` did not call it at all at the base tip while all four other call sites propagate with `?`, and that `install()` still ends in `save_manifest`, so the recovery property is real; confirmed `ComfyUiManifest::source_url` is written but never read anywhere outside this one new self-referential read, which bounds the blast radius of the change; confirmed `e2e-test-hooks` is opt-in (no `default` feature list) and that no release workflow passes it, so `docs/release-trust.md`'s claim that the ComfyUI URL override cannot exist in a production build is accurate; confirmed cucumber 0.23's default `Runner::Basic::which_scenario` maps a bare `serial` tag to `ScenarioType::Serial` and that this crate does not override `which_scenario`, so `@serial` is a real mechanism and not a decorative tag; confirmed the new PTY predicate is falsifiable (an empty or spinner-free screen fails it rather than passing vacuously); leak, prompt-injection and conflict-marker scans all clean; all 22 commits carry sign-off and a signature block. The full suite, the e2e lanes and a workspace-wide build were not run here. This review worked from a CI conclusion set of 24 success, 2 skipped, 1 pending, 1 failure; nothing below is attributed to any particular job, and the failing check was not inspected. Blocking: 1 · Non-blocking: 5.

**Prior objection: discharged.** This automation's open change request at 82f7f5a1 held that a cleanup commit in the same round added a manifest read to a command that never read it, turning a recoverable broken manifest into a hard parse-error abort. That was accurate at 82f7f5a1, where the reuse branch read `load_manifest(paths)?`. At 677d9d71 the same line reads `load_manifest(paths).ok().flatten()`, so a read or parse failure now falls back to the current source URL instead of propagating, and `install()` goes on to rewrite the manifest. The recovery property the objection named is restored, and the objection no longer holds.

**`apps/rocm/src/comfyui.rs:355` — the head commit fixes a regression this PR itself introduced, and ships no test; the fixed branch remains uncovered.**

677d9d71 ("fix(comfyui): don't abort install on unreadable manifest") touches one file, seven insertions, no test. Coverage of the two arms the PR added is split and incomplete:

- The `Ok(None)` arm (no manifest on disk, fall back to the current URL) is exercised incidentally: the pre-existing `ready_install_pending_dependencies` fixture plants a ComfyUI source folder with a `requirements.txt` but no ComfyUI manifest, so scenarios `comfyui-01` and `comfyui-02` do take the reuse branch. Neither asserts anything about `source_url`, so this is execution, not coverage.
- The `Err` arm — an unreadable or unparseable manifest, which is precisely what the prior objection was about — is exercised by nothing. No unit test reaches it: the only test that calls `install()` is the dry-run one, which returns at the `options.dry_run` early return before this code. No scenario reaches it either: the new `comfyui-04` is a fresh install, so it takes the download branch, and there is no corrupt-manifest scenario anywhere under `tests/e2e-cucumber/features/`.

The consequence is that the change survives a wholesale revert of 677d9d71 and survives the sharper mutation too: flip `.ok().flatten()` back to `?`, or swap the `map_or_else` arms, and every test in the repository still passes. That is the exact defect this PR already shipped once at 82f7f5a1 and had to correct in a follow-up commit, with nothing left in the tree to stop it recurring. Because `source_url` is write-only bookkeeping, the provenance half of this is low-stakes — but the abort half is a user-visible failure of `rocm comfyui install`, and it is the half with no coverage at all.

This also conflicts with `AGENTS.md` §3 as stated at the base tip: every bug fix ships a regression test in the same change that fails before and passes after, and user-observable behaviour — here, whether install aborts with a parse error or completes — needs a scenario or an explicit note in the PR text saying why not.

Concrete fix, and it is cheap. Extract the reuse branch into a free function:

    fn reused_source_url(paths: &AppPaths) -> String {
        load_manifest(paths)
            .ok()
            .flatten()
            .map_or_else(comfyui_source_archive_url, |manifest| manifest.source_url)
    }

then add three unit tests using the existing `test_paths` helper: no manifest on disk returns the current URL; a valid manifest recording a different URL returns that URL verbatim; a manifest file holding bytes that are not valid JSON returns the current URL rather than an error. The third is the mutation-sensitive one — it fails the instant `.ok().flatten()` becomes `?`.

## Non-blocking

- `apps/rocm/src/comfyui.rs:355` — the discarded `load_manifest` error is not logged, though the install log handle is open two lines above; a corrupt manifest silently changes recorded provenance with no trace. One `writeln!(log, ...)` in the fallback would fix it, and would also make the blocking test above easier to assert on.
- `tests/e2e-cucumber/src/paced_download.rs` — the multi-segment route test cannot distinguish any branch in this crate, since the route is a plain `format!("/{paced_file}")` literal that axum matches identically with or without slashes; it pins upstream routing behaviour, which its own comment says, but it is not coverage of this PR's code.
- Commit history — 22 commits including one `wip(e2e): ...` and two that exist only to satisfy fmt and clippy; the repo's own conduct rules ask for meaningful commits, so that noise is worth squashing before merge.
- `tests/e2e-cucumber/src/paced_download.rs` — `deterministic_payload` seeds `StdRng`, which carries no cross-version reproducibility guarantee, so a future `rand` bump would silently change the fixture bytes and the compressed size that the pacing constants were tuned against. The doc comment flags this; nothing pins it.
- `tests/e2e-cucumber/tests/e2e/comfyui_steps.rs` and `therock_steps.rs` — `grow_rows(60)` is issued immediately after spawn with no synchronisation barrier, relying on fork/exec being slower than the resize call. Both sites document the assumption honestly, but it is a race by construction; a wait-for-first-output barrier would remove it.

siloteemu (dismissed):
🔴 Automated review · pr-review-watcher · 677d9d71

Blocking: 1. This replaces our earlier change request on this pull request, which has been retired - the objection it named is discharged at this head. The objection below is new and is the only one now outstanding from this account.

**apps/rocm/src/comfyui.rs:355 - the head commit fixes a regression this pull request itself introduced, and ships no test; the fixed branch stays uncovered.**

The head commit touches one file, seven insertions, no test. Coverage of the two arms this pull request added is split and incomplete:

- The no-manifest-on-disk arm is exercised incidentally, because an existing fixture plants a source folder with no manifest, so two scenarios do take the reuse branch. Neither asserts anything about the recorded source URL, so that is execution rather than coverage.
- The unreadable-or-unparseable-manifest arm - precisely the case the retired objection was about - is exercised by nothing. The only unit test that calls install() returns at the dry-run early return before this code, and the new scenario is a fresh install, so it takes the download branch. There is no corrupt-manifest scenario anywhere in the feature files.

The consequence is that the change survives a wholesale revert of the head commit, and survives the sharper mutation too: flip the fallback back to error propagation, or swap the two arms, and every test in the repository still passes. That is the same defect this pull request already shipped once and had to correct in a follow-up commit, with nothing left in the tree to stop it recurring. The provenance half of this is low-stakes because the recorded URL is write-only bookkeeping, but the abort half is a user-visible failure of the install command, and it is the half with no coverage at all.

The contributor rules as stated on the base branch also ask that every bug fix ship a regression test in the same change that fails before and passes after, and that user-observable behaviour - here, whether install aborts with a parse error or completes - get a scenario or an explicit note in the pull request text saying why not.

Concrete fix, and it is cheap. Extract the reuse branch into a small free function that loads the manifest, discards the error, and falls back to the current source URL. Then add three unit tests using the existing test-paths helper: no manifest on disk returns the current URL; a valid manifest recording a different URL returns that URL verbatim; a manifest file holding bytes that are not valid JSON returns the current URL rather than an error. The third is the mutation-sensitive one - it fails the instant the fallback becomes error propagation again.

Everything else reviewed at this head is in good shape; the full round of findings is in the review comment posted alongside this one.

jussielo-amd:
Addressed the blocking issue from [this review](https://github.com/ROCm/rocm-cli/pull/421#issuecomment-5812985713):

`comfyui.rs:355` — confirmed the coverage gap. Fixed in 7b71f43e: extracted the reuse-branch closure into a free function `reused_source_url(paths: &AppPaths) -> String`, and added three unit tests using the existing `test_paths` helper:

- `reused_source_url_falls_back_to_current_url_when_no_manifest_on_disk`
- `reused_source_url_returns_recorded_url_from_valid_manifest`
- `reused_source_url_falls_back_to_current_url_on_unparseable_manifest`

The third test is the mutation-sensitive one — verified locally that it fails immediately if the fallback reverts to propagating `load_manifest`'s error instead of `.ok().flatten()`.

`cargo build`/`clippy --all-targets`/`cargo test -p rocm comfyui` all clean on top of the fix (32 tests passed).

Non-blocking items from the review are noted but not addressed in this pass.

siloteemu:
🔴 Automated review · pr-review-watcher · 36a59e92

## Summary

Adds PTY-driven e2e coverage for the download/extraction spinner (a paced loopback HTTP fixture plus two scenarios) and fixes the truncated progress display it exposed, by splitting the spinner's text into label + suffix and truncating only the label. No blocking findings. Verified: in a scratch copy of the tree, `cargo test -p rocm --bin rocm` for the `comfyui::tests::reused_source_url*` group (3 pass) and the `cli_progress` module (17 pass), then five single-branch mutations — confirming that reverting the label-truncation branch to whole-line tail truncation fails `assemble_status_line_keeps_the_progress_suffix_intact_when_the_label_would_overflow`, that the corrupt-manifest fallback and the recorded-URL lookup each die to their own named test, and that two sub-details of the label branch survive every test (item 1 below); also confirmed first-hand the merge base equals the base tip with identical file sets, the leak scan over the diff is clean, and no prompt-injection content is present. Not run: the full workspace suite, the e2e/cucumber suite, clippy, and the Windows/GPU lanes. Working from CI check counts at this head: 23 success, 2 failure, 2 skipped, 1 pending. Blocking: 0 · Non-blocking: 5.

## Standing objection

**DISCHARGED.**

First-hand from the tree, not from commit text: `apps/rocm/src/comfyui.rs:1003` now holds a free function `reused_source_url(paths)` that loads the manifest, discards the error via `.ok().flatten()`, and falls back to `comfyui_source_archive_url()` — exactly the extraction the objection asked for. `install()` at `apps/rocm/src/comfyui.rs:339` calls it in the reuse arm. Three unit tests exist at `apps/rocm/src/comfyui.rs:2175`, `2184`, `2196`, covering no-manifest, valid-manifest-with-a-different-URL, and bytes-that-are-not-JSON respectively.

Mutation-verified in a scratch copy, never in the reviewed tree:

- Replacing `.ok().flatten()` with `.expect(...)` — i.e. the fallback becoming error propagation again, the exact regression the objection named — fails `reused_source_url_falls_back_to_current_url_on_unparseable_manifest` and only that test. The unreadable/unparseable arm is now genuinely covered, not merely executed.
- Swapping the arms so the function always returns the current URL fails `reused_source_url_returns_recorded_url_from_valid_manifest` and only that test. The provenance half is covered too.

Every commit in the range was read whole from the raw object; the two commits after the one the objection was filed at are the extraction plus tests (`7b71f43`) and a formatting-only follow-up (`36a59e9`), and nothing in them weakens the fix.

**None.**

## Non-blocking

- `apps/rocm/src/cli_progress.rs:254` — the ordinary label-fits path has no exact-output assertion; in a scratch copy, dropping the frame/label separator space and separately shrinking `label_budget` by one each left all 17 module tests green. The PR's actual fix is covered (mutation E above), so this is a gap on sub-details rather than on the headline change; one `assert_eq!(line, format!("⠋ short{suffix}"))` at a non-boundary width closes it.
- `apps/rocm/src/cli_progress.rs:149` — the comment calls the `terminal::size()`-failure fallback "not untested", but nothing forces an `Err` from it and `FALLBACK_WIDTH` is referenced only at its single use site; reword to say the fallback *value* matches the harness width so the resulting rendering is covered, while the `Err` branch itself is not exercised — otherwise the next reader hunts for a test that does not exist.
- `apps/rocm/src/cli_progress.rs:215` — the doc's first paragraph says "frame and suffix alone" meet or exceed `max_width`, but `reserved` also includes the mandatory separator space; the later paragraph gets it right, so only a skim misleads.
- `apps/rocm/src/cli_progress.rs:229` — `suffix.filter(|s| !s.is_empty())` is unreachable from every current caller and untested; either drop it or say inline that it is defensive-only.
- `tests/e2e-cucumber/features/comfyui.feature` — comfyui-04's label is short enough never to truncate at 80 columns, so it does not exercise the truncation fix; only the tarball scenario is a regression test for it, and a one-line note would stop someone later citing comfyui-04 as that coverage.

Observation, not a finding: two checks are red at this head and I could not attribute either to a specific defect in this diff.

siloteemu (commented):
🔴 Automated review · pr-review-watcher · 36a59e92

Blocking: 0. This is a formal review record, not an approval — this account files no GitHub approvals. The change request previously held by this account has been withdrawn as discharged.

The reuse-branch extraction and its three unit tests were verified by mutation in a scratch copy: turning the fallback back into error propagation, and swapping the arms, each fail their own named test and nothing else.

Five non-blocking items and the full verification detail are in the review comment posted alongside this one.

